PR #15, opened 2018-12-01
Removed empty comp_ders [] in min_rng ders. Changes are made around line 44 and 90 (in new file)

khanh93vn:


boris-kz (approved):
Thanks Khanh, this will do for now.
Although it might be better to use r:
if r:
     e_.append((pri_p, d, m), comp_ders_)
else:
     e_.append(pri_p, d, m)
etc.
Good catch on ini_y +1!

boris-kz:
if P[5]:
          ip = e_[i][0][0]
          (pri_ip, i_d, i_m), input_ders_ = pri_e
else:
          ip = e_[i][0]
          (pri_ip, i_d, i_m), input_ders_ = pri_e, []
?

khanh93vn:
-About using r instead of comp_ders_: I added a breakpoint on line 99 with the condition: comp_ders_ and r == 0. The debugger stopped at that breakpoint more than once, meaning that there are cases where r = 0 while comp_ders_ is not empty. Or do you mean that when r = 0, adding comp_ders_ is not neccessary anymore?

-About P[5]: If com_ders_ is added according to r as mentioned above then this should be correct, I guess

PR #34, opened 2020-01-27
Implemented the C++ version native wrapper for from_p

MayukhSobo:
Implemented the faster version of `from_P_` in C++11 and integrated it with a C++/Python wrapper. We achieved performance improvement of several folds.

### DIFF
--- a/frame_2D_alg/README.md
+++ b/frame_2D_alg/README.md
@@ -1,0 +1,21 @@
+# Building the Native Libraries
+
+- Ensure you have a valid c++11 library
+- Ensure [swig](http://www.swig.org/download.html) is installed on your system
+- Numpy core libraries should be installed
+- Open terminal and give the following command
+
+```sh
+$ bash build.sh
+```
+
+This shall generate many files. 2 most important files are `_f_p.so` and `f_p.py`.
+
+- Make sure you have these two files in your local directory.
+- Now run the `frame_blob.py` file. We have modified the code to accept the native implementation of frame_blobs.
+
+
+# More things to do
+
+- Improve then memory management of the C++ code. (Now we are not deallocating the memory)
+- Implement the `scan_p` in C++ and integrate it too.
--- a/frame_2D_alg/build.sh
+++ b/frame_2D_alg/build.sh
@@ -1,0 +1,11 @@
+rm -f *.o f_p_wrap.cpp _f_p.so f_p.py
+rm -rf __pycache__
+
+g++ -std=c++11 -O3 -march=native -fPIC -c f_p.cpp
+swig -python -c++ -o f_p_wrap.cpp f_p.i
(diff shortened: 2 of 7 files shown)


boris-kz:
Thanks a lot Mayukh! Khanh will go over your code soon, he is currently busy.

MayukhSobo:
The C++ code is present in the first PR

boris-kz:
Thanks, I realized that a minute later 😀

kwcckw:
didn't realize we can just call the c++ library in python in this way.  f_p.i, f_p.h, f_p.cpp are used for the form_P_ function right? and the header file was used to import the library in python environment. I also can see there's a numpy.i, is this in used as well? I guess the complicated part is the integration, to convert the variable so that it is parse-able between c++ and python?

boris-kz:
I think we should keep this version separate for now. Mayukh, could you
post it in your repo instead, I may copy-paste it latter?

ninjakx:
@kwcckw: numpy.i is the file which we have taken from [numpy Github repo ](https://github.com/numpy/numpy/blob/master/tools/swig/numpy.i). This is used for handing N-D arrays since we are passing 2D arrays from python and using it on cpp code. So it's helping us to interface that.
Yes,  f_p.i, f_p.h, f_p.cpp are used for the form_P_ function. 

> I guess the complicated part is the integration, to convert the variable so that it is parse-able between c++ and python?

Yes.

kwcckw:
I tried and facing issue to install the SWIG. I'm using window and following the guide in your readme:

- Ensure [swig](http://www.swig.org/download.html) is installed on your system

Is the workflow supported in window? Since in window we do not have usr but C:/Users instead

img

MayukhSobo:
@khanh93vn  I am sorry that you are having issues using `shared_ptr` but let me assure you that shared_ptr is available from C++11. Only some features of `shared_ptr` like `element_type` or `weak_type` are C++17 feature. https://en.cppreference.com/w/cpp/memory/shared_ptr also confirms that it is a c++11 feature. Please check your compiler settings. 

coalg

You can see that I am able to compile the code with c++11 only.

Here is my g++ info
> g++ (Ubuntu 7.4.0-1ubuntu1~18.04.1) 7.4.0
Copyright (C) 2017 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE

MayukhSobo:
@kwcckw Please follow the following page to install `swig` on Windows machine. 

http://www.swig.org/Doc1.3/Windows.html

PS: We never build the code on windows. But there is a support for windows and swig.

khanh93vn:
Thank you, I'll try again with different settings

PR #48, opened 2021-05-18
Update comp_pixel_versions.py

goodengineer:
**If** you meant in code:

_Chaldean Numerology
The numerical value of quarant in Chaldean Numerology is: 2
Pythagorean Numerology
The numerical value of quarant in Pythagorean Numerology is: 2_

**Then** you may leave it as quarant, **while:**

Using the logic of by analogy reasoning, you have used before in the same file word of quadrant ⇒ quadrant again.

_Geometrically, I understand quadrant/quarant means here the box(rectangular) of adjacent pixels._


boris-kz:
Please don't bother with alternative versions folder.

PR #49, opened 2021-05-19
Update comp_slice_.py

goodengineer:


frame_2D_alg/comp_slice_.py
```diff
@@ -497,7 +497,7 @@ def comp_slice_full(_P, P):  # forms vertical derivatives of derP params, and co
     ddX = dX - _dX  # long axis curvature, if > ave: ortho eval per P, else per PP_dX?
     mdX = min(dX, _dX)  # dX is inversely predictive of mP?
 
-    # is this looks better? or it would better if we stick to the old code?
+    # is this looking better? or it would better if we stick to the old code?
```
kwcckw, line 500: Hi, those are my edits, right now it is just a draft and not finalized yet. So probably you can start to look at 2D algo ->frame_blobs.py, intra_blob.py , or 1D algo -> line_patterns.py, line_PPs_draft.py.

goodengineer: OK, Great.

What do you think about UML/class diagram of the software? Which way is best to study it? What can induce more clarity?
I will inspect/study as per your requests/order/needs of the Project if you would like to make great help by hiring/employing me on reciprocally suitable terms.  Which repo fork is better that I switch to:

https://kwcckw.github.io/CogAlg/

or

https://github.com/khanh93vn/CogAlg

or

https://github.com/boris-kz/CogAlg

? What is more convenient for you, that I fork selected repo or commit if I find what directly to existing master or which branch?

Kind Regards.

kwcckw: i think @boris-kz will be more suitable to provide more information here.

PR #85, opened 2023-05-04
Update README.md

standardgalactic:
A few minor grammar suggestions.


boris-kz:
Thanks, but my "mistakes" are mostly intentional: I hate clutter. Those 'the' and 'a' mean nothing in the context where either one can be used, depending on POV of a reader.

PR #88, opened 2026-03-20
Implement prediction error feedback for self-adjusting code

lasred:
Add mechanism to use prediction errors (dpTT) for parameter adjustment:

- ErrorTracker class: accumulates prediction errors across comparisons, tracks which parameters have high error (need more attention)

- fit_code() function: adjusts algorithm parameters based on errors:
  1. Parameter attention: high-error params get increased weight in wTT_
  2. Threshold adjustment: corrects systematic bias in ave/avd

- Integration: error_tracker.accumulate() called in comp_N_ where dpTT was previously computed but unused

### DIFF
--- a/frame_2D_alg/agg_recursion.py
+++ b/frame_2D_alg/agg_recursion.py
@@ -113,6 +113,52 @@ mW = dW = 9  # fb weights per dTT, adjust in agg+
 wY = wX = 64; wYX = np.hypot(wY,wX)  # focus dimensions
 decay = ave / (ave+avd)  # match decay / unit dist?
 
+# --- Prediction Error Tracking (for code fitting) ---
+class ErrorTracker:
+    """Track prediction errors to adjust algorithm parameters."""
+    def __init__(self):
+        self.reset()
+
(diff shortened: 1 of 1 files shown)


boris-kz:
This is not worked out yet, I just updated it as 

                # relative prediction error to fit the code, direction-agnostic?
                rpTT += np.abs(pTT-Link.dTT) / Link.dTT
            else:
                pL = CN(typ=-1, nt=[_N,N], dTT=pTT,m=m,d=d,c=min(N.c,_N.c),r=lr, angl=np.array([dy_dx,1],dtype=object),span=dist)
                L_+= [pL]; N.rim+=[pL]; _N.rim += [pL]; N_+=pL.nt; acc[0]+=pTT*pL.c; acc[1]+=pL.c; acc[2]+=pL.r*pL.c  # all +ve
                # ~= links in clustering
    for N in set(N_):
    rL.rpTT += rpTT  # any root -> frame: high error-> lower source code weight?
    TT,cm,rm, TTd,cd,rd = acc

still tentative

PR #89, opened 2026-03-21
Test pull request.

kwcckw:


kwcckw:
Looks like my repo is out sync and it requires the merging of prior edits into the repo as well. It's very messy now so i wouldn't think the merging is possible.

frame_2D_alg/agg_recursion.py
```diff
@@ -65,7 +65,8 @@ def __init__(n, **kwargs):
         # 2= G:  + rim,kern,mang,sub,exe
         # 3= C | Cn: +rN_| m_,d_,r_,o_
         n.dTT, n.TTn, n.TTc = [np.copy(kwargs.get(x,np.zeros((2,9)))) for x in ('dTT','TTn','TTc')]  # +wTTC per Ct.dTT, all m_,d_ [M,D,n, I,G,a, L,S,A]
-        n.wTT_ = kwargs.get('wTT_',[None,None,None,None])  # dTT: Nt+Lt+Bt, TTC, TTn, TTc
+        # should be follow the same array size?
+        n.wTT_ = kwargs.get('wTT_',[np.ones((2,9)) for _ in range(4)])  # dTT: Nt+Lt+Bt, TTC, TTn, TTc
```
kwcckw: So if we init wTT_ as 1s in CN, we should run cent_TT per G in sum2G? Else they will always be 1.

frame_2D_alg/agg_recursion.py
kwcckw: Actually i'm getting endless recursion with the current cent_TT:
        
        def cent_TT(dTT, _wTT, r):  # weight attr matches | diffs by their match to the sum, recompute to convergence
        
            coT = np.abs(dTT[0]) + np.abs(dTT[1]) + eps  # complemented vals
            wTT = []
            for fd, _wT, dT in zip((0,1), _wTT, dTT):
                vT = np.abs(dT)  # if -m: wrong or surprise value?
                while True:
                    rvT = vT/coT * _wT  # weighted normalized vals
                    mean = max(rvT.mean(), eps)  # scalar
                    invdev_ = np.minimum(rvT / mean, mean / np.maximum(rvT, eps))
                    wT = invdev_ / max(invdev_.mean(), eps)   # mean(wT)=1
                    if np.sum(np.abs(wT -_wT)) < ave * r:  # if np.linalg.norm(wT - _wT, 1) < r?
                        break
                    _wT = wT
                wTT += [wT]
            _wTT[:] = np.array(wTT)
            # single-mode dTT, extend to 2D-3D lev cycles in H, cross-level param max / centroid?

Here are a few values of `np.sum(np.abs(wT -_wT)) `  across severel while loop iterations:

        1.0474740918183132
        7.9999625482513
        1.5305618437322113
        1.8620470710177714
        1.5114817526785111
        1.6210186937742723
        1.5038748850984718
        1.5393867140525181
        1.501264645829316
        1.5127073691803141
        1.5004083584879533
        1.5040882328662195
        1.5001314031262603
        1.5013140706626091
        1.5000422293849582
        1.5004222475534006
        1.5000135589179864
        1.5001356595109232
        1.5000043452948357
        1.5000435756375259
        1.5000013848211597
        1.5000139892599138
        1.5000004336227717
        1.500004483333336
        1.5000001280080226
        1.500001429148406
        1.5000000298161922
        1.5000004478623457
        1.4999999982679106
        1.5000001325828376
        1.4999999881316979
        1.5000000312860196
        1.49999998487501
        1.499999998740154
        1.499999983828664
        1.4999999882834247
        1.4999999834924815
        1.4999999849237595
        1.4999999833844675
        1.499999983844326
        1.4999999833497626

So looks like when the difference value converges at value > ave * r， it will run indefinitely.

boris-kz:
You just need to rebase your repo

kwcckw:
> You just need to rebase your repo

Thanks, i will try that tomorrow before the commit again.

frame_2D_alg/agg_recursion.py
kwcckw: @boris-kz Not sure if it's clearer to have the discussion here? Or we can move back to the commit based discussion?

frame_2D_alg/agg_recursion.py
boris-kz, line 1: So this works fine for discussions, doesn't matter if I can't merge it, I don't want to anyway?

frame_2D_alg/agg_recursion.py
boris-kz: I hear that github is prioritizing discussions on PRs vs. commits, that's why we have visibility problems on the latter. Let's try to work this out

frame_2D_alg/agg_recursion.py
```diff
@@ -954,17 +956,18 @@ def expand_lev(_iy, _ix, elev, T):  # seed tile is pixels in 1st lev, or Fg in h
...
+            F.H += [Q2R([n for N in tile_ for n in N.N_], R=F, fN=1)]  # concat prior top lev
             if cross_comp(F.Nt, rr=elev)[0]:  # spec->tN_,tC_,tL_, proj comb N_'L_?
                 elev += 1
                 if rV > ave:
                     for dTT, wTT, ww in zip((F.dTT,F.Ct.dTT,F.TTn,F.TTc), F.wTT_, (wN,wC,1,wc)):
                         wTT[:] = cent_TT(dTT,2) * ww  # max-scope correlation weights
                 if elev == max_elev:  # fb / top lev
                     rV, wTT_ = ffeedback(F)  # update globals, rV is not used?
-                wTT_ *= F.wTT_  # F.wTT_ is effectively global
+                for wTT, wwTT in zip(wTT_, F.wTT_): wTT *= wwTT  # F.wTT_ is effectively global  (we need to multiple each item in list individually)
```
boris-kz: why individually, they have the same shape?

frame_2D_alg/agg_recursion.py
```diff
@@ -65,7 +65,8 @@ def __init__(n, **kwargs):
         # 2= G:  + rim,kern,mang,sub,exe
         # 3= C | Cn: +rN_| m_,d_,r_,o_
         n.dTT, n.TTn, n.TTc = [np.copy(kwargs.get(x,np.zeros((2,9)))) for x in ('dTT','TTn','TTc')]  # +wTTC per Ct.dTT, all m_,d_ [M,D,n, I,G,a, L,S,A]
-        n.wTT_ = kwargs.get('wTT_',[1,1,1,1])  # dTT: Nt+Lt+Bt, TTC, TTn, TTc, or ones?
+        # n.wTT_ = kwargs.get('wTT_',[1,1,1,1])  # dTT: Nt+Lt+Bt, TTC, TTn, TTc, or ones?
+        n.wTT_ = kwargs.get('wTT_',[np.ones((2,9)) for _ in range(4)])  # dTT: Nt+Lt+Bt, TTC, TTn, TTc, or ones?
```
boris-kz: we need that because wTT is nested, so the arrays must be the same shape?

frame_2D_alg/agg_recursion.py
kwcckw: Okay, i guess we can have the discussion here first. Right now i'm still trying some workarounds, but looks like it's still getting the same problem.

frame_2D_alg/agg_recursion.py
kwcckw: Since wTT_ is a list, so we can't multiply them unless we convert it into a numpy array.

frame_2D_alg/agg_recursion.py
kwcckw: Alright, i think it will clearer if we have a new pull request on each new commit? I will close this one before the next PR again.

frame_2D_alg/agg_recursion.py
boris-kz: so what's the fix?

frame_2D_alg/agg_recursion.py

frame_2D_alg/agg_recursion.py
boris-kz: ok, right

frame_2D_alg/agg_recursion.py
boris-kz: diff_value > ave * r is not a convergence.
the real problem is that this eval is per wT, it should be per wTT?

kwcckw: Probably, so cent_TT actually should be per wTT_? But i'm not sure if it's correct to update them all with the same mean?

PR #90, opened 2026-03-22
Changes (duplicated):

kwcckw:


kwcckw:
So i just managed to merge them all, this is a cleaner version now. Thanks for your suggestions earlier!

boris-kz:
> So i just managed to merge them all, this is a cleaner version now. Thanks for your suggestions earlier!

But you deleted all commits, with the comments on them

frame_2D_alg/agg_recursion.py
```diff
@@ -65,7 +65,8 @@ def __init__(n, **kwargs):
         # 2= G:  + rim,kern,mang,sub,exe
         # 3= C | Cn: +rN_| m_,d_,r_,o_
         n.dTT, n.TTn, n.TTc = [np.copy(kwargs.get(x,np.zeros((2,9)))) for x in ('dTT','TTn','TTc')]  # +wTTC per Ct.dTT, all m_,d_ [M,D,n, I,G,a, L,S,A]
-        n.wTT_ = kwargs.get('wTT_',[1,1,1,1])  # dTT: Nt+Lt+Bt, TTC, TTn, TTc, or ones?
+        # n.wTT_ = kwargs.get('wTT_',[1,1,1,1])  # dTT: Nt+Lt+Bt, TTC, TTn, TTc, or ones?
+        n.wTT_ = kwargs.get('wTT_',[np.ones((2,9)) for _ in range(4)])  # dTT: Nt+Lt+Bt, TTC, TTn, TTc, or ones?
```
kwcckw, line 69: > we need that because wTT is nested, so the arrays must be the same shape?

Actually it could be an array of [1,1,1,1], but the problem is we update G.wTT_ with 4 (2,9) arrays from cent_TT?

    for TT, wTT, ww in zip([C.dTT, C.Ct.dTT, C.TTn, C.TTc], C.wTT_, [wN, wC, 1, wc]):  # wTT_ should be C.wTT_?
        wTT[:] = cent_TT(TT, R) * ww

boris-kz: that's optional, and both versions will work?

kwcckw:
> > So i just managed to merge them all, this is a cleaner version now. Thanks for your suggestions earlier!
> 
> But you deleted all commits, with the comments on them

Yes, thanks for pointing that out. I just pushed them into another branch here:

https://github.com/kwcckw/CogAlg/tree/master2

frame_2D_alg/agg_recursion.py
kwcckw: That won't work for [1,1,1,1] because wTT will be 1, and we can't reassign them as ` cent_TT(TT, R) * ww` too.

boris-kz:
> > > So i just managed to merge them all, this is a cleaner version now. Thanks for your suggestions earlier!
> > 
> > 
> > But you deleted all commits, with the comments on them
> 
> Yes, thanks for pointing that out. I just pushed them into another branch here:
> 
> https://github.com/kwcckw/CogAlg/tree/master2

ok great.
so it should be:

def cent_TT(dTT, r):  # weight attr matches | diffs by their match to the sum, recompute to convergence

    wTT,_wTT = [],np.ones((2,9)); coT = np.abs(dTT[0]) + np.abs(dTT[1]) + eps  # complemented vals
    while True:
        for fd, _wT, dT in zip((0,1), _wTT, dTT):
            vT = np.abs(dT)  # if -m: wrong, or surprise value?
            rvT = vT / coT * _wT  # weighted normalized vals
            mean = max(rvT.mean(), eps)  # scalar
            invdev_ = np.minimum(rvT / mean, mean / np.maximum(rvT, eps))
            wT = invdev_ / max(invdev_.mean(), eps)   # mean(wT)=1
            wTT += [wT]
            break
        _wTT = wTT
    return np.array(wTT)  # single-mode dTT, extend to 2D-3D lev cycles in H, cross-level param max / centroid?
?

frame_2D_alg/agg_recursion.py
boris-kz: ok, 

    for i, TT, wTT, ww in zip([0,1,2,3], [C.dTT, C.Ct.dTT, C.TTn, C.TTc], C.wTT_, [wN, wC, 1, wc]):
        C.wTT_[i] = cent_TT(TT, R) * ww
?

frame_2D_alg/agg_recursion.py
```diff
@@ -220,7 +221,7 @@ def proj_V(_N,N, dist, dy_dx, dec):  # _N x N induction
         else: break  # beyond initial induction range, re-sort by proj_V?
     for N in set(N_):
         if N.rim: Q2R(N.rim, N.Rt, merge=0, froot=0)
-    rL.rpTT += rpTT  # any root -> frame: high error-> lower source code weight?
+    if rL is not None: setattr(rL, 'rpTT',rpTT+getattr(rL, 'rpTT',0))  # any root -> frame: high error-> lower source code weight?
```
boris-kz, line 224: this is tentative, it assumes that we always pass root, not just rL

frame_2D_alg/agg_recursion.py
```diff
@@ -542,11 +543,11 @@ def cluster_P(_C_,N_,root):  # Parallel centroid refining, _C_ from cluster_C, N
     return [c for c in out_ if c]  # or full out_?
 ''' next order: level-parallel cluster_H / multiple agg+? compress as autoencoder? '''
 
-def add_H(H,h, root):
+def add_H(H,h, root, fN=0):
     for Lev,lev in zip_longest(H, h):  # bottom-up
         if lev:
-            if Lev: Q2R([Lev,lev], Lev, merge=2,froot=0)
-            else: H.append(CopyF(lev, root))
+            if Lev: Q2R([Lev,lev], Lev, merge=2,froot=0, fN=fN)  # add_H might add CN from F.H
```
boris-kz, line 549: thanks

frame_2D_alg/agg_recursion.py
```diff
@@ -578,9 +579,9 @@ def sum2C(N_, _C, _i=None, root=None):  # fuzzy sum + base attrs for centroids
     R = 0; TT = np.zeros((2,9)); kern = np.zeros(4); span = 0; yx = np.zeros(2)
     for N, c in zip(N_,cc_):
         rc = c/ Cc; TT += N.dTT*rc; kern += N.kern*rc; span += N.span*rc; yx += N.yx*rc; R += N.r*rc
-    m,d = vt_(TT,R, TTw(root,1))  # wTTC
+    m,d = vt_(TT,R, TTw(root.root,1))  # wTTC (should be root.root here since root is Ft)
     C = CN(typ=3, Nt=CF(N_=N_), dTT=TT, m=m, d=d, c=Cc, r=R, yx=yx, kern=kern, span=span, root=root)
-    for TT, wTT, ww in zip([C.dTT, C.Ct.dTT, C.TTn, C.TTc], [C.dTT, C.Ct.dTT, C.TTn, C.TTc], [wN, wC, 1, wc]):
+    for TT, wTT, ww in zip([C.dTT, C.Ct.dTT, C.TTn, C.TTc], C.wTT_, [wN, wC, 1, wc]):  # wTT_ should be C.wTT_?
```
boris-kz, line 584: right

frame_2D_alg/agg_recursion.py
kwcckw: So could you clarify if cent_TT returns single value instead of array with size of (2,9)? Also we need to update wN, wC, wn and wc in cent_TT when we call cent_TT in frame_H after the feedback?

boris-kz: wTT, ((2,9))

wN, wC, wn, wc are separate, they are fixed now, we will need to adjust them in feedback too

kwcckw: In feedback, we are adjusting rTT_ by summing them:

        rN, rC, rn, rc = [np.abs(_tt/ (tt+eps)) for _tt,tt in zip((_N,_C,_n,_c), (N,C,n,c))]
        for rTT,rtt in zip(rTT_,(rN,rC,rn,rc)):
            rTT += rtt

So to adjust wN, wC, wn and wc, i think we can't just sum them? It should be scaled with rN, rC, rn and rc?

frame_2D_alg/agg_recursion.py
boris-kz, line 44: I've been trying to add that claude intro to https://boris-kz.github.io/CogAlg/, the damn thing can't even parse .md

frame_2D_alg/agg_recursion.py
boris-kz: those are just short names for corresponding TTs

frame_2D_alg/agg_recursion.py
kwcckw: Looks like it messed up the format and it should be based on index.md. I think LLM should be able to correctly format the text?

frame_2D_alg/agg_recursion.py
kwcckw: Yes, but we need to update `wN, wC, wn and wc` based on those dTTs?

frame_2D_alg/agg_recursion.py
boris-kz: thanks, but it only shows raw file?

kwcckw: I can see the intro is well formatted now? So it's done automatically via chatgpt codex?

boris-kz: ok, that must my settings. No, that was claude:  https://claude.ai/public/artifacts/af765400-f196-41d5-95ed-9e84f29cfd4d

frame_2D_alg/agg_recursion.py
boris-kz: That's higher-order weights, right now just relative code complexity, which is fixed

frame_2D_alg/agg_recursion.py
kwcckw: Okay, i can see there's a thumb-up from chatgpt-codex-connector bot, i thought it's from the codex.

frame_2D_alg/agg_recursion.py
kwcckw: But those higher-order weights should be updated from feedback, which is from F and their H's dTT?

boris-kz: no, code complexity doesn't change

kwcckw: I see, so when do we really adjust those higher order weights? If code complexity is fixed throughout the processing, then those weights won't be adjusted at all?

boris-kz: only if there is code-adjusting feedback, or underlying machine changes

kwcckw: I see, that code-adjusting feedback should be higher level than frame_H, so it should be constant here, thanks for clarifying.

PR #92, opened 2026-03-23
master

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -241,7 +241,7 @@ def comp_C_(C_, rr,_C_=[], fall=1, fC=0):  # simplified for centroids, trans-N_s
             for i, L in enumerate(L_):
                 if val_(L.dTT, rr+nw, wTTc,fi=0) < 0:
                     C0,C1 = L.nt
-                    if C0 is C1 or C1 in merg_: continue  # not merged
+                    if C0 is C1 or C1 in merg_ or C0 in merg_: continue  # not merged  (if C0 is already merged, skip it?)
```
kwcckw, line 244: I think C0 might be in merg_ too? So if C0 in merg_, we need to skip?

frame_2D_alg/agg_recursion.py
```diff
@@ -518,7 +519,7 @@ def cluster_P(_C_,N_,root):  # Parallel centroid refining, _C_ from cluster_C, N
         for N in N_:
             N.m_,N.d_,N.r_,N.rN_ = map(list,zip(*[vt_(base_comp(C,N)[0],C.r,wTTC) + (C.r,C) for C in _C_]))  # distance-weight match?
             N.o_ = list(np.argsort(np.argsort(N.m_)[::-1])+1)  # rank of each C_[i] = rdn of C in C_
-            dM+= sum(abs(_m-m) for _m,m in zip(N._m_,N.m_))
+            dM+= sum(abs(_m-m) for _m,m in zip(N._m_,N.m_))  # N._m_ is init empty and here it's always zero
```
kwcckw, line 522: So from the init above:

    for N in N_: N._m_,N._o_,N._d_,N._r_ = [],[],[],[]

`N._m_` is always empty and we won't get any dM here. I think we should pre-init `N._m_, _d_, _o_ and _r_`? Or the dM eval below should be true on the first loop?

    if M > Ave*O and dM > Ave*dO:

frame_2D_alg/agg_recursion.py
```diff
@@ -606,7 +607,7 @@ def sum2G(ft_, root=None, init=1, typ=None):
             V = lm - ave*(lr+1) * Nw
             if V > 0:
                 if (mdecay(L_)-decay)*V > ave*Cw:
-                    G_,r = cluster_C(G.Nt, N_,r+1)
+                    G_,r = cluster_C(G.Nt, N_,r+1)  # we parse N_ as E_ to cluster_C, so that E_ in exemplar is just N_ now?
```
kwcckw, line 610: Previously cluster_C is meant for exemplars:

    def cluster_C(Ft, E_, r):

So that `E_` now is actually `N_` only?

frame_2D_alg/agg_recursion.py
```diff
@@ -677,8 +679,9 @@ def cent_TT(dTT, r):  # weight attr matches | diffs by their match to the sum, r
             wTT += [wT]
         if np.sum(np.abs(wTT-_wTT)) < ave * r:  # if np.linalg.norm(wT - _wT, 1) < r?
             break
-        _wTT = wTT
-    return np.array(wTT)  # single-mode dTT, extend to 2D-3D lev cycles in H, cross-level param max / centroid?
+        _wTT = np.array(wTT); wTT = []  # we need to reset wTT?
```
kwcckw, line 682: We need to reset wTT and update `_wTT`? And with the update from wTT into `_wTT`, we need to return `_wTT`?

frame_2D_alg/agg_recursion.py
```diff
@@ -639,7 +640,8 @@ def add_Nt(G, Nt, merge=0):  # addition to Q2R
         add_H(G.Nt.H if merge else G.Nt.H[:-1], Nt.H, G)  # exclude top lev if not merge
     N_ = Nt.N_
     if merge: G.N_ += N_ # never 2
-    else: G.H[-1].N_ += N_  # in new level?
+    elif G.Nt.H: G.Nt.H[-1].N_ += N_  # in new level? (this H merging should be in Nt?)
```
kwcckw, line 643: The merging in H should be on G.Nt?

frame_2D_alg/agg_recursion.py
boris-kz: right

frame_2D_alg/agg_recursion.py
```diff
@@ -447,8 +448,8 @@ def cluster_C(Ft, E_, r):  # form centroids by clustering exemplar surround via
     C_,_C_ = [],[]  # form root.Ct, may call cross_comp-> cluster_N, incr rc
     for n in Ft.N_: n._C_,n.m_,n._m_,n.o_,n._o_,n.rN_ = [],[],[],[],[],[]
     for E in E_:
-        C = Copy_(E,Ft,init=2,typ=0)  # all rims in root, sequence along eigenvector?
-        for TT, wTT, ww in zip([C.dTT,C.Ct.dTT,C.TTn,C.TTc], [C.dTT,C.Ct.dTT,C.TTn,C.TTc], [wN,wC,1,wc]):
+        C = Copy_(E,Ft,init=2,typ=3)  # all rims in root, sequence along eigenvector?  (typ should be 3 here)
```
boris-kz, line 451: Chee, you don't need to leave such comments, I can see the changes
because when you do, I have to manually remove them  
If the comment is meaningful, add it in PR only

kwcckw: Alright, so next time i will include those additional comments only if it requires additional clarification.

boris-kz: add it in PR only

frame_2D_alg/agg_recursion.py
```diff
@@ -416,6 +416,7 @@ def trans_cluster(G):
             while __L_:
                 _L_ = []
                 for L in __L_:  # flood-fill via frontier links
+                    if L in in_: continue  # L could be clustered here? in_ is shared across _N_, and `__L_= N.rim`, so L might be already clustered here
```
boris-kz, line 419: right

frame_2D_alg/agg_recursion.py
boris-kz: yes, we need to add loop count:

            dM+= sum(abs(m-_m) for _m,m in zip(N._m_,N.m_)) if nloop else sum(N.m_)
            dO+= sum(abs(o-_o) for _o,o in zip(N._o_,N.o_)) if nloop else sum(N.o_)

frame_2D_alg/agg_recursion.py
boris-kz: `for L in set(__L_) - in_:`

frame_2D_alg/agg_recursion.py
boris-kz: yes, cluster_N is suppoused to get E_ too, but that's from cross_comp
this is sub+, which is already selective for strong clusters, so I don't know if we need additional selection here?

frame_2D_alg/agg_recursion.py
boris-kz: this wTT won't be used anyway, except for return, which should be updated wTT

frame_2D_alg/agg_recursion.py
```diff
@@ -755,7 +758,7 @@ def proj_focus(PV__, y,x, tile):  # radial accum of projected focus value in PV_
 
     m,d,n = tile.m, tile.d, tile.c  # add r?
     V = (m-ave*n) + (d-avd*n)
-    dy,dx = tile.angl[0]; a = dy / max(dx,eps)  # average link_ orientation, projection
+    dy,dx = tile.angl[0]; a = dy / (eps if dx == 0 else dx)  # average link_ orientation, projection (shouldn't use max here since that removes sign?)
```
boris-kz, line 761: yes, dx or eps?

frame_2D_alg/agg_recursion.py
boris-kz: right

frame_2D_alg/agg_recursion.py
kwcckw: We need to reassign `_wTT = wTT` for the next while loop if it's not break? And we need to reset wTT = [] because it may pack new wT in the next loop ` wTT += [wT]`?

frame_2D_alg/agg_recursion.py
kwcckw: The selected strong clusters are those Gs? For sub+, if we look at the G.N_, i think we may still need to select the per G's strong clusters?

frame_2D_alg/agg_recursion.py
kwcckw: > for L in set(__L_) - in_:

This looks better, thanks.

frame_2D_alg/agg_recursion.py
kwcckw: We can't use `or` here since that selects eps if dx is negative too.

frame_2D_alg/agg_recursion.py
boris-kz: ok, right

frame_2D_alg/agg_recursion.py
boris-kz: or only replaces 0, why -ve?

frame_2D_alg/agg_recursion.py
```diff
@@ -639,7 +640,8 @@ def add_Nt(G, Nt, merge=0):  # addition to Q2R
         add_H(G.Nt.H if merge else G.Nt.H[:-1], Nt.H, G)  # exclude top lev if not merge
     N_ = Nt.N_
     if merge: G.N_ += N_ # never 2
-    else: G.H[-1].N_ += N_  # in new level?
+    elif G.Nt.H: G.Nt.H[-1].N_ += N_  # in new level? (this H merging should be in Nt?)
+    else: G.Nt.H = [sum2F(N_, G.Nt.nF, G)]  # init 1st level
```
boris-kz, line 644: else: G.Nt.H += [Q2R((N_+ G.Nt.H.pop() if G.Nt.H else []), G.Nt)]  # extend or init top level

frame_2D_alg/agg_recursion.py
boris-kz: ok,

                r+=1; E_ = get_exemplars({N for L in L_ for N in L.nt},r)
                G_,r = cluster_C(G.Nt,E_,r) if (mdecay(L_)-decay)*V >ave*Cw else cluster_N(G.Nt, E_,r)

frame_2D_alg/agg_recursion.py
boris-kz, line 1: Claude: Interesting idea. Let me think through this concretely.

You currently have 4 weight levels on dTT operations: `wTTN` (node clustering), `wTTC` (centroid clustering), `wTTn` (node comp), `wTTc` (centroid comp). These weight how much each of the 9 dTT attributes matters at each stage.

The extension you're proposing: weights on finer-grained code elements (functions, blocks, operations), forming a hierarchy that mirrors the call tree. The AST gives you that tree structure for free.

A few questions:

1. **What's the weight target?** Currently wTT_ weights dTT attributes per operation type. For function-level weights, are you weighting the function's *contribution to projected match* (like your existing `nw, cw, Nw, Cw, specw` but systematically derived)? Or weighting the function's *output dTT* the way wTT_ already does, just at finer granularity?

2. **Granularity problem.** AST gives you every expression, but most are mechanically necessary (index ops, assignments). The meaningful weight targets seem to be at the level of: comp operations (base_comp, comp_derT, comp_F, comp_H), clustering operations (cluster_N, cluster_C, cluster_P), and evaluation thresholds (the `> ave*nw` checks). Is that roughly the right grain, or do you want finer?

3. **Feedback mechanism.** Your current `ffeedback` computes ratios between consecutive-level TTs. For a function weight hierarchy, the signal would be: does calling this function improve projected match relative to its cost? That's essentially `nw, cw, Nw, Cw` but derived from actual prediction error rather than hand-set. The AST just automates *discovering* which functions exist and their nesting — the hard part is still computing the weight update. Do you have a specific update rule in mind, or is `cent_TT` the template?

4. **Static vs dynamic.** AST structure is static (code doesn't change yet). But the *effective* call tree is dynamic — which branches execute depends on val checks. Are you weighting the static tree or the runtime trace?

The code self-modification angle makes this especially interesting: if weights on code elements get low enough, that's a signal to skip or replace that code path. That's a softer version of the eventual code self-modification.


>Right, rpTT is the ratio between projected and actual dTT in `comp_N_`:

```python
rpTT += np.abs(pTT-Link.dTT) / Link.dTT
```

So the logic is: a function's weight = how much it reduces prediction error relative to cost. You can only compute that if you have both the projection (pTT) and the actual result (Link.dTT) to compare.

That means:

1. **Only functions downstream of `proj_V` / `comp_N` pairs** can get empirical weights — you need the before/after to measure gain. Functions that don't have a projection stage (like initial `base_comp`) have no rpTT to derive weights from.

2. **Bootstrap problem.** On first pass there's no rpTT yet, so initial weights must be defaults (your current hand-set `nw, cw, Nw, Cw`). Weights refine only after rpTT accumulates across multiple agg+ cycles. That's consistent with your existing feedback being coarse at first.

3. **Currently rpTT is accumulated but not used** — you have the commented-out `# root.rpTT += rpTT`. So this is the missing link: rpTT feeds back into function weights.

Questions:

- rpTT is per-link (per comp_N call). To get a weight for, say, `comp_H` specifically, you'd need rpTT *with and without* comp_H's contribution — an ablation. Is that what you're thinking, or just attributing the total rpTT proportionally to each sub-function's dTT contribution?

- The proportional attribution seems more tractable: each sub-operation (base_comp, comp_derT, comp_H) produces a partial dTT. Its weight = how much its partial dTT reduced rpTT relative to its cost. That's essentially `cent_TT` applied to prediction error components rather than match components.

boris-kz: https://chatgpt.com/share/69c1c9bd-d8f0-8011-acdc-54e07abc3deb

frame_2D_alg/agg_recursion.py
kwcckw: Actually i don't get why we need to merge N_ and G.Nt.H? They are in different level?

frame_2D_alg/agg_recursion.py
kwcckw: Yes, we should only replace 0 to prevent zero division. For negative case, for example, -1 or eps, it will selects eps, which is wrong?

frame_2D_alg/agg_recursion.py
kwcckw: So this is about from the global weight to the finer weights on each operation? Right now we do have per G'sweight, and we need additional per function weights to differentiate their contribution during the feedback?

frame_2D_alg/agg_recursion.py
boris-kz: It should be N_+ G.Nt.H.pop().N_

frame_2D_alg/agg_recursion.py
boris-kz: it won't select eps for -ves

kwcckw: For example, when we use `1e-7 or -1`, we are getting `1e-7`? So the negative sign is removed.

frame_2D_alg/agg_recursion.py
kwcckw: So the merge is 0 here, and if G.Nt.H is not empty, we should just merge last level's N_?

    elif G.Nt.H: G.Nt.H[-1].N_ += N_

Then if G.Nt.H is empty, we should create new level, so that popping from G.Nt.H is not relevant since G.Nt.H is empty anyway?

frame_2D_alg/agg_recursion.py
boris-kz: -1 or 1e-7 = -1

frame_2D_alg/agg_recursion.py
boris-kz: It won't pop if H is empty
it's replacing the top level anyway, because we have to recompute the sum

frame_2D_alg/agg_recursion.py
boris-kz: It's about all process weights, they should probably map to part of AST, vs. being hand-coded flat list

frame_2D_alg/agg_recursion.py
boris-kz: You sound out of shape Chee, get some sleep?

frame_2D_alg/agg_recursion.py
kwcckw: >  they should probably map to part of AST

Does this mean it should be part of the process, which mean they should be in each of those instances and functions? Or else could you clarify how they can be mapped as part of AST?

frame_2D_alg/agg_recursion.py
kwcckw: Okay thanks. So i was testing with the python ide earlier and looks like the problem is the sequence:

`(1e-7 or -1)` returns 1e-7

`(-1 or 1e-7)` returns -1

So in this case, we should use `(dx or eps)` but not `(eps or dx)`.

frame_2D_alg/agg_recursion.py
kwcckw: Okay, right, thanks.

frame_2D_alg/agg_recursion.py
boris-kz: That's what I have on top:

>yes, dx or eps?

you don't need to test it, just think

frame_2D_alg/agg_recursion.py
boris-kz: with automated process, we would compute AST and trace it down, at each node: 
- somehow compute the value of adding wTT on it,
- if +ve: assign that wTT attr and add the process of computing it on upstream functions, continue
- else: terminate that branch

boris-kz: claude: The AST traversal mirrors your data hierarchy: selective depth-first, with val check at each node.

But the key difficulty is "somehow compute the value of adding wTT on it." Let me think about what that means concretely:

At each AST node (function/block), the value of a dedicated wTT is: **how much variance in rpTT this node's output explains that isn't already captured by its parent's wTT.** If the parent wTT already weights this node's contribution well enough, a finer-grained wTT adds no predictive value — terminate.

So the test at each AST node would be something like:

1. Compute rpTT with parent's wTT applied uniformly to this node
2. Compute rpTT with a local wTT (via `cent_TT` on this node's partial dTT)
3. Gain = reduction in rpTT. Cost = maintaining and computing the extra wTT
4. If gain > cost: assign wTT, recurse into children
5. Else: terminate, this node inherits parent's wTT

That's analogous to your data clustering: finer resolution only where it's justified by the match.

Two issues:

- **You need to isolate each node's partial dTT contribution.** For sequential operations (base_comp then comp_derT in `comp_N`) that's straightforward — each produces a separable dTT component. For nested ones (comp_H calling comp_derT) it's trickier because the child's dTT is part of the parent's.

- **When does this run?** It can't run every cycle — the AST is stable, only rpTT changes. So you'd recompute the weight tree periodically, like your other feedback. First cycle: all default weights. After enough rpTT accumulates: traverse AST, assign wTTs where justified, prune where not.

Is the partial dTT isolation the main obstacle, or do you see a simpler way to decompose contributions?

kwcckw: I see, but I'm not sure the method above is correct in order to get the added wTT ? Why we need rpTT when we can use the difference between their wTT?

Suppose the process can be run in batches since it's just the node_ and a same root?

boris-kz: it's actually rdpTT:  += np.abs(pTT-Link.dTT) / Link.dTT

that's novelty value, it increases comp_N_ weight and decrease weights of upstream processes that formed pTT, symmetrically. Because it's the error of upstream processes
But first, we need the process of forming meaningful AST, which may also improve the code without weights.
I don't think there is anything off-the-shelf?

boris-kz: Gemini: new global tracking variables:

code_H = {}  # nested dict representing the execution/AST tree
exec_stack = []  # runtime tracking

def trace(func):  # decorator to build and track AST-like code weights

    def wrapper(*args, **kwargs):
        global exec_stack
        
        # 1. Identify current path
        name = func.__name__
        parent = exec_stack[-1] if exec_stack else code_H
        if name not in parent:
            parent[name] = {'w': 1.0, 'sub': {}, 'M': 0, 'D': 0}  # init code node
        
        # 2. Push to stack
        exec_stack.append(parent[name]['sub'])
        
        # 3. Execute function
        res = func(*args, **kwargs)
        
        # 4. Pop stack
        exec_stack.pop()
        
        # Optional: bind local return vals (like m, d) to parent[name]['M'] here if standardized
        return res
    return wrapper

boris-kz: Claude:

Let me look at the actual code structure to think about this concretely.Python's `ast` module gives you the call tree directly. For your code, the relevant AST nodes are function definitions and call expressions within them. Here's what it would look like concretely:

```python
import ast, inspect

def get_call_tree(module_path):
    with open(module_path) as f:
        tree = ast.parse(f.read())
    
    func_bodies = {}  # func_name -> [called_func_names]
    for node in ast.walk(tree):
        if isinstance(node, ast.FunctionDef):
            calls = [n.func.id for n in ast.walk(node) 
                     if isinstance(n, ast.Call) and isinstance(n.func, ast.Name)]
            func_bodies[node.name] = calls
    return func_bodies
```

For `agg_recursion.py` this would give you something like:

```
cross_comp -> [comp_N_, comp_C_, vt_, get_exemplars, cluster_N, sum2F, val_, cross_comp]
comp_N_ -> [proj_V, vt_, comp_N, Q2R, sum_vt]
comp_N -> [base_comp, comp_derT, vt_, comp_H, comp_F, Q2R]
base_comp -> [comp, comp_A, comp_derT]
cluster_N -> [val_, nt_vt, sum2G, sum2F, Q2R, add_Nt]
cluster_C -> [base_comp, vt_, val_, sum2C, cent_TT, cluster_P, cross_comp]
...
```

But the raw AST has problems for your purpose:

1. **Flat calls vs. conditional calls.** The AST doesn't distinguish `comp_H` called inside `if m > ave*nw` from `base_comp` called unconditionally. You need to know which calls are behind val checks — those are the ones where wTT matters, because they're the conditional operations with measurable gain.

2. **The meaningful unit isn't always a function.** Inside `comp_N`, the sequence `base_comp` → conditional `comp_H` → conditional `comp_F` per fork — that conditional structure is the real tree, not the function boundary. An `if` block in the AST is `ast.If`, so you could detect val-gated blocks specifically.

3. **Recursive/repeated calls.** `cross_comp` calls itself, `comp_F` calls `comp_N_` or itself. The AST gives you the static structure, but the weight tree needs to handle depth via your existing `r` parameter.

So practically, the process would be:

- Parse AST, extract function defs and their calls
- For each call site, check if it's inside an `ast.If` whose test involves `ave`, `val_`, `vt_` — that marks it as conditional/weighted
- Build the tree of conditional operations only
- Each node in that tree gets a wTT slot, initialized to parent's wTT
- rpTT feedback refines nodes top-down, pruning where gain < cost

The main question: do you want this as runtime infrastructure (the code introspects itself and builds the weight tree), or as a one-time analysis that generates a static weight structure you then hardcode? The former is closer to code self-modification, the latter is more practical now.

kwcckw: > it's actually rdpTT: += np.abs(pTT-Link.dTT) / Link.dTT
> 
> that's novelty value, it increases comp_N_ weight and decrease weights of upstream processes that formed pTT, symmetrically. Because it's the error of upstream processes But first, we need the process of forming meaningful AST, which may also improve the code without weights. I don't think there is anything off-the-shelf?

This is actually better if we can skip the global weights, so we should process and apply those weights locally now?

To form the AST, i think we can explore the ast module, as suggested from LLM above? Then we can consider if we really need to implement it from scratch.

PR #93, opened 2026-03-24
3/24/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -327,7 +328,7 @@ def base_comp(_N,N):  # comp Et, kern, extT, dTT
     _pars = np.array([_M*rn,_D*rn,_c,_I*rn,_G*rn, [_Dy,_Dx],_L*rn,_N.span], dtype=object)  # Et, kern, extT
     pars  = np.array([M,D,c, (I,wI),G, [Dy,Dx], L,(N.span,wS)], dtype=object)
     mA,dA = comp_A(_N.angl[0]*_N.angl[1], N.angl[0]*N.angl[1])
-    m_,d_ = comp(_pars,pars, mA,dA)  # M,D,n, I,G,a, L,S,A
+    m_,d_ = comp(_pars,pars, mA or eps,dA or eps)  # M,D,n, I,G,a, L,S,A
```
kwcckw, line 331: So we need to use eps here since we may get 0 value from d and that is causing zero division in computing rdpTT (0s in L.dTT).

frame_2D_alg/agg_recursion.py
```diff
@@ -641,7 +642,7 @@ def add_Nt(G, Nt, merge=0):  # addition to Q2R
         add_H(G.Nt.H if merge else G.Nt.H[:-1], Nt.H, G)  # else top lev = Nt.N_
     N_ = Nt.N_
     if merge: G.N_ += N_ # never 2
-    else: G.Nt.H += [Q2R((N_+ G.Nt.H.pop().N_ if G.Nt.H else []), G.Nt)]  # extend or init top level
+    else: G.Nt.H += [Q2R((N_+ G.Nt.H.pop().N_ if G.Nt.H else []), G.Nt, merge=0)]  # extend or init top level
```
kwcckw, line 645: So we need merge = 0 here since we are packing the lev from N_ into H instead of merging it into Nt.N_.

frame_2D_alg/agg_recursion.py
```diff
@@ -701,17 +702,16 @@ def Copy_(N, root=None, init=0, typ=None):
...
     if typ:
         for attr in ['fin','span','mang','sub','exe']: setattr(C,attr, getattr(N,attr))
-        for attr in ['nt','kern','box','compared']: setattr(C,attr, copy(getattr(N,attr)))
+        for attr in ['nt','kern','box','compared','dTT','TTn','TTc','m','d','c','r']: setattr(C,attr, copy(getattr(N,attr)))
         for attr in ['Nt','Lt','Bt','Ct','Xt','Rt']: setattr(C, attr, CopyF(getattr(N,attr), root=C))
         if init:  # new G
             C.yx = [N.yx]; C.angl = np.array([copy(N.angl[0]), N.angl[1]],dtype=object)  # get mean
-            if init==1: C.L_=[l for l in N.rim if l.m>ave]; N.root=C; C.fin = 0  # else centroid
+            C.L_=[l for l in N.rim if l.m>ave]; N.root=C; C.fin = 0  # else centroid
+            C.N_ = [N]; C.m_=[]; C._m_=[]; C.o_=[]; C._o_=[]
```
kwcckw, line 710: So right now we only use init > 1 in cluster_C_,  so i guess we can remove the init = 2?

Also we should copy dTT, TTn, TTc, m, d c and r as well?

frame_2D_alg/agg_recursion.py
```diff
@@ -958,7 +958,7 @@ def expand_lev(_iy, _ix, elev, T):  # seed tile is pixels in 1st lev, or Fg in h
             F = CN(box=np.array([0,0,Y,X]), yx=np.array([Y//2, X//2]))  # same center on all levels
             F.H = []; F.N_ = tile_  # or Nt = Q2R(tile_)?
             if elev: [add_H(F.H,T.H, F,fN=1) for T in tile_]
-            F.H += [Q2R([n for N in tile_ for n in N.N_], R=F, fN=1)]  # concat prior top lev
+            F.H += [Q2R([n for N in tile_ for n in N.N_], merge=0, R=F, fN=1)]  # concat prior top lev
```
kwcckw, line 961: So this is the same as above where we shouldn't merge tile's N_ into F.N_ since they should be packed into H instead.

frame_2D_alg/agg_recursion.py
```diff
@@ -259,6 +259,7 @@ def comp_C_(C_, rr,_C_=[], fall=1, fC=0):  # simplified for centroids, trans-N_s
             comp_N(_C,C, rr,A=dy_dx,span=dist,L_=L_,N_=N_,acc=acc)  # simplified for typ=3
     for N in list(set(N_)):
         if N.rim: Q2R(N.rim, N.Rt, merge=0, froot=0)
+    # if np.any(rdpTT): accum_rdpTT('comp_N_', rdpTT)  # feed prediction error into code weights
```
kwcckw, line 262: > But first, we need the process of forming meaningful AST, which may also improve the code without weights.
I don't think there is anything off-the-shelf?

So from claude, i think we may use a dict to implement the simpler version of AST first?

1. Init the weight and rdpTT

          cw_H = {
              'comp_N':    10,  # was nw: node comparison threshold
              'comp_F':     5,  # was cw: fork/sub comparison
              'cluster_N': 15,  # was Nw: connectivity clustering
              'cluster_C': 20,  # was Cw: centroid clustering
              'proj_N':    20,  # was specw: projection specificity
          }
      
          rdpTT_H = {} 

2. Then accumulate the `rdpTT` into `rdpTT_H` after each operation, as in my line 262 above.

3. Update the cw in ffeedback based on global value of `rdpTT_H`.

Also the eval should be based on cw_H, for example, in cross_comp, eval for cluster:


        if L_:
            M, D = vt_(TT, r, TTw(Ft.root,fC+2))
            if M * ((len(L_)-1)*Lw) > ave * cw_('cluster_N'):  # was Nw
                setattr( Ft.root,('TTn','TTc')[fC], TT+TTd)
                E_ = get_exemplars({N for L in L_ for N in L.nt}, r)
                G_,r = cluster_N(Ft, E_, r)
    
I think we still need the weights, but it's in a global dict cw_H, which is clearer and better for tracking and update purpose.

frame_2D_alg/agg_recursion.py
```diff
@@ -354,11 +355,11 @@ def comp(_pars, pars, meA,deA):  # compute m_,d_ from inputs or derivatives
             p, avd = p
             d = _p - p
             m_ += [avd - abs(d)]  # + complement max(avd,ad)?
-            d_ += [d]
+            d_ += [d or eps]
```
boris-kz, line 358: that or eps should be added in division only, here it just accumulates garbage

frame_2D_alg/agg_recursion.py
boris-kz: don't add it here. you really need to improve your mental hygiene
unless it's some stupid LLM doing this for you

frame_2D_alg/agg_recursion.py
boris-kz: this is forming a single level, merge is fine

frame_2D_alg/agg_recursion.py
boris-kz: it's merging in level

frame_2D_alg/agg_recursion.py
```diff
@@ -608,7 +609,7 @@ def sum2G(ft_, root=None, init=1, typ=None):
         if lm*ld* ((len(L_)-1)*Lw) > ave*avd* (lr+1)*cw:
             V = lm - ave*(lr+1) * Nw
             if V > 0:
-                r+=1; E_= get_exemplars({N for L in L_ for N in L.nt},r)
+                r+=1; E_= get_exemplars({N for L in G.Lt.N_ for N in L.nt},r)
```
boris-kz, line 612: we get L_ in 596

frame_2D_alg/agg_recursion.py
boris-kz: I frankly don't care about these Copy functions, LLM can update them with whatever attrs we have in N, this is a brainless job

frame_2D_alg/agg_recursion.py
kwcckw: When merge is true, we are packing tile's N_'s N_ into F.Nt.N_, which is wrong since  F.Nt.N_ should contain tile_ only? That `tile's N_'s N_ ` should be in H[-1].N_ only?

frame_2D_alg/agg_recursion.py
kwcckw: Yes, this is just a cleanup since we don't really need init == 1 or init ==2 now.

frame_2D_alg/agg_recursion.py
kwcckw: Thanks for pointing this out. I think we need to standardize the usage of eps since right now it's used in init or some accumulations, which is causing the same problem. 

In cluster_C:

    C_,cnt,olp, mat,dif, DTT,Dm,Do = [],0,0,0,0,np.zeros((2,9)),0,eps

In cent_TT"

    coT = np.abs(dTT[0]) + np.abs(dTT[1]) + eps

In proj_N:

    cos_d = (N.angl[0].dot(A) / (np.hypot(*N.angl[0]) * dist + eps)) * N.angl[1]

In PP2N:

    dTT = np.array([ np.array([mM, mD, mL, mI, mG, mA, mL, mL / 2, eps]),  # extA=eps
                               np.array([dM, dD, dL, dI, dG, dA, dL, dL / 2, eps])])


So we need to remove those eps above and include eps only if it's in the division.

frame_2D_alg/agg_recursion.py
boris-kz: this whole expression is H[-1]

frame_2D_alg/agg_recursion.py
boris-kz: LLMs can do this clean-up better than you, focus on something substantial

frame_2D_alg/agg_recursion.py
kwcckw: > this whole expression is H[-1]

So the actual problem is we parse R as F, while it should be R=None since we should init new level instead of using F as lev.

frame_2D_alg/agg_recursion.py
boris-kz: you want dict just to show the function names?
that's what classes are for, we can just use CN F intead of globals?

frame_2D_alg/agg_recursion.py
boris-kz: R is root. F is a direct root for all levels, because H is just a list
F is different from G because it has H, which is a pipeline, so the levels are temporally specific.
CF.H levels are already summed in CF.N_, they are derivation-specific .

frame_2D_alg/agg_recursion.py
kwcckw: But we need to access them globally? If we use an instance, we may need to parse them around in order to retrieve the weights.

Also the name should be more straight forward to map each rdpTT and accumulate or update the weights?

For example, in the line above:

     if np.any(rdpTT): accum_rdpTT('comp_N_', rdpTT)  

We actually can use `inspect.stack()[0].function` to get `comp_N_` since it's called from `comp_N_`, then we can skip the process to manually insert each name:

     if np.any(rdpTT): accum_rdpTT(inspect.stack()[0].function, rdpTT)

boris-kz: We need to pass something anyway, and F has all the attrs, it's easier than names.
Same CN, just a bit extended, so we don't need a whole new structure

boris-kz: Or we can make frame a new class, CP (for process), which will add process attrs: the call tree with corresponding process weights?

frame_2D_alg/agg_recursion.py
kwcckw: Okay, so `Q2R` can't be used directly here since `Q2R` returns R as level, which we will pack F into F.H again.

So i guess we need to run it twice to accumulate the lev into F:

    lev = Q2R([n for N in tile_ for n in N.N_], R=None, fN=1); F.H_ += [lev]; Q2R([F, lev], merge=0, R=F)

frame_2D_alg/agg_recursion.py
boris-kz: claude:

Yes, Frame already accumulates feedback (`ffeedback`, `wTT_`), so making it a dedicated class that also holds the call tree is a natural extension.

CP would have:
- the current frame attrs (dTT, wTT_, H, N_, etc.)
- call tree parsed from AST at init
- process weights indexed by call tree nodes
- rpTT accumulation per node for feedback

But a few things to clarify:

1. **CP per module or per system?** The call tree spans `agg_recursion`, `comp_slice`, `slice_edge`, `frame_blobs`. One CP for the whole pipeline, or one per module with cross-references? The weights need to span the full tree since rpTT is computed in `comp_N_` but the operations being weighted include `comp_slice` and deeper.

2. **When does CP parse the AST?** If at init (once), the tree is static and weights update across cycles. That's your current feedback model. If the eventual code self-modification changes the code, CP would need to re-parse — but that's far off.

3. **Naming: CP is already used** in `slice_edge` for the P class. `CProc` or `CFr` (frame process) to avoid collision?

4. **Relationship to CN.** Currently frame is a CN in `frame_H`. CP would either replace CN for frames specifically, or inherit from CN and add process attrs. The latter seems right since frames still participate in the data hierarchy (they have dTT, N_, H).

So roughly:

```python
class CProc(CN):  # or CFr
    def __init__(P, module_paths):
        super().__init__()
        P.call_tree = parse_AST(module_paths)  # conditional call nodes
        P.pwTT_ = init_weights(P.call_tree)    # process wTT per node
        P.rpTT_ = {}  # accumulated prediction error per node
```

The key design question: how does the call tree index connect to runtime? Each conditional operation needs to know its node ID to accumulate rpTT and look up its weight. Is that an explicit parameter passed through, or do you derive it from the call stack?

frame_2D_alg/agg_recursion.py
boris-kz: ok, R=None is a default, how about this:

Q2R([F,Q2R([n for N in tile_ for n in N.N_], fN=1)], merge=0, R=F)

boris-kz: claude:

Yes, so line 962 should concat the lower-level nodes into the top H level, similar to how `add_Nt` does it:

```python
F.H += [Q2R([n for N in tile_ for n in N.N_], fN=1)]
```

This creates a new CN level from the flattened sub-nodes and appends it to F.H. Then the outer Q2R on line 962 accumulates that into F as before.

So replace lines 962-963 with:

```python
lev = Q2R([n for N in tile_ for n in N.N_], fN=1)
F.H += [lev]
Q2R([F, lev], merge=0, R=F)
```

Or kept compact:

```python
F.H += [lev := Q2R([n for N in tile_ for n in N.N_], fN=1)]
Q2R([F, lev], merge=0, R=F)
```

This avoids the double-overwrite and F-in-F.H problems. The `froot=1` default will set `lev.root = F`, which seems correct here.

kwcckw: Okay right, default R is None, and your version looks correct to me now, thanks.

kwcckw: > Or we can make frame a new class, CP (for process), which will add process attrs: the call tree with corresponding process weights?

Okay, this sounds better. That pwTT would be the adjusted weight and we accumulate each process's rdpTT into P.rpTT_. But what would be the call_tree? Do we really need it?

> ok, R=None is a default, how about this:
> 
> Q2R([F,Q2R([n for N in tile_ for n in N.N_], fN=1)], merge=0, R=F)

But this doesn't create the lev, so we still need to init the inner Q2R as lev and pack it into F.H instead.

frame_2D_alg/agg_recursion.py
kwcckw: > Or we can make frame a new class, CP (for process), which will add process attrs: the call tree with corresponding process weights?

Okay, this sounds better. That pwTT would be the adjusted weight and we accumulate each process's rdpTT into P.rpTT_. But what would be the call_tree? Do we really still need it?

boris-kz: call tree will represent functions, CN doesn't have them. we need to map process weights to the functions. But it's a call network, not really a call tree

frame_2D_alg/agg_recursion.py
boris-kz: right            

F.H += [lev := Q2R([n for N in tile_ for n in N.N_],fN=1)]; Q2R([F,lev], merge=0,R=F)  # concat prior top lev

frame_2D_alg/agg_recursion.py
boris-kz: we can pack that wTT_ call tree in global nested CF Omni:
dTT=wTT, N_: sub-fork call tree level, [CF] per level? 

top level O = CF(N_= [CF(nF='comp'), CF(nF='clust'), CF(nF='fback'))  

claude:
Then `comp` fork unfolds:

```
CF(nF='comp', N_= [CF(nF='comp_N_'), CF(nF='comp_C_'), CF(nF='comp_N')])
```

And `comp_N` unfolds further:

```
CF(nF='comp_N', N_= [CF(nF='base_comp'), CF(nF='comp_H'), CF(nF='comp_F')])
```

Each leaf CF's dTT = that operation's wTT. Parent dTT = sum of children's, same as data hierarchy.

But a few issues:

1. **CF.H vs CF.N_ for depth.** Your call tree has two kinds of nesting: sequential operations within a function (siblings in N_), and sub-calls within a conditional block (levels in H). Which maps to which? N_ = sibling forks within same function, H = depth of sub-calls?

2. **Where does rpTT land?** When `comp_N` computes rpTT, it needs to decompose it across its children CFs (`base_comp`, `comp_H`, `comp_F`). That decomposition is by each child's partial dTT contribution — same as `cent_TT` logic.

3. **Omni is static structure, dynamic weights.** The CF tree is built once from AST, but dTT (=wTT) updates each feedback cycle. That's consistent with how data CFs work.

The main thing I'm not sure about: `comp` and `clus` are not really parallel forks — clustering operates on comp output, sequentially. Does the top-level N_ imply they're alternatives, or just that they're the two major branches of the process? If sequential, maybe H is more appropriate than N_ at the top level?

boris-kz: the process forks clearly cluster by function and the type of processing:

Comparison decomposes (subtracts to get m,d), clustering composes (adds to get sums). They're complementary operations, alternating in each agg cycle.
And feedback is also distinct: it's multiplicative (scaling weights by ratios). So the three top forks are three fundamental arithmetic types:

comp: subtractive — _p - p, min(_a,a), decomposition into m,d
clust: additive — TT += n.dTT*rc, N_ += [N], composition into sums
fback: multiplicative — wTT *= rTT, ave *= rV, scaling by ratios

These aren't just call-graph clusters, they're operationally typed by their core arithmetic. That's a much stronger basis for typing than callee proximity.
Within each, sub-types follow the same logic. Within comp:

elem comp: direct subtraction (comp_derT: _i_ - i_)
structural comp: subtraction + normalization (base_comp: rn scaling)
aggregate comp: iterated subtraction across sets (comp_N_, comp_C_)

Within clust:

connectivity: additive by adjacency (cluster_N)
centroid: additive by similarity to center (cluster_C, cluster_P)
summation: pure accumulation (sum2G, sum2F, sum2C)

So the type hierarchy in O directly reflects arithmetic complexity: scalar → structural → aggregate, within each fundamental operation type. And the wTT per fork weights how much each arithmetic type contributes to prediction — which is exactly what rpTT measures.
Does this mean val/eval functions (val_, vt_, get_exemplars) are a fourth type — conditional/selective — or are they part of fback?

kwcckw: > The main thing I'm not sure about: comp and clus are not really parallel forks — clustering operates on comp output, sequentially. Does the top-level N_ imply they're alternatives, or just that they're the two major branches of the process? If sequential, maybe H is more appropriate than N_ at the top level?

The the usage of CF and nested N_ is clearer on each layer of comp function. For clustering, it shouldn't be parallel to comp, but they did share a same top level. 

So i think they should be sequential since clustering always after the comp and feedback always after the clustering?  And they should be in N_ (as in above), where `top level O = CF(N_= [CF(nF='comp'), CF(nF='clust'), CF(nF='fback'))` since comp, clust and feedback are in the same level. 

For more complicated cases, sub+ or agg+, we may have a transition stage where cluster (cluster_N_) ->  comp (cross_comp), so we may get combinations of comp and clust fork in N_? Something like:

    CF(nF='cluster_N_', N_= [CF(nF='sum2G'), CF(nF='trans_cluster'),CF(nF='cross_comp'))

boris-kz: clustering is parallel to comp as a coarse processing type, O will fork-out to different types and sub-types.
it's not really in calling order, but alternative sub-forks do tend be called by the same root fork.

So this is function clustering, by connectivity through calls and by centroid though operation similarity.
We can combine both criterions in defining O.N_?

kwcckw: > So this is function clustering, by connectivity through calls and by centroid though operation similarity.
We can combine both criterions in defining O.N_?

I see, yes, so i guess O.N_ should be just defined by their nature of operation? Which may include combinations of comp, clust or feedback.

boris-kz: it's all function clustering, but connectivity here defined differently from data processing.
With data, connectivity is pair-wise similarity, here it's just a call sequence, the caller and the callee are generally not similar.

kwcckw: Okay, and each sub O should be added automatically to root O within the function, unless we specify it differently?

For example, in base_comp, we should check if `CF(nF='base_comp')` in current root O's N_ and add it? In this case, those root O should be retrieve via the mapping from one global root O? 

But we may call base_comp or comp_N from multiple levels of agg+ and sub+, i'm not sure if this can be easily tracked with global root O? Or each root O should be parsed across the functions?

boris-kz: Ideally they should be clustered automatically, similarly to primary data processing, but I don't know exactly how. Manually, it looks something like this:

O = CF(N_= [CF(nF='comp_'), CF(nF='clust'), CF(nF='eval_')])  # replaces wTT_: process weights, may be also param weights?

comp_ = CF(N_= [CF(nF='cross_comp'), CF(nF='comp_N_'), CF(nF='comp_C_')])  # finer comps downstream?

clust = CF(N_= [CF(nF='exemplars'), CF(nF='cluster_N'), CF(nF='cluster_C'), CF(nF='cluster_P')])  # Q2R, cent_TT, sum functions downstream?

eval_ = CF(N_= [CF(nF='vt_'), CF(nF='val_'), CF(nF='proj_TT'), CF(nF='proj_N'), CF(nF='ffeedback')])  # or ffeedback is a clust type?

boris-kz: Claude: A few notes:

1. `cross_comp` is more of a sequencer than a comp operation — it calls both comp and clust. It might belong at O level rather than inside `comp_`. Or is it here because its core contribution is initiating comparison?

2. `ffeedback` in `eval_` makes sense — it scales weights, same type as `vt_`, `val_`. It's not clustering even though it accumulates ratios across levels.

3. `cluster_P` is arguably more `eval_` than `clust` — it refines memberships by parallel reweighting (EM-like), more multiplicative than additive. Or is the additive sum2C inside it the dominant operation?

4. Missing: `comp_N`, `base_comp`, `comp_derT`, `comp_F`, `comp_H` — these would be downstream of `comp_N_` and `comp_C_`? So:

```
comp_N_ = CF(N_= [CF(nF='comp_N'), CF(nF='proj_V')])
comp_N  = CF(N_= [CF(nF='base_comp'), CF(nF='comp_H'), CF(nF='comp_F')])
base_comp = CF(N_= [CF(nF='comp'), CF(nF='comp_A'), CF(nF='comp_derT')])
```

5. `sum2G`, `sum2F`, `sum2C`, `Q2R` — pure accumulation. Downstream of `clust`, or a fourth top-level fork? They're additive but not selective like clustering proper.

Main question: how deep do you want to define this now? The top two levels seem stable. Deeper levels depend on rpTT telling you where finer weights are needed.

kwcckw: > Ideally they should be clustered automatically, similarly to primary data processing, but I don't know exactly how. Manually, it looks something like this:
> 
> O = CF(N_= [CF(nF='comp_'), CF(nF='clust'), CF(nF='eval_')]) # replaces wTT_: process weights, may be also param weights?
> 
> comp_ = CF(N_= [CF(nF='cross_comp'), CF(nF='comp_N_'), CF(nF='comp_C_')]) # finer comps downstream?
> 
> clust = CF(N_= [CF(nF='exemplars'), CF(nF='cluster_N'), CF(nF='cluster_C'), CF(nF='cluster_P')]) # Q2R, cent_TT, sum functions downstream?
> 
> eval_ = CF(N_= [CF(nF='vt_'), CF(nF='val_'), CF(nF='proj_TT'), CF(nF='proj_N'), CF(nF='ffeedback')]) # or ffeedback is a clust type?

I think eval_ should be part of comp_ of clust? Because we may have eval within both comp and clust?

Also does this cross_comp in `comp_ = CF(N_= [CF(nF='cross_comp'), CF(nF='comp_N_'), CF(nF='comp_C_')]) ` represent all instances of cross_comp? I thought they should be nested in other instances of O?

boris-kz: O is global, it actually maps to cross_comp: common root function

>I think eval_ should be part of comp_ of clust? Because we may have eval within both comp and clust?

that's precisely why it should be clustered by similarity, not connectivity. just like centroids may be shared by multiple Gs

boris-kz: we may also add attr_ fork, also top-level, to replace all global param weights?

more on automatic function clustering in the end:

https://claude.ai/share/ccdde186-1f41-47a9-952a-560bf0d2e81f

frame_2D_alg/agg_recursion.py
boris-kz: right, please do

frame_2D_alg/agg_recursion.py
kwcckw: > we may also add attr_ fork, also top-level, to replace all global param weights?

Yes, but that should be per top O only right? 
 
> O is global, it actually maps to cross_comp: common root function

But we may call multiple cross_comp, so it maps to the top cross_comp only? And subsequent cross_comp will be mapped in nested N_?

frame_2D_alg/agg_recursion.py
```diff
@@ -259,6 +259,7 @@ def comp_C_(C_, rr,_C_=[], fall=1, fC=0):  # simplified for centroids, trans-N_s
             comp_N(_C,C, rr,A=dy_dx,span=dist,L_=L_,N_=N_,acc=acc)  # simplified for typ=3
     for N in list(set(N_)):
         if N.rim: Q2R(N.rim, N.Rt, merge=0, froot=0)
+    # if np.any(rdpTT): accum_rdpTT('comp_N_', rdpTT)  # feed prediction error into code weights
     TTm,cm,rm,TTd,cd,rd = acc
```
boris-kz, line 263: We run out of limit on that thread :)

>But we may call multiple cross_comp, so it maps to the top cross_comp only? And subsequent cross_comp will be mapped in nested N_?

doesn't matter, it's the same function

kwcckw: Oh really, looks like 19 comments is the limit. 

Also for this global O, later we should parse it into comp_slice or frame_blobs for similar process too?

boris-kz:
Don't worry about it

PR #94, opened 2026-03-25
3/25/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -122,18 +118,40 @@ def __bool__(f): return bool(f.c)  # N_ may be empty?
 wY = wX = 64; wYX = np.hypot(wY,wX)  # focus dimensions
 decay = ave / (ave+avd)  # match decay / unit dist?
 
+# omni: root process forks, wTT=wTT, dTT=rdpTT, N_=[CF]: sub-forks, cluster by calls or/and similarity? next level:
+comp_ = CF(nF='comp_', N_=[CF(nF='cross_comp',m=Nw), CF(nF='comp_N_',m=nw), CF(nF='comp_C_',m=nw), CF(nF='comp_F',m=cw)]) # finer comps downstream?
+clust = CF(nF='clust', N_=[CF(nF='exemplars',m=nw), CF(nF='cluster_N',m=Nw), CF(nF='cluster_C',m=Cw), CF(nF='cluster_P',m=Cw)])  # Q2R, cent_TT, sum functions downstream?
+eval_ = CF(nF='eval_', N_=[CF(nF='vt_',m=cw), CF(nF='val_',m=cw), CF(nF='proj_TT',m=specw), CF(nF='proj_N',m=specw), CF(nF='ffeedback',m=nw)])  # or ffeedback is a clust type?
+# something like this?
+attr_ = CF(nF='attr_', N_=[CF(nF='wTTN',wTT=wTTN,m=wN), CF(nF='wTTC',wTT=wTTC,m=wC), CF(nF='wTTn',wTT=wTTn,m=wn), CF(nF='wTTc',wTT=wTTc,m=wc)])
+Om = CF(nF='Om'); Om.comp_ = comp_; Om.clust = clust; Om.eval_ = eval_; Om.attr_ = attr_  # or  it's better pack them as Om.attr instead inside N_? It should be clearer?
```
kwcckw, line 127: So i edited the structure of O here. 

- Renamed it to Om to prevent the same var name as O in certain functions.
- For those process fork, i think we need CF.wTT for wTT, CF.dTT for rdpTT, and CF.m for cw?
- Also it's clearer to pack top level comp_, clust, eval_ and attr_ in Om's attr instead of Om.N_?

frame_2D_alg/agg_recursion.py
```diff
@@ -122,18 +118,40 @@ def __bool__(f): return bool(f.c)  # N_ may be empty?
...
+def get_index(node):  # update mapping network (map each name such as comp_ or cross_comp to their CF)
+    cw_map[node.nF] = node
+    for c in node.N_: get_index(c)
+    for attr in ('comp_','clust','eval_','attr_'):
+        if hasattr(node, attr): get_index(getattr(node, attr))
+
+get_index(Om)
+def cw_(name): cw_map.get(name).m    # w
+def aw_(name): cw_map.get(name).wTT  # wTT
+def rw_(name): cw_map.get(name).dTT  # rdpTT
```
kwcckw, line 140: So from Om, i think we can use a mapping dict to map function name (cross_comp, comp_N_, etc) to each CF fork in Om.

Then we can use additional functions such as cw_, aw_, rw_ above to get cw, wTT and rdpTT from each fork for update purpose.

boris-kz: This mapping is not a recursive structure, we will need a new one for each forking level.
just search for whatever name on the promt in N_[i].nF?

frame_2D_alg/agg_recursion.py
boris-kz: It should probably be X for cross_comp, that's the root function it maps to
rdpTT is currently per G, not per function, we need some credit assignment scheme
comp_, clust, eval_ and attr_ is just a top level, we need to make this fork nesting recursive

frame_2D_alg/agg_recursion.py
```diff
@@ -122,18 +118,40 @@ def __bool__(f): return bool(f.c)  # N_ may be empty?
 wY = wX = 64; wYX = np.hypot(wY,wX)  # focus dimensions
 decay = ave / (ave+avd)  # match decay / unit dist?
 
+# omni: root process forks, wTT=wTT, dTT=rdpTT, N_=[CF]: sub-forks, cluster by calls or/and similarity? next level:
+comp_ = CF(nF='comp_', N_=[CF(nF='cross_comp',m=Nw), CF(nF='comp_N_',m=nw), CF(nF='comp_C_',m=nw), CF(nF='comp_F',m=cw)]) # finer comps downstream?
+clust = CF(nF='clust', N_=[CF(nF='exemplars',m=nw), CF(nF='cluster_N',m=Nw), CF(nF='cluster_C',m=Cw), CF(nF='cluster_P',m=Cw)])  # Q2R, cent_TT, sum functions downstream?
+eval_ = CF(nF='eval_', N_=[CF(nF='vt_',m=cw), CF(nF='val_',m=cw), CF(nF='proj_TT',m=specw), CF(nF='proj_N',m=specw), CF(nF='ffeedback',m=nw)])  # or ffeedback is a clust type?
+# something like this?
+attr_ = CF(nF='attr_', N_=[CF(nF='wTTN',wTT=wTTN,m=wN), CF(nF='wTTC',wTT=wTTC,m=wC), CF(nF='wTTn',wTT=wTTn,m=wn), CF(nF='wTTc',wTT=wTTc,m=wc)])
```
boris-kz, line 126: this is supposed to include all the attrs above: ave,avd,  wM,wD,wc, wG,wI,wa, wL,wS,wA, etc,
with multiple nesting levels

frame_2D_alg/agg_recursion.py
kwcckw: > comp_, clust, eval_ and attr_ is just a top level, we need to make this fork nesting recursive

Okay, but could you clarify if we will get deeper level's `comp_, clust, eval_ and attr_` for deeper agg+ or sub+?

If yes, how do we really track it since the tree may grow in-breadth and in-deep at the same time?

boris-kz: This is a function call/type nesting as in AST, not G nesting as in agg+/sub+, totally different thing 
that includes attr names,
claude:

Yes, `ast` gives you everything inside a function body:

- `ast.Assign` — assignments, with targets and values
- `ast.AugAssign` — `+=`, `-=`, `*=`
- `ast.BinOp` — `+`, `-`, `*`, `/`, `@` (matmul)
- `ast.Compare` — `>`, `<`, `==`
- `ast.If` — conditionals, with test expression
- `ast.For`, `ast.While` — loops
- `ast.Return` — return values
- `ast.Subscript` — indexing `N.dTT[1]`
- `ast.Attribute` — attribute access `N.yx`, `N.rim`

So for each function you could count:

```python
op_signature = {
    'sub': count(ast.Sub),      # subtractive
    'add': count(ast.Add),      # additive  
    'mult': count(ast.Mult),    # multiplicative
    'matmul': count(ast.MatMult), # @ weighted sum
    'div': count(ast.Div),      # normalization
    'compare': count(ast.Compare), # evaluation
    'branch': count(ast.If),    # conditional complexity
    'loop': count(ast.For + ast.While),  # iteration
}
```

And attribute access tells you what types it works with — `n.dTT`, `n.rim`, `n.yx` etc. distinguish comp functions (access dTT, kern) from clustering functions (access N_, rim, fin).

The op_signature is directly comparable between functions: subtract two signatures, min is match. That's your function "latuple" equivalent.

The main limitation: AST is syntactic, not semantic. `a - b` in `comp_derT` (computing difference) and `a - b` in `sum2C` (adjusting index) look identical. Do you care about that distinction at this level, or is the statistical signature (comp_derT has many more subtractions than sum2C) sufficient?

frame_2D_alg/agg_recursion.py
boris-kz: def fetch(root, name):

    for i, fork in enumerate(root.N_):
        if fork.nF==name:
            return i,fork

frame_2D_alg/agg_recursion.py
kwcckw: From my understanding, it's still have the level structure?

So top level is Om, and then we have comp_, clust and eval_ and attr_, and then each of their N_ will be in the next level. 

However, i'm not quite clear on the recursive structure in AST here, could you explain a bit on that? Is it the mapping of deeper node will be mapped to a higher level instead?

boris-kz: recursive is the unpack / access process

comparison between functions / higher AST levels will reveal their shared elements.
if these elements are shared across sufficient number of roots, they should be elevated to the root level themselves, just like centroids

frame_2D_alg/agg_recursion.py
```diff
@@ -122,18 +118,40 @@ def __bool__(f): return bool(f.c)  # N_ may be empty?
...
+def aw_(name): cw_map.get(name).wTT  # wTT
+def rw_(name): cw_map.get(name).dTT  # rdpTT
+
+
 def vt_(TT, r, wTT=wTTN):  # brief val_ to get m,d, rc=0 to return raw vals, wTTn for comp_N
 
-    m_,d_= TT; ad_ = np.abs(d_); t_ = m_ + ad_ + eps  # ~ max comparand
-    return m_/t_ @ wTT[0] - ave*r, ad_/t_ @ wTT[1] - avd*r  # norm by co-derived variance
+    m_,d_= TT; ad_ = np.abs(d_); t_ = m_ + ad_  # ~ max comparand
+    return m_/np.where(t_==0,eps,t_) @ wTT[0] - ave*r, ad_/np.where(t_==0,eps,t_) @ wTT[1] - avd*r  # norm by co-derived variance
```
boris-kz, line 146: use:  def eps_(a): a[a == 0] = eps; return a?

frame_2D_alg/agg_recursion.py
boris-kz, line 1: So it does seem that commenting here is easier than on commits, especially typing and visibility?
except if you do it over email

frame_2D_alg/agg_recursion.py
boris-kz: claude:

On prototype promotion: yes, if a pattern appears across enough functions in a fork, it becomes a new CF in the parent fork's N_. Like if the subtract→abs→min pattern is shared across `comp_derT`, `comp_A`, `comp`, it becomes `CF(nF='diff_match')` in `comp_.N_` — a discovered operation type alongside the manually defined ones.

On credit assignment: this is the hard part. rdpTT is computed in `comp_N_` as aggregate prediction error. To assign credit to individual functions that contributed, you need to decompose it.

Two approaches:

1. **By partial dTT contribution.** Each function in the call chain produces a partial dTT change. Its credit = how much of total rdpTT its partial output explains. Like `cent_TT` logic: weight by correlation to the sum. This is cheap but coarse — it doesn't tell you if the function *caused* the error or just correlated with it.

2. **By ablation.** Run with and without each function (or with default wTT vs refined wTT), compare rdpTT. The difference is that function's credit. This is precise but expensive — needs multiple passes.

The practical middle ground: **marginal contribution along the call sequence.** Track cumulative dTT at each function boundary in the call chain. rdpTT decomposes into segments:

```
rdpTT_total = rdpTT_base_comp + rdpTT_comp_H + rdpTT_comp_F + ...
```

Where each segment = `abs(pTT_after - pTT_before)` for that function's contribution. You already have pTT (projected) and Link.dTT (actual) — you'd just need intermediate checkpoints.

That means instrumenting the call chain to record partial dTT at each step. Is that acceptable overhead, or too invasive for now?

frame_2D_alg/agg_recursion.py
kwcckw: This is replacing the array's value as eps, which we want to avoid that since that could be accumulated? I think we need to return and replace a copied instance instead?

frame_2D_alg/agg_recursion.py
kwcckw: > just search for whatever name on the promt in N_[i].nF?

Yes, but in your fetch, it should have another recursive fetch in the else section? But i'm not sure the value of `i` for the recursive structure yet.

frame_2D_alg/agg_recursion.py
kwcckw: From claude above, those are built-in AST module, and we are not using that?

> recursive is the unpack / access process
> if these elements are shared across sufficient number of roots, they should be elevated to the root level themselves, just like centroids

I see, also on `comp_, clust, eval_ and attr_ is just a top level, we need to make this fork nesting recursive`, it's just mean that the unpack and access is recursive? Instead of typical recursive structure like comp_->clust->feedback->comp_ again?

frame_2D_alg/agg_recursion.py
kwcckw: Yes, also one of the problem probably is the limitation of the reply number.

boris-kz: just start a new thread

frame_2D_alg/agg_recursion.py
boris-kz: This has nothing to with recursion in agg_recursion yet, pixels look very different from the code
We will probably get generalized process later, but need to get something to generalize from first. 

we should be using AST, this manual scheme is just stab, it should automated once we have an idea what it should look like.

frame_2D_alg/agg_recursion.py
boris-kz: ok, def eps_(a): return np.where(a==0, eps, a)

frame_2D_alg/agg_recursion.py
boris-kz: not sure, I think we only need to get top-level attrs, which may be promoted lower-level prototypes

frame_2D_alg/agg_recursion.py
kwcckw: > On prototype promotion: yes, if a pattern appears across enough functions in a fork, it becomes a new CF in the parent fork's N_. Like if the subtract→abs→min pattern is shared across comp_derT, comp_A, comp, it becomes CF(nF='diff_match') in comp_.N_ — a discovered operation type alongside the manually defined ones.

Okay, so the subsequent forks under the global O is dynamic and we may create new one during the runtime?

boris-kz: that will probably be a higher-order feedback

kwcckw: > This has nothing to with recursion in agg_recursion yet, pixels look very different from the code We will probably get generalized process later, but need to get something to generalize from first.
> 
> we should be using AST, this manual scheme is just stab, it should automated once we have an idea what it should look like.

I see, this makes sense now. 

For discovered operations, i'm not sure if that's possible since all operation should be defined explicitly in the code first? Unless the fork combination is too complex and all the subsequent deeper node should be generated from the AST?

For credit assignment, the ablation method looks more accurate since we can capture the precise difference. But i think it's not really practical to run through it with several iterations?

boris-kz: not discovered, promoted from lower AST levels to higher as prototype ops, if in multiple higher forks

not sure about credit assignment yet

kwcckw: Thanks, this is clearer now.

As in example above, if we have multiple operations of "subtract→abs→min" in different nodes, they should be promoted as a higher AST level. To determine this kind of pattern and promote them, it should be an internal process within the AST?

boris-kz: right, like cross_comp but with binary match

boris-kz: Re credit assignment, we have gain from comps: rdpTT, and estimated loss reduction from clustering.
So that gain should be fed back the process fork tree: sequential roots from current function fork, then
 
c,_c = fork, root_fork.c;  C = c+_c
root_fork.wTT = (root_fork.wTT*_c + fork.wTT*c) / C?

kwcckw: I think yes, but the fed back process should be sequentially from the deeper level to higher root level. Also for the final feedback process, that should be per each node instead of just the global O to update each level node's cw and wTT?

frame_2D_alg/agg_recursion.py
kwcckw: Okay, I guess this only can be decided once we have the finalized O structure. The promoted structure should be in root.N_ so I think we still need to unpack the deeper nodes if there's no matching name in top level. For i, if we need it from tracking purpose, I think we should pack it per node?

frame_2D_alg/agg_recursion.py
boris-kz: we have gain from comps: rdpTT, and estimated loss reduction from clustering.
So gain should be fed back up the process fork tree: sequential roots from current function fork

But to assign credit proportionally, we need to get dTT and c from the data stricture on which the process was performed. Because rdpTT is computed relative to dTT, not wTT.
So we need data_root to compute contribution, and func_root to assign it to?

something like this:
`_c = G.root.c; c = G.c; C = c+_c`
`func_root.wTT *= rdpTT * ((func_root.wTT*_c + func.wTT*c) / C)`?

boris-kz: Claude: Yes, the dual root makes sense. Each operation has:

- **data context**: the G (or N, L) it operated on → provides dTT, c for proportional credit
- **process context**: its fork in O → where the credit lands as wTT update

The formula logic: rdpTT is the raw gain, but how much of it to assign to this function vs its root depends on their relative contribution, weighted by the data they processed.

But I think the weighting is off. Let me unpack:

- `func_root.wTT * _c`: root's weight scaled by root's data scope
- `func.wTT * c`: this function's weight scaled by its data scope
- ratio = this function's share of total weighted scope

So the update would be:

```
share = (func.wTT * c) / (func_root.wTT * _c + func.wTT * c)
func.wTT *= 1 + rdpTT * share
func_root.wTT *= 1 + rdpTT * (1 - share)
```

Both get updated, proportionally to their contribution. rdpTT > 0 means prediction improved → increase weights. rdpTT < 0 → decrease.

But the question is: where do you get `func.wTT` and `func_root.wTT` from? Currently wTT_ is global, not indexed by function. The O tree gives you the indexing, but each function needs to know its position in O to look up its wTT.

That means each function call needs a reference to its CF in O — either passed as argument or looked up from call stack. Which is more consistent with your current calling convention: adding an `oF` parameter, or a global lookup by function name?

kwcckw: > Because rdpTT is computed relarive to dTT, not wTT.

Yes, and for "where do you get func.wTT and func_root.wTT from?", that should be from cent_TT?

> That means each function call needs a reference to its CF in O — either passed as argument or looked up from call stack. 

I think that could be done with dict mapping or the fetch function below?

boris-kz: it's just a func_fork.root, higher fork

func_root wTT is dTT, and we don't need wTT / rdpTT in CF, that stays in data

frame_2D_alg/agg_recursion.py
boris-kz: don't know about i. unpacking deeper nodes is specification it should be conditional

frame_2D_alg/agg_recursion.py
kwcckw: > func_root wTT is dTT,

Could you clarify on this? Each func root should have their own wTT , updated from feedback , so why it's dTT here?

boris-kz:
Because there is no dTT, we're just using the name

frame_2D_alg/agg_recursion.py
kwcckw: So for the params tree:

    CF(nF='attr_')  # packs param weight tree above

Suppose it should be pack single weight per node? Or single node for same level's weights? And their weight will be packed as m?

PR #95, opened 2026-03-26
3/26/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -104,8 +104,8 @@ def __init__(f, **kwargs):
     def __bool__(f): return bool(f.c)  # N_ may be empty?
 
 ave = .3; avd = ave*.5  # ave m,d / unit dist: the top of filter hierarchy
-wM,wD,wc, wG,wI,wa, wL,wS,wA = 10, 10, 20, 20, 5, 20, 2, 1, 1  # dTT weights = reversed relative ave
-wT = np.array([wM,wD,wc, wG,wI,wa, wL,wS,wA])
+wM,wD,wi, wG,wI,wa, wL,wS,wA = 10, 10, 20, 20, 5, 20, 2, 1, 1  # dTT weights = reversed relative ave
```
kwcckw, line 107: I changed wc to wi here because we have the same name of wc from `wC,wN,wc,wn = 4,3,2,1 ` below.

frame_2D_alg/agg_recursion.py
```diff
@@ -116,12 +116,38 @@ def __bool__(f): return bool(f.c)  # N_ may be empty?
...
 decay = ave / (ave+avd)  # match decay / unit dist?
 
+tt = CF(nF='TT',  # promoted from wT * ave
+        N_=[CF(nF='wT', 
+               N_=[CF(nF='wM'), CF(nF='wD'), CF(nF='wi'),
+                   CF(nF='wG'), CF(nF='wI'), CF(nF='wa'),
+                   CF(nF='wL'), CF(nF='wS'), CF(nF='wA')])
+])
+
+attr_ = CF(nF='attr_',
```
kwcckw, line 126: I added a draft of the param weight tree here.

So i think the most complicated part is the tree of wTT, where the hierarchy should be something like this?

    wTT_ ->  TT -> wT -> w ?

Also since decay is derived from ave/(ave+avd), it should be higher level than ave or avd?

frame_2D_alg/agg_recursion.py
```diff
@@ -659,7 +685,7 @@ def add_Nt(G, Nt, merge=0):  # addition to Q2R
         add_H(G.Nt.H if merge else G.Nt.H[:-1], Nt.H, G)  # else top lev = Nt.N_
     N_ = Nt.N_
     if merge: G.N_ += N_ # never 2
-    else: G.Nt.H += [Q2R(N_+ (G.Nt.H.pop().N_ if G.Nt.H else []), G.Nt)]  # extend or init top level
```
kwcckw, line 662: The root should be None here to create a new level n G.Nt.H. Else we are packing G.Nt to G.Nt.H, which is wrong.

frame_2D_alg/agg_recursion.py
boris-kz: yes, wTT_(  TT( wT( w, each a CF. 
we may add CF per dDert, dKern, dExt, as in your lines above, they probably behave differently
decay is just a computational short-cut, not independent 
So all this nesting needs to be unpacked in cent_TT, etc, we need a generic unpack function

frame_2D_alg/agg_recursion.py
boris-kz: G.Nt.H in G.Nt, but that's still wrong because summed H is redundant

frame_2D_alg/agg_recursion.py
```diff
@@ -116,12 +116,38 @@ def __bool__(f): return bool(f.c)  # N_ may be empty?
...
+                      CF(nF='wTTC', N_=[CF(nF='wC'), tt]),  
+                      CF(nF='wTTn', N_=[CF(nF='wn'), tt]), 
+                      CF(nF='wTTc', N_=[CF(nF='wc'), tt])]),
+
+               # secondary
+               CF(nF='secondary', N_=[CF(nF='aveB'),CF(nF='Lw'),CF(nF='distw'),CF(nF='intw'),CF(nF='wY'),CF(nF='wX')]),
+               CF(nF='wYX'),  # promoted from np.hypot(wY,wX)
+               
+               # fb weights per dTT, adjust in agg+
+               CF(nF='fb_w', N_=[CF(nF='mW'),CF(nF='dW')])
```
boris-kz, line 141: this would be in your tt, it needs another nesting level

frame_2D_alg/agg_recursion.py
boris-kz, line 1: So we don't need f.wTT= kwargs.get('wTT',np.zeros((2,9))), that's f.dTT in operation forks: oF

Normalized oF feedback sum we discussed yesterday should actually use oF.c, not G.c, that's a separate feedback path from data root feedback, we need both.

Also in oF, we need separate sub-forks for primary call tree (connectivity) and promoted type tree (centroid),
we can use oF.Lt for the second? It would have separate feedback too.

frame_2D_alg/agg_recursion.py
boris-kz: probably better wC here, and  wcc,wnc, wc,wn = 4,3,2,1: additional c for clustering

frame_2D_alg/agg_recursion.py
boris-kz: wT = CF(nF='wT', N_=[CF(nF='wM'),CF(nF='wD'),CF(nF='wC'), CF(nF='wG'),CF(nF='wI'),CF(nF='wa'), CF(nF='wL'),CF(nF='wS'),CF(nF='wA')])
TT = CF(nF='TT', N_=[CF(nF='mT',N_=wT),CF(nF='dT',N_=wT)])
TT_= CF(nF='TT_',N_=[CF(nF='wnc'),CF(nF='wcc'),CF(nF='wn'),CF(nF='wc')]), # wTTN,wTTC,wTTn,wTTc are nested

frame_2D_alg/agg_recursion.py
```diff
@@ -116,12 +116,38 @@ def __bool__(f): return bool(f.c)  # N_ may be empty?
...
+               N_=[CF(nF='wM'), CF(nF='wD'), CF(nF='wi'),
+                   CF(nF='wG'), CF(nF='wI'), CF(nF='wa'),
+                   CF(nF='wL'), CF(nF='wS'), CF(nF='wA')])
+])
+
+attr_ = CF(nF='attr_', 
+           N_=[CF(nF='ave_', N_=[CF(nF='ave'), CF(nF='avd')]),  # ave m,d / unit dist: the top of filter hierarchy
+               CF(nF='decay'),  # promoted from ave/(ave+avd)
+    
+               CF(nF='wTT_',   # promoted from TT * w
```
boris-kz, line 130: TT_, from above

frame_2D_alg/agg_recursion.py
kwcckw: > Also in oF, we need separate sub-forks for primary call tree (connectivity) and promoted type tree (centroid),
we can use oF.Lt for the second? 

Okay, i guess that's better since i was thinking to pack them (promoted) in N_, but we can't differentiate them in N_. Also from those promoted tree, do we need something like root reference to reference the primary tree?

frame_2D_alg/agg_recursion.py
kwcckw: Okay, so dDert, dKern, and dExt will be dTT[0:3], dTT[3:6] and dTT[7:10] ?

I think we need the generic unpack function to unpack them in both cent_TT and feedback, and we can use the dict method to map each name to CF? Otherwise, we can use the existing x_, but that requires the recursive unpacking process in each call, which is not really efficient.

frame_2D_alg/agg_recursion.py
kwcckw: This will be the same level as wT? Or it will be in the bottom level, same as those ws?

frame_2D_alg/agg_recursion.py
kwcckw: Okay right, so i guess we not really need this H packing?

frame_2D_alg/agg_recursion.py
kwcckw: > wT = CF(nF='wT', N_=[CF(nF='wM'),CF(nF='wD'),CF(nF='wC'), CF(nF='wG'),CF(nF='wI'),CF(nF='wa'), CF(nF='wL'),CF(nF='wS'),CF(nF='wA')]) TT = CF(nF='TT', N_=[CF(nF='mT',N_=wT),CF(nF='dT',N_=wT)]) TT_= CF(nF='TT_',N_=[CF(nF='wnc'),CF(nF='wcc'),CF(nF='wn'),CF(nF='wc')]), # wTTN,wTTC,wTTn,wTTc are nested

So the wT, TT and TT_ has the same nesting level here? Or the promoted fork should be in Lt?

frame_2D_alg/agg_recursion.py
boris-kz: that's oF.root

frame_2D_alg/agg_recursion.py
boris-kz: I really don't like dict, it's not customizable

frame_2D_alg/agg_recursion.py
boris-kz: right

frame_2D_alg/agg_recursion.py
boris-kz: no, they are nested sequentially, draft:

boris-kz: wT = CF(nF='wT', N_=[CF(nF='wM'),CF(nF='wD'),CF(nF='wC'), CF(nF='wG'),CF(nF='wI'),CF(nF='wa'), CF(nF='wL'),CF(nF='wS'),CF(nF='wA')])
    TT = CF(nF='TT', N_=[CF(nF='mT',N_=wT),CF(nF='dT',N_=wT)])
    TT_ = CF(nF='TT_', N_=[CF(nF='wnc'),CF(nF='wcc'),CF(nF='wn'),CF(nF='wc')])  # nested with wTTN,wTTC,wTTn,wTTc
    wY,wX,hyp = CF(nF='yx_', N_=[CF(nF='wY'), CF(nF='wX'), CF(nF='hyp')])
    aveB,Lw,distw,intw = CF(nF='aveB'), CF(nF='Lw'), CF(nF='distw'), CF(nF='intw')
    # process weight tree:
    X = CF(nF = 'X', # maps to cross_comp, replacing globals, dTT = process wTT, root: caller for credit assign, not similarity type?
        N_ = [  # process forks:
        CF(nF='comp_', N_=[CF(nF='comp_N_'), CF(nF='comp_C_'), CF(nF='comp_N'), CF(nF='comp_F')]),  # finer comps downstream?
        CF(nF='clust', N_=[CF(nF='exemplars'), CF(nF='cluster_N'), CF(nF='cluster_C'), CF(nF='cluster_P')]),  # Q2R, cent_TT, sum functions downstream?
        CF(nF='eval_', N_=[CF(nF='vt_'), CF(nF='val_'), CF(nF='proj_TT'), CF(nF='proj_N'), CF(nF='ffeedback')]),  # or ffeedback is a clust type?
        CF(nF='attr_', N_=[ave,avd,decay, TT_,wY,wX,hyp,aveB,Lw,distw,intw])
        ])  # replace with automatic fork clustering by call+type, || AST, wTT *= rel prediction error, _wTT /= rel prediction error?

frame_2D_alg/agg_recursion.py
kwcckw: Okay, for now i guess we can continue with the recursive function method first.

frame_2D_alg/agg_recursion.py
kwcckw: I see, so for :

    TT = CF(nF='TT', N_=[CF(nF='mT',N_=wT),CF(nF='dT',N_=wT)])

That N_ is a node instead of a list, is that intended?

Also TT is not in any of the fork?

boris-kz: TT is in TT_. No, N_ should not be nested

boris-kz: # single attr oFs: w = m: reversed relative ave, no dTT?
    ave,avd, decay = CF(nF='ave',m=.3), CF(nF='avd',m=.3), CF(nF='decay',m = ave/(ave+avd))  # ave m,d / unit dist
    wY, wX = CF(nF='wY',m=64), CF(nF='wX',m=64); YX = CF(nF='YX',m=np.hypot(wY,wX))
    aveB,Lw, distw,intw = CF(nF='aveB',m=100), CF(nF='Lw',m=.5), CF(nF='distw',m=.5), CF(nF='intw',m=.5)
    # attr weight tree:
    wT = CF(nF='wT', N_=[CF(nF='wM',m=9),CF(nF='wD',m=9),CF(nF='wC',m=9), CF(nF='wG',m=20),CF(nF='wI',m=5),CF(nF='wa',m=20), CF(nF='wL',m=2),CF(nF='wS',m=1),CF(nF='wA',m=1)])
    TT = CF(nF='TT', N_=[CF(nF='mT',N_=wT,m=9), CF(nF='dT',N_=wT,m=9)])  # m,d forks
    wnc,wcc, wn,wc = TT,TT,TT,TT; wnc.nF='wnc';wnc.m=10; wcc.nF='wcc';wcc.m=20; wn.nF='wn';wn.m=5; wc.nF='wc';wn.m=10
    TT_= CF(nF='TT_',N_=[wnc,wcc,wn,wc])  # add sub-forks, as below?  maps to Nt,Ct,Lt, no Bt: no call, no info?
    # proc weight tree:
    X = CF(nF = 'X', # maps to cross_comp, replacing globals, dTT = process wTT, root: caller for credit assign, not similarity type?
        N_ = [  # process forks:
        CF(nF='comp_', N_=[CF(nF='comp_N_'), CF(nF='comp_C_'), CF(nF='comp_N'), CF(nF='comp_F')]),  # finer comps downstream?
        CF(nF='clust', N_=[CF(nF='exemplars'), CF(nF='cluster_N'), CF(nF='cluster_C'), CF(nF='cluster_P')]),  # Q2R, cent_TT, sum functions downstream?
        CF(nF='eval_', N_=[CF(nF='vt_'), CF(nF='val_'), CF(nF='proj_TT'), CF(nF='proj_N'), CF(nF='ffeedback')]),  # or ffeedback is a clust type?
        CF(nF='attr_', N_=[ave,avd,decay,TT_, wY,wX,YX, aveB, Lw,distw,intw])
        ])  # replace with automatic fork clustering by call+type, || AST, wTT *= rel prediction error, _wTT /= rel prediction error?
    '''
    if  projecting_root: F.root.wTT *= rdpTT * (F.c/ F.root.c)  # c-weighted feedback
    elif selecting_root: (F.m - F.root.Lt.m) * (F.root.c-Ft.c)  # clustering value = loss reduction: root.Lt.m < selective F.m, add dval?
    '''

boris-kz: claude: two distinct credit paths:

**Projecting**: straightforward — rdpTT directly measures projection error, weighted by data scope ratio:
```
F.root.wTT *= rdpTT * (F.c / F.root.c)
```

**Selecting**: value = how much selectivity reduced loss. The logic: clustering selects a subset (Ft.c < F.root.c), and the selected subset has higher match density (Ft.m > F.root.Lt.m per unit). So gain is:

```
(Ft.m - F.root.Lt.m) * (F.root.c - Ft.c)
```

Match density improvement × scope of what was excluded. That's compression: same total match from fewer elements.

But the second formula gives a value, not a ratio to multiply wTT by. To make it a wTT update like the first:

```
selecting_root.wTT *= (Ft.m - F.root.Lt.m) * (F.root.c - Ft.c) / (F.root.m * F.root.c)
```

Normalized by root's total value?

Also: `F.root.Lt.m` as the baseline — that's match of links between initial nodes before clustering. So the comparison is match density of clustered output vs pre-clustering links. If clustering didn't improve on that, its wTT decreases.

Third case missing: **eval/gating** credit. When `val_` skips a comp, the counterfactual gain is approximately `pTT` of the skipped comp. If pTT was low (correctly skipped), eval gets positive credit. If later evidence shows skipped comp would have been valuable, eval gets negative credit. But "later evidence" is hard to get without actually doing the comp. Is that just left as unknown for now?

kwcckw: > TT = CF(nF='TT', N_=[CF(nF='mT',N_=wT,m=9), CF(nF='dT',N_=wT,m=9)])  # m,d forks
wnc,wcc, wn,wc = TT,TT,TT,TT; wnc.nF='wnc';wnc.m=10; wcc.nF='wcc';wcc.m=20; wn.nF='wn';wn.m=5; wc.nF='wc';wn.m=10

So all wnc, wcc, wn and wc are sharing a same TT node and it doesn't matter here? Or each higher level wnc, wcc etc should have their own TT?

> two distinct credit paths:

I'm not sure on the third case, but by comparing the projection and selection method, i would think projection is better here, since the selection probably requires a very precise threshold to select the relevant one? And we will need another feedback to adjust this threshold again.

boris-kz: projection is for comp_N_ root oF, selection is when root is clustering, these are two feedback forks.

 wnc, wcc, wn, wc shape should be TT, different instances

kwcckw: Thanks, this is clear now.

For the selection:

    selecting_root: (F.m - F.root.Lt.m) * (F.root.c-Ft.c)  # clustering value = loss reduction: root.Lt.m < selective F.m, add dval?
    
It's actually a clustering value so we shouldn't use the threshold here.

When we have feedback for both comp and clustering, they should use a same operation as in ffeedback? The only difference will be the comp or cluster value + those weights?

boris-kz: oF feedback is more like Q2R, through immediate roots, quite different from ffeedback

kwcckw: > oF feedback is more like Q2R, through immediate roots, quite different from ffeedback

Then this oF feedback should replace ffeedback now? Since we are updating the same weights and attrs in oF feedback?

boris-kz: ffeedback is mostly for ave,avd, much more sparse. but it's updated yet, maybe it should be replaced

kwcckw: I think eventually it should be replaced since the same aves are in the attr_ fork, which in under X too.

boris-kz: It's between levels in frame_H pipeline, I think this root feedback should be within a level.
Not sure

kwcckw: > It's between levels in frame_H pipeline, I think this root feedback should be within a level. Not sure

But they are sharing a same set of aves? So we might have two different level of feedback system updating a same set of aves.

boris-kz: right

PR #96, opened 2026-03-27
3/27/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -90,7 +98,7 @@ def __bool__(n): return bool(n.c)
 class CF(CBase):  # iF/ data: rim, Nt,Ct, Bt,Lt: ext|int- defined nodes, ext|int- defining links, Lt/Ft, Ct/lev, Bt/G
                   # oF/ code: primary call tree + internal Lt type tree,
     name = "fork"
-    rim = prop_F_('Nt'),
+    rim = prop_F_('Nt'); TT = prop_CF_()
```
kwcckw, line 101: For Wcn,Wcc, Wn,Wc, their wTT should be retrieved from their nodes in nested N_, so we need a function that similar to prop_F_ in order to retrieve each node's m.

To retrieve their wTT values, we need to use something like Wcn.TT now.

Also ws are ints and they are packed in Wcn,Wcc, Wn,Wc's m too, so we actually just need to use Wcn.m and this line can be skipped?

    wcn,wcc, wn,wc = 10,20,5,10

And all usage of wcn,wcc, wn,wc should be reference from W's m instead.

frame_2D_alg/agg_recursion.py
```diff
@@ -682,14 +690,14 @@ def add_Nt(G, Nt, merge=0):  # addition to Q2R
 def add_Lt(G, Lt):  # addition to Q2R
 
     L_ = Lt.N_
-    if Lt.m > ave*specw:  # comp typ -1 pre-links
```
kwcckw, line 685: For specw, if we need to retrieve it from nested wTT_, we actually just need to use wn, which is Wn.m?

frame_2D_alg/agg_recursion.py
kwcckw: > Yes, dTT is to accumulate composites weights, per function or G

I think those weights should be retrieved from lower weights.m instead?

boris-kz: weights.m is much less precise, params in wTT vary differently from each other, the sum doesn't reflect that. It's the same reason why we are summing dTT instead of m,d in clustering

frame_2D_alg/agg_recursion.py
```diff
@@ -55,6 +55,14 @@ def get(N): return getattr(N,F).N_
     def set(N, new_N): setattr(getattr(N,F),'N_', new_N)
     return property(get,set)
 
+def prop_CF_():  # factory function to get and update TT only
```
boris-kz, line 58: thanks, so this is prop_w_

frame_2D_alg/agg_recursion.py
boris-kz: So if we use oF.N_ for weights, it should be a fixed np.array, same shape for same-type oF?
and then oF.dTT can be used for rdpTT, that should be summed per oF too?

boris-kz: >And all usage of wcn,wcc, wn,wc should be reference from W's m instead.

That's too complex, oFs should be used for updating only, we probably still need to get global numerical weights out of them before data processing

frame_2D_alg/agg_recursion.py
```diff
@@ -106,17 +114,17 @@ def __bool__(f): return bool(f.c)  # N_ may be empty?
 
 # single attr oFs: w = m: reversed rel ave, no dTT?
 ave,avd = .3,.5; decay = ave/(ave+avd) # ave m,d / unit dist
-Ave,Avd = CF(nF='ave',m=ave), CF(nF='avd',m=avd)  # updatable form?
+Ave,Avd,Decay = CF(nF='ave',m=ave), CF(nF='avd',m=avd),CF(nF='decay',m=decay)  # updatable form?
```
boris-kz, line 117: decay doesn't need an independently updatable form, same as wYX

frame_2D_alg/agg_recursion.py
kwcckw: > weights.m is much less precise, params in wTT vary differently from each other, the sum doesn't reflect that. It's the same reason why we are summing dTT instead of m,d in clustering
> So if we use oF.N_ for weights, it should be a fixed np.array, same shape for same-type oF?

Yes, but we only have either array based weights from higher level or a single weight from bottom level, so i think a fixed shape of (2,9) should be fine here. 

> and then oF.dTT can be used for rdpTT, that should be summed per oF too

Yes, so i think rdpTT should be per higher level of Wcn,Wcc, Wn,Wc only, but not those 9 lowest level of CFs.

kwcckw: > That's too complex, oFs should be used for updating only, we probably still need to get global numerical weights out of them before data processing

Okay, then those oF's m (Wcn,Wcc, Wn,Wc's m) should be always referencing the actual global numerical weight such as wcn,wcc, wn,wc, we may need to create a new property method on that.

frame_2D_alg/agg_recursion.py
```diff
@@ -137,12 +145,12 @@ def x_(root, name):
         if fork.nF==name:
             return i,fork
 
-def vt_(TT, r, wTT=Wcn):  # brief val_ to get m,d, rc=0 to return raw vals, Wn for comp_N
+def vt_(TT, r, wTT=Wcn.TT):  # brief val_ to get m,d, rc=0 to return raw vals, Wn for comp_N
```
boris-kz, line 148: this is just Wcn.N_?

frame_2D_alg/agg_recursion.py
boris-kz: m w updating is a lot less frequent than using, or wTT summing, so we can get global ws in batches

frame_2D_alg/agg_recursion.py
kwcckw: `Wcn.N_` are packing mT and dT?

    def make_TT(nF, m): # m,d forks
        return CF(nF=nF, m=m, N_=[make_T('mT'), make_T('dT')])

So we need to get Wcn.N_'s N_'s m, where Wcn.N_[0] is mT, and Wcn.N_[0].N_ are those 9CFs packing the 9 params weights. I use `.TT` because it's using prop_W to get those weights via .TT method.

frame_2D_alg/agg_recursion.py
kwcckw: Okay, but i can see below is using ` decay.m `, so that's wrong?

boris-kz: yes, should be decay

frame_2D_alg/agg_recursion.py
boris-kz: like I said, for data processing all weights should be unpacked as globals, 
we only need nested oF to update them? 
All that oF processing is basically meta-code, separate higher order

frame_2D_alg/agg_recursion.py
kwcckw: > m w updating is a lot less frequent than using, or wTT summing, so we can get global ws in batches

We use wTT more frequently, so the global ws in batches should be from Wcn,Wcc, Wn,Wc?

boris-kz: yes, we still need global wTTN, wTTC, wTTn, wTTc?
also that use of oF.N_ looks wrong, it should be oF.dTT?
rdpTT should be per G only?

kwcckw: > yes, we still need global wTTN, wTTC, wTTn, wTTc?

Yes, that's actually saves a lot of work, but we need to make sure each value in wTTN is the same or mapped to Wcn's N_'s N_'s m after the feedback.

> also that use of oF.N_ looks wrong, it should be oF.dTT?

Yes, i think oF.dTT should be packing rdpTT, so and that can't be per oF.N_.

> rdpTT should be per G only?

rdpTT should be per oF as well since it's not exclusive to G?  We may need to update rdpTT after all comps and clustering, so it can't be per G?

boris-kz: rdpTT can be both per G and oF, to compute the contribution.
So, oF.dTT maps to pTT, oF.N_ would be a call sequence for that func, clustrable by results,
then we need separate oF.wTT and optional oF.rdpTT?
That becomes similar to CN, we need to either add new attrs or convert it to CN?

boris-kz: Claude:
Keep CF but add the optional attrs. Converting to CN would pull in a lot of unused baggage (`kern`, `span`, `angl`, `mang`, `box`, `yx`, `nt`, `compared`, etc.) that has no meaning for process nodes.

The additions to CF would be minimal:

```python
f.wTT = kwargs.get('wTT', np.ones((2,9)))
f.rdpTT = kwargs.get('rdpTT', np.zeros((2,9)))
```

That's it for now. CF already has `dTT`, `m`, `d`, `c`, `r`, `N_`, `H`, `root` — sufficient for accumulation, hierarchy, and feedback. The structural difference is real: oFs cluster by call sequence and result type, not by spatial proximity or parameter similarity. Forcing them into CN would blur that distinction.

frame_2D_alg/agg_recursion.py
boris-kz: So lets keep this weight updating in ffeedback only for now.
That means in all other functions oFs will just accumulate call results.

frame_2D_alg/agg_recursion.py
boris-kz: we probably need separate oF.wTT anyway, so this is not needed?

frame_2D_alg/agg_recursion.py
kwcckw: Okay, so we should remain and revert some of the.removed.global weights since some of them are packed in those oFs.m now.

frame_2D_alg/agg_recursion.py
kwcckw: > then we need separate oF.wTT and optional oF.rdpTT?
> we probably need separate oF.wTT anyway, so this is not needed?

We would need oF.rdpTT, and oF.wTT requires prop_w_ to access them and assign them, so we still need this prop_w_?

I agree with the LLM and using CF is better here since CN has too much unused attrs, and their N_ access is not direct , which i don't think it's better.

boris-kz: just add separate oF.wTT, dTT will be use to sum call results

also simplify globals, similar to last version:

frame_2D_alg/agg_recursion.py
boris-kz: right

frame_2D_alg/agg_recursion.py
boris-kz: wM,wD,wi, wG,wI,wa, wL,wS,wA = 10, 10, 20, 20, 5, 20, 2, 1, 1  # dTT weights = reversed relative ave, update from wTT_ after feedback
    def make_T(nF, w):
        wTT = [CF(nF='wM',m=wM),CF(nF='wD',m=wD),CF(nF='wi',m=wi),CF(nF='wG',m=wG),CF(nF='wI',m=wI),CF(nF='wa',m=wa),
               CF(nF='wL',m=wL),CF(nF='wS',m=wS),CF(nF='wA',m=wA)] * w  # pseudo, make np.array?
        W = sum([F.w for F in wTT])
        return CF(nF=nF, w=W, wTT=wTT)
    def make_TT(nF, w):  # m,d forks
        return CF(nF=nF, wTT =np.array([make_T('mT',w),make_T('dT'),w]))
    wN,wC, wn,wc = 10,20,5,10  # sum from wTTN,wTTC,wTTn,wTTc, maps to Nt,Ct,Lt, no Bt: no call, no info?
    WN,WC, Wn,Wc = make_TT('wN',wN), make_TT('wC',wC), make_TT('wn',wn), make_TT('wc',wc)
    wTTN, wTTC, wTTn, wTTc = WN,WC, Wn,Wc
    wTT_ = CF(nF='wTT_', N_=[WN,WC,Wn,Wc])  # add sub-forks, as below?

kwcckw: Sorry I'm a bit confused here. Using oF.wTT means we will be using WN.wTT since Ws are globals, and we still need prop_w_ to access them from deeper nodes? Or they will be accessed from oF.wTT without prop_w_ first (local to oF) and their values will be updated to those nodes in feedback?

boris-kz: yes, we just need to form globals first, and there should be no single-variable CFs?

    wT = np.array([wM,wD,wi, wG,wI,wa, wL,wS,wA])
    wN,wC, wn,wc = 10,20,5,10  # fork weights
    wTTN, wTTC, wTTn, wTTc = np.array([wT,wT*avd])*wN, np.array([wT,wT*avd])*wC, np.array([wT,wT*avd])*wn, np.array([wT,wT*avd])*wc
    WN,WC, Wn,Wc = CF(nF=wN,wTT=wTTN,w=wN), CF(nF=wC,wTT=wTTC,w=wC), CF(nF=wn,wTT=wTTn,w=wn), CF(nF=wc,wTT=wTTc,w=wc) 
    wTT_ = CF(nF='wTT_', N_=[WN,WC,Wn,Wc])  # map to Nt,Ct,Lt, no Bt: no call, no info? add sub-forks:

kwcckw: I see, then attr_ is much simpler now. We actually only need weight from those single var CF, which can be a global var anyway?

boris-kz: right, revised for both forms:

    wT = np.array([wM,wD,wi, wG,wI,wa, wL,wS,wA])
    wN,wC, wn,wc = 10,20,5,10  # fork weights
    wTTN, wTTC, wTTn, wTTc = np.array([wT,wT*avd])*wN, np.array([wT,wT*avd])*wC, np.array([wT,wT*avd])*wn, np.array([wT,wT*avd])*wc
    wTT_ = [wTTN,wTTC, wTTn,wTTc]  # map to Nt,Ct,Lt, no Bt: no call, no info?
    WN,WC, Wn,Wc = CF(nF=wN,wTT=wTTN,w=wN), CF(nF=wC,wTT=wTTC,w=wC), CF(nF=wn,wTT=wTTn,w=wn), CF(nF=wc,wTT=wTTc,w=wc)
    WTT_ = CF(nF='wTT_', wTT=[WN,WC,Wn,Wc])  #  add sub-forks:

boris-kz: class CF(CBase):  
                 
                  # iF/ data: rim, Nt,Ct, Bt,Lt: ext|int- defined nodes, ext|int- defining links, Lt/Ft, Ct/lev, Bt/G
                  # oF/ code: N_ calls, H deeper call tree, Lt type tree, dTT=results, w=m or separate sum wTT?
    name = "fork"
    def __init__(f, **kwargs):
        super().__init__()
        f.N_ = kwargs.get('N_',[])  # flat
        f.H  = kwargs.get('H', [])  # CF levs / Nt||Ct
        f.nF = kwargs.get('nF','Nt')
        f.dTT= kwargs.get('dTT',np.zeros((2,9))); f.m, f.d, f.c, f.r = [kwargs.get(x,0) for x in ('m','d','c','r')]
        f.wTT= kwargs.get('wTT',np.zeros((2,9))); f.w, f.wc, f.wr = [kwargs.get(x,0) for x in ('w','wc','wr')]  # fork coefs, no wd?
        f.Lt = kwargs.get('Lt',[])  # +|-/ cross_comp, or type tree in oF
        f.fb_ = kwargs.get('fb_',[])
        f.typ = 0  # blocks sub_comp
        f.root = kwargs.get('root',None)

kwcckw: Since all vars are globals, we only WTT_ to update rdpTT?

frame_2D_alg/agg_recursion.py
boris-kz: probably, but we also need to add redundancy

frame_2D_alg/agg_recursion.py
boris-kz: >Since all vars are globals, we only WTT_ to update rdpTT?

Finer oFs need to be accumulated from individual calls before updating?

kwcckw: If we need to accumulate finer oFs, we actually need single var CFs? Or those finer oFs should be retrieved from nested N_ via WTT_?

frame_2D_alg/agg_recursion.py
boris-kz: >if we need to accumulate finer oFs, we actually need single var CFs? Or those finer oFs should be retrieved from nested N_ via WTT_?

yes, those that are in WTT_, maybe we need to add all single var weights there?

frame_2D_alg/agg_recursion.py
kwcckw: That redundancy should be from the local Nt or Lt?

frame_2D_alg/agg_recursion.py
kwcckw: Do you mean adding another set of single var weight sin WTT_? But that's redundant if we can retrieve.it in nested N_?

frame_2D_alg/agg_recursion.py
boris-kz: total r = local + Lt.r?

frame_2D_alg/agg_recursion.py
boris-kz: weights are in wTT now, N_ is the call sequence. we can add nesting in wTT

kwcckw: I see, so those weights will be in WTT.wTT?  And it's actually same as the prior wTT, with a size of (2.9) array?

frame_2D_alg/agg_recursion.py
kwcckw: But where do we get another local r here? G.r shouldn't be local right?

frame_2D_alg/agg_recursion.py
boris-kz: right

PR #98, opened 2026-03-29
3/29/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -130,11 +130,19 @@ def F_root(_F, fC=1):
...
-        if sub: F_root(F, fC)
+        if isinstance(sub, CF): F_root(F, fC)
 F_root(Z)
 def getF(oF, name):
     for fork in oF.Lt.N_:  # current + promoted fork types
-            if fork.nF==name: return fork
+        if fork.nF==name: return fork
+    # if not found, we need to promote it to higher Lt?
+    # But we need to accumulate enough calls before promoting it, which we still need to check for nested Lt?
+    for fork in oF.Lt.N_:
```
kwcckw, line 140: > we don't need nested Lts, if promoted their oFs will be in a higher Lt, else just ignore

But lower oFs won't be in higher Lt before any accumulation and feedback? For example, in comp_N_:

    getF(Z,'comp_N_').N_ += [(L_,rdpTT)]

We need to accumulate rdpTT first, and only promote them when it has enough calls during the feedback? So we probably need conditional nested Lt  to retrieve the oF and pack their calls?

frame_2D_alg/agg_recursion.py
```diff
@@ -648,7 +656,7 @@ def sum2G(ft_, root=None, init=1, typ=None):
     if G.Bt:
         Bt = G.Bt; bd,br = Bt.d,Bt.r
         if bd > avd*br*wn and N.typ!=1: cross_comp(Bt, br,'Bt')  # no ddfork
-        Bt.brrw = Bt.m * (root.root.m * (decay * (root.root.span/G.span)))  # external lend only, subtract from root?
+        if N_[0].typ: Bt.brrw = Bt.m * (root.root.m * (decay * (root.root.span/G.span)))  # external lend only, subtract from root?
```
kwcckw, line 659: When N is PP in trace_Edge, their root is tile and tile.root is empty, so we should just skip the borrowing here?

frame_2D_alg/agg_recursion.py
```diff
@@ -130,11 +130,19 @@ def F_root(_F, fC=1):
     for F in (_F.Lt.N_ if fC else _F.N_):  # fork types or calls
         setattr(F, 'croot' if fC else 'root', _F)
         sub = F.Lt if fC else F.N_
-        if sub: F_root(F, fC)
+        if isinstance(sub, CF): F_root(F, fC)
```
kwcckw, line 133: Sorry, this edit is wrong because sub could be F.Lt or F.N_.

So i think we should change `sub = F.Lt if fC else F.N_` because F.Lt could be an empty list, and it should be:

    sub = if isinstance(F.Lt, CF) if fC else F.N_

boris-kz: it's fine, this works for empty list too:

if sub: F_root(F, fC)

frame_2D_alg/agg_recursion.py
boris-kz: the whole oF.Ct forming, including promotion, should be in ffeedback,
before that we simply record calls

frame_2D_alg/agg_recursion.py
boris-kz: check if root.root: ?

frame_2D_alg/agg_recursion.py
boris-kz, line 98: we need separate Ct here, both for clarity and because Lt can be used for cross-call comps:

        f.Lt = kwargs.get('Lt',[])  # +|-/ cross_comp, including cross-call in oF
        f.Ct = kwargs.get('Ct',[])  # promoted types, in oF only?

kwcckw: So `f.Ct` is only exclusive to oF only where Ct.N_ are packing those promoted oFs?

frame_2D_alg/agg_recursion.py
kwcckw: But it skips empty oF? We need to assign root to those empty oF?

frame_2D_alg/agg_recursion.py
kwcckw: Yes, the problem is we need to retrieve the related call CF to record those calls, and we can't do that without searching for nested Lt.N_ ?

frame_2D_alg/agg_recursion.py
kwcckw: Yes that should work the same way.

frame_2D_alg/agg_recursion.py
```diff
@@ -130,11 +130,19 @@ def F_root(_F, fC=1):
...
 def getF(oF, name):
     for fork in oF.Lt.N_:  # current + promoted fork types
-            if fork.nF==name: return fork
+        if fork.nF==name: return fork
+    # if not found, we need to promote it to higher Lt?
+    # But we need to accumulate enough calls before promoting it, which we still need to check for nested Lt?
+    for fork in oF.Lt.N_:
+        for sub_fork in fork.Lt.N_:
+            if sub_fork.nF==name:
+                oF.Lt.N_ += [sub_fork]
```
boris-kz: for (sub_,V) in oF.N_:
if V > ave: oF.Ct.N_ += [(sub_[0].nF, V)]

frame_2D_alg/agg_recursion.py
boris-kz: right

frame_2D_alg/agg_recursion.py
boris-kz: why?

frame_2D_alg/agg_recursion.py
boris-kz: it has nothing to do with oF.Ct at this point

kwcckw: Yes, i was referring to the part where we record calls, for example:

     getF(Z,'comp_N_').N_ += [(L_,rdpTT)]

In the first call, `comp_N_` won't be in promoted Ct.N_, and it's still nested in `comp_`, we will never get anything to record the calls with `getF(Z,'comp_N_').N_ ` if we don't retrieve them recursively?

frame_2D_alg/agg_recursion.py
kwcckw: This is for root assignment, so `if sub` always return false for oFs since their Lt.c is empty before any accumulation.

frame_2D_alg/agg_recursion.py
boris-kz: we can promote them from lower-oF Cts in ffeedback, they should be filled as below, but recursive

frame_2D_alg/agg_recursion.py
boris-kz: all this should be in ffeedback, top-down, so we don't need F_root at all

frame_2D_alg/agg_recursion.py
boris-kz, line 127: so this Lt (Ct now) is just an guess about how it will look after ffeedback, 
it should really be built-up from clustered calls in Z.N_

frame_2D_alg/agg_recursion.py
kwcckw: Do you mean this section?

      for (oF_,V) in Z.N_:  # call sequence
            if V > ave: 
                for oF in oF_: getF(Z,oF.nF).Ct.N_ += [oF]  # add gain per oF?

But we need to increment V first, which still requires us to retrieve the correct (oF_,V) in Z.N_ in the first place?

frame_2D_alg/agg_recursion.py
kwcckw: Okay thanks, so this is not relevant now.

frame_2D_alg/agg_recursion.py
```diff
@@ -130,11 +130,19 @@ def F_root(_F, fC=1):
...
+        if isinstance(sub, CF): F_root(F, fC)
 F_root(Z)
 def getF(oF, name):
     for fork in oF.Lt.N_:  # current + promoted fork types
-            if fork.nF==name: return fork
+        if fork.nF==name: return fork
+    # if not found, we need to promote it to higher Lt?
+    # But we need to accumulate enough calls before promoting it, which we still need to check for nested Lt?
+    for fork in oF.Lt.N_:
+        for sub_fork in fork.Lt.N_:
```
boris-kz, line 141: def ffeedback(F):  # F:frame, adjust filters: all aves *= rV, ultimately differential backprop per ave?

    rTT_ = [np.ones((2,9)),np.ones((2,9)),np.ones((2,9)),np.ones((2,9))]
    # sum ratios between consecutive-level TTs:
    _N,_C,_n,_c = F.dTT, F.Ct.dTT, F.TTn, F.TTc
    for lev in F.H:
        # top-down frame expansion levels, not lev-selective or sub-lev recursive
        N,C,n,c = lev.dTT, lev.Ct.dTT, lev.TTn, lev.TTc
        rN, rC, rn, rc = [np.abs(_tt / eps_(tt)) for _tt,tt in zip((_N,_C,_n,_c), (N,C,n,c))]
        for rTT,rtt in zip(rTT_,(rN,rC,rn,rc)): rTT += rtt
        _N,_C,_n,_c = N,C,n,c
        for (oF_,V) in Z.N_:  # call sequence
            if V > ave:
                for oF in oF_:
                    fork = getF(Z, oF.nF)
                    if fork: fork.N_ += [oF]  # in type_, add gain per oF?
                    else:  # init fork
                        Z.Ct.N_ += [CF(nF = oF.nF,N_=[oF])]
        # cluster Z.Ct.N_ here
...

frame_2D_alg/agg_recursion.py
boris-kz: That V is formed in actual function calls, and Z.N_ is just a call trace.
then we pack Z.Ct.N_ from these calls by nF, as below

kwcckw: > so this Lt (Ct now) is just an guess about how it will look after ffeedback,
it should really be built-up from clustered calls in Z.N_

> That V is formed in actual function calls, and Z.N_ is just a call trace.
then we pack Z.Ct.N_ from these calls by nF, as below

I see, so the promoted Z.Ct.N_ should be built up, and before that Z.N_ should be just a guess of all oFs and it should be flat?

boris-kz: Z.N_ is an actual call trace

frame_2D_alg/agg_recursion.py
boris-kz: Actually, that Z is global, top-level only:

def ffeedback(F):  # F:frame, adjust filters: all aves *= rV, ultimately differential backprop per ave?
    
    for (oF_, V) in Z.N_:  # global call trace, top lev only?
        if V > ave:
            for oF in oF_:
                F = getF(Z, oF.nF)
                if F: F.N_ += [oF]  # in type_, add gain per oF?
                else: Z.Ct.N_ += [CF(nF=oF.nF, N_=[oF])]  # init fork
    # cluster Z.Ct.N_ here
...

kwcckw: > Z.N_ is an actual call trace

Sorry i'm a bit confused here. If Z.N_ is the actual call trace, and Ct will be added only in ffeedback, where do we get this F from `F = getF(Z, oF.nF)`?

boris-kz: That should be for Z.Ct.N_ only:

def ffeedback(F):  # F:frame, adjust filters: all aves *= rV, ultimately differential backprop per ave?

    for (nF, oF_,V) in Z.N_:  # global call trace, top lev only?
        if V > ave:
            for oF in oF_:
                F = getF(Z, nF)
                if F: F.N_ += [oF]  # in type_, add gain per oF?
                else: Z.Ct.N_ += [CF(nF=nF, N_=[oF])]  # init fork
    # cluster Z.Ct.N_ here
...
in the actual calls:

    # call trace in arg oF:
    oF.N_ += [CF(nF='comp_N_', root=oF, N_=L_, dTT=rdpTT)]  # error lowers source code weight
    TT,cm,rm, TTd,cd,rd = acc

kwcckw: I see, so Z.Ct.N_ should be packing all possible oFs other than those promoted oFs right?

boris-kz: all salient oFs, single copy. It's initially flat, should be clustered when complete

frame_2D_alg/agg_recursion.py
boris-kz, line 1: Actually, Z / oF.N_ is only the top level of call tree, deeper levels should be stored in H, same as for G.Nts.

Claude:
Let me check sum2G's H logic:The key differences I see are mostly about what's absent in oFs vs. data Gs:

**What's the same:** H is a list of levels, each level is a CF with N_ of same-depth elements, formed bottom-up via add_H. Root pointers trace upward. sum2G wraps ft_ (N_,L_,B_ with their TTs) into a G with combined forks.

**What's different for oFs:**

1. **No spatial params.** oFs have no yx, box, kern, span, angl, mang. So comb_Ft, add_Nt, add_Lt are mostly irrelevant — those are dominated by geometric bookkeeping.

2. **No L_ between oFs (yet).** Data Gs have L_ from cross-comp. oF nodes don't get cross-compared during tree building — that happens later in ffeedback clustering. So sum2G's Lt/Bt fork logic doesn't apply at tree-building time.

3. **dTT content is different.** For data, dTT is match/diff from cross-comp. For oFs, dTT would be rdpTT: relative prediction error of the call. The summing logic (weighted by c) is the same, but the semantic is inverted — high data dTT[0] is good, high oF dTT (error) is bad.

4. **Ct replaces Nt as the primary fork?** In data Gs, Nt holds the nodes. In the commented Z structure, Ct holds the type-clustered sub-forks (comp_, clust_, eval_, attr_). So the oF tree's branching structure maps to Ct rather than Nt.

So a simplified `sum2oF` would be essentially sum2F (lines 582-599) — just dTT/c/r aggregation and H formation via add_H — without the geometric parts. The fork structure (Nt/Lt/Bt/Ct) and cross-fork comp in comb_Ft would only come after clustering oFs in ffeedback, not during tree building. Does that match your thinking, or do you want fork differentiation during building too?

frame_2D_alg/agg_recursion.py
boris-kz: Claude: Here's the updated ffeedback with Nt-wrapped H and clustering sequence:

```python
def ffeedback(F):  # F:frame, adjust filters: all aves *= rV, ultimately differential backprop per ave?

    def sum_oH(oF_, Ft):  # bottom-up H from oF tree, Ft wraps result
        for oF in oF_:
            if oF.N_: sum_oH(oF.N_, oF)  # recurse to leaves first
        H = []
        for oF in oF_:
            if H: H[0] += oF.N_
            elif oF.N_: H = [list(oF.N_)]
            if oF.Nt.H: add_H(Ft.Nt.H, oF.Nt.H, Ft)
        if H: Ft.Nt.H += [Q2R(lev, froot=0) for lev in H]

    # wrap each oF.H in Nt for comp_H compatibility:
    def wrap_H(oF):
        if not oF.Nt: oF.Nt = CF(root=oF, nF='Nt')
        oF.Nt.H = oF.H if hasattr(oF,'H') and oF.H else []
        for child in oF.N_:
            if isinstance(child, CF): wrap_H(child)

    for oF in Z.N_: wrap_H(oF)
    if not Z.Nt: Z.Nt = CF(root=Z, nF='Nt')
    Z.Nt.H = []
    sum_oH(Z.N_, Z)
    # comp & cluster oFs by nF type -> Z.Ct:
    if not Z.Ct: Z.Ct = CF(root=Z, nF='Ct')
    nFd = {}  # group by nF
    for lev in Z.Nt.H:
        for oF in lev.N_:
            if oF.nF not in nFd: nFd[oF.nF] = []
            nFd[oF.nF] += [oF]
    Z.Ct.N_ = []
    for nF, oF_ in nFd.items():
        TT,c,r = sum_vt(oF_)
        tF = CF(nF=nF, N_=oF_, dTT=TT, c=c, r=r, root=Z.Ct)
        if len(oF_) > 1:  # cross-comp within type
            L_,lTT,lc,lr,_,_,_ = comp_N_(combinations(oF_,2), r)
            if L_: tF.Lt = CF(N_=L_, dTT=lTT, c=lc, r=lr)
        Z.Ct.N_ += [tF]
    # rTT_ from frame H:
    rTT_ = [np.ones((2,9)) for _ in range(4)]
    _N,_C,_n,_c = F.dTT, F.Ct.dTT, F.TTn, F.TTc
    for lev in F.H:
        N,C,n,c = lev.dTT, lev.Ct.dTT, lev.TTn, lev.TTc
        rN,rC,rn,rc = [np.abs(_tt / eps_(tt)) for _tt,tt in zip((_N,_C,_n,_c),(N,C,n,c))]
        for rTT,rtt in zip(rTT_,(rN,rC,rn,rc)): rTT += rtt
        _N,_C,_n,_c = N,C,n,c
    rM = rD = 0
        rm,rd = vt_(rTT, F.r, wTT_[i]); rM+=rm; rD+=rd
    return rM+rD, rTT_
```

Two things worth confirming:

**1. comp_N_ for oFs.** This calls base_comp which uses kern, angl, span etc. — oFs don't have meaningful values there. Should oF cross-comp just use comp_derT on dTT[1] (the rdpTT derivatives) instead of full base_comp?

**2. Feedback direction.** The rTT_ from F.H adjusts data weights. But Z.Ct now has per-type Lt from oF cross-comp — should those Lt matches/diffs also feed back into wTT_, adjusting code weights? That would close the loop between process error (Z) and filter adjustment.

kwcckw: > Actually, Z / oF.N_ is only the top level of call tree, deeper levels should be stored in H, same as for G.Nts.

Previously we had it as nested in N_ , and to pack it in H, it could be just Z.H? Where Z.N_ packed the flat version of calls and Z.H packed each layered of oFs or promoted oFs.

I think sum2F can't be applied directly here since Z.Ct is different and we need to retrieve those oFs from Z.Ct.N_ and pack it in different levels of Z.H?

boris-kz: Z.N_ is top level, Z.H is lower levels, same as for Nt.
Z.Ct is empty before clustering, so it's not relevant for initial packing

kwcckw: > Z.N_ is top level, Z.H is lower levels, same as for Nt. Z.Ct is empty before clustering, so it's not relevant

I see, so based on the actual call below:

> in the actual calls:
> 
> ```
> # call trace in arg oF:
> oF.N_ += [CF(nF='comp_N_', root=oF, N_=L_, dTT=rdpTT)]  # error lowers source code weight
> TT,cm,rm, TTd,cd,rd = acc
> return L_,TT,cm,rm/(cm or eps), TTd,cd,rd/(cd or eps)
> ```

That oF should be Z? And initially it's packed flat in Z.N_ and that should be updated to H in ffeedback?

But the clustering into H and N_ should have same nesting in Z.Ct? Which mean Z.Ct needs H too?

boris-kz: Z is a top oF, then we get one for each callee function
Clustering should probably be secondary to some light version of cross_comp in oF_N_
yes, oF.Ct.H is probably aligned with oF.H, similar to Nt

kwcckw: > Z is a top oF, then we get one for each callee function

But in comp_N_:

    oF.N_ += [CF(nF='comp_N_', root=oF, N_=L_, dTT=rdpTT)]  # error lowers source code weight

How do we get this oF if it's not Z? Or it will be created along each function call? And we need to pack that in Z.N_ in the runtime?

> Clustering should probably be secondary to some light version of cross_comp in oF_N_

Does this mean clustering is optional and we need compare those oFs first before grouping them into H?

boris-kz: yes, func name in each function call, the default is Z, pack as the above in comp_N_ 
something like 

    def getF(oF, name):
        for fork in oF.Ct.N_:  # promoted fork types only
            if fork.nF == name: return fork

    Z = sum2F(Z.N_,'Nt', Z).Nt  # -> Z.Nt with, sum_H nested calls, not redundant
    for oF in Z.N_:  # init Z.Ct.N_ as type_, not selective, in random order?
        F = getF(Z.Ct, oF.nF)
        if F: F.N_ += [oF]
        else: Z.Ct.N_ += [CF(nF=oF.nF, N_=[oF])]
    # draft:
    cross_comp(Z.Ct, frame.r+1)  # need a light version, with comp_derT and cluster_C only?
    # promote lower forks subcomp-> tFs
    rTT_ = [np.ones((2,9)),np.ones((2,9)),np.ones((2,9)),np.ones((2,9))]
    # sum ratios between consecutive-level TTs:
    _N,_C,_n,_c = frame.dTT, frame.Ct.dTT, frame.TTn, frame.TTc
    for lev in frame.H:
        # top-down frame expansion levels, not lev-selective or sub-lev recursive
        N,C,n,c = lev.dTT, lev.Ct.dTT, lev.TTn, lev.TTc
        rN, rC, rn, rc = [np.abs(_tt / eps_(tt)) for _tt,tt in zip((_N,_C,_n,_c), (N,C,n,c))]
        for rTT,rtt in zip(rTT_,(rN,rC,rn,rc)): rTT += rtt
        _N,_C,_n,_c = N,C,n,c
    rM = rD = 0
        rm, rd = vt_(rTT, frame.r, wTT_[i]); rM+=rm; rD+=rd
    return rM+rD, rTT_

kwcckw: So H will be added only in cross_comp of Z.Ct? Which means we need to parse Z.Nt to align their H? 

Also  all oFs should be unique in Z.N_ right?

boris-kz:
Yes, unique. Initial Ct.H maps to oF.H already, and cross_comp starts from
the top level

kwcckw:
> Yes, unique. Initial Ct.H maps to oF.H already, and cross_comp starts from the top level

If cross_conp should starts from the top level, that's mean the lower level H should be added in cross_comp?

PR #99, opened 2026-03-30
3/30/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -181,20 +181,37 @@ def TTw(G, wi=0): return wTT_[wi] * G.wTT_[wi]  # * secondary wweights, average=
...
+            if nF == 'Z': G_,r= cluster_C(Ft,E_,r)
+            else:
+                G_,r = cluster_N(Ft, E_, r)  # -> cluster_C, _P
+                if G_:
+                    Ft = sum2F(G_, nF, Ft.root); fC = G_[0].N_[0].typ==3  # not sure, G.N_= spliced C_ if more valuable?
+                    if val_(TT*wn, r, TTw(Ft.root,fC+2), (len(G_)-1)*Lw,1,TTd, rd/(r+rd)) > 0:
+                        G_,r = cross_comp(Ft,r,nF)  # agg+, trans-comp
     return G_, r  # G_ is recursion flag
 
+def comp_derT_(oF_, r, oF=Z):
```
kwcckw, line 200: For the cross_comp of oFs, i think we can create another 'Z' fork, and reuse the cross_comp with additional 'Z" fork and comp_N with full = 0 as below.

frame_2D_alg/agg_recursion.py
```diff
@@ -237,21 +254,22 @@ def proj_V(_N,N, dist, dy_dx, dec):  # _N x N induction
...
     if fall:
         pairs = product(C_,_C_) if _C_ else combinations(C_,r=2)  # comp between | within list
         for _C, C in pairs:
             if _C is C:
                 dtt = np.array([C.dTT[1],np.zeros(9)]); acc[0]+=dtt; acc[1]+=1; acc[2]+=1  # overlap=match
             else:
                 dy_dx = _C.yx-C.yx; dist = np.hypot(*dy_dx)
-                comp_N(_C,C, rr, A=dy_dx,span=dist,L_=L_,N_=N_,acc=acc)
+                Link = comp_N(_C,C, rr, A=dy_dx,span=dist,L_=L_,N_=N_,acc=acc)
+                pTT = comp_derT(_C.dTT[1], C.dTT[1]); rdpTT += np.abs(pTT-Link.dTT) / eps_(Link.dTT)
```
kwcckw, line 272: I think we need to add rdpTT for comp_C_ too? But they don't have projected TT, so that will be a direct comp_derT?

frame_2D_alg/agg_recursion.py
```diff
@@ -237,21 +254,22 @@ def proj_V(_N,N, dist, dy_dx, dec):  # _N x N induction
     for N in set(N_):
         if N.rim: Q2R(N.rim, N.Rt, merge=0, froot=0)
     # call trace:
-    oF.N_ += [CF(nF='comp_N_',root=oF, N_=L_, dTT=rdpTT)]  # error lowers source code weight
+    oF.N_ += [CF(nF='comp_N_',root=oF, N_=L_, c=len(L_), dTT=rdpTT)]  # error lowers source code weight
```
kwcckw, line 257: We might have multiple comp_N_ from both agg+ and sub+ (same with all the other oFs), and since they should be unique, do we need to check the existing oF with nF == 'comp_N_' and add the related params only? Or they are all unique because they are added from different composition, which we need some additional params (such as nest or level) to represent them?

frame_2D_alg/agg_recursion.py
```diff
@@ -986,11 +1006,11 @@ def getF(oF, name):
 
     sum2F(Z.N_,'Nt', Z)  # fset Nt, sum nested calls in H
     for oF in Z.N_:  # init Z.Ct.N_= type_, not selective, in random order?
-        F = getF(Z.Ct, oF.nF)
-        if F: F.N_ += [oF]
-        else: Z.Ct.N_ += [CF(nF=oF.nF, N_=[oF])]
+        F = getF(Z, oF.nF)
+        if F: F.N_ += [oF]; F.m += oF.m; F.c += oF.c; F.dTT += oF.dTT
```
boris-kz, line 1010: this will be Q2R after all appends

frame_2D_alg/agg_recursion.py
boris-kz: we need a new version of cross_comp

kwcckw: Okay, but for comp_derT of oFs, we can create a new func here and call comp_N with full == 0?

frame_2D_alg/agg_recursion.py
kwcckw: We may need to pack all the F here and run Q2R after the oF loops since not all oF in Z.Ct.N_ has added oF.

frame_2D_alg/agg_recursion.py
boris-kz: I think we need to rewrite most of them

frame_2D_alg/agg_recursion.py
boris-kz: they will have different root oF, and we need to assign root G to it, definiting the data scope

frame_2D_alg/agg_recursion.py
boris-kz: rdpTT is the only thing we need to compare here?

frame_2D_alg/agg_recursion.py
boris-kz: that's what I said

frame_2D_alg/agg_recursion.py
kwcckw: Okay, so Link.dTT is actually the rdpTT here and there's no need to compare with pTT again here?

frame_2D_alg/agg_recursion.py
kwcckw: Okay, that's mean we might need to preinit those oF (for example, comp_N_) before packing it into Z, so that we can parse them as root to those sub functions (such as comp_N, comp_F, etc).

frame_2D_alg/agg_recursion.py
kwcckw: For comp, it should be comp_derT between those oFs, and i think we still need link between them for centroid clustering?

After comp, the centroid clustering process should be the same for oFs to group those similar oFs?

frame_2D_alg/agg_recursion.py
boris-kz: yes this comp_derT is just for clustering, by the gain in predictive value

frame_2D_alg/agg_recursion.py
boris-kz: use arg root as data root?

oF.N_ += [CF(nF='comp_N_',root=(oF,root), N_=L_, dTT=rdpTT)]

frame_2D_alg/agg_recursion.py
boris-kz: I think this clustering is just sorting by gain at this point, we don't even need comparison? 
Claude:
oFs are already typed by nF, and the main feedback is adjusting weights by their rdpTT. Cross-comp between oFs of the same type would tell you how much their prediction errors vary, but that's second-order and premature.

frame_2D_alg/agg_recursion.py
kwcckw: So each oF in different composition should have different data root, but it's better to parse the actual oF root and assign them as root? I think we will need that to pack them into H later.

frame_2D_alg/agg_recursion.py
kwcckw: But how exactly we group those similar oFs by sorting gain? Sorting is just arranging them in order, and clustering needs more than that?

boris-kz: this is for promotion to higher oF.Ct.N_, simply promote if gain>ave

frame_2D_alg/agg_recursion.py
boris-kz: oF should be processed differently anyway, we probably need typ=4 or something

frame_2D_alg/agg_recursion.py
```diff
@@ -460,7 +480,7 @@ def trans_cluster(G):
         # clustering value = selectivity or loss reduction vs root.Lt: all comps, add dval?
         selV = (Ft.m - Ft.root.Lt.m) * (Ft.root.Lt.c - Ft.c)  # select in G.Lt
         # not updated:
-        Z.N_+= [('cluster_N', G_,selV)]  # feedback
+        Z.N_+= [CF(nF='cluster_N',root=Z, N_=G_, c=len(G_), m=selV)]  # feedback
```
boris-kz, line 483: sel_TT = Ft.dTT*Ft.c - Ft.root.Lt.dTT* Ft.root.Lt.c  # ?
        Z.N_+= [CF(nF='cluster_N',root=(oF,Ft.root), N_=G_, dTT=sel_TT)]  # feedback

frame_2D_alg/agg_recursion.py
boris-kz: ok, that data root = CF.N_[0].root

frame_2D_alg/agg_recursion.py
boris-kz: The sel_TT idea is right — clustering value is selectivity: how much the clustered subset improves over the full set of comparisons. But a few issues:

**1. root is a tuple.** `root=(oF, Ft.root)` — CF.root should be a single reference (caller oF), not a pair. Ft.root is the data context, which is already recoverable from G_ contents. Keep it:

```python
Z.N_ += [CF(nF='cluster_N', root=oF, N_=G_, dTT=sel_TT)]
```

**2. sel_TT sign/meaning.** `Ft.dTT*Ft.c` is total clustered value, `Ft.root.Lt.dTT*Ft.root.Lt.c` is total comparison value (all links). The difference is positive when clustering compresses well — removes low-match links. But this is absolute, not relative. For feedback you probably want relative:

```python
sel_TT = (Ft.dTT*Ft.c - Ft.root.Lt.dTT*Ft.root.Lt.c) / eps_(Ft.root.Lt.dTT*Ft.root.Lt.c)
```

This gives you the fractional gain per attribute, same form as rdpTT in comp_N_.

**3. Consistency with comp_N_ trace.** comp_N_ stores `rdpTT` (prediction error ratio), cluster_N would store `sel_TT` (selectivity ratio). Both are (2,9) relative measures, but semantically different — one is comp accuracy, the other is clustering gain. They get summed in sum2F though. Is that intended, or should they be in separate dTT slots (like m row vs d row)?

kwcckw: This changes:

    sel_TT = Ft.dTT*Ft.c - Ft.root.Lt.dTT* Ft.root.Lt.c  # ?
    Z.N_+= [CF(nF='cluster_N',root=(oF,Ft.root), N_=G_, dTT=sel_TT)]  # feedback

Shouldn't be the same as the original selV computation? Because (a-b) * (c-d) is different with (a*c) - (b*d)? The final value will be either scaled to larger or lower value. So it's more accurate with the difference after the scaling?

frame_2D_alg/agg_recursion.py
kwcckw: > ok, that data root = CF.N_[0].root

Yes, so we actually don't need that data root too. We should parse the actual oF root, which we need to create them first.

frame_2D_alg/agg_recursion.py
kwcckw: I see, if this is just promotion, we don't group them by similarity? Or that will be another separated process later?

frame_2D_alg/agg_recursion.py
boris-kz: probably this:
`sel_TT = (Ft.dTT*Ft.c - Ft.root.Lt.dTT*Ft.root.Lt.c) / eps_(Ft.root.Lt.dTT*Ft.root.Lt.c)`

frame_2D_alg/agg_recursion.py
boris-kz: it's just a caller name?

frame_2D_alg/agg_recursion.py
boris-kz: oF similarity only exist if their code differs, that's after code modification

frame_2D_alg/agg_recursion.py
boris-kz, line 1: Claude: 
Let me check all the conditionally called functions:Now let me map all conditionally called functions and their gain metrics. Here's the draft — additions only, showing what to insert at each call site:

```python
# cross_comp: add _oF param, create oF, pass to children
def cross_comp(Ft, rr, nF='Nt', _oF=None):
    oF = CF(nF='cross_comp', root=_oF, dTT=np.zeros((2,9)))
    if _oF: _oF.N_ += [oF]
    else:   Z.N_ += [oF]  # top-level call
    ...
    L_, TT,c,r,TTd,cd,rd = comp_C_(N_,rr,fC=1,_oF=oF) if fC else comp_N_(combinations(N_,2),rr,_oF=oF)
    ...
            G_,r = cluster_N(Ft, E_, r, _oF=oF)
    ...
                G_,r = cross_comp(Ft,r,nF, _oF=oF)  # agg+ recursion
    # gain: output G_ vs input N_
    if G_: oF.dTT = (Ft.dTT*Ft.c - TT*c) / eps_(TT*c)
    return G_, r

# comp_N_: add _oF param
def comp_N_(_pairs, r, tnF=None, rL=None, _oF=None):
    ...
    # existing rdpTT computation
    oF = CF(nF='comp_N_', root=_oF, N_=L_, dTT=rdpTT)
    if _oF: _oF.N_ += [oF]
    else:   Z.N_ += [oF]
    ...  # remove old getF(Z,'comp_N_') line

# comp_C_: add _oF param
def comp_C_(C_, rr, _C_=[], fall=1, fC=0, _oF=None):
    ...
    oF = CF(nF='comp_C_', root=_oF, N_=L_, dTT=TTm)
    if _oF: _oF.N_ += [oF]
    else:   Z.N_ += [oF]
    ...

# cluster_N: add _oF param
def cluster_N(Ft, _N_, r, _oF=None):
    ...
    sel_TT = (Ft.dTT*Ft.c - Ft.root.Lt.dTT*Ft.root.Lt.c) / eps_(Ft.root.Lt.dTT*Ft.root.Lt.c)
    oF = CF(nF='cluster_N', root=_oF, N_=G_, dTT=sel_TT)
    if _oF: _oF.N_ += [oF]
    else:   Z.N_ += [oF]
    ...  # remove old getF(Z,'cluster_N') line
    return G_, r

# cluster_C: add _oF param
def cluster_C(Ft, E_, r, _oF=None):
    ...
    if oC_:
        ...
        sel_TT = (Ct.dTT*Ct.c - Ft.dTT*Ft.c) / eps_(Ft.dTT*Ft.c)
        oF = CF(nF='cluster_C', root=_oF, N_=oC_, dTT=sel_TT)
        if _oF: _oF.N_ += [oF]
        else:   Z.N_ += [oF]
        ...
            cross_comp(Ct, r, _oF=oF)  # pass oF down
    ...  # remove old getF(Z,'cluster_C') line
    return oC_, r

# cluster_P: add _oF param, called from cluster_C
def cluster_P(_C_, N_, root, _oF=None):
    ...
    # gain: convergence delta
    oF = CF(nF='cluster_P', root=_oF, N_=[c for c in out_ if c], dTT=np.zeros((2,9)))
    if _oF: _oF.N_ += [oF]
    else:   Z.N_ += [oF]
    return [c for c in out_ if c]

# sum2G: conditional sub+ calls, add _oF param
def sum2G(ft_, root=None, init=1, typ=None, _oF=None):
    ...
    if G.Lt:  # sub+
        ...
        if V > 0:
            oF = CF(nF='sum2G_sub', root=_oF, dTT=np.zeros((2,9)))
            if _oF: _oF.N_ += [oF]
            else:   Z.N_ += [oF]
            ...
            G_,r = cluster_C(G.Nt,E_,r,_oF=oF) if ... else cluster_N(G.Nt,E_,r,_oF=oF)
            ...
                cross_comp(G.Nt,r,'Nt',_oF=oF)
    if G.Bt:
        ...
        if bd > avd*br*wn and N.typ!=1:
            cross_comp(Bt, br,'Bt', _oF=_oF)  # Bt fork, same parent
    return G
```

And the callers in `vect_edge` and `frame_H` init the top-level oFs:

```python
# vect_edge, trace_edge calls:
    cross_comp(T.Nt, T.r, _oF=None)  # already handled: _oF=None -> Z.N_

# frame_H:
    if cross_comp(F.Nt, rr=elev, _oF=None)[0]:  # top-level
```

Two open questions:

**1. cluster_P gain metric.** It doesn't have a clean before/after TT like cluster_N/C. The gain is convergence — `dM` from the while loop. Should `oF.dTT` just be the final C_ dTT, or the delta from last iteration?

**2. comp_N (single pair).** It's called many times per comp_N_ loop. Tracing each individual comp_N would bloat Z.N_ enormously. I left it out — the comp_N_ oF with rdpTT already summarizes. Should comp_N get its own trace, or is comp_N_ granularity sufficient?

frame_2D_alg/agg_recursion.py
```diff
@@ -986,11 +1006,11 @@ def getF(oF, name):
 
     sum2F(Z.N_,'Nt', Z)  # fset Nt, sum nested calls in H
```
boris-kz, line 1007: def ffeedback(frame):  # adjust filters: all aves *= rV, ultimately differential backprop per ave?

    def getF(typ_, name):
        for fork in typ_:  # current + promoted fork types per level
            if fork.nF==name: return fork
    def init_C_(N_):
        typF_ = []  # init Ct.N_: all calls per type per level, in random order?
        for F in N_:
            typF = getF(typF_, F.nF)
            if typF: typF.N_ += [F]  # calls / function type
            else:    Z.Ct.N_ += [CF(nF=F.nF, N_=[F])]  # add new typF
        return typF_

    sum2F(Z.N_,'Nt', Z)  # fset Nt, sum nested calls in H
    for N_ in Z.H: Z.Ct.H += [init_C_(N_)]  # Z.Ct.H maps to Z.H
    _H = []
    for lev in Z.Ct.H:  # bottom-up
        _H += [Q2R(typ_) for typ_ in lev]  # may be empty list?
    H, _prom_ = [],[]
    r = frame.r + 1  # tentative
    for _lev in _H:  # select and promote high-gain forks, bottom-up
        lev = [typ for typ in _lev if typ.m > ave*r]  # bottom lev
        H += [Q2R(lev, Z.Ct) if lev else None]  # lev.r += r?
        r+=1  # increase for promotion
        prom_ = [typ for typ in lev+_prom_ if typ.m > ave*r]
        _prom_ = prom_
    Z.Ct.N_ = Q2R([typ_ for typ_ in init_C_(Z.N_+_prom_)])  # top lev + promoted types
    Z.Ct.H = H
    rTT_ = [np.ones((2,9)),np.ones((2,9)),np.ones((2,9)),np.ones((2,9))]
    # sum ratios between consecutive-level TTs:
    _N,_C,_n,_c = frame.dTT, frame.Ct.dTT, frame.TTn, frame.TTc
    for lev in frame.H:
        # top-down frame expansion levels, not lev-selective or sub-lev recursive
        N,C,n,c = lev.dTT, lev.Ct.dTT, lev.TTn, lev.TTc
        rN, rC, rn, rc = [np.abs(_tt / eps_(tt)) for _tt,tt in zip((_N,_C,_n,_c), (N,C,n,c))]
        for rTT,rtt in zip(rTT_,(rN,rC,rn,rc)): rTT += rtt
        _N,_C,_n,_c = N,C,n,c
    rM = rD = 0
        rm, rd = vt_(rTT, frame.r, wTT_[i]); rM+=rm; rD+=rd
    return rM+rD, rTT_

boris-kz: Claude: Issues I see:

**1. init_C_ returns typF_ but appends to Z.Ct.N_.** The function builds `typF_` locally but never returns it — it appends to `Z.Ct.N_` instead. Then returns `typF_` which is always empty since new types go to `Z.Ct.N_`. Should be:

```python
    def init_C_(N_):
        typF_ = []
        for F in N_:
            typF = getF(typF_, F.nF)
            if typF: typF.N_ += [F]
            else:    typF_ += [CF(nF=F.nF, N_=[F])]  # not Z.Ct.N_
        return typF_
```

**2. Z.H structure mismatch.** `sum2F(Z.N_,'Nt', Z)` builds `Z.Nt.H`, not `Z.H`. Then `for N_ in Z.H` iterates the wrong thing. Should be:

```python
    for lev in Z.Nt.H: Z.Ct.H += [init_C_(lev.N_)]  # lev is CF with N_
```

**3. Z.Ct not initialized.** Same issue as before — `Z.Ct` defaults to `[]`, so `Z.Ct.H`, `Z.Ct.N_` will fail. Need `Z.Ct = CF(root=Z, nF='Ct')` at Z init or top of ffeedback.

**4. Q2R on init_C_ result.** `Z.Ct.N_ = Q2R([typ_ for typ_ in init_C_(Z.N_+_prom_)])` — Q2R expects a list of CN/CF with `.c`, `.dTT`, etc. But init_C_ returns a list of CFs, and wrapping in another list makes it `[[CF,...]]`. Should be:

```python
    Z.Ct.N_ = init_C_(Z.N_ + _prom_)  # already a list of typFs
    Q2R(Z.Ct.N_, Z.Ct, froot=0)  # sum into Z.Ct
```

**5. Promotion accumulates across levels but re-filters same types.** `_prom_` carries forward, but `lev+_prom_` mixes current level types with promoted ones — a promoted type could appear twice if it exists at multiple levels. Need dedup or merge:

```python
        for typ in _prom_:
            existing = getF(lev, typ.nF)
            if existing: existing.N_ += typ.N_  # merge
            else: lev += [typ]
```

frame_2D_alg/agg_recursion.py
kwcckw: Okay, thanks, then we really just need the promotion here since we don't have code modification yet.

frame_2D_alg/agg_recursion.py
kwcckw: Yes, but we need the reference of root oF? Because there might be multiple calls with the same caller name?

frame_2D_alg/agg_recursion.py
kwcckw: > probably this: `sel_TT = (Ft.dTT*Ft.c - Ft.root.Lt.dTT*Ft.root.Lt.c) / eps_(Ft.root.Lt.dTT*Ft.root.Lt.c)`

Actually why we are not scale the result by the sum of both Cs?

frame_2D_alg/agg_recursion.py
kwcckw: For the first problem, is that a typo?

      def init_C_(N_):
          typF_ = []  # init Ct.N_: all calls per type per level, in random order?
          for F in N_:
              typF = getF(typF_, F.nF)
              if typF: typF.N_ += [F]  # calls / function type
              else:    Z.Ct.N_ += [CF(nF=F.nF, N_=[F])]  # add new typF
          return typF_

`typF_` is init empty but we are getting `typF` from `typF_` again, which will never return anything？We packing new `typF_` to `Z.Ct.N_`, and `typF_` is always empty here? So we need to pack typF into both `typF_` and `Z.Ct.N_`? 


For the second problem:

    for N_ in Z.H: Z.Ct.H += [init_C_(N_)]  # Z.Ct.H maps to Z.

init_C_(N_) return typF_, which is a list, so Z.Ct.H packs nested list now?

frame_2D_alg/agg_recursion.py
kwcckw: Yes, i think we need something like this to init current function's oF first, and parse it as oF root to the calling function.

frame_2D_alg/agg_recursion.py
boris-kz: right, need to add oF.N_, that's the data scope

frame_2D_alg/agg_recursion.py
boris-kz: that's not comparable to rdpTT?

frame_2D_alg/agg_recursion.py
boris-kz: edited:

    def getF(typ_, name):
        for fork in typ_:  # current + promoted fork types per level
            if fork.nF==name: return fork
    def init_C_(N_):
        typ_ = []  # group calls by nF type
        for oF in N_:
            tF = getF(typ_, oF.nF)
            if tF: tF.N_ += [oF]  # calls / function type
        return typ_

    sum2F(Z.N_,'Nt', Z)  # fset Z.Nt, sum nested calls in H
    Z.Ct.H = [init_C_(lev.N_) for lev in Z.Nt.H]  # maps to Z.H
    H, _prom_ = [],[]; r = frame.r+1
    for lev in Z.Ct.H:  # bottom-up: select and promote high-gain types
        for typ in _prom_:  # merge lower typ_s
            _typ = getF(lev, typ.nF)
            if _typ: Q2R(list(set(_typ.N_+typ.N_)), _typ, merge=0,froot=0)
            else:    lev += [typ]
        H+= [Q2R([tF for tF in lev if tF.m > ave*r], froot=0) if lev else CF()]
        _prom_ = [tF for tF in lev if tF.m > ave*(r+1)]  # promote to next level
        r += 1
    Z.Ct.H = H  # lower levs
    Z.Ct.N_ = init_C_(Z.N_ + [oF for tF in _prom_ for oF in tF.N_])  # top lev
    Q2R(Z.Ct.N_, Z.Ct, froot=0)
    # sum ratios between consecutive-level TTs:
    _N,_C,_n,_c = frame.dTT, frame.Ct.dTT, frame.TTn, frame.TTc; rTT_ = [np.ones((2,9)) for _ in range(4)]
    for lev in frame.H:
        # top-down frame expansion levels, not lev-selective or sub-lev recursive
        N,C,n,c = lev.dTT, lev.Ct.dTT, lev.TTn, lev.TTc
        rN, rC, rn, rc = [np.abs(_tt / eps_(tt)) for _tt,tt in zip((_N,_C,_n,_c), (N,C,n,c))]
        for rTT,rtt in zip(rTT_,(rN,rC,rn,rc)): rTT += rtt
        _N,_C,_n,_c = N,C,n,c
    rM = rD = 0
        rm, rd = vt_(rTT, frame.r, wTT_[i]); rM+=rm; rD+=rd
    return rM+rD, rTT_

frame_2D_alg/agg_recursion.py
boris-kz: Right.
Claude review:

Z is a valued AST: same tree structure as the code's call graph, but each node carries execution-specific dTT (gain/error per attribute), c (data scope), and r (redundancy). Regular AST is static structure; Z is that structure filled with empirical results from a specific input.

The key differences from a standard AST:

**What Z has that AST doesn't:**
- dTT per node: quantified gain of each call on actual data
- H levels: depth-grouped aggregation, enabling cross-level comparison
- Ct type-clustering with promotion: functions grouped by operational similarity, not just call position

**What AST has that Z doesn't (yet):**
- Full branching structure (Z only traces conditionally executed paths, not skipped branches)
- Parameter signatures (Z nodes have nF names but not the actual args that selected the branch)

The second point is interesting — a skipped branch (condition failed, function not called) has zero gain but also zero cost. Currently Z has no record of those. For feedback, knowing that a branch was evaluated but rejected (val_ < threshold) could be as informative as knowing what executed. That would be negative entries: cost of evaluation without gain.

The data scope aspect is what makes this more than profiling — each oF's N_ contains the actual L_ or G_ produced, so you can trace which data patterns drove which code paths. That's the basis for the self-modification loop: Z cross-comp discovers which code patterns correlate with which data patterns.

frame_2D_alg/agg_recursion.py
kwcckw: For this merging part:
      
      for typ in _prom_:  # merge lower typ_s
              _typ = getF(lev, typ.nF)
              if _typ: Q2R(list(set(_typ.N_+typ.N_)), _typ, merge=0,froot=0)
              else:    lev += [typ]

It's a bit unique because we shouldn't remove the lower level `typ` from lower level if they are merged to higher level `_typ`?

boris-kz: how is it removed, this from `_prom_`, it doesn't affect the lower level?

frame_2D_alg/agg_recursion.py
kwcckw: Okay right, ` eps_(Ft.root.Lt.dTT*Ft.root.Lt.c)` should be fine here, thanks.

frame_2D_alg/agg_recursion.py
kwcckw: > Parameter signatures 

Do we need something like that? A method to directly select a fork or a oF from the nested Z? I think we only can trace them step by step via root and nF now?

frame_2D_alg/agg_recursion.py
kwcckw: Yes, the lower level remained the same, i meant we shouldn't remove it right? Since usually we remove the merged element.

boris-kz: it's not removed here

frame_2D_alg/agg_recursion.py
boris-kz: That would be added to function call eval process?

Right — Z collects gains but doesn't feed them back into the val_/vt_ calls that gate each function. The loop would be: Z.Ct.N_ has per-type aggregate gain (typF.dTT), which should adjust the threshold for that function's eval.

Currently, val_ uses static `ave*r` or `ave*wN` etc. The feedback would multiply those by a type-specific weight derived from the typF's dTT — high-gain functions get lower thresholds (called more), low-gain ones get higher thresholds (called less).

Something like, in the rTT_ section of ffeedback:

```python
    for typF in Z.Ct.N_:
        rg = typF.m / eps_(typF.c)  # gain rate per type
        typF.w = rg / (ave * r)  # relative to baseline, >1: lower threshold, <1: higher
```

Then at each call site, the eval references the type weight:

```python
    # in cross_comp, before calling cluster_N:
    cw = getF(Z.Ct.N_, 'cluster_N')
    if M * ((len(L_)-1)*Lw) * wn * (cw.w if cw else 1) > ave:
```

Or more uniformly, a single function that wraps val_:

```python
def oval_(nF, TT, r, wTT, **kw):  # oF-weighted val_
    cw = getF(Z.Ct.N_, nF)
    w = cw.w if cw else 1
    return val_(TT, r/w, wTT, **kw)  # lower r = lower threshold for high-gain types
```

But this means every conditional call needs to know its own nF string. Is that acceptable, or do you want the weight lookup tied to the _oF that's already being threaded through?

kwcckw: Yes, this should be included, but when we have multiple Hs, how do we decide if we need to select oF of 'cluster_N' from N_ or H?

boris-kz: It's a single branching H, root.H is simply merged N.Hs, which are more specific.
So please add that arg oF to the functions, at least the main ones

kwcckw: Alright, I guess  we need that oF for most of the functions, except those eval based function such as val_ or vt_.

PR #100, opened 2026-03-31
3/31/2026

kwcckw:


kwcckw:
From the draft of call tree, looks like we still need eval based oFs such as vt_ or val_, and that's actually requires additional oFs in their calling function such as sum2G, Q2R, add_Nt, add_Lt, etc. So we need to add oF for all of them?

frame_2D_alg/agg_recursion.py
```diff
@@ -235,23 +241,24 @@ def proj_V(_N,N, dist, dy_dx, dec):  # _N x N induction
                 # ~= links in clustering
         else: break  # beyond initial induction range, re-sort by proj_V?
     for N in set(N_):
-        if N.rim: Q2R(N.rim, N.Rt, merge=0, froot=0)
+        if N.rim: Q2R(N.rim, N.Rt, merge=0, froot=0,_oF=oF)
     # call trace:
-    oF.N_ += [CF(nF='comp_N_',root=oF, N_=L_, dTT=rdpTT)]  # assign data root for scope, error lowers source code weight
```
kwcckw, line 240: I think root `_oF.N_` should be packing `oF` ? Then maybe `L_` should be packed in `oF.L_` since `oF.N_` should be packing deeper level's oFs only?

frame_2D_alg/agg_recursion.py
```diff
@@ -130,13 +130,15 @@ def __bool__(f): return bool(f.c)  # N_ may be empty?
 '''
 Z = CF(nF ='Z', Ct=CF(nF='Ct'))  # global meta code, Z.N_= call trace, root=caller, Ct: function sub-types in ffeedback
 
-def vt_(TT, r, wTT=wTTn):  # brief val_ to get m,d, rc=0 to return raw vals, Wn for comp_N
+def vt_(TT, r, wTT=wTTn, _oF=None):  # brief val_ to get m,d, rc=0 to return raw vals, Wn for comp_N
 
+    oF = CF(nF='vt_', root=_oF); _oF.N_ += [oF]
```
boris-kz, line 135: so we are adding such fork with each function call, that can be a factory function in CF?

frame_2D_alg/agg_recursion.py
boris-kz: that's a default per call, so it should be in the caller, not the callee?

frame_2D_alg/agg_recursion.py
boris-kz: Let me look at the current code to understand the CF structure and how function calls work.Now I see the pattern clearly. Every function has this boilerplate:

```python
oF = CF(nF='func_name', root=_oF); _oF.N_ += [oF]
```

Two options:

**1. Factory on CF** — a classmethod or staticmethod:
```python
@staticmethod
def call(name, _oF):
    oF = CF(nF=name, root=_oF); _oF.N_ += [oF]
    return oF
```
Usage: `oF = CF.call('vt_', _oF)`

**2. Decorator / meta-function** — wraps any function definition, auto-creates oF from function name:
```python
def traced(func):
    def wrapper(*args, _oF=None, **kwargs):
        oF = CF(nF=func.__name__, root=_oF); _oF.N_ += [oF]
        return func(*args, _oF=oF, **kwargs)
    return wrapper
```
Then `_oF` disappears from each function's signature and body — the decorator handles it. But the inner function now receives `_oF` which is actually the *new* `oF` (the child), and needs to pass it down as `_oF` to callees.

Wait — there's a subtlety. In your current code, some functions pass `_oF=oF` to callees (e.g., `sum_vt` passes `_oF=oF` to `vt_`), while the function itself uses `oF` locally. The decorator approach would rename `_oF` in the signature to mean the *local* oF, which changes semantics.

A cleaner decorator:
```python
def traced(func):
    def wrapper(*args, _oF=None, **kwargs):
        oF = CF(nF=func.__name__, root=_oF); _oF.N_ += [oF]
        return func(*args, _oF=oF, **kwargs)  # callee sees its _oF = caller's oF
    return wrapper
```
This works if every function's `_oF` param is only used to create its own `oF` and pass down — which is your current pattern. The decorator replaces the two-line boilerplate, and the decorated function no longer needs the `oF = CF(...)` line. But `_oF` in the decorated function body would now already be the local oF.

Issue: some functions like `cross_comp` also set `oF.dTT` in the CF constructor. The decorator would need kwargs for that:
```python
def traced(func):
    def wrapper(*args, _oF=None, **kw):
        oF = CF(nF=func.__name__, root=_oF, **{k:kw.pop(k) for k in list(kw) if k.startswith('oF_')})
        _oF.N_ += [oF]
        return func(*args, _oF=oF, **kw)
    return wrapper
```
That's getting ugly. The `cross_comp` case with `dTT=np.zeros((2,9))` is the exception though.

**My recommendation**: option 1, `CF.call(name, _oF)` — minimal, no magic, one line instead of two, keeps explicit control for exceptions like `cross_comp` where you init extra attrs on oF. The decorator saves more boilerplate but obscures the call-tree construction and makes exceptions harder.

Which direction do you prefer, or do you want a hybrid (decorator for simple cases, explicit for complex ones)?

boris-kz: Good question. The key distinction: iF (data forks) accumulate `dTT` from comparisons of their `N_`, while oF (code forks) accumulate `dTT` from *results* of function calls in their `N_`. Same structure, different semantics for what N_ contains and how dTT is produced.

Arguments for a separate class:

1. oF doesn't need `Lt`, `Ct` in the same sense — `Lt` as cross-comp links between calls is speculative, `Ct` as sub-types is different from data Ct
2. oF.N_ contains call instances (other oFs), not CN nodes — type mixing in N_ is messy
3. `wTT` in oF means process weight (how much to invest in this function), vs. iF where it means attribute weight — same name, different semantics
4. A leaner oF class avoids carrying dead iF attrs (`kern`, `span`, `mang`, etc. don't apply)

Arguments against:

1. The whole point of the architecture is uniform cross-comp and clustering across levels — if you eventually cross-comp and cluster *code elements* the same way as data elements, they need the same interface
2. `dTT`, `m`, `d`, `c`, `r`, `root`, `N_`, `H`, `wTT`, `fb_` are shared — that's most of CF already
3. A subclass would preserve uniformity while allowing oF-specific attrs

So: **subclass**.

```python
class oF(CF):  # code fork: function call trace
    name = "call"
    def __init__(f, **kwargs):
        super().__init__(**kwargs)
        # drop iF-only defaults, add oF-specific:
        f.rd = kwargs.get('rd', 0)  # rdpTT credit
```

With `CF.call` factory:
```python
@staticmethod
def call(name, _oF, cls=None, **kw):
    o = (cls or oF)(nF=name, root=_oF, **kw)
    _oF.N_ += [o]; return o
```

This keeps cross-comp/clustering uniform (oF inherits the interface), while making the code/data distinction explicit in type. And `isinstance(f, oF)` cleanly separates them where needed.

frame_2D_alg/agg_recursion.py
kwcckw: Yes, but the caller `_oF.N_` should be packing nested `oF`s only? While the data `L_` should be packed separately?

frame_2D_alg/agg_recursion.py
kwcckw: The second method is a built-in wrapper, but we still need to define it before each function, so i guess we can continue with the first method since both methods requires us to include some code in each function.

For the first method, we can control to accumulate dTT from comparison or the result of clustering or eval, which is better in our use case?

boris-kz: the whole convo:  https://claude.ai/share/01497a32-7c62-495e-8156-853750c6499d

kwcckw: Okay, the new class should be more precise, but actually the existing CF can do the same thing too?

And the static caller method:

      @staticmethod
      def call(typ_oF, _oF, **kw):  # typ_oF = the definition singleton
          o = oF(nF=typ_oF.nF, root=_oF, typ=typ_oF, **kw)
          _oF.N_ += [o]; typ_oF.c += 1
          return o

This still requires us to init oF in every function, which is the same?

       def func(_oF):
           oF = oF.call('comp', _oF)

frame_2D_alg/agg_recursion.py
boris-kz: ok, right:

The call instance `o` is already created before the function body runs. So either:

**1. Return oF from factory, use it directly** (current pattern):
```python
def comp_N_(_pairs, r, _oF=None):
    oF = oF.call(typ_comp_N_, _oF)
    # oF.N_ is available for packing data
    oF.N_ += [L]  # pack results
```
But here `oF.N_` mixes data nodes with sub-call oFs — same list for both.

**2. Separate attrs** — data N_ vs call N_:
```python
class oF(CF):
    def __init__(f, **kw):
        super().__init__(**kw)
        f.call_ = []  # sub-call oFs (code tree)
        f.N_    # data nodes packed per call
```
Then factory does `_oF.call_ += [o]` instead of `_oF.N_ += [o]`. Clean separation.

**3. Pass oF as implicit context** — thread it through without explicit param:
```python
import contextvars
_current_oF = contextvars.ContextVar('oF')
```
Avoids `_oF` param everywhere but adds hidden state — probably not your style.

I think option 2 is right: `oF.N_` holds data (inputs/outputs of this call), `oF.call_` holds sub-calls (code tree). Mirrors the data/code duality within each call instance. The function accesses its oF directly since `call()` returns it:

```python
def comp_N_(_pairs, r, _oF=None):
    o = oF.call(typ_comp_N_, _oF)
    # o.N_ for data, o.call_ for sub-calls (auto-filled by callee oF.call)
```

boris-kz: I think we need this contextvars, option C:

Yes, `ContextVar` is per-thread and can be scoped per call via token-based reset:

```python

def call(typ_oF, **kw):
    _oF = _cur.get()  # caller's oF
    o = oF(nF=typ_oF.nF, root=_oF, typ=typ_oF, **kw)
    _oF.call_ += [o]
    _cur.set(o)  # now callees see this oF
    return o
```

But the problem: you need to *restore* the parent oF when the function returns. That requires either:

**A. Manual reset:**
```python
def comp_N_(_pairs, r):
    o = oF.call(typ_comp_N_)
    # ... body, callees see o via _cur ...
    _cur.set(o.root)  # restore parent
```
Error-prone, same boilerplate you're trying to eliminate.

**B. Context manager:**
```python
@contextmanager
def call(typ_oF, **kw):
    _oF = _cur.get()
    o = oF(nF=typ_oF.nF, root=_oF, typ=typ_oF, **kw)
    _oF.call_ += [o]
    token = _cur.set(o)
    yield o
    _cur.reset(token)
```
```python
def comp_N_(_pairs, r):
    with oF.call(typ_comp_N_) as o:
        # ... body
```
Clean restore, but forces indentation of entire body.

**C. Decorator** — wraps the whole function, auto push/pop:
```python
def traced(typ_oF):
    def wrap(func):
        def inner(*a, **kw):
            _oF = _cur.get()
            o = oF(nF=typ_oF.nF, root=_oF, typ=typ_oF)
            _oF.call_ += [o]; _cur.set(o)
            result = func(*a, **kw)
            _cur.set(_oF)  # restore
            return result
        return inner
    return wrap
```
Then `_oF` param disappears entirely from all signatures. Callee just does `_cur.get()` if it needs its own oF for data packing.

Option C is cleanest but most hidden. Your preference?

frame_2D_alg/agg_recursion.py
boris-kz: We can use the decorator, option C below

frame_2D_alg/agg_recursion.py
kwcckw: If i'm not mistaken, I think the C method is still the same since it's a decorator, which needs us to place it infront of each function? And we still need to manually insert the type (nF)?

boris-kz: Not with contextvars:

In option C the decorator handles everything, the function gets its own oF via `_cur.get()`:

```python
_cur = contextvars.ContextVar('oF', default=Z)  # root

def traced(typ_oF):
    def wrap(func):
        def inner(*a, **kw):
            _oF = _cur.get()
            o = oF(nF=typ_oF.nF, root=_oF, typ=typ_oF)
            _oF.call_ += [o]; _cur.set(o)
            result = func(*a, **kw)
            _cur.set(_oF)
            return result
        return inner
    return wrap

@traced(typ_comp_N_)
def comp_N_(_pairs, r):
    o = _cur.get()    # this call's oF
    o.N_ += [L]       # pack data
    # callees auto-nested via _cur
```

So `_cur.get()` inside the body replaces both the `_oF` parameter and the `oF = CF(...)` boilerplate. The `call()` static method is no longer needed — the decorator *is* the call factory.

One convenience shortcut to avoid `_cur.get()` everywhere:

```python
def oF_(): return _cur.get()  # alias
```

Then in body: `oF_().N_ += [L]`

kwcckw: In this section:

      @traced(typ_comp_N_)
      def comp_N_(_pairs, r):
          o = _cur.get()    # this call's oF
          o.N_ += [L]       # pack data
          # callees auto-nested via _cur

That `typ_comp_N_` should be varied across different functions, which needs us to parse the unique typ?

boris-kz: It can grab function name in

    def traced(func):
        def inner(*a, **kw):
            name = func.__name__
            if name not in oF._typ:
                Z.Ct.N_ += [t]  # register type


```python
import contextvars

class CoF(CF):  # code fork: function call trace

    name = "func"
    _typ = {}  # func_name -> typ_oF singleton in Z.Ct
    def __init__(f, **kw):
        super().__init__(**kw)
        f.call_ = []  # sub-call oFs (code tree)
        f.rd = kw.get('rd', 0)  # credit TT, keep dTT for data?
    @staticmethod
    def get(): return CoF._cur.get()
    @staticmethod
    def traced(func):
        def inner(*a, **kw):
            name = func.__name__
            if name not in CoF._typ:
                Z.Ct.N_ += [t]  # register type
            typ = CoF._typ[name]
            _CoF = CoF._cur.get()
            oF = CoF(nF=name, root=_CoF, typ=typ)
            _CoF.call_ += [oF]; typ.c += 1  # use Q2R?
            CoF._cur.set(oF)
            result = func(*a, **kw)
            CoF._cur.set(_CoF)  # restore
            return result
        return inner
```

Usage:
```python
@CoF.traced
def comp_N_(_pairs, r):

    oF = CoF.get()      # this call's CoF, oF.nF = 'comp_N_'
    oF.N_ += [L]       # pack data
    # sub-calls auto-nested
```

Open question: should `typ.c += 1` also accumulate `typ.dTT += o.dTT`? That needs to happen after the function returns, so it would go between `result = func(...)` and `_cur.set(_oF)`:

```python
            result = func(*a, **kw)
            typ.dTT += oF.dTT  # aggregate into type
            CoF._cur.set(_CoF)
```

frame_2D_alg/agg_recursion.py
boris-kz, line 1: oF.Ct.call_ now is typ singletons, and H is just call nesting.
Now we need to flatten that H into call_ and group similar functions into types in real H, in feedback

So we need a way to compare the functions, by constituent operations?

frame_2D_alg/agg_recursion.py
```diff
@@ -235,23 +241,24 @@ def proj_V(_N,N, dist, dy_dx, dec):  # _N x N induction
                 # ~= links in clustering
         else: break  # beyond initial induction range, re-sort by proj_V?
     for N in set(N_):
-        if N.rim: Q2R(N.rim, N.Rt, merge=0, froot=0)
+        if N.rim: Q2R(N.rim, N.Rt, merge=0, froot=0,_oF=oF)
     # call trace:
-    oF.N_ += [CF(nF='comp_N_',root=oF, N_=L_, dTT=rdpTT)]  # assign data root for scope, error lowers source code weight
+    oF.dTT = rdpTT; oF.L_=L_; _oF.N_ += [oF]  # assign data root for scope, error lowers source code weight
```
boris-kz, line 246: so:

    oF = CoF.get()  # this call's CoF, oF.nF = 'comp_N_'
    oF.N_ += [L_]; oF.c=cm; oF.dTT=rdpTT  # or Q2R? for data scope, no oF.call_+=[oF]: sub-calls auto-nested

frame_2D_alg/agg_recursion.py
kwcckw: > Open question: should typ.c += 1 also accumulate typ.dTT += o.dTT?

So `_CoF` is the root and we pack sub oF into call_ instead of N_ now. 

            _CoF = CoF._cur.get()
            oF = CoF(nF=name, root=_CoF, typ=typ)
            _CoF.call_ += [oF]

 For c, i think it should be +1 for every added sub oF?

> oF.Ct.call_ now is typ singletons, and H is just call nesting.

So oF.Ct.N_ should be still packing types oF right?

> So we need a way to compare the functions, by constituent operations?

To do this, we may need to define each group of operation first? Or that should be done via the clustering of oFs? Something like compare all-> group by operation -> compare between and within operation -> group again?

boris-kz: no, c must correspond to rdpTT: oF.N_ += [L_]; oF.c=cm; oF.dTT=rdpTT 

>So oF.Ct.N_ should be still packing types oF right?

I think its CoF now, so oF.Ct.call_, in feedback only

we can unpack operations as long as they are different, down to arithmetic, which should be comparable?

kwcckw: > no, c must correspond to rdpTT: oF.N_ += [L_]; oF.c=cm; oF.dTT=rdpTT

Okay, so this oF.c is the same as the data.c. Also for eval's based oFs, we may not have their dTT or c, so they will be empty?

> we can unpack operations as long as they are different, down to arithmetic, which should be comparable?

Yes, but we don't really pack this exact operation related information in each CoF? They probably only have different nF, dTT or c?

boris-kz: oF = CoF.get()  # this call oF.nF = 'comp_N_', sub-calls are auto-nested in oF.call_
    oF.Lt = Q2R(L_)  # data scope, c-proportional, or get separate oL_ with L.dTT=rdpTT?

then we can use oF.N_ instead of call_?

>Yes, but we don't really pack this exact operation related information in each CoF? They probably only have different nF, dTT or c?

it's AST per function

kwcckw: > then we can use oF.N_ instead of call_?

Yes, so if oF.Lt.N_ packs data nodes now, it's actually neater with oF.N_.

> it's AST per function

Okay right, so as you mentioned before, we just need a way to compare the function. 

Since each function is a AST, i think we can first compare by using their data Lt.N_, and then maybe dTT and c?

boris-kz: oF.Lt = Q2R(L_)  # data scope, c-proportional, or get separate oL_ with L.dTT=rdpTT?

frame_2D_alg/agg_recursion.py
boris-kz: To skip putting @CoF.traced before function defs:

**1. Module-level `__getattr__`** — intercept function calls, but only works for module attribute access, not local calls. Won't work.

**2. Metaclass / decorator on the whole module** — wrap all functions at import time:
```python
def trace_all(module_dict):
    for name, obj in module_dict.items():
        if callable(obj) and not isinstance(obj, type):
            module_dict[name] = CoF.traced(obj)

# at end of module:
trace_all(vars())
```
But fragile — catches things you don't want (imported functions, lambdas).

**3. `sys.settrace`** — intercept every call at runtime:
```python
def _trace(frame, event, arg):
    if event == 'call' and frame.f_code.co_filename == __file__:
        # create oF, push context
    elif event == 'return':
        # pop context
    return _trace
sys.settrace(_trace)
```
Works universally but heavy overhead, and you lose access to `oF.get()` inside the function body without extra plumbing.

**4. Import hook** — AST-transform the source at import, inject `@CoF.traced` on every `def`. Most magic, hardest to debug.

None of these are clean. The explicit `@CoF.traced` is one line per function, self-documenting, and lets you exclude functions that shouldn't be traced (utilities, `__init__`, etc.). I'd keep it explicit.

frame_2D_alg/agg_recursion.py
kwcckw: I think we probably can add a numeric types for comparison purpose? All comp's type = 1, clust 's type = 2 and eval's type = 3? Using a string nF might not be feasible to compare them.

frame_2D_alg/agg_recursion.py
kwcckw: I think the difficulty in debugging is a big problem, but i think we can try the third method first since we can skip the init of traced and get oF when needed.

boris-kz: ok please try that

frame_2D_alg/agg_recursion.py
boris-kz: that's totally backwards, the typs must be defined bottom-up

kwcckw: Okay, then I think the best method is from their data node_ and nested data node_ now since Lt.N_ should carry enough information to differentiate and cluster them.

boris-kz: it should be comparing operations

boris-kz: Gemini review:

Here is a review of the `CoF` implementation and its related logic.

### 1. Context Restoration Safety
If a traced function raises an exception or exits unexpectedly, the context remains stuck on the failed `oF`. This will corrupt the call tree for all subsequent operations. You need a `try...finally` block to guarantee the context is popped.

### 2. Mismatched `dTT` Accumulation
In the `traced` decorator, you accumulate `typ.dTT += oF.dTT`. However, looking at the traced functions (e.g., `comp_N_`), you are storing the results in the data scope `oF.Lt` (`oF.Lt = Q2R(L_)`), not directly in `oF.dTT`. Since `oF` inherits from `CF`, its own `dTT` remains `np.zeros((2,9))` unless you explicitly update it inside the function. 

You either need the functions to update `oF.dTT` directly, or the decorator needs to pull from `oF.Lt.dTT` (or sum `oF`'s forks).

### 3. Unbounded Call Trace Memory
`_CoF.N_ += [oF]` appends a new code fork instance for every single call to `cross_comp` and `comp_N_`. Since these are recursive and loop heavily, `Z.N_` will grow indefinitely. Unless `ffeedback()` or another meta-function is actively pruning or compressing `Z.N_` at the end of a cycle, this will cause memory issues on larger inputs.

Here is the suggested implementation to fix the first two issues:

```diff
@@ -107,10 +107,13 @@
             oF = CoF(nF=name, root=_CoF, typ=typ)  # call instance
             _CoF.N_ += [oF]; typ.c += 1  # use Q2R?
             CoF._cur.set(oF)
-            result = func(*a, **kw)
-            typ.dTT += oF.dTT
-            CoF._cur.set(_CoF)  # restore after
+            try:
+                result = func(*a, **kw)
+            finally:
+                # update typ.dTT from oF.Lt if oF.dTT isn't directly updated
+                typ.dTT += (oF.Lt.dTT if oF.Lt else oF.dTT)
+                CoF._cur.set(_CoF)  # restore after
             return result
         return inner
```

**Additional Note on `CoF._typ` Initialization:**
`Z.Ct.N_ += [t]` executes on the *first call* of a traced function. It relies on `Z` being defined globally before any traced function executes. This works fine given python's execution model, but it tightly couples the class method to the global `Z` instance defined later in the file.

kwcckw: > it should be comparing operations

But each operation is defined by their dTT, c , data N_ , etc?  So comparing operations should involve those params ?

boris-kz: this is comparing functions, not data

kwcckw: Okay right, maybe we can use the structure of their tree here? Such as the tree forking depth and number for comparison?

boris-kz: that's not functionality

kwcckw: In that case, how should we really cluster them when all existing params are not directly related? Do we need another new param in CoF to represent their functionality?

boris-kz: we need to compare AST, many component functions overlap

kwcckw: Right now we don't have any var to pack their component functions, and that should be included in the actual AST?

boris-kz: you don't know what AST is

kwcckw: From my understanding, AST is a tree that shows order of those caller and callee functions?

https://en.wikipedia.org/wiki/Abstract_syntax_tree

In our case, for "component functions overlap", it's just overlapping of nested oFs in call_?

boris-kz: oFs are actual calls, AST is all possible calls

PR #101, opened 2026-04-01
4/1/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -249,7 +273,7 @@ def proj_V(_N,N, dist, dy_dx, dec):  # _N x N induction
     acc = [TT,cm,rm, TTd,cd,rd]
     for pL in sorted(pairs, key=lambda x: x[0]):  # proximity prior, test compared?
         dist,dy_dx,_N,N = pL # rim angl is not canonic
-        pTT = proj_V(_N,N, dist, dy_dx, root.m if root<2 else decay** (dist/((_N.span+N.span)/2)))  # based on current rim
+        pTT = proj_V(_N,N, dist, dy_dx, root.m if root!=2 else decay** (dist/((_N.span+N.span)/2)))  # based on current rim
```
kwcckw, line 276: This should be !=2 since root could be an instance and we can't perform <2 check with the instance.

frame_2D_alg/agg_recursion.py
```diff
@@ -1056,6 +1082,14 @@ def init_C_(N_):
 
 if __name__ == "__main__":  # './images/toucan_small.jpg' './images/raccoon_eye.jpeg', add larger global image
 
+    trace_all_module_functions(vars(),exclude= {'trace_all_module_functions',
```
kwcckw, line 1085: So i end up using method 2 here because method 3 messed up the debugging tool by using `sys.settrace `. With this method, we run the function to trace all required functions here, and we need to explicitly insert the excluded function names.

frame_2D_alg/agg_recursion.py
```diff
@@ -103,30 +107,34 @@ def __init__(f, **kwargs):
...
     def __bool__(f): return bool(f.c)  # N_ may be empty?
 
-import contextvars
 class CoF(CF):  # code fork: call_: function call trace, H deeper call tree, Ct: types, dTT=results, w=m or separate sum wTT?
     name = "func"
     _cur = contextvars.ContextVar('oF')
     _typ = {}  # func_name -> typ_oF singleton in Z.Ct
     def __init__(f, **kw):
         super().__init__(**kw)
+        f.typ = kw.get('typ', None)
```
boris-kz, line 116: CF has typ

kwcckw: Okay right, CoF is inherited from CF, thanks.

frame_2D_alg/agg_recursion.py
```diff
@@ -103,30 +107,34 @@ def __init__(f, **kwargs):
...
+            return func
+        @wraps(func)
         def inner(*a, **kw):
             name = func.__name__
             if name not in CoF._typ:
                 t = CoF(nF=name); CoF._typ[name] = t
                 Z.Ct.N_ += [t]  # register type
             typ = CoF._typ[name]; _CoF = CoF._cur.get()
-            oF = CoF(nF=name, root=_CoF, typ=typ)  # call instance
+            oF = CoF(nF=name, root=_CoF, typ=typ, Lt=CF()); oF.Lt.root=oF  # call instance
```
boris-kz, line 130: we don't actually need Lt for data, C0F already has data fields that are not used otherwise.
we just need separate `_CoF.call_`, below

frame_2D_alg/agg_recursion.py
```diff
@@ -267,7 +291,7 @@ def proj_V(_N,N, dist, dy_dx, dec):  # _N x N induction
         if N.rim: Q2R(N.rim, N.Rt, merge=0, froot=0)
     # call trace:
     oF = CoF.get()  # this call oF.nF = 'comp_N_', sub-calls are auto-nested in oF.call_
-    oF.Lt = Q2R(L_)  # data scope, c-proportional, or get separate oL_ with L.dTT=rdpTT?
+    oF.Lt = Q2R(L_, R=oF)  # data scope, c-proportional, or get separate oL_ with L.dTT=rdpTT?
```
boris-kz, line 294: so how do we modify Q2R to change data fields in CoF?

kwcckw: We probably need another nested Q2R in Q2R to pack and accumulate those data in each call_'s N_?

frame_2D_alg/agg_recursion.py
kwcckw: So we revert back to prior structure where .call_ is packing oFs, while N_ is packing data ?

frame_2D_alg/agg_recursion.py
boris-kz: I think Q2R is fine as is, just skip the return:  Q2R(L_,R=oF)

frame_2D_alg/agg_recursion.py
boris-kz: yes,

            typ = CoF._typ[name]
            _CoF = CoF._cur.get()
            oF = CoF(nF=name, root=_CoF, typ=typ)  # call instance
            _CoF.call_ += [oF]  # Q2R(call_)  before evals

boris-kz: So `Q2R(L_, R=CoF.get())` writes `m,d,dTT,c,r` directly into the oF, and `oF.N_ = L_` (via merge in Q2R). No separate `Lt`.

Changes:

**In `CoF.traced`**, remove `Lt=CF()` line:
```python
oF = CoF(nF=name, root=_CoF, typ=typ)  # was: + Lt=CF(); oF.Lt.root=oF
_CoF.call_ += [oF]
```

And `call_` needs init — add to `CoF.__init__`:
```python
def __init__(f, **kw):
    super().__init__(**kw)
    f.call_ = kw.get('call_', [])
```

**In `comp_N_`**, replace:
```python
oF = CoF.get()
oF.Lt = Q2R(L_, R=oF)
```
with:
```python
Q2R(L_, R=CoF.get())
```

**In `comp_C_`**, same pattern:
```python
Q2R(L_, R=CoF.get())
```

**In `cluster_N`**, replace:
```python
oF = CoF.get()
oF.Lt.N_ = G_; oF.Lt.dTT=sel_TT; oF.Lt.c = sel_c
```
with:
```python
oF = CoF.get()
oF.N_ = G_; oF.dTT = sel_TT; oF.c = sel_c
```

**In `cluster_C`** and **`cluster_P`**, same — write directly to `oF` instead of `oF.Lt`.

One question: in `ffeedback` you access `Z.Nt.H` via `sum2F(Z.N_, 'Nt', Z)`. With this change, `Z.N_` is the call trace (from `_CoF.call_ += [oF]`). But `sum2F` expects data nodes. Should `ffeedback` use `Z.call_` for the call tree and `Z.N_` for data? Or is `Z.N_` always the call trace (never gets data via `Q2R`)?

frame_2D_alg/agg_recursion.py
kwcckw: Okay thanks, looks like R will be returned. 

If we need to replace oF.Lt with new CF from L_, we should init new R , and R should be None?

    oF.Lt = Q2R(L_)

boris-kz: you don't need that

frame_2D_alg/agg_recursion.py
boris-kz: in cluster_P:

    dCt = Q2R(set(_C_)-set(C_))  # compression
    oF = CoF.get(); oF.N_=C_; oF.dTT=dCt.dTT; oF.c = dCt.c  # data
?

frame_2D_alg/agg_recursion.py
boris-kz, line 280: each link should get `L.rdpTT` (relative prediction error), and `Q2R` gets a flag to sum from `n.rdpTT` instead of `n.dTT`:

In `comp_N_`, after comp_N:
```python
Link.rdpTT = np.abs(pTT - Link.dTT) / eps_(Link.dTT)
```

In `Q2R`, add `frd=0` param:
```python
def Q2R(N_, R=None, merge=1, froot=1, fN=0, frd=0):
    ...
    R.m, R.d, R.dTT, R.c, R.r = sum_vt(N_, fm=1, frd=frd)
```

And in `sum_vt`, add `frd=0`:
```python
def sum_vt(N_, fm=0, wTT=wTTn, frd=0):
    for n in N_:
        rc = n.c / C; TT += (n.rd if frd else n.dTT) * rc; R += n.r * rc
    return (*vt_(TT,R,wTT), TT,C,R) if fm else (TT,C,R)
```

Then in `comp_N_`:
```python
Q2R(L_, R=CoF.get(), frd=1)
```

But wait — `rd` is only set on links formed by full `comp_N` (where we had `pTT`). The predictive links (`pL`, typ=-1) skip `comp_N` and won't have `rd`. Should those get `rd = np.zeros((2,9))` by default, or be excluded from the oF sum?

frame_2D_alg/agg_recursion.py
kwcckw: Sorry, i confused it with the Lt version earlier.

frame_2D_alg/agg_recursion.py
kwcckw: Or we can use if `hasattr`? If N has rdpTT, we sum TT from rdpTT instead of dTT?

frame_2D_alg/agg_recursion.py
kwcckw: This `_C_` are those Cs from prior loop, i think it should be the first input `_C_` - final C_?

boris-kz: ok, right

frame_2D_alg/agg_recursion.py
boris-kz: that would mean operation has 0 value if there was no projection, that wrong?

boris-kz: Right, `rd` should represent prediction error relative to the result. If there's no projection, the whole `dTT` is "surprising" — so `rd` should be high, not zero.

Default `rd` for links without projection should be `np.ones((2,9))` — 100% error, meaning no prediction was available. Or maybe just skip `frd` for those and use `dTT` directly?

Actually, the cleaner option: pLs (typ=-1) already have `dTT = pTT`, so their `rd` would be undefined. They shouldn't be in the oF sum at all since they weren't actually computed — they're projections that skipped comp_N. Only real comp_N links contribute to the operation's value.

So: `Q2R(L_, R=CoF.get(), frd=1)` where `L_` only contains actual comp_N outputs (typ=1), not pLs. That's already the case since pLs are appended separately and L_ in comp_N_ only gets `+= [Link]` from comp_N calls, or `+= [pL]` for predictive ones.

Wait — pLs *are* added to the same `L_`. Should we split into `L_` (comp_N results) and `pL_` (projections)?

boris-kz, line 199: def sum_vt(N_, fr=0, fm=0, wTT=wTTn):  # basic weighted sum of CN|CF list, wTTn for comp_N

    for n in N_:
        rc = n.c / C; TT += (n.dTT,n.rTT)[fr] *rc; R += n.r*rc  # * weight
    return (*vt_(TT,R,wTT), TT,C,R) if fm else (TT,C,R)

def Q2R(N_, R=None, merge=1, froot=1, fN=0, fr=0):  # update root with N_

    if R is None: R = CN() if fN else CF()  # root
    if fr: N_ = [n for n in N_ if n.typ>-1]  # skip pLs if proj
    R.m, R.d, R.dTT, R.c, R.r = sum_vt(N_, fr,fm=1)  # fr: local surprise
    if merge:
        for N in [n for N in N_ for n in N.N_] if merge==3 else (N_[1].N_ if merge == 2 else N_):  # 2: pair merge
            R.N_ += [N]
            if froot: N.root=R
            if fN: R.C_ += N.C_; R.TTn += N.TTn; R.TTc += N.TTc  # frame_H only?
    if fN and R.C_: R.Ct = Q2R(R.C_)
    return R

kwcckw: > that would mean operation has 0 value if there was no projection, that wrong?

From ` (n.dTT,n.rTT)[fr]` above, it selects dTT or rTT, so there shouldn't be any zero value?

> Wait — pLs are added to the same L_. Should we split into L_ (comp_N results) and pL_ (projections)?

We shouldn't need this for pL since they are positive and should be processed the same way as L?

kwcckw:
> oFs are actual calls, AST is all possible calls

Okay thanks, i was thinking each oF.call_ packs something like their own nested sub AST earlier.

To compare those calls and group them, could you clarify what would be the suitable params? Data's related params are not directly related, and maybe just typ, dTT and c?

frame_2D_alg/agg_recursion.py
boris-kz: pLs are no op for comp_N_, so they should be excluded

boris-kz:
> > oFs are actual calls, AST is all possible calls
> 
> Okay thanks, i was thinking each oF.call_ packs something like their own nested sub AST earlier.
> 
> To compare those calls and group them, could you clarify what would be the suitable params? Data's related params are not directly related, and maybe just typ, dTT and c?

we need to compare the core process itself, as well as callees.
comp_ functions are subtractive, clust_ functions are additive, there must be a way to determine that automatically

frame_2D_alg/agg_recursion.py
boris-kz, line 1: `ffeedback` should first map calls to to data H: frame.H[i].H, they have specific data scope?
Or that should be done in traced():

```
_CoF.call_ += [oF]  # add Q2R(call_) before evals? nest by data scope?

```

so call_  is in each level of CoF.H?

Claude

Right, each oF's data scope corresponds to a specific level in the data hierarchy. So nesting calls by data scope during tracing makes sense — then `Z.H` mirrors `frame.H` with oFs at each level.

But how does `traced()` know which data level the call belongs to? The data level is only known after the function executes and produces output. So it can't be determined at call entry.

Maybe on exit — after the function returns, we know the oF's data scope (from `oF.N_` set by Q2R). Then nest into the appropriate H level:

```python
result = func(*a, **kw)
typ.dTT += oF.dTT
# nest by data scope:
depth = data_depth(oF)  # how to determine?
while len(_CoF.H) <= depth: _CoF.H.append([])
_CoF.H[depth].append(oF)
```

But `data_depth` — is that the root chain length from oF's data to frame? Or should it be simpler: the recursion depth tracked by a counter in the context?

Or skip auto-nesting in `traced()` and let `ffeedback` map `flat call_` to `frame.H` by matching `oF.N_` elements to level nodes?

kwcckw: > pLs are no op for comp_N_, so they should be excluded

They should be excluded only in the oF data summation? So when fr is true, N should be skipped if their typ = =1?

kwcckw:
> > > oFs are actual calls, AST is all possible calls
> > 
> > 
> > Okay thanks, i was thinking each oF.call_ packs something like their own nested sub AST earlier.
> > To compare those calls and group them, could you clarify what would be the suitable params? Data's related params are not directly related, and maybe just typ, dTT and c?
> 
> we need to compare the core process itself, as well as callees. comp_ functions are subtractive, clust_ functions are additive, there must be a way to determine that automatically

I see, at this point i don't see there any method to do that yet unless we explicitly define the function as comp or clust.

frame_2D_alg/agg_recursion.py
boris-kz: yes:

    Q2R([n for n in N_ if n.typ>-1], R=CoF.get(),fr=1)  # skip pLs, oF.call_+=[oF], oF.nF='comp_N_', adds data

boris-kz:
ask claude :)

frame_2D_alg/agg_recursion.py
kwcckw: > Or skip auto-nesting in `traced()` and let `ffeedback` map `flat call_` to `frame.H` by matching `oF.N_` elements to level nodes?

I think we can try this? So we should just record calls in traced, and then only sort them to H in ffeedback, and the nesting could be traced by their root?

boris-kz: auto-nesting in traced() should be simpler, we need to form Z.H along with frame.H[i].H, it should be 1/1 mapping. There are multiple nesting orders, it will be pretty hard to map after the fact

kwcckw: So the nesting in Z.H should be automatically added when we add H in frame?

But why Z.H maps to frame.H[i].H but not frame.H? At this point Z is per frame?

boris-kz: Claude — CN has `Nt.H`, `Ct.H`, and within each level there's sub-clustering forming deeper nesting. So a single depth counter can't capture this.

The nesting path is more like a multi-dimensional address: which fork (Nt/Ct/Bt), which level in that fork's H, which sub-level...

That's too complex for auto-nesting in `traced()`. Maybe `call_` should just stay flat per oF, and `ffeedback` maps them to the data hierarchy by matching `oF.N_` to nodes in `frame.H[i].Nt.H[j]...`?

Or simpler: each oF already has `root` pointing to its caller oF. So the call tree is implicit in `root` chains. `ffeedback` can walk `oF.root` chains and match against data H by checking which data nodes `oF.N_` belongs to.

Is the mapping even needed at init, or is it sufficient to do it lazily in `ffeedback`?

boris-kz: > So the nesting in Z.H should be automatically added when we add H in frame?
> 
> But why Z.H maps to frame.H[i].H but not frame.H? At this point Z is per frame?

because frame is a pipeline of expansion, temporal in addition to spatial.
each level has it's own H

boris-kz: ok, we can trace it through roots. 
then oF.H is only added in clustering functions by similarity, which is not implemented yet?

kwcckw: > ok, we can trace it through roots. then oF.H is only added in clustering functions by similarity, which is not implemented yet?

Actually we have both comp and cluster in data and calls running independently, and they should have the same H hierarchy? Or we only need one of them?

boris-kz: Z.H should be from clustering *functions*, not data, by similarity of their operations. We just discussed that.

boris-kz:
Yes — you can inspect the function's operations: what arithmetic it applies to inputs. Subtractive ops produce derivatives (differences), additive ops produce sums (clusters).

Simplest automatic way: track the sign of accumulation in oF. Comp functions have `oF.dTT` dominated by differences (from `comp_derT`, subtraction), clustering functions have `oF.dTT` dominated by sums (from `Q2R`, `sum_vt`).

Or more directly: comp functions output links (typ=1, with nt pair), clustering functions output nodes (typ=2, with N_). So:

```python
oF.fd = 1 if any(n.typ == 1 for n in oF.N_) else 0  # 1: comp (subtractive), 0: clust (additive)
```

But that's post-hoc detection, not really "automatic determination of operation type."

The deeper question is whether you can derive the arithmetic signature from the actual Python operations — inspecting AST or bytecode for ratio of `+` vs `-` ops. That's possible but heavy.

Or is it enough to just tag `comp_` and `clust_` by naming convention in `CoF._typ`?

```python
typ.fd = 1 if 'comp' in name else 0
```

That's brittle but trivial. What granularity do you actually need for the clustering-by-similarity step?

frame_2D_alg/agg_recursion.py
kwcckw: > Z.H should be from clustering _functions_, not data, by similarity of their operations. We just discussed that.

Yes, but if Z.H needs to be mapped 1 to 1 to frame.H[i].H, how is that feasible when we cluster their H differently ?

kwcckw:
> Yes — you can inspect the function's operations: what arithmetic it applies to inputs. Subtractive ops produce derivatives (differences), additive ops produce sums (clusters).
> 
> Simplest automatic way: track the sign of accumulation in oF. Comp functions have `oF.dTT` dominated by differences (from `comp_derT`, subtraction), clustering functions have `oF.dTT` dominated by sums (from `Q2R`, `sum_vt`).
> 
> Or more directly: comp functions output links (typ=1, with nt pair), clustering functions output nodes (typ=2, with N_). So:
> 
> ```python
> oF.fd = 1 if any(n.typ == 1 for n in oF.N_) else 0  # 1: comp (subtractive), 0: clust (additive)
> ```
> 
> But that's post-hoc detection, not really "automatic determination of operation type."
> 
> The deeper question is whether you can derive the arithmetic signature from the actual Python operations — inspecting AST or bytecode for ratio of `+` vs `-` ops. That's possible but heavy.
> 
> Or is it enough to just tag `comp_` and `clust_` by naming convention in `CoF._typ`?
> 
> ```python
> typ.fd = 1 if 'comp' in name else 0
> ```
> 
> That's brittle but trivial. What granularity do you actually need for the clustering-by-similarity step?

This is almost similar with the response from the claude from my end:

      1. Same function? — oF1.nF == oF2.nF
      2. Same op type? — both comp_ or both clust_? (determines how you interpret dTT)
      3. Value similarity — compare oF1.dTT vs oF2.dTT using the same comp_ logic used for data (i.e., pass them through vt_)
      4. Structural similarity — recursively compare oF1.call_ vs oF2.call_: same sub-functions called in the same order? This is essentially a call-tree diff.

I think the structural similarity above could be useful? Since if sub-functions are called in same order, they should be similar in operation?

kwcckw:
Suggestions from GPT 5.4:

Best recommendation by stage
Minimal and useful now

Use:

1. function name family (comp, cluster, proj, eval, etc.)
2. output type of oF.N_
3. dTT magnitude profile

That’s enough to start.
Slightly better

Add:

4. parent oF.root.nF
5. data level / depth
6. output count and total c

Later, if needed

Add:

8. AST/bytecode operator histogram
9. call-subtree pattern (oF.call_ composition)

I think the non data related features would be just "AST/bytecode operator histogram" and "call-subtree pattern (oF.call_ composition)"? Which we can start with them first?

frame_2D_alg/agg_recursion.py
boris-kz: that's just initial call mapping when the calls are recorded, it won't be in Z.H

boris-kz:
Chat is useless here. Bytecode is too

PR #102, opened 2026-04-02
4/2/2026

kwcckw:


kwcckw:
> Z.H should be from clustering functions, not data, by similarity of their operations. We just discussed that.

Actually Z.H should be the code operations on the data from frame.H[i].H right? If we form their Z.H correctly using their root, there's no additional steps requires to align it with frame.H[i].H since it should aligned the same way.

For convenience, we probably can include depth in CoF and apply depth+1 to new CoF whenever CoF is initialized, or the depth could be retrieved by recursively getting root.root.root... too.

boris-kz:
Z.H is clustering over the code itself, completely different from frame.H

frame_2D_alg/agg_recursion.py
```diff
@@ -122,7 +122,7 @@ def traced(func):
         def inner(*a, **kw):
             name = func.__name__
             if name not in CoF._typ:
-                t = CoF(nF=name); CoF._typ[name] = t
+                t = CoF(nF=name); CoF._typ[name] = t  # not root in typ CoF until clustering?
```
boris-kz, line 125: this is not real clustering, it can actually be in call_, we don't need flat call list?
actually, this call_ nesting should be in ffeedback: 
```
if name not in CoF._typ:
    Z.call_ += [t]  # register type, or in ffeedback only?

```

kwcckw: This `t` is a typ CoF and packed flat in Z.Ct.N_, while `call_` are those call instances, and we need both? Each typ in Z.Ct.N_ is unique but we may have multiple calls with the same nF or typ in call_, so i don't see why we don't need the flat call list.

boris-kz: It's the same thing, just nested. We can trace the calls through the roots anyway

boris-kz: @staticmethod
    def traced(func):
        @wraps(func)
        def inner(*a, **kw):
            _CoF = CoF._cur.get()
            _CoF.call_ += [oF]
            CoF._cur.set(_CoF)
            return result
        inner.wrapped = True  # flag
        return inner

boris-kz: def ffeedback(frame):  # adjust filters: all aves *= rV, ultimately differential backprop per ave?

    typ_ = []  # group calls by function name
    for oF in Z.call_:
        tF = next((fork for fork in typ_ if fork.nF==oF.nF), None)
        if tF: tF.N_ += [oF]
    Z.call_ = [Q2R(typ.N_, typ, froot=0) for typ in typ_]

...

frame_2D_alg/agg_recursion.py
```diff
@@ -180,7 +180,7 @@ def sum_vt(N_, fr=0, fm=0, wTT=wTTn):  # basic weighted sum of CN|CF list, wTTn
 
     C = sum(n.c for n in N_); R = 0; TT = np.zeros((2,9))
     for n in N_:
-        rc = n.c / C; TT += (n.dTT,n.rTT)[fr] *rc; R += n.r*rc  # * weight
+        rc = n.c / C; TT += (n.rTT if fr else n.dTT)*rc; R += n.r*rc  # * weight
```
boris-kz, line 183: how is that different?

frame_2D_alg/agg_recursion.py
kwcckw: > It's the same thing, just nested. We can trace the calls through the roots anyway

I see, so actually this is removing the typ oF in Ct.N_ and it will be computed in ffeedback and will be referenced in oF.typ only?

frame_2D_alg/agg_recursion.py
kwcckw: We may not have rTT across all Ns when we call sum_vt, so `n.rTT` should be used only when fr is true. `(n.dTT,n.rTT)[fr]` will call rTT in the first place, which causing error in nodes.

boris-kz: ok right

frame_2D_alg/agg_recursion.py
boris-kz: it's oF.nF, and we don't need Ct at this point. May be for function clustering

kwcckw: Okay right, it's nF now and from this line:

    Z.call_ = [Q2R(typ.N_, typ, froot=0) for typ in typ_]

Each call's root is their typ oF instead of higher level call oF？

boris-kz: that's typF, call's root is still the caller

kwcckw: Okay right, sorry i overlooked the froot = 0 earlier.

kwcckw: > it's oF.nF, and we don't need Ct at this point. May be for function clustering

So we don't have a direct H for typ CoF, and we only can retrieve them via Z.H now. Also without Lt, CoF.c should be data's c or code's c? Previously the data's c is packed in Lt.c.

boris-kz: it's all data.
Right now we need to figure how to compare and cluster the code

boris-kz: Within comp_: `comp_N` does full base_comp (subtraction + division via rn), `comp_F` does derT comp (subtraction only), `comp_C_` adds centroid normalization (division). Different arithmetic signatures.

Within clust_: `cluster_N` is connectivity-based (binary in/out), `cluster_C` is centroid-based (weighted sum), `cluster_P` is parallel refine (iterative reweighting). Different aggregation patterns.

And eval functions: `vt_`, `val_`, `proj_TT` are multiplicative (normalization, weighting).

So three coarse types by dominant operation: subtractive, additive, multiplicative. Finer differentiation within each.

The automatic way: each oF accumulates `dTT` through Q2R. The *ratio* between `oF.dTT` and `sum(child.dTT for child in oF.call_)` reveals the operation type:
- ratio ≈ -1: subtractive (comp)
- ratio ≈ +1: additive (clust) 
- ratio ≈ multiplicative factor: eval/projection

```python
oF.sig = oF.dTT / eps_(sum(c.dTT for c in oF.call_))
```

Then clustering oFs by similarity of `sig` gives you operation types without naming conventions.

kwcckw: Actually i'm not sure on this part:

      The automatic way: each oF accumulates dTT through Q2R. The ratio between oF.dTT and sum(child.dTT for child in oF.call_) reveals the operation type:
      
          ratio ≈ -1: subtractive (comp)
          ratio ≈ +1: additive (clust)
          ratio ≈ multiplicative factor: eval/projection

This is assuming dividing the sum of all child's dTT should be equal to 1? But i think this may not be true since we might skip some sub N's dTT when eval is false, which makes this method not accurate.

boris-kz: I don't think it means very much, what really matters is the order of operations: subtraction-first or summation-first. Sub-typing should be by complexity: full-scope in comp_C_ vs. selective in comp_N_, recusive refining in cluster_C, non-selective in cluster_P.

boris-kz: Minor correction: not complexity per se, but scope / selectivity of the operation. `comp_C_` vs `comp_N_` is full-scope vs selective comp, while `cluster_C` vs `cluster_P` is recursive refine vs parallel/global update. So the type axes seem to be: 1) subtraction-first vs summation-first, 2) selective vs full-scope, 3) recursive vs parallel refinement. "Complexity" is secondary, mostly a consequence of these.

boris-kz: That wasn't me, I had chat GPT review the PR and post the comment above, with my permission :). Claude and Gemini can't do that yet

frame_2D_alg/agg_recursion.py
```diff
@@ -490,7 +491,7 @@ def trans_cluster(G):
                 r += 1
         # clust_val = selectivity vs root.Lt: all comps, add dval?
         sel_TT = (Ft.dTT*Ft.c - Ft.root.Lt.dTT*Ft.root.Lt.c) / eps_(Ft.root.Lt.dTT * Ft.root.Lt.c)
-        oF = CoF.get(); oF.Lt.N_=G_; oF.dTT=sel_TT; oF.c = Ft.root.Lt.c - Ft.c  # add data
+        oF = CoF.get(); oF.N_=G_; oF.dTT=sel_TT; oF.c = Ft.root.c - Ft.c  # add data
         # combine C_:
         Q2R([C for N in (G_ if G_ else _N_) for C in N.Ct.N_], Ft.root.Ct, froot=0)
     return G_, r
```
boris-kz, line 497: So first, we need to cross-compare functions, bottom-up in composition, same as pixels. 
Base functions are atomic operations:

Level 0: Python primitives — `+, -, *, /, min, abs, np.hypot, cos, atan2`

Level 1: functions composed only of level 0 ops — `comp_derT`, `comp_A`, `vt_`, `eps_`

Level 2: functions calling level 1 — `comp_N`, `base_comp`, `cent_TT`

Level 3: functions calling level 2 — `comp_N_`, `comp_C_`, `cluster_N`

Level 4: `cross_comp`, `frame_H`

Cross-comp at each level compares functions of same composition order — just like comparing nodes of same sub depth. Match between `comp_derT` and `vt_` at level 1: both do element-wise subtraction and absolute value, but `vt_` adds normalization (division). So partial match + derivative.

The call tree already encodes this hierarchy, and base functions are leaf oF typs, with empty call_.

From `trace_func`, excluded functions are untraced (primitives for our purpose). So leaf typs are traced functions that only call untraced ops:

- `eps_`, `comp_A`, `comp_derT`, `vt_`, `extend_box`, `mdecay`...

Wait — some of these are already in the `exclude` list of `trace_func`:
```python
trace_func(vars(), exclude={'trace_functions','eps_','prop_F_','extend_box','TTw','Copy_','CopyF'})
```

So those aren't traced at all. The actual leaf typs are traced functions whose `call_` only contains untraced calls — detectable at runtime: after execution, `oF.call_` is empty (no child oFs).

```python
leaf_typs = [t for t in CoF._typ.values() if not any(c.call_ for c in t.call_)]
```

Then compare leaf typs by their source AST, cluster them, and work up to typs that call leaf typs, and so on. Same bottom-up hierarchy as data.

But some functions like `comp_A` are called from both `comp_slice` and `agg_recursion` — same typ, different contexts. Does context matter for the comparison, or just the static code?

frame_2D_alg/agg_recursion.py
kwcckw: > That wasn't me, I had chat GPT review the PR and post the comment above, with my permission :). Claude and Gemini can't do that yet

Wow, didn't know they can do that now. But to determine those orders, we still need to automatically identify each operation type first?

boris-kz: just pass the link, ask it to read the comments and reply on PR. Github should be in Chat's connected apps

boris-kz: >But to determine those orders, we still need to automatically identify each operation type first?

Uh, what is that supposed to mean? Order is the sequence, that's the first thing AST decodes

kwcckw: > > But to determine those orders, we still need to automatically identify each operation type first?
> 
> Uh, what is that supposed to mean? Order is the sequence, that's the first thing AST decodes

Okay, so by following root, we can retrieve sequence from top oF along the call tree, and only we drill down to the additional params such as their types? I think order alone may not be sufficient to cluster them?

frame_2D_alg/agg_recursion.py
boris-kz: claude:
Right. So the primitives are (operation, operand_type) pairs:

- `(+, scalar)` vs `(+, 2x9_array)` — same op, different complexity
- `(-, float)` vs `(-, int)` — same op, different precision

And sequence matters: `[sub, abs, min]` in `comp_derT` vs `[abs, sub, div]` in `vt_` — different process distance even though similar ops.

So comparison between two functions is like comp_slice: ordered sequence of (op, operand_complexity) tuples, with process distance = position difference between matched ops.

```
comp_derT: [(sub, 2x9), (abs, 2x9), (mul_sign, 2x9)]
vt_:       [(abs, 1x9), (add, 1x9), (div, 1x9), (matmul, 1x9), (sub, scalar)]
```

Match: shared (op, type) pairs. Miss: different ops or types at similar positions. Process distance: positional offset of matched pairs, analogous to spatial distance in pixel comp.

This is essentially 1D comp_P on code sequences — each op is a "dert" with parameters (op_type, operand_shape, operand_dtype). Consecutive matching ops form code "Ps", breaks form boundaries.

The operand complexity dimension: `scalar < 1D_array < 2D_array`, roughly maps to composition order in data. So a function doing `(sub, 2x9)` is higher-complexity than `(sub, scalar)` even though the operation is identical.

Should operand complexity be a separate comparison dimension, or combined with op type into a single distance?

kwcckw: I see, but those primitives are not recorded in each call, and we need to add that?

In actual case, it will be combinations of primitives and deeper oFs ?

boris-kz: they are recorded in AST, and level is defined by the absence of higher-level components, whatever the mixture

kwcckw: Okay, I think we can temporary include them in CoF, there should be a way to do that. 

> compares functions of same composition order

Also for the comparison, it always compare the same level functions only? Primitives with primitives, function level 1 with level 1 and so on? Or it doesn't matter?

boris-kz: Don't include them in CoF, I don't wan't cratches. Just try to undrstand the principles

kwcckw: Alright, and for the comparison:

> Match: shared (op, type) pairs. Miss: different ops or types at similar positions. Process distance: positional offset of matched pairs, analogous to spatial distance in pixel comp.

The match and miss is straightforward, but for process distance, that will be similar with rng+ in 1D algorithm? It's still considered a match with scaled value with distance when there's offset?

boris-kz: let's compare the same positions only, in AST sequence.
comparison should contunue only is all prior positions matched?

boris-kz: Claude — strict sequential match, terminate on first mismatch. Same as form_P in slice_edge: extend while matching, break on miss.

So two functions' ASTs are aligned position by position:

```
func_A: [op0, op1, op2, op3, op4]
func_B: [op0, op1, op2, op5, op6]
                        ^ break here
match length = 3
```

Each position compares `(op_type, operand_type, operand_shape)`. Match at position i requires all three to match (or within tolerance for shape?).

This is the simplest possible comp — no skipping, no alignment search. The match length directly measures shared prefix = shared initial process. Functions that start differently are immediately dissimilar.

But function order in AST depends on code layout — `if/else` branches create different sequences for the same logic. Should branching be flattened, or compared as a separate dimension?

And loops: `for n in N_: TT += n.dTT` — is the loop body one op (repeated) or N ops? Probably one op with a multiplier (like L in comp_slice)?

kwcckw: > let's compare the same positions only, in AST sequence.
> comparison should contunue only is all prior positions matched?

Yes, i think we should stop when there's discrepancy in their function since a single different function may means a lot? Unless we want to capture that information too.

For example:

`'frame_H', 'vect_edge', 'cross_comp', 'comp_N_'` should be different with `'frame_H', 'vect_edge', 'cross_comp', 'comp_C_'` although they might be matched in the subsequent functions.

boris-kz: this should go through clusters of functions first, such as comp_ and clust_, we don't have them yet

kwcckw: > this should go through clusters of functions first, such as comp_ and clust_, we don't have them yet

That should be promoted typs? If we need that information in their sequence, then we will need to cluster those function first before the sequence assignment.

boris-kz: it's completely new types, after func cross_comp, we don't have that process yet.

PR #103, opened 2026-04-03
4/3/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -1004,14 +1004,38 @@ def expand_lev(_iy, _ix, elev, T):  # seed tile is pixels in 1st lev, or Fg in h
         else: break
     return F  # for intra-lev feedback
 
+# draft, pack typ oF from tree of calls, also create H within the unpacking process?
+def pack_call_tree(oF, typ_, depth=1, sequence=['Z']):
```
kwcckw, line 1008: `Z.call_` is nested now since we pack each call into the root instead of the same Z, so we will need additional process to unpack all calls and group them into typ oF.

I also added depth and sequence as new params into each call during the unpacking process, and it should be something like this:

    ['Z', 'frame_H', 'vect_edge', 'cross_comp', 'val_', 'proj_N', 'val_', 'sum_vt', 'Q2R', 'Q2R', 'sum_vt', 'vt_']

I think this should be useful for the comp code later.

Also for Z.H, i think we can add it along the unpacking process here ?

boris-kz: it should be packed nested to begin with, instead of ffeedback:

    def traced(func):  # _oF.call_ += call oF
        @wraps(func)
        def inner(*a, **kw):
            _CoF = CoF._cur.get()
            _CoF.call_ += [oF]
            tF = next((fork for fork in Z.call_ if fork.nF==oF.nF), None)
            if tF: tF.N_ += [oF]  # top-down in data depth?
            else:  Z.call_ += [CF(nF=oF.nF, N_=[oF])]
            CoF._cur.set(_CoF)
            return result
        inner.wrapped = True
        return inner

boris-kz: Actually, Z.call_ is already used for nested call tree.
We can use Z.N_ for typ list, Z doesn't need a data context, it's redundant to frame?

            tF = next((fork for fork in Z.N_ if fork.nF==oF.nF), None)
            if tF: tF.N_ += [oF]  # typ list top-down in data depth?
            else:  Z.N_ += [CF(nF=oF.nF, N_=[oF])]

frame_2D_alg/agg_recursion.py
```diff
@@ -260,7 +260,7 @@ def proj_V(_N,N, dist, dy_dx, dec):  # _N x N induction
     for N in set(N_):
         if N.rim: Q2R(N.rim, N.Rt, merge=0, froot=0)
     # call trace:
-    L_ = [n for n in L_ if n.typ>-1]; if L_: Q2R(L_,R=CoF.get(),fr=1)  # skip pLs, oF.call_+=[oF], oF.nF='comp_N_', adds data
+    L_ = [n for n in L_ if n.typ>-1]; Q2R(L_,R=CoF.get(),fr=1) if L_ else [] # skip pLs, oF.call_+=[oF], oF.nF='comp_N_', adds data
```
boris-kz, line 263: that doesn't make any sense

if L_ := [n for n in L_ if n.typ>-1]: Q2R(L_, R=CoF.get(), fr=1)

frame_2D_alg/agg_recursion.py
```diff
@@ -1059,5 +1083,6 @@ def trace_func(module_dict, module_name=None, exclude=()):
     trace_func(vars(),exclude= {'trace_functions','eps_','prop_F_','extend_box','TTw','Copy_','CopyF'})
     Y,X = imread('./images/toucan.jpg').shape
     # frame = agg_frame(0, image=imread('./images/toucan.jpg'), iY=Y, iX=X)
+    CoF._cur.set(Z)  # set root call to Z
```
boris-kz, line 1086: we don't need this with the edits above, Z should not store data, it's redundant to frame?
actually it's the same for cross_comp oF?

frame_2D_alg/agg_recursion.py
```diff
@@ -1004,14 +1004,38 @@ def expand_lev(_iy, _ix, elev, T):  # seed tile is pixels in 1st lev, or Fg in h
...
+    typ_ = pack_call_tree(Z, [])
+    Z.call_ = [Q2R(typ.N_, typ, merge=0, froot=0) for typ in typ_]
+    '''
     typ_ = []  # group calls by function name
     for oF in Z.call_:
         tF = next((fork for fork in typ_ if fork.nF==oF.nF), None)
         if tF: tF.N_ += [oF]  # top-down in data depth?
         else:  typ_ += [CF(nF=oF.nF, N_=[oF])]
-    Z.call_ = [Q2R(typ.N_, typ, froot=0) for typ in typ_]  # no Q2R(Z.call_,Z): redundant to frame?
+    Z.call_ = [Q2R(typ.N_, typ, merge=0, froot=0) for typ in typ_]  # no Q2R(Z.call_,Z): redundant to frame?
```
boris-kz, line 1037: thanks, 

    Z.call_ = [Q2R(typ.N_, typ, merge=0, froot=0) for typ in Z.call_]

kwcckw: So merge should be 0 here, else we are packing back typ.N_ into root typ.

frame_2D_alg/agg_recursion.py
kwcckw: Thanks, your version is better.

frame_2D_alg/agg_recursion.py
kwcckw: > We can use Z.N_ for typ list, Z doesn't need a data context, it's redundant to frame?

We still need Z.call_ to pack the nested calls? For type, if it's packed in Z.N_, we may not able to retrieve the data from oF in case we need it later?

boris-kz: that Z.N_ is for Z only, its data is redundant to frame

kwcckw: So do you mean only Z.N_ is exclusive for types, while the other oF.N_ is still packing those data?

boris-kz: yes

frame_2D_alg/agg_recursion.py
```diff
@@ -1059,5 +1083,6 @@ def trace_func(module_dict, module_name=None, exclude=()):
     trace_func(vars(),exclude= {'trace_functions','eps_','prop_F_','extend_box','TTw','Copy_','CopyF'})
     Y,X = imread('./images/toucan.jpg').shape
     # frame = agg_frame(0, image=imread('./images/toucan.jpg'), iY=Y, iX=X)
+    CoF._cur.set(Z)  # set root call to Z
     frame = frame_H(image=imread('./images/toucan.jpg'), iY=Y//2 -31, iX=X//2 -31, Ly=64,Lx=64, Y=Y, X=X, rV=1)
     # search frames ( tiles inside image, at this size it should be 4K, or 256K panorama, won't actually work on toucan
```
boris-kz, line 1088: AST extractor for python primitives, Claude:

import ast, inspect, textwrap
import numpy as np

'''
Extract sequential operation signatures from function AST for cross-comparison.
Each op is (op_type, op_name, operand_info) tuple.
Branches are flattened, loops are single ops with nested body.
Comparison is sequential: terminate on first mismatch (like form_P).
'''

operand type inference from AST node names and patterns

NUMPY_ARRAY_FUNCS = {'np.zeros','np.ones','np.array','np.abs','np.sum','np.mean',
                     'np.minimum','np.maximum','np.hypot','np.diff','np.copy',
                     'np.add','np.subtract','np.sign','np.stack','np.unravel_index'}
SCALAR_FUNCS = {'min','max','abs','sum','len','round','int','float'}
TRIG_FUNCS = {'cos','sin','atan2','np.hypot'}

def get_op_name(node):

    """Get readable name from Call func node"""
    if isinstance(node, ast.Attribute):
        val = ast.unparse(node.value) if hasattr(ast,'unparse') else ''
        return f"{val}.{node.attr}"
    elif isinstance(node, ast.Name):
        return node.id
    return ast.dump(node)

def infer_operand(node):

    """Infer operand type from AST node, rough static analysis"""
    if isinstance(node, ast.Constant):
        return ('scalar', type(node.value).__name__)
    elif isinstance(node, ast.Name):
        name = node.id
        # heuristic: capitalized short names are usually arrays (TT, N_, etc)
        if name in ('eps',): return ('scalar','float')
        if name.endswith('_') or name.isupper() or 'TT' in name: return ('array','inferred')
        return ('unknown', name)
    elif isinstance(node, ast.Subscript):
        base = infer_operand(node.value)
        return ('element', base[1])  # indexing into array
    elif isinstance(node, ast.Call):
        fname = get_op_name(node.func)
        if fname in NUMPY_ARRAY_FUNCS: return ('array', fname)
        if fname in SCALAR_FUNCS: return ('scalar', fname)
        if fname in TRIG_FUNCS: return ('scalar', fname)
        return ('call_result', fname)
    elif isinstance(node, ast.BinOp):
        left = infer_operand(node.left)
        right = infer_operand(node.right)
        # array + anything = array
        if left[0]=='array' or right[0]=='array': return ('array','computed')
        return ('scalar','computed')
    elif isinstance(node, ast.UnaryOp):
        return infer_operand(node.operand)
    elif isinstance(node, ast.List):
        return ('list', f'len={len(node.elts)}')
    elif isinstance(node, ast.Tuple):
        return ('tuple', f'len={len(node.elts)}')
    return ('unknown', type(node).__name__)

def extract_ops(node):

    """Extract sequential ops from AST node, flattening branches, wrapping loops"""
    ops = []

    if isinstance(node, (ast.Module, ast.FunctionDef)):
        for stmt in node.body:
            ops.extend(extract_ops(stmt))

    elif isinstance(node, ast.Assign):
        # extract ops from the value side
        ops.extend(extract_expr_ops(node.value))

    elif isinstance(node, ast.AugAssign):  # +=, -=, *=
        op_name = type(node.op).__name__
        left_type = infer_operand(node.target)
        right_type = infer_operand(node.value)
        ops.append(('AugOp', op_name, left_type, right_type))
        ops.extend(extract_expr_ops(node.value))

    elif isinstance(node, ast.Return):
        if node.value:
            ops.extend(extract_expr_ops(node.value))

    elif isinstance(node, (ast.For, ast.While)):
        body_ops = []
        for stmt in node.body:
            body_ops.extend(extract_ops(stmt))
        ops.append(('Loop', len(body_ops), body_ops))

    elif isinstance(node, ast.If):
        # flatten: include both branches
        for stmt in node.body:
            ops.extend(extract_ops(stmt))
        for stmt in (node.orelse or []):
            ops.extend(extract_ops(stmt))

    elif isinstance(node, ast.Expr):
        ops.extend(extract_expr_ops(node.value))

    elif isinstance(node, (ast.Try,)):
        for stmt in node.body:
            ops.extend(extract_ops(stmt))

    elif isinstance(node, ast.With):
        for stmt in node.body:
            ops.extend(extract_ops(stmt))

    return ops

def extract_expr_ops(node):

    """Extract ops from expression nodes"""
    ops = []

    if isinstance(node, ast.BinOp):
        # recurse into operands first (depth-first = execution order)
        ops.extend(extract_expr_ops(node.left))
        ops.extend(extract_expr_ops(node.right))
        op_name = type(node.op).__name__
        left_type = infer_operand(node.left)
        right_type = infer_operand(node.right)
        ops.append(('BinOp', op_name, left_type, right_type))

    elif isinstance(node, ast.UnaryOp):
        ops.extend(extract_expr_ops(node.operand))
        op_name = type(node.op).__name__
        operand_type = infer_operand(node.operand)
        ops.append(('UnaryOp', op_name, operand_type))

    elif isinstance(node, ast.Compare):
        ops.extend(extract_expr_ops(node.left))
        for comp in node.comparators:
            ops.extend(extract_expr_ops(comp))
        op_names = [type(op).__name__ for op in node.ops]
        ops.append(('Compare', op_names, infer_operand(node.left)))

    elif isinstance(node, ast.Call):
        # recurse into args first
        for arg in node.args:
            ops.extend(extract_expr_ops(arg))
        fname = get_op_name(node.func)
        arg_types = [infer_operand(a) for a in node.args]
        ops.append(('Call', fname, arg_types))

    elif isinstance(node, ast.ListComp):
        ops.extend(extract_expr_ops(node.elt))
        ops.append(('ListComp', infer_operand(node.elt)))

    elif isinstance(node, ast.IfExp):
        # flatten: include both
        ops.extend(extract_expr_ops(node.body))
        ops.extend(extract_expr_ops(node.orelse))

    elif isinstance(node, ast.Tuple):
        for elt in node.elts:
            ops.extend(extract_expr_ops(elt))

    elif isinstance(node, ast.List):
        for elt in node.elts:
            ops.extend(extract_expr_ops(elt))

    elif isinstance(node, ast.Subscript):
        ops.extend(extract_expr_ops(node.value))

    elif isinstance(node, ast.Starred):
        ops.extend(extract_expr_ops(node.value))

    return ops


def func_signature(func):

    """Extract operation signature from a function"""
    source = textwrap.dedent(inspect.getsource(func))
    tree = ast.parse(source)
    return extract_ops(tree)


def comp_sig(sig_a, sig_b):

    """Compare two signatures sequentially, terminate on first mismatch.
    Returns match length and list of (m,d) per matched position."""

    matches = []
    for i, (a, b) in enumerate(zip(sig_a, sig_b)):
        m, d = comp_op(a, b)
        if m <= 0:
            break  # first mismatch terminates
        matches.append((m, d))
    return len(matches), matches


def comp_op(a, b):

    """Compare two op tuples, return (match, diff)"""
    # must match op category
    if a[0] != b[0]:
        return 0, 1

    m, d = 1, 0  # base match for same category

    if a[0] == 'BinOp':
        # compare op type: Add, Sub, Mult, Div
        if a[1] == b[1]: m += 1
        else: return 0, 1  # different arithmetic = mismatch
        # compare operand types
        om = comp_operand(a[2], b[2]) + comp_operand(a[3], b[3])
        m += om; d += (2 - om)

    elif a[0] == 'Call':
        if a[1] == b[1]: m += 2  # same function
        else:
            # partial match: same category of function?
            if ({a[1], b[1]} <= NUMPY_ARRAY_FUNCS or
                {a[1], b[1]} <= SCALAR_FUNCS or
                {a[1], b[1]} <= TRIG_FUNCS):
                m += 1; d += 1
            else:
                return 0, 1  # different function category

    elif a[0] == 'AugOp':
        if a[1] == b[1]: m += 1
        else: return 0, 1

    elif a[0] == 'Compare':
        if a[1] == b[1]: m += 1
        else: d += 1

    elif a[0] == 'Loop':
        # compare loop bodies recursively
        inner_len, inner_matches = comp_sig(a[2], b[2])
        m += inner_len; d += max(len(a[2]), len(b[2])) - inner_len

    return m, d


def comp_operand(a, b):
    """Compare operand type tuples, return match in 0:1"""
    if a[0] == b[0]:  # same category: scalar, array, etc
        return 1.0 if a[1] == b[1] else 0.5
    return 0


# ---------- demo ----------

if __name__ == "__main__":

    # test on simple functions
    def func_a(x, y):
        d = x - y
        m = min(abs(x), abs(y))
        return m, d

    def func_b(x, y):
        d = x - y
        m = min(abs(x), abs(y))
        m *= -1 if (x < 0) != (y < 0) else 1
        return m, d

    def func_c(x, y):
        s = x + y
        r = s / 2
        return r

    print("=== func_a signature ===")
    sig_a = func_signature(func_a)
    for op in sig_a: print(f"  {op}")

    print("\n=== func_b signature ===")
    sig_b = func_signature(func_b)
    for op in sig_b: print(f"  {op}")

    print("\n=== func_c signature ===")
    sig_c = func_signature(func_c)
    for op in sig_c: print(f"  {op}")

    print(f"\ncomp(a,b): match_len={comp_sig(sig_a, sig_b)[0]}")
    print(f"comp(a,c): match_len={comp_sig(sig_a, sig_c)[0]}")
    print(f"comp(b,c): match_len={comp_sig(sig_b, sig_c)[0]}")

frame_2D_alg/agg_recursion.py
kwcckw: Okay, then we not really need H because Z.call_ is already nested with the proper hierarchy, and we just need to add the flat typ CF in Z.N_.

boris-kz: H is for function clustering in ffeedback

frame_2D_alg/agg_recursion.py
boris-kz: Updated in the end:

https://claude.ai/share/d4db1d96-cab0-4bae-ad2c-006fac8e83fe

frame_2D_alg/agg_recursion.py
kwcckw: > it's completely new types, after func cross_comp, we don't have that process yet.

Okay right, and the function clustering should be within the same level functions? Or it could be clustering functions across multiple levels in a sequence?

boris-kz: initial Z.N_ is 1st level, func clustering will create incrementally higher levels, same as agg+

kwcckw: I see, so the initial call_ hierarchy is actually no longer useful? That's only useful to trace the sequence right?

frame_2D_alg/agg_recursion.py
boris-kz: The problem: `if ...: a; b; c; elif ...: d; e; f` — the `elif` gets placed at same indent as `if`, but separated by the intermediate `acc[1]`, `acc[2]`, `L_` lines which are also at same indent. So `elif` appears after non-block statements.

This is a fundamental issue with one-liner `if/elif` chains with semicolons. Need to detect `elif`/`else` in the parts and keep them properly connected:Lines 343-344 are separate lines. Line 343 `if ...: a;b;c;d;` splits to `if: a` then `b`, `c`, `d` at same indent. Then line 344 `elif` appears — but `b,c,d` broke the if/elif chain.

The cleanest fix: for `if cond: a; b; c`, expand to a proper block:

```python
if cond:
    a
    b
    c
```It parses. Level 7 has a cycle (mutual recursion). Let me see the signatures and cross-comp:Working. Key results:

- **`val_` x `vt_`: match=5, M=16.5** — strongest match among leaves, they share the normalization pattern
- **`comp_H` x `comp_N`: match=6, M=81.0** — strong, both do recursive sub-comp with `comp_derT` + `vt_`
- **`inner` x `traced`: match=11, M=33.0** — decorator wrapper functions, naturally similar
- **`comp_N_` x `comp_C_`: match=1** — low prefix match despite being conceptually similar (they diverge early in structure)

The strict prefix-only rule is very selective — most matches are 1 (just the first op). Functions that do similar things but start differently get 0.Notable observation: `comp_N_` x `comp_C_` only matches 1 despite being structurally analogous. That's because they start differently — `comp_C_` begins with `product/combinations` setup while `comp_N_` starts with a loop over `_pairs`. The shared comp logic is deeper in the sequence.

This suggests the strict prefix-only rule may be too conservative for higher-level functions. For data (pixels, Ps), adjacency makes prefix comparison natural. For code, the meaningful similarity might be in subsequences rather than prefixes — similar to how `comp_slice` handles non-contiguous links via `prelink__`.

Should we add a secondary comparison mode that allows gaps (like der+ links between non-adjacent matched ops), evaluated only if prefix match is below threshold?

frame_2D_alg/agg_recursion.py
boris-kz: It can be used to assign oF filters at each data composition level, we need that in each call eval?

kwcckw: > It can be used to assign oF filters at each data composition level, we need that in each call eval?

Yes, that assignment should be done automatically with CoF now, but where do we have this call eval?

boris-kz: for example in cross_comp:
if M * ((len(L_)-1)*Lw) * wn > ave:

that wn should actually be wN, assigned to oF 'cluster_N' in feedback up the call chain.
in agg+, oF.root chain maps to lower operand composition, and oF.call_ to higher operand composition?

boris-kz: claude:

Yes, that's the mapping. `oF.root` points to the caller (which operates on simpler operands), so following the root chain goes down in operand composition. `oF.call_` holds sub-calls (which operate on more complex/composed operands), so following call_ goes up in operand composition.

But there's a subtlety worth clarifying: in your current code, `oF.root` is set to `_CoF` (the caller context), and `oF.call_` accumulates the sub-calls made within that function. So the tree is:

- `Z` (root) → `Z.call_` contains top-level function calls
- Each `oF.call_` contains functions called within that function
- Each `oF.root` points back to the enclosing caller

This means operand composition increases going deeper into `call_` because inner functions receive already-composed results from outer ones? Or is it the reverse — outer functions like `cross_comp` operate on higher-composition nodes while inner functions like `comp_N` operate on their lower-composition elements?

If the latter, then `call_` maps to *lower* operand composition (decomposition into sub-operations on simpler elements), and `root` maps to *higher* operand composition (the enclosing operation that works with composed results). Which direction do you intend?

frame_2D_alg/agg_recursion.py
kwcckw: > val_ x vt_: match=5, M=16.5 — strongest match among leaves, they share the normalization pattern
> comp_H x comp_N: match=6, M=81.0 — strong, both do recursive sub-comp with comp_derT + vt_
> inner x traced: match=11, M=33.0 — decorator wrapper functions, naturally similar
> comp_N_ x comp_C_: match=1 — low prefix match despite being conceptually similar (they diverge early in structure)

So this is matching between functions using only their primitives right? Which is comparing only typ? But in actual case, we may need to match multiple comp_N in the call, which may need their calling sequence as well?

> This suggests the strict prefix-only rule may be too conservative for higher-level functions. For data (pixels, Ps), adjacency makes prefix comparison natural. For code, the meaningful similarity might be in subsequences rather than prefixes — similar to how `comp_slice` handles non-contiguous links via `prelink__`.

Maybe we need an additional selective comp to compare those inner operations regardless of their sequence? Since it maybe too strict for higher level operations here.

boris-kz: >So this is matching between functions using only their primitives right?

no, check the extended version above. I don't think it's correct, but in the right direction

frame_2D_alg/agg_recursion.py
boris-kz: ok, the caller is higher-composition, prior calls are lower-composition, but they are not direct callers, they only feed back to the caller, which may increase in operand composition with the the current call

boris-kz: Currently `wn,wc,wN,wC` are true globals, but they should vary by composition level. Each `oF` in the root chain operates at a different composition level, so `oF.w` should be the level-specific weight that replaces the corresponding global when that function executes.

The feedback would propagate up the root chain, setting `oF.w` based on the `rdpTT` (relative prediction error) of that oF's results. Then during execution, instead of using global `wn`, a function would use `CoF.get().w` or similar — the weight from the current call context.

But there's a question: the current globals map to fork types (`wn` for comp_N, `wC` for clustering, etc.), not composition levels. So `oF.w` would need to combine both:
- which fork type (comp vs cluster vs eval)
- which composition level (depth in root chain)

Is the intent that each `oF` has a single scalar `w`, or a structured weight like `wTT` that covers all its sub-operations? The current `CF.wTT` seems closer to what's needed — per-attribute weights specific to that fork's context.

frame_2D_alg/agg_recursion.py
kwcckw: > The problem: `if ...: a; b; c; elif ...: d; e; f` — the `elif` gets placed at same indent as `if`, but separated by the intermediate `acc[1]`, `acc[2]`, `L_` lines which are also at same indent. So `elif` appears after non-block statements.
> 
> This is a fundamental issue with one-liner `if/elif` chains with semicolons. Need to detect `elif`/`else` in the parts and keep them properly connected:Lines 343-344 are separate lines. Line 343 `if ...: a;b;c;d;` splits to `if: a` then `b`, `c`, `d` at same indent. Then line 344 `elif` appears — but `b,c,d` broke the if/elif chain.
> 
> The cleanest fix: for `if cond: a; b; c`, expand to a proper block:
> 
> if cond:
>     a
>     b
>     c
> ```It parses. Level 7 has a cycle (mutual recursion). Let me see the signatures and cross-comp:Working. Key results:
> 
> - **`val_` x `vt_`: match=5, M=16.5** — strongest match among leaves, they share the normalization pattern
> - **`comp_H` x `comp_N`: match=6, M=81.0** — strong, both do recursive sub-comp with `comp_derT` + `vt_`
> - **`inner` x `traced`: match=11, M=33.0** — decorator wrapper functions, naturally similar
> - **`comp_N_` x `comp_C_`: match=1** — low prefix match despite being conceptually similar (they diverge early in structure)
> 
> The strict prefix-only rule is very selective — most matches are 1 (just the first op). Functions that do similar things but start differently get 0.Notable observation: `comp_N_` x `comp_C_` only matches 1 despite being structurally analogous. That's because they start differently — `comp_C_` begins with `product/combinations` setup while `comp_N_` starts with a loop over `_pairs`. The shared comp logic is deeper in the sequence.
> 
> This suggests the strict prefix-only rule may be too conservative for higher-level functions. For data (pixels, Ps), adjacency makes prefix comparison natural. For code, the meaningful similarity might be in subsequences rather than prefixes — similar to how `comp_slice` handles non-contiguous links via `prelink__`.
> 
> Should we add a secondary comparison mode that allows gaps (like der+ links between non-adjacent matched ops), evaluated only if prefix match is below threshold?

This is the latest conversation i can see from the link above, which extended version you referring to here?

frame_2D_alg/agg_recursion.py
kwcckw: > assigned to oF 'cluster_N' in feedback up the call chain.

I see, then we need the level-specific weight from those oFs, but we are creating new oF in each H hierarchy, then the prior feedback updated weight may not be updated to those new oF? So we need another process to update each oF's weights to the prior feedback updated weights when we init new oF?

frame_2D_alg/agg_recursion.py
boris-kz: yes, that can deal with our functions, not just primitives

frame_2D_alg/agg_recursion.py
boris-kz: Probably, so we need to sum that oF.w from oF.call_, which should be updated at this point, that's how Python works?

boris-kz: Right, it's just bottom-up accumulation. Inner calls complete first (Python call stack), so by the time an oF returns, all its `call_` entries already have their `w` set. Then:

`oF.w = sum(sub.w for sub in oF.call_)`

Each oF's `w` is the total weight of its sub-operations, which naturally reflects composition level — deeper calls accumulate less, higher callers accumulate more. No type clustering needed for this, no separate feedback pass.

And this replaces the globals because at any point during execution, `CoF.get().w` gives you the accumulated weight of the current call context, which you use instead of `wn,wN` etc. in filter expressions like `ave * CoF.get().w`.

what sets `w` at the leaves — oFs with no `call_`, is `rdpTT` or direct match contribution?

kwcckw: But we have so many calls, and each call may get their own weights, so we need to map them all when we init new oF after the feedback?

boris-kz: it's automatic in Python, see above

frame_2D_alg/agg_recursion.py
kwcckw: Yes, that's using prefixes, so it should work for all primitives and functions. But do we really to strictly follow the order? I think It could be selective in higher level comp and not just following the order?

frame_2D_alg/agg_recursion.py
kwcckw: > Right, it's just bottom-up accumulation. Inner calls complete first (Python call stack), so by the time an oF returns, all its call_ entries already have their w set. Then:

This is after the function is executed (after inner functions are completed), but we need the weight before that for eval purpose?

    if M * ((len(L_)-1)*Lw) * wn > ave:

boris-kz:
It's a different weight for prior agg+

kwcckw:
> It's a different weight for prior agg+

Okay thanks , i was confused earlier. So the weight should be already updated during the feedback in prior agg+.

PR #104, opened 2026-04-04
4/4/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -120,12 +121,16 @@ def traced(func):  # _oF.call_ += call oF
         def inner(*a, **kw):
             _CoF = CoF._cur.get()  # not for Z
             oF = CoF(nF=func.__name__, root=_CoF)
+            # sum from CoF's calls' w, which should be updated at this point
+            oF.w = sum([call.w for call in _CoF.call_]) if _CoF.call_ else 1
```
kwcckw, line 125: > so we need to sum that oF.w from oF.call_, which should be updated at this point, that's how Python works?

So from your prior comment, we should init new oF.w from all ws that up to this point, and actually that should be getting `_CoF` recursively and get their `w` until no higher `_CoF`?

> what sets w at the leaves — oFs with no call_, is rdpTT or direct match contribution?

For bottom leaves, oF's call_ is empty, so we should compute w from their rTT or dTT?

            TT = getattr(oF, 'rTT', oF.dTT)
            if np.any(TT): oF.w = vt_(TT, oF.r)[0]  # weight from TT
            else:          oF.w = 1  # default weight
      
And if both `rTT` and `dTT` are empty, weight is default to 1?

frame_2D_alg/agg_recursion.py
```diff
@@ -1007,9 +1012,27 @@ def expand_lev(_iy, _ix, elev, T):  # seed tile is pixels in 1st lev, or Fg in h
...
+        if np.any(TT): oF.w = vt_(TT, oF.r)[0]  # weight from TT
+        else:          oF.w = 1  # default weight
+            
+
 def ffeedback(frame):  # adjust filters: all aves *= rV, ultimately differential backprop per ave?
 
-    Z.call_ = [Q2R(typ.N_, typ, merge=0, froot=0) for typ in Z.call_]  # no Q2R(Z.call_,Z): redundant to frame?
+    Z.N_ = [Q2R(typ.N_, typ, merge=0, froot=0) for typ in Z.N_]  # no Q2R(Z.call_,Z): redundant to frame?
+ 
+    for typ in Z.N_:  # update weights bottom up
```
kwcckw, line 1033: Sorry, this is probably wrong, the w update should be from CoF instead of in ffeedback.

frame_2D_alg/agg_recursion.py
```diff
@@ -111,6 +111,7 @@ class CoF(CF):  # oF/ code fork, Ct: types, dTT=results, w=m or separate sum wTT
     def __init__(f, **kw):
         super().__init__(**kw)
         f.call_ = kw.get('call_',[])  # function call trace, nest by functions in ffeedback
+        f.w = kw.get('w', 0)        # total weight of nested operations
```
boris-kz, line 114: this w is in CF

frame_2D_alg/agg_recursion.py
boris-kz: Looks right. So that w should be in 0-2, averaging 1?

also, for example in cross_comp:

if M * ((len(L_)-1)*Lw) * wN > ave:
that wN should really be typ 'get_exemplars' oF.w?
how do we access it from current cross_comp call?

boris-kz: The current call is `CoF.get()`, which is the `cross_comp` oF. The next call will be `get_exemplars`, which will create its own oF in `call_`. But you need its `w` *before* calling it, to decide whether to call it.

So you need the `w` from the *previous* instance of that call — the type singleton in `Z.Ct`, or just the last oF with `nF=='get_exemplars'` in `Z.call_`.

Something like:

```python
_typ = next((f for f in Z.call_ if f.nF=='get_exemplars'), None)
if M * ((len(L_)-1)*Lw) * (_typ.w if _typ else wN) > ave:
```

Or cache in a dict for faster lookup. The globals `wN,wC,wn,wc` become fallback defaults for first-ever calls.

kwcckw: > that wN should really be typ 'get_exemplars' oF.w?
> how do we access it from current cross_comp call?

Actually that's my concern too. So if we need oF.w from oF.call_, which should be updated up to the point where we call the function, we won't have that unless we retrieve it from typ oF in Z? That typ oF of 'get_exemplars' should be updated with weights from prior ffeedback.

frame_2D_alg/agg_recursion.py
```diff
@@ -120,12 +121,16 @@ def traced(func):  # _oF.call_ += call oF
...
             _CoF.call_ += [oF]
             tF = next((fork for fork in Z.N_ if fork.nF==oF.nF), None)  # Z.N_ is not used for data: redundant to frame, same for cross_comp oF?
-            if tF: tF.N_ += [oF]  # typ_, top-down in data depth?
-            else:  Z.N_ += [CF(nF=oF.nF, N_=[oF])]  # group calls by function name
+            if tF is None: Z.N_ += [CF(nF=oF.nF, N_=[oF])]  # group calls by function name
+            else:          tF.N_ += [oF]  # typ_, top-down in data depth?
             CoF._cur.set(oF); result = func(*a, **kw)
             CoF._cur.set(_CoF)
+            if oF.call_: oF.w = sum([call.w for call in oF.call_])  # sum from all calls after function is executed             
+            else:        oF.w = 1
```
boris-kz, line 133: TT = getattr(oF, 'rTT', oF.dTT)
    if np.any(TT):
        m,d = vt_(TT, oF.r)
        oF.w = abs(m) / (abs(d) or eps)  # ratio, 0:inf, mean~1
    else:
        oF.w = 1

frame_2D_alg/agg_recursion.py
boris-kz: yes, 

`_typ = next((f for f in Z.call_ if f.nF=='get_exemplars'), None)`
`if M * ((len(L_)-1)*Lw) * (_typ.w if _typ else wN) > ave:`

?

frame_2D_alg/agg_recursion.py
kwcckw: Sorry, i overlooked that.

frame_2D_alg/agg_recursion.py
kwcckw: >  So that w should be in 0-2, averaging 1?

So that w shouldn't be ranged from 0 - 2 but could be any value from 0 to inf now?

frame_2D_alg/agg_recursion.py
kwcckw: > _typ = next((f for f in Z.call_ if f.nF=='get_exemplars'), None)

But this is getting `get_exemplars`'s from call instead of typ? Those typ should be in Z.N_?

boris-kz: Z.N_ is global, we need to oF.w in the current scope?

frame_2D_alg/agg_recursion.py
boris-kz: yes, 0:2 was wrong, w is multiplicative

boris-kz: Here's claude's rewrite:

```python
Right, sorry. `N_` holds data results, `call_` holds sub-calls — different things:

```python
class CoF(CF):

    name = "func"
    _wF = {}

    def __init__(f, **kw):
        super().__init__(**kw)
        f.call_ = kw.get('call_',[])
        f.w = CoF._wF.get(kw.get('nF',''), 1)

    @staticmethod

    @staticmethod
    def traced(func):
        @wraps(func)
        def inner(*a, **kw):
            _CoF = CoF._cur.get()
            _CoF.call_ += [oF]
            if oF.call_:
                typ_ = []
                for sub in oF.call_:
                    tF = next((f for f in typ_ if f.nF==sub.nF), None)
                    if tF: tF.N_ += [sub]
                    else:  typ_ += [CF(nF=sub.nF, N_=[sub], w=sub.w)]
                for tF in typ_: tF.w = np.mean([sub.w for sub in tF.N_])
                oF.call_ = typ_
                oF.w = np.mean([tF.w for tF in typ_])
            else:
                TT = getattr(oF,'rTT', oF.dTT)
                m,d = (vt_(TT,oF.r) if np.any(TT) else (0,0))
                oF.w = abs(m) / (abs(d) or eps)
            CoF._wF[oF.nF] = oF.w
            CoF._cur.set(_CoF)
            return result
        inner.wrapped = True
        return inner
```

Then usage replaces globals:
```python
if M * ((len(L_)-1)*Lw) * CoF._wF.get('get_exemplars',1) > ave:
```

boris-kz: >tF = next((f for f in Z.call_ if f.nF==oF.nF), None)

this should be for _CoF instead of Z, and only once _CoF is completed?

kwcckw: > > tF = next((f for f in Z.call_ if f.nF==oF.nF), None)
> 
> this should be for _CoF instead of Z, and only once _CoF is completed?

This is getting typ oF, which we should getting it from Z.N_? Those `_CoF` shouldn't pack any typ oF except it is a Z.

frame_2D_alg/agg_recursion.py
kwcckw: For current scope, we may have multiple `get_exemplars` from both agg+ or sub+ (similar for all other functions), i don't see how we can get an exact matching oF.w for current `get_exemplars` here?

boris-kz: it doesn't matter, they all get summed in root oF?
but this w summing should be c-weighted, right now it's not? use Q2R for that?

boris-kz: Right, functions operating on more data should weigh more. But `Q2R` does a lot more than needed here — it sets `m,d,dTT,c,r` and manages `N_` merging.

Simpler to just weight directly:

```python
            if oF.call_:
                typ_ = []
                for sub in oF.call_:
                    tF = next((f for f in typ_ if f.nF==sub.nF), None)
                    if tF: tF.N_ += [sub]
                    else:  typ_ += [CF(nF=sub.nF, N_=[sub], w=sub.w, c=sub.c)]
                C = sum(tF.c for tF in typ_) or eps
                for tF in typ_:
                    tc = sum(sub.c for sub in tF.N_) or eps
                    tF.w = sum(sub.w * sub.c/tc for sub in tF.N_)
                    tF.c = tc
                oF.call_ = typ_
                oF.w = sum(tF.w * tF.c/C for tF in typ_)
```

Each level: c-weighted mean of sub-ws, stays in 0:inf range.

frame_2D_alg/agg_recursion.py
boris-kz: that's what I thought, but we need local weights

frame_2D_alg/agg_recursion.py
kwcckw: > it doesn't matter, they all get summed in root oF?

Yes, so based on the code above:

            if oF.call_:
                typ_ = []
                for sub in oF.call_:
                    tF = next((f for f in typ_ if f.nF==sub.nF), None)
                    if tF: tF.N_ += [sub]
                    else:  typ_ += [CF(nF=sub.nF, N_=[sub], w=sub.w, c=sub.c)]
                C = sum(tF.c for tF in typ_) or eps
                for tF in typ_:
                    tc = sum(sub.c for sub in tF.N_) or eps
                    tF.w = sum(sub.w * sub.c/tc for sub in tF.N_)
                    tF.c = tc
                oF.call_ = typ_
                oF.w = sum(tF.w * tF.c/C for tF in typ_)
                
 This section should be after the `CoF._cur.set(oF); result = func(*a, **kw)`, where the function is executed? Else this oF.call_ is always empty anyway.
 
 So before that, oF.w is empty?

frame_2D_alg/agg_recursion.py
kwcckw: > that's what I thought, but we need local weights

But the local here is local to current call, or it's local to current function type?

boris-kz: both?

frame_2D_alg/agg_recursion.py
boris-kz: CoF.call_ should be complete by the time _CoF calls it

frame_2D_alg/agg_recursion.py
boris-kz: def traced(func):

        @wraps(func)
        def _sum_vt(F, fN):
            N_ = F.N_ if fN else F.call_
            F.c = C = sum(n.c for n in N_) or eps
            for a in 'r','w': setattr(F,a, sum(getattr(f,a) * f.c / C for f in N_))
        def inner(*a, **kw):
            _CoF = CoF._cur.get()
            _CoF.call_ += [oF]
                typ_ = []
                for sub in oF.call_:
                    tF = next((f for f in typ_ if f.nF==sub.nF), None)
                    if tF: tF.N_+= [sub]
                    else:  typ_ += [CoF(nF=sub.nF, N_=[sub], w=sub.w, c=sub.c)]
                for tF in typ_: _sum_vt(tF, fN=1)
                oF.call_= typ_; _sum_vt(oF, fN=0)
            else:
                m,d = (vt_(TT,oF.r) if np.any(TT) else (0,0))
                oF.w = abs(m) / (abs(d) or eps)
            CoF._wdt[oF.nF] = oF.w
            CoF._cur.set(_CoF)
            return out
        inner.wrapped = True
        return inner

kwcckw: > both?

But we only can select one? If local to call, we need to get the same call oF's w? And if it is local to typ, we need to get the same type oF from Z.N_? 

In code above:

    oF.w = abs(m) / (abs(d) or eps)

We only assign oF.w in the end of the function, after this line `out = func(*a, **kw)`, so before that we won't have w in oF?

frame_2D_alg/agg_recursion.py
kwcckw: > CoF.call_ should be complete by the time _CoF calls it

I think it may not be completed yet?

When we call `get_exemplars`, `_CoF` is `cross_comp`, and we might add `cluster_N` to the `_CoF.call_` later?

boris-kz: this is about CoF, not _CoF

frame_2D_alg/agg_recursion.py
boris-kz: that's a callee, all callees should terminate by the time the caller is evaluated

frame_2D_alg/agg_recursion.py
boris-kz: Claude:
each oF.w already reflects c-weighted average performance of its callee tree via `_sum_vt`. So `_wdt` is just a running prior for the next invocation's initial `.w` — last-write is a recency-biased estimate rather than a true average, but functional.

One concern: `_wdt` is used as a threshold multiplier at line 221:
```python
if M * ((len(L_)-1)*Lw) * CoF._wdt.get('get_exemplars',1) > ave:
```

Here it gates whether clustering proceeds. Last-write means this threshold fluctuates with whatever data `get_exemplars` last processed, which could be noisy. Would a decayed running average be worth the complexity, something like:

```python
CoF._wdt[oF.nF] = CoF._wdt.get(oF.nF,1) * decay + oF.w * (1-decay)
```

Or is that premature given the current state of the system?

kwcckw: > this is about CoF, not _CoF

Could you clarify if the CoF here is referring to the oF instance? `_CoF` is the caller and `oF` is the callee, and `CoF` is the class?

>  Here it gates whether clustering proceeds. Last-write means this threshold fluctuates with whatever data get_exemplars last processed, which could be noisy.

So if `CoF._wdt.get(oF.nF,1) ` gets the last processed same oF , i think it's not accurate? Since it could be from sum2G or cross_comp? We may need to get the last similar sequence oF?

frame_2D_alg/agg_recursion.py
kwcckw: Yes, so if we are retrieving last nF 's oF, i think there's no more issue here.

frame_2D_alg/agg_recursion.py
boris-kz: CoF is the callee, new instance, but it must be completed before the caller is evaluated

boris-kz: It's populated *during* `func(*a, **kw)`. Any traced sub-function inside `func` does:

```python
_CoF = CoF._cur.get()  # gets this oF
_CoF.call_ += [oF]     # appends to this oF's call_
```

So between `oF = CoF(...)` and `if oF.call_:`, the line `out = func(*a, **kw)` executes, and sub-calls append into `oF.call_` via the context var. The comment "complete at this point" is correct — it means func returned, so all sub-calls have finished appending.

kwcckw: > CoF is the callee, new instance, but it must be completed before the caller is evaluated

Yes, thanks for clarifying, this is clear now.

> Last-write means this threshold fluctuates with whatever data get_exemplars last processed, which could be noisy

Instead of using the last write same nF's oF, i think we can get the average w from all same typ's call in Z.N_?

kwcckw: > It's populated _during_ `func(*a, **kw)`. Any traced sub-function inside `func` does:
> 
> ```python
> _CoF = CoF._cur.get()  # gets this oF
> _CoF.call_ += [oF]     # appends to this oF's call_
> ```
> 
> So between `oF = CoF(...)` and `if oF.call_:`, the line `out = func(*a, **kw)` executes, and sub-calls append into `oF.call_` via the context var. The comment "complete at this point" is correct — it means func returned, so all sub-calls have finished appending.

Yes, so callee only will be completed after this line `result = func(*a, **kw)`, where the caller function ended.

If we retrieve caller's params in the caller function, no all callee is completed and packed to caller yet.

boris-kz: this is Python, dammit, a caller can't evaluate the callee untill all sub-calls terminate. We discussed that a bunch of times

kwcckw: > this is Python, dammit, a caller can't evaluate the callee untill all sub-calls terminate. We discussed that a bunch of times

Yes, sorry, i misunderstood earlier, this is clear now.

PR #105, opened 2026-04-05
4/5/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -138,8 +137,13 @@ def inner(*a, **kw):
             else:
                 TT = getattr(oF,'rTT', oF.dTT)  # rTT includes cluster compression
                 m,d = (vt_(TT,oF.r) if np.any(TT) else (0,0))
-                oF.w = abs(m) / (abs(d) or eps)
-            CoF._wdt[oF.nF] = oF.w
+                oF.w = (abs(m) / (abs(d) or eps)) or 1      
+            # from prior discussion, we need Z.N_ exclusively for types, and it should be global?
+            ztF = next((f for f in Z.N_ if f.nF==oF.nF), None)
```
kwcckw, line 142: So we need Z.N_ exclusively for typ oF? `Z` is not init within traced, so we won't have typs in Z unless we pack them here.

frame_2D_alg/agg_recursion.py
```diff
@@ -138,8 +137,13 @@ def inner(*a, **kw):
...
+            if ztF: ztF.N_ += [oF]
+            else:   ztF = CoF(nF=oF.nF, N_=[oF], w=oF.w, c=oF.c); Z.N_ += [ztF]
+            _sum_vt(ztF, fN=1); _sum_vt(Z, fN=1)  # or run this in batch during feedback for Z?
+            CoF._wdt[oF.nF] = ztF.w  # instead of update from last oF, update from typ's oF.w?
```
kwcckw, line 146: The existing `CoF._wdt[oF.nF] = oF.w` update weight using last oF.w, which may not be accurate. I think we should use the same typ oF from Z.N_? That should contain weight from all same typ oF, which is more accurate.

frame_2D_alg/agg_recursion.py
boris-kz: Do we need this exception for Z now, or just use Z.call_ for oFs?

frame_2D_alg/agg_recursion.py
boris-kz: How is that accurate, it's a different scope?

frame_2D_alg/agg_recursion.py
```diff
@@ -124,7 +123,7 @@ def _sum_vt(F, fN):
             F.c = C = sum(n.c for n in N_); F.r = sum(f.r* (f.c/C) for f in N_); F.w = sum(f.w* (f.c/C) for f in N_)
         def inner(*a, **kw):
             _CoF = CoF._cur.get()
-            oF = CoF(nF=func.__name__, root=_CoF)
+            oF = CoF(nF=func.__name__, root=_CoF,c=1)
```
boris-kz, line 126: c should be summed from N_, why c=1

kwcckw: Actually i'm not so sure on this part yet. Most of the eval' oF's N_ is empty, so they will never get their c filled, then i guess we should follow a same method to include eps in the division only.

Also for `F.w = sum(f.w* (f.c/C) for f in N_)`, if `f.c` is 0, `w` is default to 0?

frame_2D_alg/agg_recursion.py
kwcckw: Okay, then we would need H here? For a same scope typ, we should retrieve each scope's typ from Z.H[i]?

frame_2D_alg/agg_recursion.py
kwcckw: We can have that exception for Z.call_, because all other oF.call_ packs only their callee, but Z.call_ should contains all nested callee along with their typ?

frame_2D_alg/agg_recursion.py
boris-kz: if oF.N_ is empty, then oF is empty, just skip it?

frame_2D_alg/agg_recursion.py
boris-kz: it's the same for both

frame_2D_alg/agg_recursion.py
boris-kz: the scope is the callees, it's automatic

frame_2D_alg/agg_recursion.py
kwcckw: Right now we use `_CoF.call_ += [oF]` in CoF, and Z will never be the oF within the `inner` method. So Z.call_ will no longer be global and pack only single `frame_H` oF since there's only 1 callee under main.

frame_2D_alg/agg_recursion.py
kwcckw: Okay, but this raises another question because then next time we call `CoF._wdt[oF.nF]`, that w is also from different scope since we pack it with prior scope's callees?

frame_2D_alg/agg_recursion.py
kwcckw: Yes, this should resolve the problem, thanks.

frame_2D_alg/agg_recursion.py
boris-kz: yes, why is that a problem?

frame_2D_alg/agg_recursion.py
boris-kz: we can unpack globals from it, not sure that's even necessary

frame_2D_alg/agg_recursion.py
```diff
@@ -138,8 +137,13 @@ def inner(*a, **kw):
             else:
                 TT = getattr(oF,'rTT', oF.dTT)  # rTT includes cluster compression
                 m,d = (vt_(TT,oF.r) if np.any(TT) else (0,0))
```
boris-kz, line 139: this w is really just m:

                oF.w = vt_(TT,oF.r)[0] if np.any(TT) else 0  # =m?

frame_2D_alg/agg_recursion.py
kwcckw: Because they are from different scope? For example, we may call cross_comp for B fork in sum2G, and then using the w from there to eval the nested cross_comp for agg+?

frame_2D_alg/agg_recursion.py
kwcckw: Okay, right, previously i was thinking we may need the global typ based values, but since we have dict based weight using global `CoF._wdt[oF.nF]`, then global N_ per Z is no longer necessary.

frame_2D_alg/agg_recursion.py
boris-kz: Ultimately it should be an additional coef for B fork, and all other forks of a function, but any local performance of that core function is still relevant.
This is related to fork clustering we discussed before, when both match and difference between the functions are significant enough to create a new function category. That's oF cross_comp, it has to be resolved anyway.

boris-kz: Claude: The hand-coded function names are the initial type level. oF cross-comp works both directions:

- **Sub-types:** significant difference within same `nF` — `cross_comp` instances with divergent `call_` trees split. Solves the `_wdt` scope problem: each sub-type gets its own weight.

- **Super-types:** significant match across different `nF` — `comp_N` and `comp_F` cluster together as a "comp" super-type, `cluster_N` and `cluster_C` as "cluster" super-type. These are the rough categories already in the commented `Z.Ct.N_` block (`comp_`, `clust`, `eval_`, `attr_`).

So the commented-out type tree in `ffeedback` is a manual approximation of what oF cross-comp should derive. And the AST comparison provides the static basis, while runtime `call_` divergence provides the dynamic refinement for sub-typing.

The `_wdt` then indexes by derived type ID rather than function name — sub-types get separate entries, super-types share a combined entry for higher-level eval.

Let me think through the concrete steps, mapped to existing structures:

**Input:** All oF instances accumulated during execution, plus static AST signatures from `op_extract`.

**Level 0 — leaf oFs:**
Traced functions with empty `call_` (only call untraced primitives). Their comparable params are AST op signatures: `[(op, operand_type), ...]` sequences.

Cross-comp pairs of leaf oF types by op signature → match (shared subsequences), miss (different ops at same position), process distance (positional offset).

Cluster by match → leaf super-types emerge (e.g., `comp_derT` and `comp_A` share sub→abs pattern, cluster into "pairwise_comp" super-type). If instances of same `nF` diverge (different branches taken) → sub-types.

**Level 1 — calls level 0:**
Functions whose `call_` contains only level 0 types. Their signature = sequence of level 0 *type IDs* (not function names — already clustered). Own non-call ops from AST interleave between calls.

So signature of `comp_N` at this level: `[own_ops..., leaf_type_3, own_ops..., leaf_type_1, ...]`

Cross-comp level 1 signatures the same way. `comp_N` and `comp_F` match on shared leaf-type calls, differ on complexity.

**Level 2+:** same recursion upward.

**Concrete implementation in `traced`/`ffeedback`:**

```python
# in ffeedback, after frame completes:

# 1. collect all oF instances by nF
typ_d = {}  # nF -> [oF, oF, ...]
for oF in flatten(Z.call_):  # recursive unpack
    typ_d.setdefault(oF.nF, []).append(oF)

# 2. bottom-up by composition level
def oF_level(oF):
    return 0 if not oF.call_ else 1 + max(oF_level(sub) for tF in oF.call_ for sub in tF.N_)

levels = {}  # level -> [nF groups]
for nF, oFs in typ_d.items():
    lev = oF_level(oFs[0])  # same nF = same level
    levels.setdefault(lev, []).append((nF, oFs))

# 3. cross-comp within each level
for lev in sorted(levels):
    groups = levels[lev]
    # a. sub-typing: within same nF, comp oF instances
    for nF, oFs in groups:
        if len(oFs) > 1:
            # compare call_ structures: sequence of child tF.nF (or type ID after clustering)
            sigs = [call_sig(oF) for oF in oFs]
            # cluster by sig match -> sub-types
            sub_types = cluster_sigs(sigs, oFs)
            # assign _wdt keys: nF:0, nF:1, ...
    
    # b. super-typing: across nFs at same level
    nF_sigs = [(nF, ast_sig(nF)) for nF, _ in groups]  # static AST
    # cluster by match -> super-types
    super_types = cluster_sigs(nF_sigs)
    # store in Z.Ct
```

**Where call_sig extracts the comparable signature:**
```python
def call_sig(oF):
    # sequence of (child_type_id, child.w, child.c) per call
    return [(tF.nF, tF.w, tF.c) for tF in oF.call_]
```

**The match/miss between two sigs** is ordered comparison — same as comp_P on code "slices". Matching positions (same child type) → match from w,c similarity. Different positions → miss.

**Key question:** the comparison params between oF instances of same nF — is it just `call_` structure, or also `dTT` (data processed) and `rTT` (prediction error)? The data params would differentiate by input context (B fork vs agg+), while call_ structure differentiates by code path taken.

I think both: `call_` for structural sub-typing, `dTT/rTT` for contextual sub-typing. They may co-vary (different data triggers different branches), or diverge (same branches, different data). Thoughts?

kwcckw: > Ultimately it should be an additional coef for B fork, and all other forks of a function, but any local performance of that core function is still relevant. This is related to fork clustering we discussed before, when both match and difference between the functions are significant enough to create a new function category. That's oF cross_comp, it has to be resolved anyway.

I see, so actually it should retrieve the weight from the clustered type or "super-type" above.

> The _wdt then indexes by derived type ID rather than function name — sub-types get separate entries, super-types share a combined entry for higher-level eval.

From the code above, it's only compare the same level or depth, and i think depth shouldn't matter here? Because most of the similar operations might not in the same depth?

boris-kz: yes, we have comp_ functions of different depth, but they should probably be compared with incremental depth difference.
so sub-typing is a version of sub+ and super-typing of agg+

boris-kz: Claude:
The comp chain: `comp_derT` → `comp_N` → `comp_F` → `comp_N_` / `comp_C_` — each adds a layer of composition around the same core operation. Same for clustering: `cluster_N` → `cluster_C` → `cluster_P` adds fuzzy membership.

So the cross-comp between depth-adjacent functions (comp_derT vs comp_N) is like rng+ on code: same core op, incremental composition distance. The match is the shared core, the miss is the added wrapper.

And agg+ across function types (comp vs cluster vs eval) compares different operations at same composition level.

This means the comparison shouldn't be strictly within-level as I outlined. Instead:

- **sub+ (sub-typing):** compare instances of same nF, or adjacent-depth functions sharing a core — `comp_N` and `comp_F` share `comp_derT` as core, differ by wrapper complexity. The depth difference is the "range" analog.

- **agg+ (super-typing):** compare across fundamentally different operations at any level — `comp_N` vs `cluster_N` at same depth, yielding "process node" super-type.

The depth-incremental comparison needs a normalization, like `rn` in data comp — deeper functions have more params, so the shared core is a fraction of the total. Match = core overlap, normalized by composition order difference.

This also means the level-by-level bottom-up I described is wrong. It should be more like the data agg cycle: cross-comp all oF types, with distance = depth difference + signature difference, cluster by combined match. Depth is just another dimension of process distance.

Does this also apply to the cluster chain? `cluster_N` → `cluster_C` → `cluster_P` have shared core (membership assignment) with incremental refinement wrapper. So comp_chain and cluster_chain are two "edges" in code space, analogous to edge blobs in image space?

kwcckw: > so sub-typing is a version of sub+ and super-typing of agg+
> sub+ (sub-typing): compare instances of same nF, or adjacent-depth functions sharing a core


Actually i don't quite understand the sub-typing here. It's comparing oFs of same nF, but recluster them into different group of sub oFs again although all of them have the same nF? In that case, how do we define their nF?

boris-kz: We have different flag args and if-else forks in the functions, if the process difference is significant enough we can assign different nF and ws to these forks. It's like automatically unpacking and repacking the functions

boris-kz: Concretely:

A function like `comp_N_` has `fC` flag → two code paths. `val_` has `fi` → two paths. `cluster_N` has internal conditionals on `m > ave` vs `d > avd`.

Each if-else fork is a potential split point. The AST gives you the branch structure statically. Runtime oF instances record which branches were taken (implicitly via `call_` — different branches call different sub-functions).

So the process:

1. AST: identify branch points per function (if-else, flag args)
2. Runtime: group oF instances by which branches they took (detectable from `call_` structure divergence)
3. If difference between groups > ave: split into sub-nFs, each with own `w` in `_wdt`
4. If difference < ave: keep merged, the branch is insignificant

And repacking: if two originally separate functions (e.g. `comp_N_` and `comp_C_`) have sub-forks that converge (same effective code path after stripping their unique wrappers), merge those sub-forks into a shared super-type.

The `_wdt` scope problem resolves naturally: `cross_comp` called from B fork takes a different branch path than from agg+, so they split into `cross_comp:0` and `cross_comp:1` with distinct weights.

One practical question: do we detect branch divergence from `call_` alone (which sub-functions were called), or also need to track which `if` branches were entered? The former is already available, the latter would need explicit branch logging in `traced`.

kwcckw: I see that's actually within a function, so is it right to understand that sub+ is recluster functions within a function only, while agg+ is is between those functions? What makes agg+ different here is it could've across different depth, while sub+ must be from a same depth?

frame_2D_alg/agg_recursion.py
```diff
@@ -138,8 +137,13 @@ def inner(*a, **kw):
...
-                oF.w = abs(m) / (abs(d) or eps)
-            CoF._wdt[oF.nF] = oF.w
+                oF.w = (abs(m) / (abs(d) or eps)) or 1      
+            # from prior discussion, we need Z.N_ exclusively for types, and it should be global?
+            ztF = next((f for f in Z.N_ if f.nF==oF.nF), None)
+            if ztF: ztF.N_ += [oF]
+            else:   ztF = CoF(nF=oF.nF, N_=[oF], w=oF.w, c=oF.c); Z.N_ += [ztF]
+            _sum_vt(ztF, fN=1); _sum_vt(Z, fN=1)  # or run this in batch during feedback for Z?
+            CoF._wdt[oF.nF] = ztF.w  # instead of update from last oF, update from typ's oF.w?
             CoF._cur.set(_CoF)
```
boris-kz, line 147: >sub+ reclusters functions within a function only, while agg+ is is between those functions? What makes agg+ different here is it could've across different depth, while sub+ must be from a same depth?

there is no substantive differentence, just different starting point, we manually unpack and merge those base functions all the time.
same as for regular sub+ and agg+, it's really the same process with different framing

kwcckw: Okay, but for agg+ and sub+, the difference would be the input? Because sub+ recluster nodes, while agg+ clusters the new clusters. For functions, is that the same?

boris-kz: yes, but the initial input here is hand-coded functions

boris-kz: Right, two independent mechanisms:

**Static (AST, before runtime):** parse if-else branches, flag args. Each branch is a code sub-sequence with its own op signature. Compare branch signatures — if difference > ave, pre-assign sub-nFs. This is deterministic, done once at module load in `trace_func`.

**Dynamic (call_, during/after runtime):** oF instances of same nF accumulate different `w, c` from their callees via `_sum_vt`. Cluster by performance divergence in `ffeedback`. This discovers context-dependent sub-types that static analysis can't see.

Static splits the code, dynamic splits the usage. They complement:

- Static split + dynamic convergence → over-split, merge back
- No static split + dynamic divergence → under-split, flag for inspection

So `op_extract.py` handles the static part: parse each function, extract branch sub-signatures, compare. The CoF runtime handles the dynamic part: cluster oF instances by `w` divergence.

For `_wdt`: static sub-types get keys at load time, dynamic sub-types get keys as they're discovered. Both feed the same dict.

Should static splitting be aggressive (split all if-else) and then prune by dynamic convergence, or conservative (split only major branches like `fd` flags)?

kwcckw: So this is a two steps clustering process? Where we have initial clustering based on the static info, then recluster them when we have the runtime w and c?

boris-kz: initial static analysis is a separate process, it won't be in the code. It's actually what LLM is doing, probably better than any hard-coded AST

kwcckw: > initial static analysis is a separate process, it won't be in the code. It's actually what LLM is doing, probably better than any hard-coded AST

Okay, then we only need to cluster them in runtime. At this point, roughly how should we start the comp and clustering process? 

From traced, after each function ended, we update w and c, and then we will unpack them in ffeedback to comp and cluster with cross_comp similar function? The entire process should starts with Z.call_ and flatten the nested calls ? Or it should unpack, comp and cluster at the same time?

boris-kz:
Start with sub+ within each typ.N_, if len(N_)>ave*Le: compare rTTs between
flags/ if forks?

kwcckw:
> Start with sub+ within each typ.N_, if len(N_)>ave*Le: compare rTTs between flags/ if forks?

This should be also within traced right? After all callee ended? And typ.N_ is local but not global?

> compare rTTs between flags/ if forks?

Actually what would bes the flags here? We only get rTTs from comp or clust funcs, but not evals based oF?

PR #106, opened 2026-04-06
4/6/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -135,18 +135,38 @@ def inner(*a, **kw):
             else:
                 TT = getattr(oF,'rTT', oF.dTT)[0]  # rTT includes cluster compression
                 oF.w = vt_(TT,oF.r)[0] if np.any(TT) else 0  # =m
+                
+            for typ in oF.call_:
+                tL_ = []
+                if len(typ.N_)>ave*Le:
+                    for call in typ.N_: call.rim = []; call.fin = 0
+                    for _call, call in combinations(typ.N_, 2):
+                        if hasattr(_call, 'rTT_') and  hasattr(call, 'rTT'):  # func under flags/ if forks should have rTT?
```
kwcckw, line 144: For those functions under flags or if forks, right now we do not have a distinct var to differentiate them yet?

I guess the closest method is by checking if they have added rTT?

frame_2D_alg/agg_recursion.py
```diff
@@ -135,18 +135,38 @@ def inner(*a, **kw):
...
             CoF._cur.set(_CoF)
             return out
         inner.wrapped = True
         return inner
+
+    @staticmethod
+    def _comp_rTT(_oF, oF):
+        _rTT = getattr(_oF, 'rTT'); rTT  = getattr(oF, 'rTT')
+        dTT = comp_derT(_rTT[1], rTT[1]);  m, d = vt_(dTT, (_oF.r + oF.r)/2)
+        L = CoF(nF='rTT', dTT=dTT, m=m, d=d, c=(_oF.c+oF.c)/2, r=(_oF.r+oF.r)/2)  # not sure
```
kwcckw, line 159: I'm not sure if we need L when comparing calls, but if we need to cluster them the same way as cluster_N, it should have L for linking purpose?

frame_2D_alg/agg_recursion.py
```diff
@@ -118,7 +118,7 @@ def get(): return CoF._cur.get(Z)
     def traced(func):
         if getattr(func, 'wrapped', False): return func
         @wraps(func)
-        def sum_crw(F): N_ = F.call_; F.c = C = sum(n.c for n in N_); F.r = sum(f.r* (f.c/C) for f in N_); F.w = sum(f.w* (f.c/C) for f in N_)
+        def sum_crw(F): N_ = F.call_; F.c = C = sum(n.c for n in N_); F.r = sum(f.r* (f.c/(C or eps)) for f in N_); F.w = sum(f.w* (f.c/(C or eps)) for f in N_)
```
boris-kz, line 121: C can't be 0

frame_2D_alg/agg_recursion.py
```diff
@@ -135,18 +135,38 @@ def inner(*a, **kw):
             else:
                 TT = getattr(oF,'rTT', oF.dTT)[0]  # rTT includes cluster compression
                 oF.w = vt_(TT,oF.r)[0] if np.any(TT) else 0  # =m
+                
+            for typ in oF.call_:
+                tL_ = []
+                if len(typ.N_)>ave*Le:
```
boris-kz, line 141: what's the hell is that Le? copy-pasta? it should be (len(typ.N_)-1)*Lw

frame_2D_alg/agg_recursion.py
boris-kz: they all have oF.w. The simplest potential separator is arg flag, then it's some difference in oF components, through AST

frame_2D_alg/agg_recursion.py
boris-kz: right

frame_2D_alg/agg_recursion.py
kwcckw: > The simplest potential separator is arg flag

This should be a variable flag in the oF? Which we need to manually add it after those flags or if forks? I'm not sure if this can be done internally within the actual ast module.

frame_2D_alg/agg_recursion.py
kwcckw: From this line:

     ` for tF in typ_: sum_crw(tF)`

When tF is a typ oF from eval based oF such as vt_, their oF.c is 0?

frame_2D_alg/agg_recursion.py
kwcckw: Okay thanks, i was thinking if that's a new ave.

frame_2D_alg/agg_recursion.py
boris-kz: C = sum(n.c for n in N_). each n must have c. 
If N_ is empty, the oF was never called, there is nothing to trace

frame_2D_alg/agg_recursion.py
boris-kz: You really need to use Claude:

Yes — AST gives you function signatures including default values:

```python
def comp_N_(_pairs, r, tnF=None, root=2):
```

AST extracts `tnF=None`, `root=2` as keyword args with defaults. And inside the body, any `if tnF:` or `if fC:` branch is an `ast.If` node whose test references that arg.

So you can trace: flag arg → which `if` branches it gates → the two sub-signatures under each branch. The link from arg to branch is explicit in AST.

What AST doesn't tell you is what values the flag actually takes at call sites — that's the dynamic part. But it does tell you *which code paths are flag-controlled*, which is enough for static sub-forking.

frame_2D_alg/agg_recursion.py
boris-kz: `traced()` is called at module load for all non-excluded functions — it wraps them. But the oF is created inside `inner`, which only runs when the function is actually called:

```python
def inner(*a, **kw):
    oF = CoF(nF=func.__name__, root=_CoF)  # only here
```

So functions that are never called during a run get no oF, no `_wdt` entry, no performance data. They're invisible to the dynamic part.

But they're still visible to static AST analysis — `op_extract` can parse them regardless. So static comparison covers all functions, dynamic only covers executed ones. Another reason both are needed.

frame_2D_alg/agg_recursion.py
boris-kz: So if call_[typ] is long and valuable enough, we need to segment typ.N_ by same-flag calls and check if oF.w is significantly different between these same-flag segments?
That's pretty expensive, it should be in ffeedback only?

boris-kz: Right, only in `ffeedback`. Per-call overhead in `traced()` should stay minimal.

But there's a gap: `traced()` doesn't record what flag values were passed. The oF only has `nF, call_, w, c`. To segment by flags you'd need to either:

- Record flag args in oF during `inner()`: `oF.args = kw` or just the flag subset. Cheap per-call, but need to know which args are flags.
- Or infer from `call_` structure: different flags → different branches → different sub-call sequences. No extra recording needed, just cluster by `call_` similarity in `ffeedback`.

The second is more consistent — you're clustering by actual process difference, not by assumed flag significance. A flag that doesn't change the call_ tree doesn't matter, and a non-flag arg that does matter gets caught.

So in `ffeedback`:
```python
for typ in Z.call_:
    if len(typ.N_) > ave_N and typ.w * typ.c > ave:  # long and valuable
        # cluster typ.N_ by call_ similarity
        # if >1 cluster with significantly different w: split typ
```

The "call_ similarity" comparison is itself a cross-comp of oF sequences — which brings it back to the same core operation. Does the comparison need to be full cross-comp of call_ trees, or just shallow: compare top-level child nF sequences?

frame_2D_alg/agg_recursion.py
boris-kz: So the pipeline:

**At module load, in `trace_func`:**
1. AST parse each function → find all if-else nodes
2. Compare op signatures of each branch pair
3. Mark forks where sig difference > ave → store as `func.sig_forks`

**Selective instrumentation:** for marked forks only, rewrite the AST to inject fork tracking:

```python
# original:
if fC:
    L_ = comp_C_(N_, rr)
else:
    L_ = comp_N_(combinations(N_,2), rr)

# rewritten by AST:
if fC:
    CoF._cur.get().fk_ += [0]  # fork 0, branch 0
    L_ = comp_C_(N_, rr)
else:
    CoF._cur.get().fk_ += [1]  # fork 0, branch 1
    L_ = comp_N_(combinations(N_,2), rr)
```

Then `oF.fk_` accumulates a sequence of branch IDs taken during that call. In `ffeedback`, segment `typ.N_` by `fk_` patterns, check `w` divergence.

Two questions:

- `oF.fk_` needs to be initialized in `CoF.__init__`: `f.fk_ = []`. Trivial addition.

Three separate stages:

1. **`op_extract.py` (static, offline):** AST parse → find forks → compare branch signatures → mark significant ones → rewrite source with `fk_` injections → output instrumented module

2. **`trace_func` (module load):** wraps already-instrumented functions with `traced()`

3. **`traced()` (runtime):** lightweight — just collects `call_` and `fk_` per oF, computes `w`

So `op_extract.py` is run manually when code changes, like a compiler pass. Its output is the instrumented `agg_recursion.py` with fork tracking baked in. Not run every time.

This also means `op_extract` can be as expensive as needed — full cross-comp of all function signatures, deep branch analysis — since it's a one-time cost. And its output (which forks are significant, what their signatures are) is also usable for the super-typing: the static function similarity data persists independently of runtime.

Is `op_extract.py` already set up for this, or does it currently only extract op counts without the branch comparison and rewriting?

kwcckw: > What AST doesn't tell you is what values the flag actually takes at call sites — that's the dynamic part. But it does tell you which code paths are flag-controlled, which is enough for static sub-forking.

Thanks, so it only contains information on the possible flag's controlled forks, but not which fork is taken. But i think that should be sufficient to let us check and compare those calls if they are in the flag's controlled forks.

Also this requires manual input of flag, which may not feasible to include it in all functions:

      # rewritten by AST:
      if fC:
          CoF._cur.get().fk_ += [0]  # fork 0, branch 0
          L_ = comp_C_(N_, rr)
      else:
          CoF._cur.get().fk_ += [1]  # fork 0, branch 1
          L_ = comp_N_(combinations(N_,2), rr)
          
          
So as the comment below suggested:
  
      clustering by actual process difference, not by assumed flag significance. A flag that doesn't change the call_ tree doesn't matter, and a non-flag arg that does matter gets caught.
      
I think it should be fine since the similarity should be captured by the subsequent calls?

frame_2D_alg/agg_recursion.py
kwcckw: >  Does the comparison need to be full cross-comp of call_ trees, or just shallow: compare top-level child nF sequences?

Since Z.call_ is not global, it only has a single typ for a single call, which is frame_H from main section, so this clustering should be global (recursively unpack Z.call_)?

frame_2D_alg/agg_recursion.py
kwcckw: > C = sum(n.c for n in N_). each n must have c. If N_ is empty, the oF was never called, there is nothing to trace

But i'm referring to the case where n.c is also empty since right now we only assign c in comp or cluster, and for eval based oF such as `vt_`, `val_`, and `sum_vt`, we didn't update those eval based oF.c? And not all of them have data's c too.

For example, if oF is  `sum_vt`, and their call is only a single `vt_`, so both oF.c and their call.c are actually empty.

boris-kz: you can't have opearation without data

frame_2D_alg/agg_recursion.py
boris-kz: I added another reply above

frame_2D_alg/agg_recursion.py
boris-kz: for any typing, we need to flatten the call_ first

kwcckw: > for any typing, we need to flatten the call_ first

That's including the types in oF.call_ within CoF right? Which mean the existing `for sub in oF.call_` is wrong and we need to flatten oF.call_ and all of their nested calls first.

frame_2D_alg/agg_recursion.py
kwcckw: Yes, so i think we probably can skip the usage of that `fk_`?

Because we should cluster by actual process difference, which should be covered by the rest of the calls.

To determine functions under the flags, that should be retrieved from ast since that info is static?

frame_2D_alg/agg_recursion.py
kwcckw: Yes, but the problem is right now we still need to assign their c manually from data, and not all function have data's c? For example, if there's no c assignment from the data in vt_, we should default it to some value? 

      def vt_(TT, r, wTT=wTTn):  # brief val_ to get m,d, rc=0 to return raw vals, Wn for comp_N
      
frame_2D_alg/agg_recursion.py
boris-kz: right

frame_2D_alg/agg_recursion.py
boris-kz: yes, I missed that

frame_2D_alg/agg_recursion.py
boris-kz: actually we only need oF for conditional calls, the evals themselves are not conditional

frame_2D_alg/agg_recursion.py
boris-kz: def flat_(oF):
        call_ = []
        for sub in oF.call_:
            call_ += [sub]
            if sub.call_: call_ += flat_(sub)
        return call_

boris-kz: def flat_(oF, call_=None):

    if call_ is None: call_ = []
    for sub in oF.call_:
        call_ += [sub]
        if sub.call_: flat_(sub, call_)
    return call_

conditional_ = ['add_H','add_Lt','cluster_C','cluster_N','cluster_P','comp_F','ffeedback','frame_H','get_exemplars','mdecay','proj_N','proj_focus','proj_TT',
                'sum2G','sum_H','trace_edge','Q2R','add_Nt','cent_TT','comp_N','comp_N_','cross_comp','sum2C','sum2F']

    name = "func"
    _wdt = {}  # weights
    def __init__(f, **kw):
        super().__init__(**kw)
        f.typ_ = kw.get('typ_', [])  # flattened, nested with tFs, if any
    @staticmethod
    @staticmethod
    def traced(func):
        @wraps(func)
        def sum_crw(F): N_ = F.typ_; F.c = C = sum(n.c for n in N_); F.r = sum(f.r* (f.c/C) for f in N_); F.w = sum(f.w* (f.c/C) for f in N_)
        def inner(*a, **kw):
            _CoF = CoF._cur.get()
            _CoF.call_ += [oF]
                call_,typ_ = flat_(oF),[]
                for F in call_:
                    if F.nF not in conditional_: continue
                    tF = next((f for f in typ_ if f.nF==F.nF), None)
                    if tF: tF.typ_+= [F]
                    else:  typ_ += [CoF(nF=F.nF, typ_=[F])]
                for tF in typ_: sum_crw(tF)
                oF.typ_ = [tF for tF in typ_ if tF.w * tF.c > ave]  # add coef
                if oF.typ_: sum_crw(oF)  # typ_ vals only?
            else:
                oF.w = vt_(TT,oF.r)[0] if np.any(TT) else 0  # m
            CoF._wdt[oF.nF] = oF.w
            CoF._cur.set(_CoF)
            return out
        inner.wrapped = True
        return inner

frame_2D_alg/agg_recursion.py
kwcckw: Okay right, those static information should be in ast anyway, so in actual case they shouldn't be included. I think right now we only can manually exclude them since we can't really differentiate them with the existing CoF.

frame_2D_alg/agg_recursion.py
kwcckw: Since nested callees may have their calls converted to types, we may need to merge deeper typ?

That's mean only the top sub `for sub in oF.call_:` is call oF, while all other deeper sub.call_ should be already converted to typ, and we need to merge them.

boris-kz: in the scheme above call_ is always a tree, the tFs are in oF.typ

kwcckw: I see, thanks. It's actually clustering all same typ's call regardless of their depth now. So eventually Z.typ_ should be a flat list packing all typs, while Z.call_ is remain as a tree.

boris-kz: yes, that should be conditional, oF processing is pretty expensive

frame_2D_alg/agg_recursion.py
kwcckw: Since that branching info only exists in actual ast, and right now we don't have that in CoF, at this point we will just ignore that and compare them all using combination ? 

    for _call, call in combinations(typ.N_, 2):

boris-kz: this is supposed to compare call_ between functions, how do you do that without AST?
we already have all the code, doing it manually is redundant.

kwcckw: Yes, so i guess we can use placeholder here for drafting the logic first.

PR #108, opened 2026-04-08
4/8/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -320,14 +319,15 @@ def comp_C_(C_,_r,_c, _C_=[], fall=1, fC=0):  # simplified for centroids, trans-
             _C = C_[j]; C = C_[j+1]
             if _C in C.compared: continue
             dy_dx = _C.yx-C.yx; dist = np.hypot(*dy_dx)
-            comp_N(_C,C,_r,A=dy_dx,span=dist,L_=L_,N_=N_,acc=acc)  # simplified for typ=3
+            comp_N(_C,C,_r,_c,A=dy_dx,span=dist,L_=L_,N_=N_,acc=acc)  # simplified for typ=3
     for N in list(set(N_)):
         if N.rim: Q2R(N.rim, N.Rt, merge=0, froot=0)
     if L_: Q2R(L_,R=CoF.get())
+    CoF.get().c += _c
```
kwcckw, line 326: So we may get c from L_ above, so input `_c` should be added on top of that? Or we should use input `_c` only if L_ is empty?

frame_2D_alg/agg_recursion.py
```diff
@@ -245,7 +245,6 @@ def cross_comp(Ft, rr, nF='Nt'):  # core function mediating recursive rng+ and d
 '''
 def comp_N_(_pairs, r,_c, tnF=None, root=2):  # incremental-distance cross_comp, max dist depends on prior match
 
-    CoF.get().c += _c
```
boris-kz, line 248: oF.c should be an operand c, not c computed in the function

boris-kz: Different things, both useful:

- **Operand c** = "this function processed N units of data" — measures *scope* of the call. Used for weighting parent aggregation: a function that ran on more data should contribute more to the parent's average w.

- **Computed c** = "this function produced N units of output" — measures *yield*. Compression ratio = computed/operand. That's the actual predictive signal: did the function reduce the data (good clustering) or just pass it through.

For `_wdt` weighting (which gates future calls via `get_exemplars` lookup), you want **yield-relative** — a function that produces a lot from little input is more valuable than one that processes a lot and outputs the same. So `w` should reflect compression, and `c` the output quantity.

Current code already does this in the cluster_N/C/P sites: `oF.c = Ft.root.c - Ft.c` (input minus output = compression amount), `oF.rTT = sel_TT` (selectivity ratio). That's neither pure operand nor pure output — it's the *delta*, which is the right thing for measuring function value.

So the operand-c-via-wrapper approach is wrong for cluster_*. It's right for pure comp functions (comp_N_, comp_C_) where there's no compression, just measurement — operand c = data examined, w = how predictable it was.

Recommendation: don't make it uniform via the wrapper. Let each function set its own c based on what's meaningful for it. The wrapper-injection convenience isn't worth conflating two different semantics.

frame_2D_alg/agg_recursion.py
```diff
@@ -703,7 +707,7 @@ def comb_Ft(Nt, Lt, Bt, Ct, root):  # from sum2G, default Nt
     T = CopyF(Nt)  # temporary accumulator
     dF_ = []
     for Ft in Lt, Bt:  # connectivity forks, Ct is not directly combined and compared, rdn only?
-        if Ft: dF_ += [comp_F(T,Ft, root.r,G)]; T.dTT,T.c,T.r = sum_vt([T,Ft])  # Bt*brrw?
+        if Ft: dF_ += [comp_F(T,Ft,(T.c+Ft.c)/2,root.r,G)]; T.dTT,T.c,T.r = sum_vt([T,Ft])  # Bt*brrw?
```
boris-kz, line 710: in all comps, it should be min c?

frame_2D_alg/agg_recursion.py
boris-kz: it's the same thing, that min c in the input is represented in all links?

frame_2D_alg/agg_recursion.py
kwcckw: If this is operand c, i think we need average since both c will be used?

frame_2D_alg/agg_recursion.py
kwcckw: > That's neither pure operand nor pure output — it's the delta, which is the right thing for measuring function value.

I see, so i guess we need `oF._c` and `oF.c` to differentiate between operand c and computed c? But for most of the computation, i think we only need computed c? Or we actually need their diference `_c - c` in most of the computation since that's the delta?

frame_2D_alg/agg_recursion.py
kwcckw: > it's the same thing, that min c in the input is represented in all links?

As described in the LLM below, there might be some delta? So we actually need 2 vars to pack them here? Or just pack the delta `_c - c`?

boris-kz: that's for clustering, this is comp

frame_2D_alg/agg_recursion.py
boris-kz: it's comp, only shared components are compared, hence min

frame_2D_alg/agg_recursion.py
boris-kz: c is not value, it's the base from which the value is derived

frame_2D_alg/agg_recursion.py
boris-kz: for cluster_, _c is the root.c - resulting Ct.c

claude:
Then it's set *after* clustering, not at the top — same as the existing pattern in cluster_N/C/P:

```python
oF = CoF.get(); oF.c = Ft.root.c - Ft.c  # already there, ~line 494
```

So nothing changes for cluster_*. Only comp_ functions need a top-of-function `CoF.get().c = sum(min(_n.c,n.c) for ...)`.

Two distinct semantics, two distinct write sites, both correct for their fork:

- **comp_**: c = shared base examined (set at entry, before processing)
- **cluster_**: c = compression delta = root.c − output.c (set at exit, after processing)

The wrapper stays generic. Each function knows which kind it is and writes accordingly. No uniform extraction.

One consequence: for cluster_ functions whose gate fails and they exit early without setting `oF.c`, c stays 0 and they look like phantoms to the parent. That's actually correct — a cluster call that produced nothing has zero compression, zero value, should not contribute. So the "no phantom leaves" rule needs softening: phantoms from failed gates are legitimate zero contributors, parent aggregation should skip them (or the c=0 guard in sum_crw is needed after all).

frame_2D_alg/agg_recursion.py
kwcckw: > c is not value, it's the base from which the value is derived

So for comp_N_ or comp_C_ here, when the min c in the input is already represent by all links, we only need the c added via Q2R of L_?

frame_2D_alg/agg_recursion.py
kwcckw: > a cluster call that produced nothing has zero compression, zero value, should not contribute. So the "no phantom leaves" rule needs softening: phantoms from failed gates are legitimate zero contributors, parent aggregation should skip them (or the c=0 guard in sum_crw is needed after all).

This makes sense, so actually their c could be 0 if they didn't cluster anything.

frame_2D_alg/agg_recursion.py
boris-kz: this is the c we need:

frame_2D_alg/agg_recursion.py
boris-kz: that's still wrong, no clustering means that the process is wasted, we have the cost without gain.
that's root.c-0, val ddTT = 0, we compressed everithn, gained nothing

boris-kz: claude— failed cluster is `c = Ft.root.c - 0 = Ft.root.c` (full operand wasted), `w = 0` (no compression achieved). Maximum cost, zero yield. That's a strong negative signal for `_wdt`, not a skip.

So the early-exit branches in cluster_N/C/P need to set c on the way out:

```python
def cluster_N(Ft, _N_, r):
    oF = CoF.get(); oF.c = Ft.root.c  # default: assume failure
    ...
    if G_:
        ...
        oF.c = Ft.root.c - Ft.c  # success: actual compression delta
    return G_, r
```

Or set it once at entry as the worst case, then overwrite on success. Same shape for cluster_C, cluster_P.

Then a "phantom leaf with c=0" only happens for genuinely unreached oFs (gate failed before any work) — and those should probably be pruned from `_CoF.call_` rather than counted as max-loss, since no cost was actually paid. The distinction:

- **Gate fails before processing**: no cost, prune from call_
- **Processing runs, produces nothing**: full cost, c = operand, w = 0

Currently the wrapper can't distinguish these because both paths leave `oF.c=0`. The function itself has to set c to operand-size on entry to mark "I started work" — then if it succeeds it overwrites with the delta, if it fails the operand value persists as max-cost signal.
</details>

boris-kz: Two orthogonal axes:

- **gain** = `w` from dTT/rTT, ≥ 0, measures predictive yield
- **cost** = process expense, ideally wall-clock, in Python unreliable (GC, interpreter noise, JIT-less)

For Python, the practical proxies:
1. **Static AST cost**: count ops per function (loops, comparisons, array ops), weight by estimated unit cost. Constant per function, doesn't reflect runtime branching.
2. **Operand-scaled static cost**: AST op count × `c` (or × `c²` for pairwise comps like comp_N_). This is what you'd get in C anyway — the inner loop dominates, and its trip count is c-derived.
3. **Call count × static**: cheap, ignores per-call data size.

Option 2 is the c-proportional estimate you mentioned. For comp_N_: cost ≈ k_pairs × c_avg, where k_pairs = O(n²) in `combinations`. For cluster_N: cost ≈ k_fill × n × avg_rim_size. Each function has a different scaling law in its operand, derivable from AST loop nesting + the operand it iterates.

Net value = `w · c − cost`. With cost c-proportional, this becomes `(w − k_func) · c` — so the per-function constant `k_func` is just a comp/cluster-specific overhead that subtracts from `w` before c-weighting. That's effectively the existing `ave` subtraction in `vt_`, just calibrated per function instead of global.

Which means: **per-function `ave` is the static cost estimate.** Currently `vt_` uses one global `ave=.3`. If you want cost-aware feedback, replace with `ave_dict[oF.nF]`, derived once from AST analysis (loop depth × operand exponent). No runtime measurement needed.

For C-lang projection later, the same AST analysis maps cleanly to cycle counts since C has no interpreter overhead. The Python prototype doesn't measure real cost but estimates it the same way the C version would, so the value-function stays consistent across languages.

Caveat: pairwise functions (comp_N_, cluster_N's flood-fill) scale superlinearly in operand, not linearly. Cost ≠ k·c, it's k·c^p where p depends on the function. AST can extract p from loop nesting over the same operand. Then `cost = k · c^p`, and net value is `w·c − k·c^p`. Still computable, just not a simple subtraction.

boris-kz: Right — `wc` is already in CF's __init__ (`f.wc = kwargs.get('wc',0)`) but never assigned anywhere. The slot's there waiting.

Usage pattern: `vt_(TT, r, wTT)` currently uses global `ave`. Replace with per-oF effective ave:

```python
def vt_(TT, r, wTT=wTTn, wc=1):
    return m_/t_ @ wTT[0] - ave*wc*r, ad_/t_ @ wTT[1] - avd*wc*r
```

And in `inner` leaf branch:
```python
oF.w = vt_(TT, oF.r, wc=oF.wc)[0] if np.any(TT) else 0
```

Then `wc` per function comes from one-time AST analysis: loop nesting depth + operand exponent. Stored as a class-level dict on CoF, populated once in `trace_func` or alongside it:

```python
CoF._wc = {'comp_N_': 2.0, 'comp_N': 1.0, 'cluster_N': 1.5, ...}
```

And `inner` reads it: `oF.wc = CoF._wc.get(func.__name__, 1)`.

Net value per oF becomes `w·c − ave·wc·r·c` implicitly via vt_, which is the cost-aware comparison you wanted. Cheap functions get low `wc`, expensive ones high — feedback then naturally favors high-yield/low-cost functions over high-yield/high-cost ones.

One thing to confirm: should `wc` also scale with operand exponent at call time (c^p) or stay constant (c^1, baked into ave·wc·r)? Constant is simpler and matches how `ave` already works. Exponent-scaling would need `ave·wc·r·c^(p-1)` which is awkward in vt_'s signature. Probably keep wc as flat per-function multiplier and ignore the exponent — close enough for feedback purposes.

kwcckw: I see, so no clustering is actually full cost with zero gain. Those functions should have full root.c when init, and only replaced with Ft.root.c - Ft.c when there's clustering.

> Then wc per function comes from one-time AST analysis: loop nesting depth + operand exponent. Stored as a class-level dict on CoF, populated once in trace_func or alongside it:

So this wc should be added from the static analysis from the function, and do we need to update it again with feedback later? Or it should be a constant to represent the native weight c of the function?

frame_2D_alg/agg_recursion.py
```diff
@@ -1006,14 +1013,15 @@ def expand_lev(_iy, _ix, elev, T):  # seed tile is pixels in 1st lev, or Fg in h
             else: break
         if T_:
             TT,C,R = sum_vt(T_); R += elev
-            if val_(TT, R/len(T_) + elev, wTT_[0], mw=(len(T_)-1)*Lw) > 0:
-                return T_
+            if val_(TT, R/len(T_) + elev, wTT_[0], mw=(len(T_)-1)*Lw) <= 0: T_ = []; C+0
```
boris-kz, line 1016: C+0?

frame_2D_alg/agg_recursion.py
boris-kz: so, 

    pairs, TT,cm,rm = [],np.zeros((2,9)), 0,0
    for pair in _pairs:  # get all-to-all pre-links
        _N, N = pair
        c = min(N.c,_N.c)  # comp weight
        if N is _N:  # overlap = unit match, no miss, or skip?
            tt = np.array([N.dTT[1],np.zeros(9)]); TT+=tt; cm+=c; rm+=(N.r+_N.r)/2 *c
        else:
            dy_dx = _N.yx-N.yx; dist = np.hypot(*dy_dx)
            pairs += [[dist, dy_dx,_N,N,c]]
    if not pairs: return 0,0,0,r,0,0,0,0

        iTT = (_N.dTT + N.dTT) * Dec
        if abs(vt_(eTT,r)[0]) > ave*wn*r*(_N.r+N.r)/2:  # spec N links
            eTT = proj_N(N, dist, dy_dx, r, dec)  # proj N L_,B_,rim, if pV>0: eTT += pTT?
            eTT+= proj_N(_N,dist, -dy_dx, r, dec)  # reverse direction
        return iTT+eTT

    N_,L_,C,TTd,cd,rd = [],[],0,np.zeros((2,9)),0,0  # any global use of dLs, rd?
    acc = [TT,cm,rm, TTd,cd,rd]
        dist,dy_dx,_N,N, c = pL # rim angl is not canonic
        lr = r+ (N.r+_N.r)/2; m,d = vt_(pTT,lr)  # +|-match certainty
        if m > 0:
            C += c  # from all Ls+ pLs?
            if abs(m) < ave*wn:  # comp if marginally predictable, update N.Rt pair eval, ave / proj surprise value?
            else:
                pL = CN(typ=-1, nt=[_N,N], dTT=pTT,m=m,d=d,c=c,r=lr, angl=np.array([dy_dx,1],dtype=object),span=dist)
                L_+= [pL]; N.rim+=[pL]; _N.rim += [pL]; N_+=pL.nt; acc[0]+=pTT*pL.c; acc[1]+=pL.c; acc[2]+=pL.r*pL.c  # all +ve
                # no oF val, ~= links in clustering
    CoF.get().c += C
    for N in set(N_):
    # call trace:
    if L_ := [n for n in L_ if n.typ>-1]: Q2R(L_, R=CoF.get(), fr=1)  # skip pLs, oF.call_+=[oF], oF.nF='comp_N_', adds data
    TT,cm,rm, TTd,cd,rd = acc

frame_2D_alg/agg_recursion.py
```diff
@@ -320,14 +319,15 @@ def comp_C_(C_,_r,_c, _C_=[], fall=1, fC=0):  # simplified for centroids, trans-
...
             if _C in C.compared: continue
             dy_dx = _C.yx-C.yx; dist = np.hypot(*dy_dx)
-            comp_N(_C,C,_r,A=dy_dx,span=dist,L_=L_,N_=N_,acc=acc)  # simplified for typ=3
+            comp_N(_C,C,_r,_c,A=dy_dx,span=dist,L_=L_,N_=N_,acc=acc)  # simplified for typ=3
     for N in list(set(N_)):
         if N.rim: Q2R(N.rim, N.Rt, merge=0, froot=0)
     if L_: Q2R(L_,R=CoF.get())
+    CoF.get().c += _c
     TTm,cm,rm,TTd,cd,rd = acc
     return L_,TTm,cm, rm/(cm or eps), TTd,cd, rd/(cd or eps)  # L_ is Lm now
```
boris-kz, line 328: def comp_C_(C_,_r, _C_=[], fall=1, fC=0):  # simplified for centroids, trans-N_s, levels

    N_,L_,tc = [],[],0; acc = [np.zeros((2,9)),0,0,np.zeros((2,9)),0,0]
    if fall:
        for _C, C in pairs:
            c = min(C.c,_C.c); tc += c  # comp weight
            if _C is C:                
            else:
        if fC:
            # merge similar distant centroids, they are non-local
            for i, L in enumerate(L_):
                if val_(L.dTT,_r+wn, wTTc,fi=0) < 0:
                    C0,C1 = L.nt
                    Q2R([C0,C1], C0,merge=2)
                    add_Nt(C0, C1.Nt); merg_ += [C1]
                    for l in C1.rim: l.nt = [C0 if n is C1 else n for n in l.nt]
                else: L_ = L_[i:]; break
    else:
        for C in C_: C.compared=set()
        C_ = sorted(C_, key=lambda C: C.dTT[0][np.argmax(wTTc[0])])  # max weight defines eigenvector
        for j in range( len(C_)-1):
            _C = C_[j]; C = C_[j+1]
            if _C in C.compared: continue
            c = min(C.c,_C.c); tc += c; acc[1]+=c  # ?
            comp_N(_C,C,_r,c,A=dy_dx,span=dist,L_=L_,N_=N_,acc=acc)  # simplified for typ=3
    CoF.get().c += tc
    for N in list(set(N_)):
    if L_: Q2R(L_,R=CoF.get())
    TTm,cm,rm,TTd,cd,rd = acc
    return L_,TTm,cm, rm/(cm or eps), TTd,cd, rd/(cd or eps)  # L_ is Lm now

frame_2D_alg/agg_recursion.py
```diff
@@ -320,14 +319,15 @@ def comp_C_(C_,_r,_c, _C_=[], fall=1, fC=0):  # simplified for centroids, trans-
...
 
-def comp_N(_N,N, r, full=1, A=np.zeros(2),span=None, rL=None, L_=None, N_=None, acc=None):
+def comp_N(_N,N, r, _c, full=1, A=np.zeros(2),span=None, rL=None, L_=None, N_=None, acc=None):
```
boris-kz, line 330: `def comp_N(_N,N, r,_c, full=1, A=np.zeros(2),span=None, rL=None, L_=None, N_=None, acc=None):`

    def comp_H(_Nt,Nt, Link):  # tentative pre-comp
        dH, tt, C,R = [],np.zeros((2,9)),0,0
        for _lev, lev in zip([_Nt]+_Nt.H, [Nt]+Nt.H):  # should be top-down
            ltt = comp_derT(_lev.dTT[1],lev.dTT[1])
            lc = min(_lev.c,lev.c); lr = (_lev.r+lev.r)/2; m,d = vt_(ltt,lr)
            dH += [CF(dTT=ltt,m=m,d=d,c=lc,r=lr,root=Link)]
            tt += ltt*lc; C+=lc; R+=lr*lc

    TT= base_comp(_N,N)[0] if full else comp_derT(_N.dTT[1],N.dTT[1])
    CoF.get().c+=_c; m,d = vt_(TT,r)
    L = CN(typ=1, nt=[_N,N], dTT=TT,m=m,d=d,c=_c,r=r, root=rL, exe=1)
    if N.typ > 0 and m > ave*wn:
        dH,htt,C,R = comp_H(_N.Nt, N.Nt, L); r=R  # tentative subcomp
        if m + vt_(htt,R/C)[0] > ave*wn:  # subcomp -> tFs in L:
            if abs(N.typ) ==1:
                for _n,n in product(_N.nt,N.nt): L.Nt.fb_ += [comp_N(_n,n,r,_c)]  # link sub-comp
            else:
                for i,(_Ft,Ft,tnF) in enumerate(zip((_N.Nt,_N.Lt,_N.Bt,_N.Ct),(N.Nt,N.Lt,N.Bt,N.Ct),('Nt','Lt','Bt','Ct'))):
                    if _Ft and Ft:  # sub-comp
                        fc = min(_Ft.c,Ft.c); CoF.get().c += fc  # _c should be base only?
                        dFt = comp_F(_Ft,Ft,fc,r,L); getattr(L,tnF).fb_ += dFt.N_  # tFt feedback
                if fb_: Ft = Q2R(fb_,getattr(L,nF)); Q2R([L,Ft], L,merge=0)
                    # +=trans-links, python-batched bottom-up
    if full:
        angl = [A, np.sign(TT[1] @ wTTn[1])]
    for n, _n in (_N,N),(N,_N):
    if L_ is not None:
        N_+= [_N,N]  # for sum_vt(rim)
        elif L.typ==1 and L.d > avd: acc[3]+=L.dTT*L.c; acc[4]+=L.c; acc[5]+=L.r*L.c  # no pLs?
    return L

frame_2D_alg/agg_recursion.py
```diff
@@ -367,9 +367,10 @@ def comp_H(_Nt,Nt, Link):  # tentative pre-comp
         if   L.m > ave: acc[0]+=L.dTT*L.c; acc[1]+=L.c; acc[2]+=L.r*L.c; L_ += [L];
         elif L.typ==1 and L.d > avd: acc[3]+=L.dTT*L.c; acc[4]+=L.c; acc[5]+=L.r*L.c  # no pLs?
     # add call trace?
+    CoF.get().c += _c + c
     return L
 
-def comp_F(_F, F, ir=0, rL=None):
+def comp_F(_F, F, _c, ir=0, rL=None):
```
boris-kz, line 373: def comp_F(_F, F, _c, ir=0, rL=None):

    ddTT = comp_derT(_F.dTT[1],F.dTT[1]); r=(_F.r+F.r)/2; m,d = vt_(ddTT,r); r+=ir
    dF = CF(dTT=ddTT,m=m,d=d,r=r)
    if _F.nF == F.nF:  # sub-comp
        _N_,N_=_F.N_,F.N_; nF=F.nF; f = nF =='Lt'
        if  _N_ and N_:
            if f: Np_ = [[_n,n] for _n,n in zip(_N_,N_) if _n and n]  # same forks
            if (rL.m+m)/2 * ((len(Np_)-1)*Lw) > ave * wn:
                if f: L_= [comp_F(*Np, min(Np[0].c,Np[1].c), r,rL=dF) for Np in Np_]; TT,C,R = sum_vt(L_)
                else: L_,TT,C,R,_,_,_= comp_N_(Np_,r,nF,rL)
                if L_:
                    Q2R([dF,CF(N_=L_,dTT=TT,c=C,r=R)], dF, merge=2)
                    Q2R([rL,dF], rL, merge=0)
    CoF.get().c += _c + dF.c
    return dF  # no cross-fork N_, no L ext updates?

frame_2D_alg/agg_recursion.py
boris-kz: just read the above

frame_2D_alg/agg_recursion.py
```diff
@@ -514,11 +516,12 @@ def trans_cluster(G):
         oF = CoF.get(); oF.N_=G_; oF.rTT=sel_TT; oF.c = Ft.root.c - Ft.c  # add data
         # combine C_:
         Q2R([C for N in (G_ if G_ else _N_) for C in N.Ct.N_], Ft.root.Ct, froot=0)
+    CoF.get().c += _c
```
boris-kz, line 519: oF.c = Ft.root.c - Ft.c, above

frame_2D_alg/agg_recursion.py
```diff
@@ -686,8 +690,8 @@ def sum2G(ft_, root=None, init=1, typ=None):
         if lm*ld* ((len(L_)-1)*Lw) > ave*avd* (lr+1)*wN:
             V = lm - ave*(lr+1) * wN
             if V > 0:
-                r+=1; E_= get_exemplars({N for L in L_ for N in L.nt},r)
-                G_,r= cluster_C(G.Nt,E_,r) if (mdecay(L_)-decay)*V > ave*wC else cluster_N(G.Nt, E_,r)
+                r+=1; E_= get_exemplars({N for L in L_ for N in L.nt},r,G.c)
```
boris-kz, line 693: c = G.Lt.c

frame_2D_alg/agg_recursion.py
```diff
@@ -619,7 +623,7 @@ def cluster_P(__C_,N_,root):  # Parallel centroid refining, _C_ from cluster_C,
 
     C_ = [c for c in out_ if c]
     dCt= Q2R(set(__C_)-set(C_))  # compression
-    oF = CoF.get(); oF.N_=C_; oF.rTT=dCt.dTT; oF.c = dCt.c  # data
+    oF = CoF.get(); oF.N_=C_; oF.rTT=dCt.dTT; oF.c = dCt.c+_c  # data
```
boris-kz, line 626: added compression = oF.c = _c - dCt.c

frame_2D_alg/agg_recursion.py
```diff
@@ -571,7 +574,7 @@ def cluster_C(Ft, E_, r,_c):  # form centroids by clustering exemplar surround v
         for n in Ft.N_:
             n._C_ = n.rN_; n._m_= n.m_; n._o_= n.o_; n.rN_,n.m_,n.o_ = [],[],[]  # new n.Ct.N_s, combine with v_ in Ct_?
         if oC_+C_ and mat*dif*olp > ave*wC*2:  # if val_(DTT,len(oC_+C_)?
-            oC_ = cluster_P(oC_+C_, Ft.N_, Ft)  # refine all memberships in parallel by global backprop|EM
+            oC_ = cluster_P(oC_+C_, Ft.N_, c, Ft)  # refine all memberships in parallel by global backprop|EM
```
boris-kz, line 577: c? you can't just plug in random shit here

frame_2D_alg/agg_recursion.py
kwcckw: For both comp_N_ and comp_C_. we have `CoF.get().c += C` before the Q2R of L_ into oF, and we actually will replace the comp weight C of oF.

I think we only need either one of them? So `CoF.get().c` should be assigned with `C` and it should be placed under 

     if L_ := [n for n in L_ if n.typ>-1]: Q2R(L_, R=CoF.get(), fr=1)  # skip pLs, oF.call_+=[oF], oF.nF='comp_N_', adds data`
     CoF.get().c = C

frame_2D_alg/agg_recursion.py
kwcckw: Sorry, it should be C = 0 to reset both T_ and C.

frame_2D_alg/agg_recursion.py
kwcckw: Yes, thanks. Clustering should be getting the delta c.

frame_2D_alg/agg_recursion.py
kwcckw: Sorry, this is a typo, it should be the input `_c`.

frame_2D_alg/agg_recursion.py
boris-kz: why under, C is set before that?
but it's not just C, we need to add all data: c,r,TT, so we need to call sum_vt, or even sum2F, for oF?

frame_2D_alg/agg_recursion.py
boris-kz: it's not '_c', we need C computed in cluster_C

frame_2D_alg/agg_recursion.py
boris-kz: claude — `vt_` reads `oF.dTT` (line 140), and CF default is zeros, so without dTT leaf `w=0` regardless of c,r.

You don't need `sum2F` (that's full graph construction with N_ merge, link wiring, etc. — overkill). `sum_vt` is the right primitive, and you already wrap it via `Q2R`. Look at line 290:
```python
if L_ := [n for n in L_ if n.typ>-1]: Q2R(L_, R=CoF.get(), fr=1)
```
That single call already writes `m,d,dTT,c,r` into the oF via `sum_vt`. `comp_N_` is conceptually complete — the other comp/cluster functions just need the same one-liner over whatever local list represents their output (links, centroids, clusters), instead of the manual `oF.c = ...` scattered around.

The cluster functions at lines 519, 590, 628 already set `oF.rTT` directly (which `vt_` prefers via `getattr(oF,'rTT',oF.dTT)`) plus `oF.c` — those are fine, they just also need `oF.r` set the same way.

So the rule reduces to: **every leaf traced function must leave (dTT|rTT, c, r) on `CoF.get()` before returning.** Use `Q2R(N_, R=CoF.get(), fr=?)` when you have a list to aggregate, direct assignment when you've already computed the TT locally (cluster case). Then `vt_` in `traced`'s leaf branch produces a real w, and the rest of the chain works.

kwcckw: I see, thanks. Then your version is correct since Q2R is already aggregated all c from Ls, and it should replace the prior oF.c if L_ is not empty.

frame_2D_alg/agg_recursion.py
kwcckw: So that should be getting C accumulated in each while loop? 

It's a bit unique here since we have c (cnt) from comp and C.c from sum2C. So we need both of them or just the clustered C.c for clustering purpose here?

frame_2D_alg/agg_recursion.py
boris-kz: also we don't need oF.c+=c, that's already done in comp_N_/comp_C_ although for different oFs
not sure we need separate oF for cross_comp?

boris-kz: Keep it. cross_comp's oF is the per-cycle container — without it, recursive cross_comp calls collapse into siblings of comp_N_/cluster_N under whatever called the outermost cross_comp, and you lose the per-cycle aggregation that distinguishes "cycle 1 was productive, cycle 2 wasn't". Also `add_typ_` groups by name, so recursive cross_comp calls naturally bucket together as one tF only if each has its own oF frame.

The point of the previous answer stands: cross_comp's oF gets its (c,r,dTT,w) *automatically* from sum_crw over its direct children (comp_N_, get_exemplars, cluster_N, inner cross_comp). No manual `oF.dTT+=...; oF.c+=...` needed in the body. The oF exists; it just fills itself via the traced subtree rather than via hand-written accumulation.

kwcckw: > not sure we need separate oF for cross_comp?
> without it, recursive cross_comp calls collapse into siblings of comp_N_/cluster_N under whatever called the outermost cross_comp, and you lose the per-cycle aggregation that distinguishes "cycle 1 was productive, cycle 2 wasn't". Also add_typ_ groups by name, so recursive cross_comp calls naturally bucket together as one tF only if each has its own oF frame.

I would think the same, i think we still need it on each new depth since it should preserve each per depth information. 

> also we don't need oF.c+=c, that's already done in comp_N_/comp_C_ although for different oFs

Or the oF.c from cross_comp should be accumulated from their callees via sum_crw only?

boris-kz: right

frame_2D_alg/agg_recursion.py
boris-kz: It's the same as in cluster_N:

oF = CoF.get(); oF.N_=G_; oF.rTT=sel_TT; oF.c = Ft.root.c - Ft.c  # add data

only the root here should be Ct formed by cluster_C, which is Ft here?

boris-kz: if oC_+C_ and mat*dif*olp > ave*wC*2:  # if val_(DTT,len(oC_+C_)?
            oC_+= C_; cc = sum([f.c for f in oC_])
            break

kwcckw: This Ft is the parsed G.Nt but not the Ct clustered from Cs?

boris-kz: I only use it to get base N_ and root, above?

kwcckw: > ```
>     if oC_+C_ and mat*dif*olp > ave*wC*2:  # if val_(DTT,len(oC_+C_)?
>         oC_+= C_; cc = sum([f.c for f in oC_])
>         oC_ = cluster_P(oC_, Ft.N_, cc, Ft)  # refine all memberships in parallel by global backprop|EM
>         break
> ```

Okay, so it should be getting the clustered C.c instead.

boris-kz: so there is still the issue of getting correct oF.r, that's not fully resolved yet

kwcckw: > so there is still the issue of getting correct oF.r, that's not fully resolved yet

Okay, let me go through that later.

PR #109, opened 2026-04-09
4/9/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -234,6 +234,7 @@ def add_oF(name, N_,TT,c,r):
 
 def cross_comp(Ft, rr, nF='Nt'):  # core function mediating recursive rng+ and der+ cross-comp and clustering
 
+    oF=CoF.get(); oF.r=Ft.r ; oF.c=Ft.c  # oF.r and oF,c is empty initially, so direct assignment without * c /C
```
kwcckw, line 237: So looks like we still need to assign oF.r and oF.c here because their callees is empty if L_ is empty, and we won't be summing anything in sum_crw.

frame_2D_alg/agg_recursion.py
```diff
@@ -329,10 +332,12 @@ def comp_C_(C_,_r, _C_=[], fall=1, fC=0):  # simplified for centroids, trans-N_s
...
             _C = C_[j]; C = C_[j+1]
             if _C in C.compared: continue
-            dy_dx = _C.yx-C.yx; dist = np.hypot(*dy_dx)  # tc += c; acc[1]+=c
-            comp_N(_C,C,_r, min(C.c,_C.c),A=dy_dx,span=dist,L_=L_,N_=N_,acc=acc)  # simplified for typ=3
+            dy_dx = _C.yx-C.yx; dist = np.hypot(*dy_dx); c = min(C.c,_C.c); r = (C.r+_C.r)/2  # tc += c; acc[1]+=c
+            tc += c; tr += r*c
+            comp_N(_C,C,_r, c ,A=dy_dx,span=dist,L_=L_,N_=N_,acc=acc)  # simplified for typ=3
     for N in list(set(N_)):
         if N.rim: Q2R(N.rim, N.Rt, merge=0, froot=0)
+    oF = CoF.get(); oF.c = tc; oF.r = tr/tc  # r and c will be replaced if L_ is not empty
```
kwcckw, line 340: > o there is still the issue of getting correct oF.r, that's not fully resolved yet

So oF.r should be added the same as oF.c, except it should be scaled by c during the accumulation  (r*c) and normalized by C (divide by total C) again in the final assignment?

boris-kz:
Right, and same for r/dTT, so we really need to use sun_vt?

kwcckw:
> Right, and same for r/dTT, so we really need to use sun_vt?

So we need dTT in oF? But why sum_vt? We can use Q2R to enable to parameter assignment? Since Q2R calls sum_vt, and sum_vt alone doesn't update the value back to oF.

frame_2D_alg/agg_recursion.py
boris-kz: cross_comp is the root of comp_s below, they set all the data and pass it up to their root?

frame_2D_alg/agg_recursion.py
boris-kz: >So we need dTT in oF? But why sum_vt? We can use Q2R to enable to parameter assignment? Since Q2R calls sum_vt, and sum_vt alone doesn't update the value back to oF.

yes, r/dTT is data value, and summation should be c-proportional
Probably Q2R, that will replace all your edits here?
or this summing is inline, we just need weighted add?

frame_2D_alg/agg_recursion.py
boris-kz: All ancestors get updated automatically, but lazily. Cascade:

- leaf returns → its (c,r,dTT,w) finalized
- leaf's parent returns → sum_crw reads leaf + siblings → parent's (c,r,w) set
- grandparent returns → sum_crw reads parent + siblings → grandparent's set
- … up to Z

By the time the outermost traced call returns, every ancestor has been updated exactly once, in order. No ancestor is skipped; none is updated twice. You don't need to walk the root chain manually.

The only thing you *don't* get is mid-execution visibility — if function F is still running and wants to read its grandparent's current c, the grandparent hasn't aggregated yet. Is that the case you're hitting? If so, what reads an ancestor before its own return?

frame_2D_alg/agg_recursion.py
kwcckw: Actually my edit here is meant for the case when L_ is empty. When L_ is not empty, we should use Q2R, but when L_ is empty (no positive Ls), we should use the all the comp weight and their r instead?

frame_2D_alg/agg_recursion.py
kwcckw: Yes, but that sum_crw is optional:

           if (len(flat_(oF))-1)*Lw > ave: sum_crw(oF, oF.call_) 

So it may not have those data in callees. So it should be default for all or just cross_comp?

frame_2D_alg/agg_recursion.py
boris-kz: we probably need separate op_ for all comps, and we don't need to resum, just

`def add2F(F,N_,TT,c,r): F.N_+=N_; F.c=C=F.c+c; rc=c/C; F.dTT+=TT*rc; F.r+=r*rc `

frame_2D_alg/agg_recursion.py
boris-kz: this sum_crw is currently not adding the process costs: wc, initally wN,wC,wn,wc = 10,20,5,10
we need to add that, then sum_crw can be default?

frame_2D_alg/agg_recursion.py
boris-kz: Conceptually fine as a leaf-write helper, but the weighting is off for an *incremental* update.

`rc = c/C` where `C = F.c + c` is the new chunk's share of the new total. But you're applying it only to the new chunk (`TT*rc`) and leaving the existing `F.dTT` un-rescaled. Result: old contributions stay at full weight, new ones get downweighted. Same for r.

Two coherent options:

1. **Plain accumulation** (what `Q2R`/`sum_vt` effectively do via summing then dividing once): 
```python
def add2F(F,N_,TT,c,r): F.N_+=N_; F.dTT+=TT; F.c+=c; F.r+=r
```
Normalize at the end if needed (or never — dTT stays as a sum, vt_ handles it).

2. **Running weighted mean** (both sides rescaled):
```python
def add2F(F,N_,TT,c,r):
    F.N_+=N_; C=F.c+c; rc_old=F.c/C; rc_new=c/C
    F.dTT = F.dTT*rc_old + TT*rc_new; F.r = F.r*rc_old + r*rc_new; F.c = C
```

Which you want depends on whether dTT/r are sums or means in your scheme. Looking at sum_vt (line 192–195): it sums `(rTT|dTT)*rc` where `rc=n.c/C` after computing `C` over the whole list — that's option 2 done in one batch. Q2R writes `F.dTT` from that, so dTT-as-mean is the convention. Use option 2.

kwcckw: I see, i would think method 2 looks better too since it uses weighted mean on both sides, which is more consistent with our current method.

frame_2D_alg/agg_recursion.py
kwcckw: > we need to add that, then sum_crw can be default?

Okay right, so we should include oF.wc now, and we need to include that when normalizing r?

boris-kz: that should be ave*wc, normalization is for data only

frame_2D_alg/agg_recursion.py
boris-kz: probably can fold it in Q2R:
make add2F the primitive, Q2R becomes a loop over it plus the merge/root bookkeeping:

```python
def Q2R(N_, R=None, merge=1, froot=1, fN=0, fr=0):
    if R is None: R = CN() if fN else CF()
    for n in N_: add2F(R, [], n.rTT if fr else n.dTT, n.c, n.r)
    if merge: ...  # unchanged N_/root/C_ wiring
    return R
```

Then sum_crw is the same loop with `n.w` folded in, sum_vt disappears (callers either use Q2R or add2F directly), and the manual `oF.dTT+=...; oF.c+=...` patterns scattered through cluster code collapse to single add2F calls. One aggregation primitive, three call sites differing only in what bookkeeping wraps it.

The `m,d` outputs that current sum_vt returns alongside (TT,C,R) just become a `vt_(R.dTT, R.r)` call at whichever site needs them — they're derived, not stored separately.

boris-kz: or:

keep sum2F for construction, drop Q2R, callers either use sum2F (when they need a new node) or add2F directly (when updating an existing oF). Q2R's current callers split naturally — the ones with `R=CoF.get()` want add2F; the ones with `R=None` want a lightweight sum2F variant or just a fresh CF + add2F loop.

kwcckw: The second method looks neater, but we would need change all existing Q2R since sum2F or add2F can't be replacing directly? So i would prefer the 1st method since we just need to plugin add2F in the existing Q2R.

frame_2D_alg/agg_recursion.py
kwcckw: Alright, but this cost shouldn't be replacing r right? We still need something like ave * r * wc?

boris-kz: right

frame_2D_alg/agg_recursion.py
boris-kz: Q2R is pretty messy, and we need to combine them anyway, thay are mostly redundant

kwcckw: Okay, in some cases we only need add2F (when fmerge and froot is false), but in some cases we still need something similar to Q2R since we need to merge and update their root, which we will need to create a new function too.

boris-kz: that's sum2F?

frame_2D_alg/agg_recursion.py
```diff
@@ -194,7 +194,7 @@ def sum_crw(F, call_=None):  # no sum wTT?
 def sum2F(N_, nF, root, TT=np.zeros((2,9)), C=0, R=0, fset=1, fCF=1):  # -> CF/CN
```
boris-kz, line 194: def add2F(F, n, fr=0, merge=0):

    a = 'rTT' if fr else 'dTT'
    if F.c:
        C=F.c+n.c; _rc,rc = F.c/C, n.c/C
        F.r = F.r*_rc + n.r*rc; F.c = C
    else:
        setattr(F, a, getattr(n,a)); F.c = n.c; F.r = n.r
    F.N_ += n.N_ if merge else [n]
    if n.H: add_H(F.H, n.H, F)

kwcckw: > that's sum2F?

But we don't need H related accumulation in prior Q2R?

So in the add2F above, it should replace Q2R? But in some cases, we need merge = 0 to not pack anything at all?

boris-kz: this is generic, H should be summed if there is any
merge is packing n.N_, that should be done anyway? 

to replace Q2R and sum_crw we need to extend sum2F, which is calling add2F

boris-kz: def add2F(F, n, fr=0, merge=0, fN=0):

    if F.c:
        C=F.c+n.c; _rc,rc = F.c/C, n.c/C; F.r = F.r*_rc+n.r*rc; F.c = C
        setattr(F,a, getattr(F,a)*_rc + getattr(n,a)*rc)
    else:
        setattr(F,a, getattr(n,a)); F.c = n.c; F.r = n.r
    F.N_ += n.N_ if merge else [n]
    if n.H: add_H(F.H, n.H, F)
    if fN and hasattr(n, 'C_'): F.C_ += n.C_  # same for L_?

def sum2F(N_, nF=None, root=None, R=None, fset=1, fCF=1, fr=0, fN=0, merge=1):  # -> CF/CN

    if R is None:
        R = (CN,CF)[fCF](nF=nF, root=root)
        if fset and root and nF: setattr(root, nF, R)
    for n in N_: add2F(R, n, fr, merge)
    if  fN and R.C_: 
        R.Ct = sum2F(R.C_,fCF=0)  # same for L_?
    return R

kwcckw: > merge is packing n.N_, that should be done anyway?

We may not merge them, for example:

comp_N_:

comp_N:

     Q2R([L,Ft], L,merge=0)
     
 > to replace Q2R and sum_crw we need to extend sum2F, which is calling add2F

So to run the same sum_crw, their call_ or typ_ should be parsed into sum2F above as N_, and oF as R, but this line `F.N_ += n.N_ if merge else [n]` shouldn't be run at all?

boris-kz: yes, that will be merge=2

kwcckw: > yes, that will be merge=2

We probably need something like merge == 0 to not merge at all, 1 = merge [n], 2 = merge N_[1].N_ only, 3 = n.N_?

Because we may need to just merge the second element, the merging in comp_C_:

     Q2R([C0,C1], C0,merge=2)

boris-kz: that Q2R([C0,C1], C0,merge=2) is actually add2F, not sum2F

so 
def add2F(F, n, fr=0, merge=0, fN=0):

    if F.c:
        C=F.c+n.c; _rc,rc = F.c/C, n.c/C; F.r = F.r*_rc+n.r*rc; F.c = C
        setattr(F,a, getattr(F,a)*_rc + getattr(n,a)*rc)
    else:
        setattr(F,a, getattr(n,a)); F.c = n.c; F.r = n.r
    if n.H: add_H(F.H, n.H, F)
    if fN and hasattr(n, 'C_'): F.C_ += n.C_  # same for L_?

?

kwcckw: Okay right. This is actual better since it init c and other params when it's empty, thanks.

boris-kz: so I pushed a bunch of clean-ups, incomplete

boris-kz: didn't review this yet:

One pervasive bug dominates everything else.

---

**1. `sum2F` 2nd-positional-arg bug — ~10 call sites broken**

`sum2F(N_, nF=None, root=None, R=None, ...)` — the 2nd positional maps to `nF` (a string), but many callers pass an existing object intending it as `R` (the accumulator). Result: a new throwaway CF is created, the intended target is never modified, the return value is discarded.

Affected lines:
- **136**: `sum2F(oF.call_, oF, merge=2)` — oF→nF, should be `R=oF`
- **153**: `sum2F(tF.N_, tF, merge=2)` — tF→nF, should be `R=tF`
- **155**: `sum2F(oF.typ_, oF, merge=2)` — oF→nF, should be `R=oF`
- **287, 329, 482**: `sum2F(N.rim, N.Rt, froot=0)` — N.Rt→nF, should be `R=N.Rt`
- **456**: `sum2F(E_, CoF.get(), froot=0)` — same pattern
- **520**: `sum2F([...], Ft.root.Ct, froot=0)` — same
- **705**: `sum2F(dF_, G.Xt)` — same
- **735**: `sum2F(L_, Lt)` — same
- **915**: `sum2F(PPd_, N.Bt)` — same

Lines 289, 330, 513 use `R=` keyword correctly. Lines 238, 588, 644, 921 correctly pass a string as nF.

Fix is mechanical: add `R=` keyword at each site. Or reorder the signature to `sum2F(N_, R=None, nF=None, ...)` so the common case (pass existing accumulator) is the natural 2nd positional.

---

**2. `add_typ_` line 153: summing empty list**

Even after fixing the R= issue, `tF.N_` is `[]` at that point. `tF.typ_` holds the grouped oFs. Should be:
```python
for tF in typ_: sum2F(tF.typ_, R=tF, merge=2)
```

---

**3. Leaf `c` still unset** (from last convo, issue #1)

Lines 137–139 set `oF.w` for leaves but never `oF.c`. Every downstream consumer that divides by `C = sum(n.c)` gets zero. The whole crw pipeline remains a no-op. Still needs a decision on source.

---

**4. `fN` plumbed through add2F/sum2F despite last convo agreeing to drop it**

Lines 199, 209–210 add C_ folding logic. Last convo concluded: handle TTn/TTc/C_ inline at the one call site (frame_H line 1017) rather than plumbing fN through. Low priority but adds dead complexity — `fN=1` is only passed at line 1017 and the `sum2F(R.C_, fCF=0)` at 210 would crash if R is a CF (no `C_` attr). The `hasattr` guard in add2F line 199 papers over this.

---

Issue #1 is the showstopper — it silently makes every traced non-leaf and every rim/typ accumulation a no-op.

kwcckw: Thanks, i will go through them again later.

PR #111, opened 2026-04-11
4/11/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -135,9 +135,13 @@ def inner(*a, **kw):
             CoF._cur.set(oF); out = func(*a, **kw)
             if oF.call_:  # complete at this point
                 if (len(flat_(oF))-1)*Lw > ave*(_CoF.wc+oF.r):
-                    sum2F(oF.call_,oF)  # represents all nested call_s
+                    sum2F(oF.call_,oF)  # represents all nested call_s   
+                    oF.w  = vt_(oF.dTT, oF.wc+oF.r)[0] if np.any(oF.dTT) else 0  # recomputed from summed dTT
```
kwcckw, line 139: If oF.w should be recomputed from summed dTT, then it's actually should be could be default here since it's using dTT or rTT regardless of empty call_?

Also for oF.wc, it won't be using the same dTT for their computation? So we need to accumulate it using *c / C (rc) from their callees?

But if oF.call_ is empty, oF.wc should be default to global wc?

frame_2D_alg/agg_recursion.py
```diff
@@ -193,12 +197,13 @@ def add2F(F, n, fr=0, merge=0):
 
     a = 'rTT' if fr else 'dTT'  # or wTT?
     if F.c:
-        C=F.c+n.c; _rc,rc = F.c/C, n.c/C; F.r = F.r*_rc+n.r*rc; F.c = C
+        C=F.c+n.c; _rc,rc = F.c/C, n.c/C; F.r = F.r*_rc+n.r*rc; F.c = C; 
+        if hasattr(n, 'wc'): F.wc = (F.wc*_rc + n.wc*rc) if F.wc != 0 else n.wc
```
kwcckw, line 201: We probably can include wc accumulation here instead of calling it in CoF above.

frame_2D_alg/agg_recursion.py
```diff
@@ -508,9 +513,9 @@ def trans_cluster(G):
                     G_ += [sum2G(ft_,CF())]; Gt_+= [[tt,c,r]]  # _C=C+c; _rc=C/_C; rc=c/_C; TT=TT*_rc+tt*rc; R=R*_rc+r*rc; C=_C
         if G_:
             for G in G_: trans_cluster(G)  # splice trans_links, merge L.nt.roots
-            C = sum([g[1] for g in Gt_]); tt,_,R= Gt_[0]; TT=deepcopy(tt)
-            for tt,c,r in Gt_[1:]: rc=c/C; TT+=tt*rc; R+=r*rc
```
kwcckw, line 512: So this is skipping the first tt, c and r, and i think we need to include it and apply the same rc too?

frame_2D_alg/agg_recursion.py
boris-kz: oF.wc is a completely separate thing, it should refined by AST analysis
yes, w should be default:

                    sum2F(oF.call_,oF)  # represents all nested call_s, add wc?
            oF.w = vt_(TT, oF.wc+oF.r)[0] if np.any(TT) else 0  # recompute

frame_2D_alg/agg_recursion.py
boris-kz: right

frame_2D_alg/agg_recursion.py
kwcckw: But AST analysis is static, and process cost should be adjusted by runtime process?

boris-kz: yes, but Python delays are inconsistent and don't accurately represent process complexity.
So we need to use static values in Tw_: initially wN,wC,wn,wc, but it should have an entry for each oF typ.
So oF.nF should really be just an index in Tw_? 
oF.wc will be dynamically summed from static Tw_[oF.nF] in its call_

kwcckw: > So oF.nF should really be just an index in Tw_?
> oF.wc will be dynamically summed from static Tw_[oF.nF] in its call_

Okay, so using this method, Tw_ will be served as an incremented value when there's a matching oF.nF?

Also  oF.nF is a string, which mean Tw_ should be a dict to link each function name to their wc?

boris-kz: oF.nF should be integers now, mapping to onF_

kwcckw: Okay, and right now Tw_ has wN,wC,wn,wc:

    Tw_= [wN,wC,wn,wc]
    
So cross_comp, comp_N_ should be mapped to wn, comp_C should be mapped to wc and so on.

For those functions like proj_N, ffeedback, vect_edge, trace_edge, they will be mapped to the default wn?

boris-kz: we need to set different values for them, just by guessing their complexity

kwcckw: > we need to set different values for them, just by guessing their complexity

Okay, i guess that should be feasible by using ast , probably compute based on the number of operations in each function, such as number of if branch, mult, div or sub functions.

boris-kz: so to use batching we need to unpack add2F in sum2F:


def add2F(F, n, fr=0, merge=0):  # unpack for batching in sum2F

    if F.c:
        C=F.c+n.c; _rc,rc = F.c/C, n.c/C; F.r = F.r*_rc + n.r*rc; F.c = C
        if hasattr(n,'wc'): F.wc = (F.wc*_rc + n.wc*rc) if F.wc else n.wc
    else: setattr(F,a,getattr(n,a)); F.c=n.c; F.r=n.r
    if getattr(n,'H',None): add_H(F.H, n.H, F)
    if hasattr(n,'C_'): F.C_ = getattr(F, 'C_', []) + n.C_  # same for L_?

def sum2F(N_, root, fr=0, merge=0, froot=0):  # -> CF/CN

    _C=root.c; n=N_[0]; C=n.c; TT,R,wC = np.zeros((2,9)),0,0 
    vt_ = [[n.rTT if fr else n.dTT], n.c, n.r, n.wc, n.H]
    for n in N_:
        vt_ += [[n.rTT if fr else n.dTT], n.c, n.r, n.wc, n.H]
        if merge: root.N_ += n.N_ if merge>1 else [n]
        if froot: n.root = root
    for tt,c,r,wc,H, r in N_:
        rc = c/C; TT += tt*rc; R+=r*rc; wC+=wc*rc; add_H(root.H, H, root, rc)
    new = CF(N_=N_, dTT=TT, c=C, r=R/C, wc=wC, H=H)
    if _C: add2F(root,new)
    else: root=new
    if getattr(root,'C_',None):  # same for L_,H, etc?
        root.Ct = sum2F(root.C_,CF(root=root))  # external init R, setattr(root, nF,R)
    return root

frame_2D_alg/agg_recursion.py
boris-kz: right

frame_2D_alg/agg_recursion.py
boris-kz: sum2F is broken in several ways, and it doesn't actually batch. Issues:

**1. C is wrong.** `C=n.c` takes only the first node; never summed across `N_`. Should be `C = sum(n.c for n in N_)` like in `sum_vt`.

**2. Double-count of N_[0].** `vt_` is pre-seeded with `N_[0]` then the first loop appends `N_[0]` again.

**3. `vt_` shadows the global `vt_` function** in this scope. Bug waiting to happen.

**4. The unpack loop iterates the wrong list.** `for tt,c,r,wc,H in N_:` iterates `N_` (nodes), not `vt_`. Nodes aren't 5-tuples → crash, or silent garbage if any node is iterable.

**5. Even if it iterated `vt_`, the structure is wrong.** `vt_ += [[tt], c, r, wc, H]` flattens into a single list, not a list of tuples. Should be `vt_ += [(tt,c,r,wc,H)]` (or just skip the staging entirely).

**6. R is divided twice.** Inside the loop `R += r*rc` already produces the c-weighted mean. Then `r=R/C` divides by C again. `sum_vt` correctly leaves it as `R`.

**7. `H=H` in the new CF** is the last node's H, not the accumulated history. And `add_H(root.H, H, root, rc)` mutates `root.H` *before* `root` gets reassigned to `new` in the `_C=0` branch — so the accumulation lands on the wrong object.

**8. Not actually batched.** Building `vt_` as a Python list and then looping over it is the same cost as looping over `N_` directly. Real batching means stacking into ndarrays once:

```python
C_ = np.array([n.c for n in N_]); C = C_.sum(); rc = C_/C
TT = np.einsum('i,ijk->jk', rc, np.stack([(n.rTT if fr else n.dTT) for n in N_]))
R  = rc @ np.array([n.r  for n in N_])
wC = rc @ np.array([n.wc for n in N_])
```

That's one numpy reduction per field instead of a Python accumulator. For small `N_` it won't matter, but it's the only thing that justifies calling it "batched" vs `sum_vt`'s straightforward loop.

**Suggestion:** the cleanest fix is to make `sum2F` structurally identical to `sum_vt` (compute C first, single pass with `rc = n.c/C`) and only switch to the np.stack form once you actually want vectorization. The current intermediate `vt_` staging buys nothing and introduces the bugs above.

One question before I touch it: do you want me to (a) just fix the bugs while keeping the current shape, or (b) rewrite it in the true-batched np.stack form?

kwcckw: I think the 1 and 2 are relevant. So we can init C and vt_ as empty and accumulate them in the loop? We don't need n anyway.

boris-kz: most of them this is what I got:

def sum2F(N_, root, fr=0, merge=0, froot=0):  # -> CF/CN

    a = 'rTT' if fr else 'dTT'
    c_ = np.array([n.c for n in N_], dtype=float); C = c_.sum(); rc_ = c_/C
    wC = rc_ @ np.array([n.wc for n in N_])
    if merge:
        for n in N_: root.N_ += n.N_ if merge>1 else [n]
    if froot:
        for n in N_: n.root = root
    new = CF(N_=N_, c=C, r=R, wc=wC, **{a: TT})
    sum_H(N_, new)  # add weighting?
    if root.c: add2F(root,new)
    else:      root = new
    if getattr(root,'C_',None):
    return root

boris-kz: claude:
def sum_H(N_, Ft):

    maxL = max((len(N.Nt.H) for N in N_ if hasattr(N,'Nt')), default=0)
    Ft.H = []
    for i in range(maxL):
        lev_, c_ = zip(*[(N.Nt.H[i], N.c) for N in N_ if hasattr(N,'Nt') and i<len(N.Nt.H)])
        c_ = np.array(c_,dtype=float); rc_ = c_/c_.sum()
        Ft.H += [sum2F(list(lev_), CF(), rc_=rc_)]
    # top lev from N_'s own N_:
    top = [n for N in N_ for n in N.N_]
    if top: Ft.H = [sum2F(top, CF())] + Ft.H

frame_2D_alg/agg_recursion.py
```diff
@@ -135,9 +135,13 @@ def inner(*a, **kw):
...
-                    sum2F(oF.call_,oF)  # represents all nested call_s
+                    sum2F(oF.call_,oF)  # represents all nested call_s   
+                    oF.w  = vt_(oF.dTT, oF.wc+oF.r)[0] if np.any(oF.dTT) else 0  # recomputed from summed dTT
+                    C = sum(f.c for f in oF.call_)
+                    oF.wc = sum(f.wc * (f.c/C) for f in oF.call_)  # not sure
             else:
                 TT = getattr(oF,'rTT', oF.dTT)  # rTT includes cluster compression
+                oF.wc =  sum(f.wc * (f.c/C) for f in oF.call_)  # or should be default to global wc?
                 oF.w = vt_(TT, _CoF.wc+oF.r)[0] if np.any(TT) else 0  # ~m
             CoF._wdt[oF.nF] = oF.w
```
boris-kz, line 146: onF_ = ['cross_comp','comp_N_','comp_C_','comp_N','comp_F','proj_N',  # comp_ functions
        'get_exemplars','cluster_N','cluster_C','cluster_P','frame_H','vect_edge','trace_edge','ffeedback'  # clust_ functions
        ]
Fw_ = [5,10,20,10,8,  5,10,...]  # maps to onF_
Fwc_= copy(Fw_)  # init as oF.wc_, accum from call_?

frame_2D_alg/agg_recursion.py
kwcckw: Actually what is the exact batched operation we need here? 

This ' wC = rc_ @ np.array([n.wc for n in N_]) ' should be the same where we multiple each rc with wc before summing them?

frame_2D_alg/agg_recursion.py
kwcckw: I think we need another section to init Fw_ from ast analysis of their function?

boris-kz: yes, in terms of relative complexity, try that.
edited:

onF_ = ['cross_comp','comp_N_','comp_C_','comp_N','comp_F','proj_N',  # comp_ functions
        'get_exemplars','cluster_N','cluster_C','cluster_P','frame_H','vect_edge','trace_edge','ffeedback'  # clust_ functions
        ]
Fw_ = [5,10,20,10,8,5, 5,10,20,30,30,10,15,10]  # maps to onF_
Fc_ = copy(Fw_)  # oF.wc_, init oF.w_, accum from call_?

frame_2D_alg/agg_recursion.py
boris-kz: yes, weighted summation for TT, r, wc

kwcckw: Okay, so we not really need this changes? I think the prior code is actually clearer.

frame_2D_alg/agg_recursion.py
kwcckw: Okay, initially I will try to check based on the number of operations , and each static analysis of function wc should be independent of deeper callees wc?

frame_2D_alg/agg_recursion.py
boris-kz: it was wrong, much longer and slower

frame_2D_alg/agg_recursion.py
boris-kz: yes, and express it as a multiple of vt_, that's what basic ave should be defined for

kwcckw: > yes, and express it as a multiple of vt_, that's what basic ave should be defined for

Okay, so vt_ will be served as the based wc with value of 1 right? So the other will be approximated based on the complexity in vt_.

boris-kz: right.

PR #112, opened 2026-04-12
4/12/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -112,10 +117,47 @@ def flat_(oF, call_=None):
         if sub.call_: flat_(sub, call_)
     return call_
 
+
+# init function‘s wc, using vt_'s complexity as base unit
+def init_wc():
+    # compute weights based on operations in function, independent of deeper callees
+    def get_ops(node):
+        # suggested by LLM
+        weights = {
```
kwcckw, line 126: So using suggestions from LLM, i'm using operations below with variable weights. Or it's actually better with fixed weight?

frame_2D_alg/agg_recursion.py
```diff
@@ -128,15 +170,15 @@ def traced(func):
...
             CoF._cur.set(oF); out = func(*a, **kw)
             if oF.call_:  # complete at this point
                 if (len(flat_(oF))-1)*Lw > ave*(_CoF.wc+oF.r):
                     sum2F(oF.call_,oF)  # represents all nested call_s
             TT = getattr(oF,'rTT', oF.dTT)  # rTT includes cluster compression
             oF.w = vt_(TT, oF.wc+oF.r)[0] if np.any(TT) else 0  # recompute
-            _CoF._W_[oF.nF] = [oF.w]  # nF is index
-            _CoF._C_[oF.nF] = [oF.wc]
+            CoF._W_[oF.nF] += oF.w  # nF is index
+            CoF._C_[oF.nF] += oF.wc
```
kwcckw, line 181: To pack it into the var and access it globally, i think it should be packed into `CoF` instead of `_CoF`?

frame_2D_alg/agg_recursion.py
```diff
@@ -188,6 +225,7 @@ def cross_comp(Ft, rr, nF='Nt'):  # core function mediating recursive rng+ and d
     L_,TT,c,r,TTd,cd,rd = comp_C_(N_,rr,fC=1) if fC else comp_N_(combinations(N_,2),rr)
     if L_:  # Lm_, no +|- Ft.Lt?
         M, D = vt_(TT, r, TTw(Ft.root, fC+2))
+        # there's a problem with the usage of IGE and CoF._W_ here
         if M * ((len(L_)-1)*Lw) * CoF._W_[iGE] > ave:  # or CoF._wdt[onF_.index('get_exemplars')]?
```
kwcckw, line 229: This is getting weight of `get_exemplars`, but `CoF_W_` of `get_exemplars` won't be added until we call it. So we need to preinit all `CoF._W_` through ast analysis first?

frame_2D_alg/agg_recursion.py
boris-kz: How are they variable? There must be generic Python estimates of delay per op, that's the only thing we need in Fw_?

frame_2D_alg/agg_recursion.py
boris-kz: this is accumulated from `_CoF.call_`, CoF is oF, you are assigning it to itself? 
How is any of that global?

frame_2D_alg/agg_recursion.py
boris-kz: Fc_ is fixed complexity weights from AST 
Fw_ is average gain from actual calls
So: `if M * ((len(L_)-1)*Lw) * Fw_[iGE] > ave * Fc_[iGE]`

frame_2D_alg/agg_recursion.py
boris-kz: Yes, several sources, with varying precision:

**Python-level (most relevant for you):**
- `dis` module + per-bytecode timing: nobody publishes a canonical table, but people have measured it. CPython's `ceval.c` has rough cost classes, and there are blog-post benchmarks (Victor Stinner, pyperformance) timing individual bytecodes.
- `timeit` micro-benchmarks per AST node type — what most "Python performance" tables in books are built from. Order-of-magnitude only; varies 2-5× across CPython versions and 10×+ between CPython/PyPy.
- For numpy-heavy code (yours), per-op cost is dominated by array size, not the AST node — a single `np.einsum` line can outweigh hundreds of pure-Python ops. AST-counting will mislead you here.

**C-level:**
- Agner Fog's instruction tables (https://www.agner.org/optimize/) — cycle counts per x86 instruction, the standard reference. Too low-level for AST mapping but it's the ground truth.
- LLVM's cost model (`TargetTransformInfo`) — used by the optimizer to estimate instruction costs. Accessible via `llvm-mca`.
- Compiler textbooks (Muchnick, Cooper & Torczon) give generic "cost = 1 for arithmetic, 3-5 for memory, 10-30 for branch mispredict, 100+ for cache miss" tables.

**Honest assessment for your use case:** static AST cost estimation is unreliable for numpy code because the dynamic cost is dominated by array dimensions known only at runtime. If `Fc_` is meant to predict actual delay, you'd get better numbers from a one-time runtime profile (`cProfile` or wrapping `traced()` to record wall time) than from AST analysis. AST gives you *structural* complexity (good for "how much work might this do"), runtime gives you *measured* complexity (good for "how much work did this do on representative inputs").

If you want the AST route anyway, the simplest defensible scheme is: count nodes weighted by class — `BinOp=1, Call=5, For/While=10×body, Comprehension=8×body, Attribute=2` — calibrated once against `timeit` on a few representative functions. That's what `op_extract.py` is presumably building toward; want me to look at it?

boris-kz: So we only need ast to get ops count per function, and then get an estimated combined delay in C, not Python
it's static, we don't need to put it agg_recursion

boris-kz: Sketch:

```python
import ast, inspect
COST = {'BinOp':1,'UnaryOp':1,'Compare':1,'BoolOp':1,
        'Subscript':3,'Attribute':2,'Name':0,
        'Call':5,'For':10,'While':10,'ListComp':8,'DictComp':8,'GeneratorExp':8,
        'If':1,'IfExp':1,'Return':0,'Assign':1,'AugAssign':2}

def func_cost(func):
    tree = ast.parse(inspect.getsource(func))
    return sum(COST.get(type(n).__name__, 0) for n in ast.walk(tree))
```

That gives you a scalar per function in ~10 lines. Run it once over `onF_` at startup to populate `Fc_`.

Limitations of this naïve version:
- Doesn't multiply loop body cost by trip count (treats `For` as cost 10 regardless of body) — fix by recursing into `node.body` and multiplying.
- Doesn't distinguish numpy calls from pure-Python calls (`np.einsum` and `len` both count as `Call=5`) — fix by inspecting `node.func` for `Attribute` with `value.id=='np'`.
- No symbolic `n` — gives one fixed scalar, not a function of input size.

For "rough estimates" the 10-line version is probably enough. The fancier version with loop-body multiplication is maybe 30 lines. Want that one instead?

frame_2D_alg/agg_recursion.py
```diff
@@ -112,10 +117,47 @@ def flat_(oF, call_=None):
...
+    return [(get_ops(funcs[name]) /base) for name in onF_]
+ 
+Fw_ = init_wc()  # maps to onF_
+Fc_ = Fw_.copy()  # oF.wc_, init oF.w_, accum from call_?
+
 class CoF(CF):  # oF/ code fork, N_,dTT: data scope, w = vt_(wTT)[0]?
     name = "func"
     _cur = contextvars.ContextVar('oF')
-    _W_,_C_ = [],[]  # sum called oF weights and costs, global Fw_ is not summed
+    _W_,_C_ = np.zeros(len(onF_)),np.zeros(len(onF_))  # sum called oF weights and costs, global Fw_ is not summed
```
boris-kz, line 160: we just need to copy global Fw_ and Fc_ here

boris-kz: no, this is from actual calls, init [] is fine

frame_2D_alg/agg_recursion.py
boris-kz: _CoF._W_ += [Fw_[oF.nF]]  # if nF is index in onF_
            _CoF._C_ += [Fc_[oF.nF]]

frame_2D_alg/agg_recursion.py
```diff
@@ -645,7 +683,7 @@ def sum2F(N_, root, fr=0, merge=0, froot=0):  # -> CF/CN
     c_ = np.array([n.c for n in N_], dtype=float); C = c_.sum(); rc_ = c_/C
     TT = np.einsum('i,ijk->jk', rc_, np.stack([getattr(n,a) for n in N_]))  # weighted sum
     R  = rc_ @ np.array([n.r  for n in N_])
-    wC = rc_ @ np.array([n.wc for n in N_])
+    wC = rc_ @ np.array([getattr(n,'wc',0) for n in N_])
```
boris-kz, line 686: this is a multiplier, so default  = 1

frame_2D_alg/agg_recursion.py
kwcckw: > So we only need ast to get ops count per function, and then get an estimated combined delay in C, not Python
it's static, we don't need to put it agg_recursion


I see, so among the options here:

      C-level:
      
          Agner Fog's instruction tables (https://www.agner.org/optimize/) — cycle counts per x86 instruction, the standard reference. Too low-level for AST mapping but it's the ground truth.
          LLVM's cost model (TargetTransformInfo) — used by the optimizer to estimate instruction costs. Accessible via llvm-mca.
          Compiler textbooks (Muchnick, Cooper & Torczon) give generic "cost = 1 for arithmetic, 3-5 for memory, 10-30 for branch mispredict, 100+ for cache miss" tables.

I think LLVM's cost model or Compiler textbooks should be more relevant, so we only need to retrieve the cost for those ops such as BinOp, UnaryOp, etc from there and define it in Python right?

frame_2D_alg/agg_recursion.py
kwcckw: > ```
>         _CoF._W_ += [Fw_[oF.nF]]  # if nF is index in onF_
>         _CoF._C_ += [Fc_[oF.nF]]
> ```

> no, this is from actual calls, init [] is fine

In the `inner` function, `CoF` is the class, `oF` is the current function's CoF, while the `_CoF` is the root to the `oF`?

This `_CoF._W_ += [Fw_[oF.nF]]` is packing `Fw_[oF.nF]` to `oF.root._W_`, which we can't access it globally later.

Also `_CoF._W_ += [Fw_[oF.nF]]` is packing the Fw_ value into `_W_`, but this doesn't ensure a correct index?

frame_2D_alg/agg_recursion.py
kwcckw: Okay right, thanks.

frame_2D_alg/agg_recursion.py
boris-kz: right, and add ast only when we have code modification, which will be a higher module

frame_2D_alg/agg_recursion.py
```diff
@@ -128,15 +170,15 @@ def traced(func):
         @wraps(func)
         def inner(*a, **kw):
             _CoF = CoF._cur.get()
-            oF = CoF(nF=onF_.index(func.__name__), root=_CoF); _CoF.call_+=[oF]
+            oF = CoF(nF=onF_.index(func.__name__), root=_CoF); oF.wc = Fc_[onF_.index(func.__name__)]; _CoF.call_+=[oF]
```
boris-kz, line 173: oF.wc = Fc_[oF.nF]

frame_2D_alg/agg_recursion.py
```diff
@@ -128,15 +170,15 @@ def traced(func):
...
             CoF._cur.set(oF); out = func(*a, **kw)
             if oF.call_:  # complete at this point
                 if (len(flat_(oF))-1)*Lw > ave*(_CoF.wc+oF.r):
                     sum2F(oF.call_,oF)  # represents all nested call_s
             TT = getattr(oF,'rTT', oF.dTT)  # rTT includes cluster compression
             oF.w = vt_(TT, oF.wc+oF.r)[0] if np.any(TT) else 0  # recompute
```
boris-kz, line 179: oF.w += vt_(TT, oF.wc+oF.r)[0] if np.any(TT) else 0  # default Fw_[oF.nF]?

frame_2D_alg/agg_recursion.py
boris-kz: it all propagates up to Z, we can get globals from it
ok, so if _W_ is mapped to Fw_, init np.zeros is correct,

            _CoF._W_[oF.nF] += [Fw_[oF.nF]]  # if nF is index in onF_
            _CoF._C_[oF.nF] += [Fc_[oF.nF]]

frame_2D_alg/agg_recursion.py
kwcckw: So we have default Fc_ from static analysis, and Fw_ should be init the same default value as Fc_?

frame_2D_alg/agg_recursion.py
kwcckw: Thanks, your version above is better.

frame_2D_alg/agg_recursion.py
boris-kz: right, because average value is neutral

frame_2D_alg/agg_recursion.py
boris-kz: chat:

Yes. Use `python.ast` as a **weighted op counter**, not as a literal timing model.

This only makes sense if you are treating the Python source as a scalar, typed, C-like subset. Python’s `ast` has separate nodes for `BinOp`, `UnaryOp`, `BoolOp`, `Compare`, `Call`, `Attribute`, `Subscript`, and `Name`, and the grammar can change across Python versions, so pin the interpreter version. A `Compare` node may contain several comparisons, a `BoolOp` several joined terms, and `Name` records `Load/Store` context. ([[Python documentation](https://docs.python.org/3/library/ast.html)][1])

Take `1 unit` = one cheap scalar ALU op on hot locals. A workable first pass:

* `Add, Sub, BitAnd, BitOr, BitXor, UAdd, USub, Invert, Eq, NotEq, Lt, LtE, Gt, GtE, Is, IsNot`: `1`
* `LShift, RShift, Not`: `1`
* `Mult`: `3`
* `Div, FloorDiv, Mod`: `20`
* `Pow, MatMult, In, NotIn`: treat as opaque/high-cost unless you know the lowering
* `Name Load/Store`: `0–1`
* `Attribute`: `2–4`
* `Subscript`: `4–8`
* `Assign`: `1`
* `AugAssign`: `1 + op_weight`
* `BoolOp(And/Or)`: `1` per join + branch penalty
* `If / IfExp`: test cost + branch penalty `~3`
* `For / While / comprehension`: body cost per iteration + loop overhead `~2`
* `Call`: `25` minimum, or just mark opaque
* container creation (`List`, `Dict`, `Set`, `Tuple` materialization): `20+` plus element costs

So the rough model is:

`cost ≈ Σ dynamic_count(op) * weight(op)`

not raw `ast.walk()` node count.

Count from AST like this:

* `BinOp`: count the operator once.
* `Compare`: count `len(node.ops)`, not `1`.
* `BoolOp`: count `len(node.values) - 1`.
* `Call`: count one call site, plus argument-expression costs.
* `Attribute` and `Subscript`: count every occurrence as memory-ish access.
* `Name`: split by `ctx=Load/Store` if useful.
* Ignore wrapper/syntax nodes like `Module`, `Expr`, `Load`, `Store`.

Two main failure modes:

* Static AST counts are almost useless without **trip counts** for loops/comprehensions.
* `ast` does not tell you types, cache locality, inlining, vectorization, or branch predictability.

If you want the coarsest usable formula, start with:

`1*(add + sub + cmp + bitop + shift + not) + 3*mult + 20*(div + floordiv + mod) + 4*(subscript + attribute) + 3*branches + 25*calls`

then multiply by estimated loop iteration counts.

The next refinement that usually matters is **cost per loop iteration**, not whole-program static cost.

[1]: https://docs.python.org/3/library/ast.html "ast — Abstract syntax trees — Python 3.14.4 documentation"

frame_2D_alg/agg_recursion.py
kwcckw: Okay, this looks correct , but could you clarify if `_CoF` is meant to map into Fw_, but it's the accumulated values?

If yes, we really should use `CoF` instead of `_CoF` here since from `_CoF = CoF._cur.get()`, `_CoF` is the root oF function that calls the current oF.

boris-kz: yes, _CoF sums from specific calls, it should be a root
but we have a bunch of bugs:

Issues:

1. **`_W_,_C_` shared mutable class attrs.** As `np.zeros(...)` at class level they're shared across all CoF instances. `_CoF._W_[oF.nF] += ...` mutates the single shared array — every instance sees every increment. If you want per-call-tree accumulators, init in `__init__`. If global, drop the `_CoF.` and just write `CoF._W_[oF.nF] += ...`.

2. **`+= [Fw_[oF.nF]]` is leftover list syntax.** `_W_` is now an ndarray, so element `+=` takes a scalar: `+= Fw_[oF.nF]`. As written it'll raise on the list.

3. **`oF.w += ...`** — `w` starts at 0 from CF init and you only assign here, so `+=` and `=` are equivalent on first pass. But if `traced` ever re-wraps or an oF gets revisited, `+=` silently double-counts. Keep `=` unless there's a reason.

4. **`oF.wc = Fc_[oF.nF]` then `oF.wc+oF.r` in the gate** — fine, but note `_CoF.wc` in the gate condition is the *parent's* wc, which now also starts from `Fc_[parent.nF]` rather than 0. Intentional? It changes the threshold meaningfully on the first nested call.

5. Minor: comment "default Fw_[oF.nF], get ave?" — if TT is empty you get `oF.w = 0`, not `Fw_[oF.nF]`. If you want the static prior as fallback, write it explicitly: `else Fw_[oF.nF]`.

kwcckw: Yes, for 1, i think we just need to init `_W_` and `_C_` per instance：

        name = "func"
        def __init__(f, **kw):
            super().__init__(**kw)
            f.typ_ = kw.get('typ_', [])  # flattened and nested with tFs
            f._W_,f._C_ = np.zeros(len(onF_)),np.zeros(len(onF_)) 
            
 For 2, it should increase the scalar value, so why we need the bracket?
 
      _CoF._W_[oF.nF] += Fw_[oF.nF]

boris-kz: right, we don't

frame_2D_alg/agg_recursion.py
```diff
@@ -645,7 +683,7 @@ def sum2F(N_, root, fr=0, merge=0, froot=0):  # -> CF/CN
     c_ = np.array([n.c for n in N_], dtype=float); C = c_.sum(); rc_ = c_/C
     TT = np.einsum('i,ijk->jk', rc_, np.stack([getattr(n,a) for n in N_]))  # weighted sum
     R  = rc_ @ np.array([n.r  for n in N_])
-    wC = rc_ @ np.array([n.wc for n in N_])
+    wC = rc_ @ np.array([getattr(n,'wc',0) for n in N_])
     if merge:
         for n in N_: root.N_ += n.N_ if merge>1 else [n]
     if froot:
```
boris-kz, line 689: So cluster_P is actually similar to FCM, not EM, and FCM is more logical for parallelization.
I think we need to adopt its normalized membership, instead of direct centroid competition.
The discussion: https://claude.ai/share/a391a1f8-ded0-46ad-8e02-26588ef75ec9

boris-kz: Chat on that Claude-drafted revision:

Closer, but not yet.

Standard FCM has three core properties: it updates centers from `u^q`, recomputes memberships from **nonnegative distances** with exponent `-2/(q-1)`, and stops on a norm of successive membership matrices. Your draft only matches the fixed-shape alternating loop; the membership law and centroid update are still different. ([[Scikit-Fuzzy](https://scikit-fuzzy.github.io/scikit-fuzzy/_modules/skfuzzy/cluster/_cmeans.html)][1])

1. **`sum2C` should read `u_`, but it should use `u_^q`, not raw `u_`.**
   In your current file, `sum2C` weights nodes by `N.c * (m/(ave*o) * _C.m)`, so centroid motion depends on match, rank-competition, and prior centroid strength. That is exactly what makes the present `cluster_P` non-FCM. For an FCM-like core, the update weight should be just `N.c * u_ik**q`, with `_C.m` removed from the update itself. `_C.m` can stay as a **final keep score**, but it should not bias centroid recomputation. 

2. **Using only `dU` is the right simplification, but your threshold is on the wrong scale.**
   With row-normalized memberships, each node has `sum(N.u_) = 1`, so on the first pass `dU ≈ len(N_)`. But your threshold is `ave * (root.r + wC) * len(N_)`. In the file, `ave=.3` and `wC=20`, so this threshold will usually be much larger than `dU`, meaning you often stop immediately after one iteration. Use a true matrix criterion instead: mean absolute `ΔU`, Frobenius norm of `U-U_prev`, or an average per entry over `N×C`. 

3. **`o_` is not a local detail.**
   Replacing it with `u_` is not just a `cluster_P` edit. `sum2C` currently consumes `m,o`, and later `add_Nt` folds `N.o_` into graph `r`. The relevant dependencies are in `sum2C` itself and later aggregation, not at the line you pointed to. So if you remove `o_` from refinement, either keep it as a separate post-hoc rank signal, or replace those downstream uses explicitly. 

4. **The real blocker is negative `m_`.**
   FCM memberships are computed from distances, and distances are nonnegative. Normalizing signed `m_` values by `sum(N.m_)` can produce invalid memberships even if the row sums to 1. So you need one of two choices:

   * make a true nonnegative dissimilarity `d_ik` from `base_comp`, then use FCM proper;
   * or make a nonnegative affinity `a_ik = max(eps, f(match))`, normalize that, and accept that this is **FCM-like fuzzy prototype refinement**, not literal FCM.
     The present draft is in the second camp unless you redefine the scoring. ([[Scikit-Fuzzy](https://scikit-fuzzy.github.io/scikit-fuzzy/_modules/skfuzzy/cluster/_cmeans.html)][1]) 

5. **Your note about “implicit `m=1`” is not correct.**
   In FCM, the fuzzifier `q` enters both places: center update uses `u^q`, and membership update uses the distance power `-2/(q-1)`. Linear normalization of similarities is not the `q=1` case; the `q→1` limit tends toward hard assignments. `q=2` is the common default. ([[Scikit-Fuzzy](https://scikit-fuzzy.github.io/scikit-fuzzy/_modules/skfuzzy/cluster/_cmeans.html)][1])

The minimal version that is genuinely closer to FCM is:

```python
d_ik = positive_distance(C_k, N_i)
u_ik = d_ik**(-2/(q-1)) / sum_j d_ij**(-2/(q-1))
w_ik = N_i.c * u_ik**q
C_k  = weighted_sum(N_i, w_ik)
```

Then keep your current end-stage pruning as a separate step. That matches your fixed `[N,C]` dense loop, keeps GPU-friendly shape, and pushes all explicit competition out of the refinement core. The main carry-over from your current design would then be the **distance/similarity definition**, not the update mechanics. 

So the shortest diagnosis is: **good direction, but the one decisive change is not `u_` normalization by itself; it is replacing `m,o,_C.m` in `sum2C` with `u_^q`, and basing `u_` on a nonnegative distance or affinity.**

[1]: https://scikit-fuzzy.github.io/scikit-fuzzy/_modules/skfuzzy/cluster/_cmeans.html "skfuzzy.cluster._cmeans — skfuzzy v0.4.2 docs"

kwcckw: > So cluster_P is actually similar to FCM, not EM, and FCM is more logical for parallelization. I think we need to adopt its normalized membership, instead of direct centroid competition. The discussion: https://claude.ai/share/a391a1f8-ded0-46ad-8e02-26588ef75ec9

Okay, and i think the u_ membership method looks simpler and better since that argsort is not really correct for parallelization. But in the code, this `N.u_ = [m/s for m in N.m_]` may get negative value and causing troubles, so we probably need a better method to normalize it first before using m in computing the membership.

boris-kz: that m should be absolute, so we need to pass r=0 here:

def vt_(TT, r, wTT=wTT):  # brief val_ to get m,d, rc=0 to return raw vals, Wn for comp_N

kwcckw: Without r, that m is just pure ratio of m to m + ad, which mean their r is actually redundant here? Or we can have another divisive method to reduce m with any additional r value?

boris-kz: we can use / (ave*r) instead of subtraction, that's a rational deviation?

boris-kz: Claude: that's the right move for FCM — and it's conceptually cleaner than subtraction anyway. A few observations:

**It's a rational deviation, same family as cent_TT.** `cent_TT` already uses `rvT / mean` → `invdev_ = min(rvT/mean, mean/rvT)`, which is a ratio-based deviation measure. Using `m / (ave*r)` for match-over-cost is the same idea: dimensionless, nonneg, multiplicative-friendly. Consistency point in favor.

**It changes what `vt_` returns.** Currently `vt_` ends with `m_/t_ @ wTT[0] - ave*r` (line 611) — subtractive. If you switch to division inside `vt_`, every caller changes meaning: the zero point moves from "breakeven" to "one" (ratio=1 means match equals cost). Callers that test `> 0` become `> 1`, callers that test `> ave*x` become `> 1 + x/something`. Worth grepping for `vt_(` usage before flipping globally.

**Alternative: local to cluster_P only.** Don't touch `vt_`; compute ratio inline where you need nonneg memberships:

```python
N.m_,N.d_,N.r_,N.rN_ = map(list,zip(*[vt_(base_comp(C,N)[0],C.r,wTTC) + (C.r,C) for C in _C_]))
u_raw = [max(m,0) + eps for m in N.m_]  # or: [m + ave*r for m,r in zip(N.m_,N.r_)]
s = sum(u_raw)
N.u_ = [u/s for u in u_raw]
```

Or better, bypass `vt_`'s subtraction entirely for this purpose — get raw `m_/t_` without the `-ave*r` term, then divide. But that means a second code path through vt_.

**The cleanest version** is what you're proposing: make `vt_` return ratio-form everywhere. Scan callers first, because the semantics shift is global. Want me to do that scan?

kwcckw: Yes, that should be better since redundancy should contribute to the m, and with a same scale of r, all C.m should be scaled down equally with their r.

boris-kz: ok I didn't add that yet, not sure it's the right form for general val.
so could you review cluster_FCM, and edit Fc_ vals in a separate meta_code module?

kwcckw: > ok I didn't add that yet, not sure it's the right form for general val.

Probably we can have that exclusively for cluster_FCM only?

> so could you review cluster_FCM, and edit Fc_ vals in a separate meta_code module?

Alright, but even with those defined ops weights, we still need the ast.walk to retrieve each func's ops to compute their fc, so we actually still need this function or something similar:

    def func_cost(func):
        tree = ast.parse(inspect.getsource(func))
        return sum(COST.get(type(n).__name__, 0) for n in ast.walk(tree))

boris-kz: right, but it should be outside agg_recursion

kwcckw: I see, so this `meta_code` should be file name, where we run it to get the fc_ weight from agg_recursion right?

boris-kz: right, that's a start, we will extend it later

boris-kz: by analogy with proj_N, we can have proj_C, projecting the links between C and it's members, as an extension of cluster_P, is that similar to EM?

Claude: Closer to EM than plain FCM, but still not quite EM.

**The analogy.** EM's E-step computes `p(C|N)` under a generative model `p(N|C)` — i.e., "given the centroid's model, how likely is N to be generated from it?" That's a *forward projection* from C to N. Your `proj_C` would project the C→N link (predict what N should look like from C's current state), then membership becomes "how well does actual N match projected N?" That's structurally the E-step: membership as model-predicted fit, not raw similarity.

**Differences from EM proper:**

1. **No probability, no likelihood.** EM needs a density `p(N|C)` — Gaussian, mixture, whatever. You'd be projecting deterministic feature values and measuring match via `base_comp`, same as everywhere else in CogAlg. That's the principled disanalogy: EM assumes a generative stochastic model, CogAlg derives prediction from cross-comparison. Same role in the algorithm, different epistemology.

2. **M-step is still your `sum2C`.** EM's M-step maximizes expected likelihood; yours is weighted summation of members. If memberships are FCM-style normalized, the update rule is identical to FCM's. Adding proj_C only changes the E-step (how membership is computed), not the M-step.

3. **Projection cost.** `proj_N` has meaningful cost because it extrapolates along a link's derivatives. `proj_C` would need C to have derivatives w.r.t. each member — which comes from the C→N link formed during cluster_C/sum2C. So the machinery exists; it's projecting the C member-link the same way proj_N projects an N→N link.

**Why it's a real upgrade over plain FCM here.** Raw `base_comp(C,N)` match treats C as a static point. Projection uses C's *structure* (its derivatives, its spread over current members) to predict N, which is strictly more information. Membership becomes "fit to C's predictive model" rather than "similarity to C's centroid value." That's the EM-ish part.

**One thing to think about.** In EM the generative model's parameters (covariances, etc.) are updated in the M-step alongside the means. Your equivalent would be C's dTT / spread getting updated by sum2C each iter, which I think already happens. So proj_C reads C.dTT to project, sum2C rewrites C.dTT from weighted members — the loop closes naturally.

So: yes, extending cluster_P with proj_C moves it from FCM toward EM-analog, with membership = match-to-projection instead of match-to-centroid-value. The gain is using C's derivatives predictively. Worth doing if the per-iter cost of projection is justified by faster convergence or better final clusters — probably yes when C has non-trivial spread.

boris-kz: https://chatgpt.com/share/69dc58d2-7e44-8329-906a-28cc0958d6b6

kwcckw: > The main caveat is self-confirmation. If N helped form C, then C projecting N back to itself can become tautological.

The additional proj_C should be helpful, but projecting between C and their existing links is actually cyclic? So a meaningful one is probably those excluded node in current iteration?

boris-kz:
This is all fully fuzzy, we have links between all Cs and all Ns, but that
u weights are different

PR #113, opened 2026-04-13
4/13/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -598,9 +599,10 @@ def cluster_FCM(__C_, N_, _c, root):  # FCM-style parallel centroid refine, repl
         dU = 0  # total membership change per loop
         for N in N_:
             N.m_,N.d_,N.r_,N.rN_ = map(list, zip(*[ vt_( base_comp(C,N)[0], C.r, wTTC) + (C.r,C) for C in _C_]))
-            s = sum(N.m_) or eps; N.u_ = [m/s for m in N.m_]  # normalize matches so Σ_C u = 1
+            dist_ = [np.hypot(*(N.yx - C.yx)) / (N.span + C.span) for C in _C_]  # FCM usually have distance? Or use the feature space distance: TT[1]/(TT[0]+TT[1]) @wTT?
```
kwcckw, line 602: So FCM usually needs distance and we can include them in the membership computation?

frame_2D_alg/agg_recursion.py
```diff
@@ -610,13 +612,13 @@ def cluster_FCM(__C_, N_, _c, root):  # FCM-style parallel centroid refine, repl
     for i, C in enumerate(C_):
         if C.m > ave*(wC+C.r):  # final pruning
             C.N_ = [n for n in C.N_ if n.u_[i] * C.m > ave * n.r_[i]]  # membership-weighted match > redundancy
-            if C.N_: out_ += [C]
+            if C.N_: out_ += [sum2C(C.N_,C, i, root=root, fu=1)]  # we need to resum N_ into C?
```
kwcckw, line 615: After pruning, we need to resum N_'s params into new C here?

frame_2D_alg/agg_recursion.py
```diff
@@ -704,13 +708,17 @@ def sum_H(N_, Ft):
...
+def sum2C(N_, _C, _i=None, root=None, fu=0):  # fuzzy sum + base attrs for centroids
 
     cc_ = []
     for N in N_:
-        if _i is None: i = N._C_.index(_C); m,o = N._m_[i], N._o_[i]  # cluster_C
-        else:          m,o = N.m_[_i],N.o_[_i]  # current m_,o_ in cluster_P
-        cc_ += [N.c * (m/(ave*o) * _C.m)]  # *_C.m: proj survival?
+        if fu:
+            u = N.u_[_i]  # membership
+            cc_ += [N.c * u ]  # apply * c when computing u, then we can skip it here?
```
kwcckw, line 717: For FCM method, i think it's just c * u here? Since we don't add m or o in the prior process.

frame_2D_alg/meta_code.py
```diff
@@ -0,0 +1,41 @@
+
+import ast
+from agg_recursion import onF_
+
+def init_wc(path):
+    # compute weights based on operations in function, independent of deeper callees
+    def get_ops(node):
+        weights = { 
+            ast.Assign:       2,    # bind name: trivial
```
kwcckw, line 9: The weights are computed with `llvm-mca` here，using the code below from claude, and `llvm-mca` should be installed first: https://llvm.org/docs/CommandGuide/llvm-mca.html

It's targeting x86_64 and skylake processor now:

      import os
      import re
      import shutil
      import subprocess
      import tempfile
      import textwrap
      
      # ──────────────────────────────────────
      # Config
      # ──────────────────────────────────────
      MTRIPLE = "x86_64"       # adjust for your OS
      MCPU    = "skylake"                        # or znver3, haswell, etc.
      
      LLVM_MCA = shutil.which("llvm-mca") \
          or r"C:\Program Files\LLVM\bin\llvm-mca.exe"
      
      if not os.path.isfile(LLVM_MCA):
          raise FileNotFoundError(
              f"llvm-mca not found at: {LLVM_MCA}\n"
              "Install LLVM or set LLVM_MCA to the correct path."
          )
      
      # ──────────────────────────────────────
      # AST node  →  surrogate snippet name
      # ──────────────────────────────────────
      AST_TO_SURROGATE = {
          "Assign":       "mov_reg",
          "Attribute":    "load_field",
          "UnaryOp":      "neg_reg",
          "BoolOp":       "test_jcc",
          "Compare":      "cmp_setcc",
          "If":           "cmp_jcc",
          "IfExp":        "cmp_cmov",
          "BinOp":        "add_reg",
          "AugAssign":    "add_reg",
          "Subscript":    "load_indexed",
          "GeneratorExp": "call_only",
          "While":        "loop_dec_jne",
          "For":          "loop_inc_cmp_jl",
          "ListComp":     "loop_store",
          "SetComp":      "loop_call",
          "Call":         "call_only",
      }
      
      # ──────────────────────────────────────
      # Assembly snippets
      #
      # Rules:
      #   - use LLVM-MCA-BEGIN / END markers
      #   - build DEPENDENCY CHAINS so that
      #     total_cycles / iterations ≈ true latency
      #   - keep helper code (ret, labels for
      #     call targets) OUTSIDE the markers
      # ──────────────────────────────────────
      SNIPPETS = {
      
          # ---------- Assign: dependent register moves ----------
          "mov_reg": """
              .intel_syntax noprefix
              .text
              # LLVM-MCA-BEGIN mov_reg
              mov rax, rbx
              mov rbx, rax
              mov rax, rbx
              mov rbx, rax
              # LLVM-MCA-END mov_reg
          """,
      
          # ---------- Attribute: dependent loads (pointer chase) ----------
          "load_field": """
              .intel_syntax noprefix
              .text
              # LLVM-MCA-BEGIN load_field
              mov rax, qword ptr [rax + 8]
              mov rax, qword ptr [rax + 8]
              mov rax, qword ptr [rax + 8]
              mov rax, qword ptr [rax + 8]
              # LLVM-MCA-END load_field
          """,
      
          # ---------- UnaryOp: dependent negation ----------
          "neg_reg": """
              .intel_syntax noprefix
              .text
              # LLVM-MCA-BEGIN neg_reg
              neg rax
              neg rax
              neg rax
              neg rax
              # LLVM-MCA-END neg_reg
          """,
      
          # ---------- BoolOp: test + conditional jump (straight-line cost) ----------
          "test_jcc": """
              .intel_syntax noprefix
              .text
              # LLVM-MCA-BEGIN test_jcc
              test rax, rax
              je .Lskip1
              .Lskip1:
              test rbx, rbx
              je .Lskip2
              .Lskip2:
              # LLVM-MCA-END test_jcc
          """,
      
          # ---------- Compare: cmp + setcc (dependent chain via flags→reg) ----------
          "cmp_setcc": """
              .intel_syntax noprefix
              .text
              # LLVM-MCA-BEGIN cmp_setcc
              cmp rax, rbx
              setl al
              movzx rax, al
              cmp rax, rbx
              setl al
              movzx rax, al
              # LLVM-MCA-END cmp_setcc
          """,
      
          # ---------- If: compare + branch ----------
          "cmp_jcc": """
              .intel_syntax noprefix
              .text
              # LLVM-MCA-BEGIN cmp_jcc
              cmp rax, rbx
              je .Lthen
              .Lthen:
              cmp rax, rbx
              je .Lthen2
              .Lthen2:
              # LLVM-MCA-END cmp_jcc
          """,
      
          # ---------- IfExp: compare + cmov (dependent) ----------
          "cmp_cmov": """
              .intel_syntax noprefix
              .text
              # LLVM-MCA-BEGIN cmp_cmov
              cmp rax, rbx
              cmovl rax, rcx
              cmp rax, rbx
              cmovl rax, rcx
              cmp rax, rbx
              cmovl rax, rcx
              # LLVM-MCA-END cmp_cmov
          """,
      
          # ---------- BinOp: dependent add chain ----------
          "add_reg": """
              .intel_syntax noprefix
              .text
              # LLVM-MCA-BEGIN add_reg
              add rax, rbx
              add rax, rbx
              add rax, rbx
              add rax, rbx
              # LLVM-MCA-END add_reg
          """,
      
          # ---------- Subscript: dependent indexed load ----------
          "load_indexed": """
              .intel_syntax noprefix
              .text
              # LLVM-MCA-BEGIN load_indexed
              mov rax, qword ptr [rbx + rax*8]
              mov rax, qword ptr [rbx + rax*8]
              mov rax, qword ptr [rbx + rax*8]
              mov rax, qword ptr [rbx + rax*8]
              # LLVM-MCA-END load_indexed
          """,
      
          # ---------- Call: call instruction only (ret outside markers) ----------
          "call_only": """
              .intel_syntax noprefix
              .text
              # LLVM-MCA-BEGIN call_only
              call .Lfoo
              call .Lfoo
              call .Lfoo
              call .Lfoo
              # LLVM-MCA-END call_only
              .Lfoo:
              ret
          """,
      
          # ---------- While: loop overhead (dec + branch) ----------
          "loop_dec_jne": """
              .intel_syntax noprefix
              .text
              # LLVM-MCA-BEGIN loop_dec_jne
              .Lw0:
              dec rcx
              jne .Lw0
              # LLVM-MCA-END loop_dec_jne
          """,
      
          # ---------- For: loop overhead (inc + cmp + branch) ----------
          "loop_inc_cmp_jl": """
              .intel_syntax noprefix
              .text
              # LLVM-MCA-BEGIN loop_inc_cmp_jl
              .Lf0:
              inc rcx
              cmp rcx, rdx
              jl .Lf0
              # LLVM-MCA-END loop_inc_cmp_jl
          """,
      
          # ---------- ListComp: loop + store ----------
          "loop_store": """
              .intel_syntax noprefix
              .text
              # LLVM-MCA-BEGIN loop_store
              .Llc0:
              mov qword ptr [rdi + rcx*8], rax
              inc rcx
              cmp rcx, rdx
              jl .Llc0
              # LLVM-MCA-END loop_store
          """,
      
          # ---------- SetComp: loop + call (ret outside markers) ----------
          "loop_call": """
              .intel_syntax noprefix
              .text
              # LLVM-MCA-BEGIN loop_call
              .Lsc0:
              call .Lhash
              inc rcx
              cmp rcx, rdx
              jl .Lsc0
              # LLVM-MCA-END loop_call
              .Lhash:
              ret
          """,
      }
      
      
      # ──────────────────────────────────────
      # Run llvm-mca on a snippet
      # ──────────────────────────────────────
      def run_llvm_mca(asm_text):
          with tempfile.NamedTemporaryFile(
              mode="w", suffix=".s", delete=False
          ) as f:
              f.write(textwrap.dedent(asm_text))
              path = f.name
          try:
              cmd = [
                  LLVM_MCA,
                  "-mtriple", MTRIPLE,
                  "-mcpu", MCPU,
                  "-iterations", "100",
                  path,
              ]
              result = subprocess.run(
                  cmd,
                  capture_output=True,
                  text=True,
              )
              if result.returncode != 0:
                  raise RuntimeError(
                      f"llvm-mca failed:\n{result.stderr}"
                  )
              return result.stdout + result.stderr   # some info on stderr
          finally:
              os.unlink(path)
      
      
      # ──────────────────────────────────────
      # Parse llvm-mca summary output
      # ──────────────────────────────────────
      PATTERNS = {
          "iterations":        re.compile(r"Iterations:\s+(\d+)"),
          "instructions":      re.compile(r"Instructions:\s+(\d+)"),
          "total_cycles":      re.compile(r"Total Cycles:\s+(\d+)"),
          "total_uops":        re.compile(r"Total uOps:\s+(\d+)"),
          "ipc":               re.compile(r"IPC:\s+([0-9.]+)"),
          "uops_per_cycle":    re.compile(r"uOps Per Cycle:\s+([0-9.]+)"),
          "block_rthroughput": re.compile(r"Block RThroughput:\s+([0-9.]+)"),
      }
      
      
      def parse_output(text):
          metrics = {}
          for key, pat in PATTERNS.items():
              m = pat.search(text)
              if m:
                  v = m.group(1)
                  metrics[key] = float(v) if "." in v else int(v)
              else:
                  metrics[key] = None
          return metrics
      
      
      # ──────────────────────────────────────
      # Compute two scores from metrics
      # ──────────────────────────────────────
      def compute_scores(metrics):
          """
          Returns (latency_score, throughput_score).
      
          latency_score    = total_cycles / iterations
                             → cycles per ONE execution of the block
                             → accounts for dependency chains
      
          throughput_score  = block_rthroughput
                             → minimum cycles per block based on
                               resource (port) bottlenecks
          """
          latency_score = None
          throughput_score = None
      
          iters  = metrics.get("iterations")
          cycles = metrics.get("total_cycles")
          brt    = metrics.get("block_rthroughput")
      
          # latency: cycles per block execution
          if cycles is not None and iters not in (None, 0):
              latency_score = round(cycles / iters, 2)
      
          # throughput: resource-limited minimum cycles per block
          if brt is not None:
              throughput_score = brt
      
          return latency_score, throughput_score
      
      
      # ──────────────────────────────────────
      # Normalise raw scores → integer weights 1..5
      # ──────────────────────────────────────
      def normalize(raw, lo=1, hi=5):
          vals = [v for v in raw.values() if v is not None]
          if not vals:
              return raw
          mn, mx = min(vals), max(vals)
          out = {}
          for k, v in raw.items():
              if v is None:
                  out[k] = None
              elif mx == mn:
                  out[k] = lo
              else:
                  out[k] = lo + round((hi - lo) * (v - mn) / (mx - mn))
          return out
      
      
      # ──────────────────────────────────────
      # Main
      # ──────────────────────────────────────
      def main():
          print(f"llvm-mca : {LLVM_MCA}")
          print(f"target   : {MTRIPLE}  cpu={MCPU}\n")
      
          raw_latency   = {}
          raw_throughput = {}
      
          # ---- run each surrogate snippet ----
          for name, asm in SNIPPETS.items():
              print(f"── {name} ──")
              try:
                  out = run_llvm_mca(asm)
                  metrics = parse_output(out)
                  lat, thr = compute_scores(metrics)
                  raw_latency[name]   = lat
                  raw_throughput[name] = thr
      
                  n_instr = metrics.get("instructions")
                  n_iter  = metrics.get("iterations")
                  instr_per_block = (
                      n_instr // n_iter
                      if n_instr and n_iter else "?"
                  )
      
                  print(f"  instructions/block : {instr_per_block}")
                  print(f"  total cycles       : {metrics.get('total_cycles')}")
                  print(f"  latency score      : {lat}  (cycles / iteration)")
                  print(f"  throughput score    : {thr}  (block rthroughput)")
                  print()
              except Exception as e:
                  print(f"  ERROR: {e}\n")
                  raw_latency[name]   = None
                  raw_throughput[name] = None
      
          # ---- map AST nodes → surrogate scores ----
          ast_latency   = {}
          ast_throughput = {}
          for ast_name, surr in AST_TO_SURROGATE.items():
              ast_latency[ast_name]   = raw_latency.get(surr)
              ast_throughput[ast_name] = raw_throughput.get(surr)
      
          # ---- normalise to 1..5 integer weights ----
          norm_lat = normalize(ast_latency)
          norm_thr = normalize(ast_throughput)
      
          # ---- print final table ----
          hdr = f"{'AST Node':<16} {'Latency':>8} {'Throughput':>11} {'Norm Lat':>9} {'Norm Thr':>9}"
          print("=" * len(hdr))
          print(hdr)
          print("=" * len(hdr))
          for ast_name in AST_TO_SURROGATE:
              lat = ast_latency.get(ast_name, "?")
              thr = ast_throughput.get(ast_name, "?")
              nl  = norm_lat.get(ast_name, "?")
              nt  = norm_thr.get(ast_name, "?")
              print(f"{ast_name:<16} {str(lat):>8} {str(thr):>11} {str(nl):>9} {str(nt):>9}")
      
          # ---- emit copy-pasteable dict ----
          print("\n# Copy-paste cost dict (latency-based, 1-5):")
          print("COSTS = {")
          for ast_name in AST_TO_SURROGATE:
              v = norm_lat.get(ast_name, 1)
              print(f'    ast.{ast_name + ":":<16} {v},')
          print("}")
      
      
      if __name__ == "__main__":
          main()

kwcckw: Output:

# AST Cost Model (Latency / Throughput Normalization)

| AST Node        | Latency | Throughput | Norm Lat | Norm Thr |
|----------------|--------:|------------:|---------:|---------:|
| Assign         | 4.03    | 1.0         | 2        | 2        |
| Attribute      | 20.03   | 2.0         | 5        | 3        |
| UnaryOp        | 4.03    | 1.0         | 2        | 2        |
| BoolOp         | 1.04    | 1.0         | 1        | 2        |
| Compare        | 6.03    | 1.0         | 2        | 2        |
| If             | 1.04    | 1.0         | 1        | 2        |
| IfExp          | 6.03    | 1.5         | 2        | 2        |
| BinOp          | 4.03    | 1.0         | 2        | 2        |
| AugAssign      | 4.03    | 1.0         | 2        | 2        |
| Subscript      | 20.03   | 2.0         | 5        | 3        |
| GeneratorExp   | 8.24    | 4.0         | 3        | 5        |
| While          | 1.04    | 0.5         | 1        | 1        |
| For            | 1.05    | 0.5         | 1        | 1        |
| ListComp       | 1.05    | 1.0         | 1        | 2        |
| SetComp        | 4.13    | 1.2         | 2        | 2        |
| Call           | 8.24    | 4.0         | 3        | 5        |

cost dict (latency-based, 1–5 scale)
COSTS = {
    ast.Assign:          2,
    ast.Attribute:       5,
    ast.UnaryOp:         2,
    ast.BoolOp:          1,
    ast.Compare:         2,
    ast.If:              1,
    ast.IfExp:           2,
    ast.BinOp:           2,
    ast.AugAssign:       2,
    ast.Subscript:       5,
    ast.GeneratorExp:    3,
    ast.While:           1,
    ast.For:             1,
    ast.ListComp:        1,
    ast.SetComp:         2,
    ast.Call:            3,
}

kwcckw:
> This is all fully fuzzy, we have links between all Cs and all Ns, but that
u weights are different

Okay, so the moment we have proj_C, all Ns are not confirmed/included yet right?

Also for this fully fuzzy scheme, we will get N with multiple Cs （overlapping between Cs)?

frame_2D_alg/agg_recursion.py
```diff
@@ -193,7 +193,7 @@ def cross_comp(Ft, rr, nF='Nt'):  # core function mediating recursive rng+ and d
             E_ = get_exemplars({N for L in L_ for N in L.nt}, r,c)
             G_,r = cluster_N(Ft, E_,r,c)  # -> cluster_C,_P, eval?
             if G_:
-                Ft = sum2F(G_, Ft.root,froot=1); fC = G_[0].N_[0].typ==3  # or G.N_ = spliced C_ if more valuable?
+                Ft = sum2F(G_,CF(root=Ft.root),froot=1); fC = G_[0].N_[0].typ==3  # or G.N_ = spliced C_ if more valuable?
```
boris-kz, line 196: Ft.root.Nt = sum2F(G_,CF(root=Ft.root),froot=1)?

frame_2D_alg/agg_recursion.py
```diff
@@ -436,11 +436,11 @@ def trans_cluster(G):
     G_= []  # add prelink pL_,pN_? include merged Cs, in feature space for Cs
     if _N_ and val_(Ft.dTT, r+wN, TTw(Ft.root), mw=(len(_N_)-1)*Lw) > 0:  #| fL?
         for N in _N_:
-            N.fin=0; N.exe=1; sum2F(N.rim, R=N.Rt, froot=0)  # only if N was added in trans-cluster?
+            N.fin=0; N.exe=1; sum2F(N.rim, N.Rt, froot=0)  # only if N was added in trans-cluster?
         G_=[]; Gt_=[]; in_= set()  # root attrs
         for N in _N_:  # form G per remaining N
             if N.fin or (Ft.root.root and not N.exe): continue  # no exemplars in Fg
-            N_ = [N]; L_,B_,C_ = [],[],[]; N.fin=1  # init G, no C_?
+            N_ = [N]; L_,B_,C_ = [],[],[]; N.fin=1  # init G, no C_? (C_ is actually redundant now? We will combine them anyway)
```
boris-kz, line 443: Ok this is lower-level C_, a new level will be selected from new N_ in sub+

frame_2D_alg/agg_recursion.py
boris-kz: we only need similarity: m|u, it works the same

frame_2D_alg/agg_recursion.py
boris-kz: right

frame_2D_alg/agg_recursion.py
boris-kz: we can use m for u, it works the same

frame_2D_alg/agg_recursion.py
boris-kz: Claude — FCM's "distance" is whatever dissimilarity measure you plug in. Euclidean is just the textbook default because FCM is usually taught on vectors in ℝⁿ. The algorithm only requires:

1. A nonneg dissimilarity `d(N,C)` (or equivalently a nonneg similarity),
2. Membership normalized across C's per N,
3. A centroid update that's consistent with the dissimilarity (weighted mean is consistent with squared Euclidean; other metrics need matching updates, e.g., weighted median for L1).

Your `base_comp` → `vt_` match already plays the role of similarity. The only FCM-specific requirement is **nonneg** (so normalization makes sense), which is what the ratio-form fix addresses. No Euclidean anywhere.

The one subtlety: FCM's convergence proof assumes the centroid update minimizes the same objective the memberships are computed from. With Euclidean + weighted mean, that's guaranteed. With `base_comp` + your `sum2C`, it's not formally guaranteed — but in practice it converges because sum2C is doing the analogous "weighted aggregate of members" operation. Just be aware you're outside the textbook convergence proof; empirical convergence is what matters.

kwcckw: Okay, so u (membership) should be scaled by N.c too?

frame_2D_alg/agg_recursion.py
kwcckw: Yes, but we don't have H for Cs, so they will be just replaced?

frame_2D_alg/agg_recursion.py
kwcckw: Yes thanks, then that edit `new.root=root.root; root = new` in sum2F is no long relevant.

frame_2D_alg/meta_code.py
boris-kz: ok, looks good

frame_2D_alg/agg_recursion.py
boris-kz: There should be H in Ct, mapping to Nt.H

frame_2D_alg/agg_recursion.py
boris-kz: why?

boris-kz:
> > This is all fully fuzzy, we have links between all Cs and all Ns, but that
> > u weights are different
> 
> Okay, so the moment we have proj_C, all Ns are not confirmed/included yet right?
> 
> Also for this fully fuzzy scheme, we will get N with multiple Cs （overlapping between Cs)?

You don't seem to understand either cluster_P or FCM, ask Claude

frame_2D_alg/agg_recursion.py
```diff
@@ -558,7 +559,7 @@ def cluster_P(__C_, N_, _c, root):  # Parallel centroid refining, _C_ from clust
     while True:
         M = O = dM = dO = 0
         for N in N_:
-            N.m_,N.d_,N.r_,N.rN_ = map(list,zip(*[vt_(base_comp(C,N)[0],C.r,wTTC) + (C.r,C) for C in _C_]))  # distance-weight match?
+            N.m_,N.d_,N.r_,N.rN_ = map(list,zip(*[vt_(base_comp(C,N)[0],C.r,wTTC,1) + (C.r,C) for C in _C_]))  # distance-weight match?
```
boris-kz, line 562: what is that 1?

frame_2D_alg/agg_recursion.py
boris-kz: our distances are actually distinct from similarity: d_ vs m_, because they don't cancel-out across the sign in summation?

claude:

Right, and that's a real structural difference from standard FCM, worth being explicit about.

**Textbook FCM conflates them by construction.** Distance *is* dissimilarity is `1/similarity`. One scalar per (N,C) pair, and membership is derived from that one number. The sign question never arises because Euclidean distance is always nonneg and symmetric in its treatment of feature deviations.

**CogAlg splits them because `base_comp` produces both m and d as independent derivatives.** `m_` is match (similarity, averages toward zero on noise), `d_` is diff (magnitude of deviation, accumulates). They're not reciprocals — they're orthogonal measurements of the same comparison. Summing `d_` across members gives you spread/variance-like info that summing `m_` would cancel out, and vice versa.

**What this means for FCM-in-cluster_P:**

- Membership normalization uses `m_` only. That's the "similarity" channel, and that's what competition among C's is over. Fine, standard.
- `d_` is not redundant with `m_` — it's the spread channel, used by sum2C to update C's derivatives (the C.dTT that proj_C would then read). So `d_` feeds the M-step's model-update, not the E-step's membership.
- This is actually *more* EM-like than FCM: EM tracks means and covariances separately; you track m (mean-like aggregate) and d (spread-like aggregate) separately. Same split, different math.

**Practical consequence for the rewrite.** `N.d_` in cluster_P isn't dead weight even though it doesn't go into `u_`. It's what sum2C uses to update C's spread, which (if you add proj_C) feeds the next iteration's projection. So keep computing it per iter, even though the membership line only touches `m_`.

The conceptual cleanup: in cluster_P, `m_` → competition (E-step), `d_` → model update (M-step). They're not two versions of the same thing, they're two halves of the cycle.

frame_2D_alg/agg_recursion.py
kwcckw: Okay, so if Cs are not empty from `N.Ct.N_`,  we need to pack this `C_` into Ft.root.Ct.H first, else if C_ here is empty , this C_ should be summed into Ft.root.Ct?

frame_2D_alg/agg_recursion.py
kwcckw: Sorry, that's for fdiv, which i added below：

    def vt_(TT, r, wTT=wTT, fdiv=0):  # brief val_ to get m,d, rc=0 to return raw vals, Wn for comp_N
    
        m = m_/t_ @ wTT[0] /ave*r if fdiv else m_/t_ @ wTT[0] - ave*r
        d = ad_/t_ @ wTT[1]/avd*r if fdiv else ad_/t_ @ wTT[1] - avd*r
        return m, d   # norm by co-derived variance

But it should be for the vt_ in cluster_FCM instead.

frame_2D_alg/agg_recursion.py
kwcckw: > why?

We have *c in cluster_P, so we should include that info too?

So d_ is still meaningful even we don't use it in the membership since it's the spread? Or we can include it in the final membership eval or pruning?

frame_2D_alg/agg_recursion.py
kwcckw: Okay, but FCM can have o_ too? So there's actually no difference here or we just need a different m computation here.

frame_2D_alg/agg_recursion.py
boris-kz: no, it's the same as Nt, Ct.N_ is moved to Ct.H only when new N_ is added

frame_2D_alg/agg_recursion.py
boris-kz: we don't need *c for N.membership values, they are all relative and should add-up to 1
we only need them in sum2C, for proportional contribution and total C value
not really sure about d_, it should matter for defining membership too.

frame_2D_alg/agg_recursion.py
boris-kz: I don't think we need o_ here, m_ is already normalized

frame_2D_alg/agg_recursion.py
boris-kz: Good question, and there's a real design choice here that I was glossing over.

**The case for d_ affecting membership.** Two N's can have the same `m_` to C but very different `d_`. High-m-low-d means "matches C and is close to it" — strong cluster member. High-m-high-d means "matches C on average but deviates a lot on individual features" — loose fit, maybe should belong less. If membership only uses `m_`, you're throwing away the deviation signal that says "this match is noisy."

**This is exactly what EM does with covariance.** Mahalanobis distance is `(N-C)ᵀ Σ⁻¹ (N-C)` — the per-feature deviation weighted by spread. Features where C has tight spread penalize deviation more. FCM's Euclidean ignores this; EM with full-covariance Gaussians uses it. Your `m_ / (ave * r)` ratio is distance-free, but you could bring `d_` in as the spread term.

**Two ways to fold d_ in:**

1. **Penalize membership by d_:** `u_raw = m_ / (eps + d_·wD)` — match discounted by deviation magnitude. Simple, and it's the direct analog of "similarity over dissimilarity." Loses the nonneg guarantee unless you clip m_.

2. **Use the combined val** your `val_` function already computes: it returns `rv = m_/t_ @ wTT[0]` with an option to mix in the d channel via `cr`. That's already in the codebase as the principled m/d combiner. Use `val_` instead of `vt_[0]` for the membership basis.

**The CogAlg-consistent answer** is probably #2: you already have a tool for weighting m against d (`val_` with `cr`), and it's the same tool used elsewhere. Membership should use whatever `val_` gives you, not raw `m_`. Then `d_` contributes to competition naturally, not just to the M-step.

**Caveat on what d_ means here.** In EM's covariance, spread is measured *across members of C* — it's C's property, not the individual N's. In your code, `d_` from `base_comp(C,N)` is the pairwise diff between C and this specific N — it's N's deviation from C, not C's overall spread. Those aren't the same thing:

- Pairwise d (what you have): "how far is N from C on each feature" — this N's fit quality.
- Member-spread (what EM uses): "how much do C's members vary on each feature" — C's tightness as a cluster.

Both are legitimate signals for membership, and they're complementary. Pairwise d is in `N.d_` already. Member-spread is in `C.dTT` after sum2C runs. `proj_C` would effectively combine them: project C's expected N using C.dTT, compare to actual N, get a residual that factors in C's spread.

So the layered answer:

- **Minimal fix now:** use `val_` not `vt_[0]` for `u_` basis, so pairwise d_ affects membership.
- **proj_C extension later:** brings C's member-spread into the membership calculation, which is the full EM analog.

frame_2D_alg/agg_recursion.py
```diff
@@ -665,11 +669,11 @@ def add2F(F, n, fr=0, merge=0):  # unpack for batching in sum2F
     a = 'rTT' if fr else 'dTT'  # or wTT?
     if F.c:
         C=F.c+n.c; _rc,rc = F.c/C, n.c/C; F.r = F.r*_rc + n.r*rc; F.c = C
-        if hasattr(n,'wc'): F.wc = (F.wc*_rc + n.wc*rc) if F.wc else n.wc
+        if hasattr(F,'wc') and hasattr(n,'wc'): F.wc = (F.wc*_rc + n.wc*rc) if F.wc else n.wc
         setattr(F, a, getattr(F,a)*_rc + getattr(n,a)*rc)
     else: setattr(F,a,getattr(n,a)); F.c=n.c; F.r=n.r
     if merge: F.N_ += n.N_ if merge>1 else [n]
-    if getattr(n,'H',None): add_H(F.H, n.H, F)
+    if hasattr(F, 'H') and getattr(n,'H',None): add_H(F.H, n.H, F)
```
boris-kz, line 676: thanks

frame_2D_alg/agg_recursion.py
```diff
@@ -610,13 +612,13 @@ def cluster_FCM(__C_, N_, _c, root):  # FCM-style parallel centroid refine, repl
...
             C.N_ = [n for n in C.N_ if n.u_[i] * C.m > ave * n.r_[i]]  # membership-weighted match > redundancy
-            if C.N_: out_ += [C]
+            if C.N_: out_ += [sum2C(C.N_,C, i, root=root, fu=1)]  # we need to resum N_ into C?
     for N in N_:  # compact per-N vectors to surviving Cs
         for v_ in (N.m_,N.d_,N.r_,N.u_):
             v_[:] = [v for v,c in zip(v_,out_) if c]
         N._u_ = []
         N.rN_ = [c for c,keep in zip(N.rN_,out_) if keep]
-    C_ = [c for c in out_ if c]
+    C_ = out_  # [c for c in out_ if c] (all Cs in out_ shouldn't be empty?)
```
boris-kz, line 621: yes, keep above is obsolete

frame_2D_alg/agg_recursion.py
```diff
@@ -598,9 +599,10 @@ def cluster_FCM(__C_, N_, _c, root):  # FCM-style parallel centroid refine, repl
         dU = 0  # total membership change per loop
         for N in N_:
             N.m_,N.d_,N.r_,N.rN_ = map(list, zip(*[ vt_( base_comp(C,N)[0], C.r, wTTC) + (C.r,C) for C in _C_]))
-            s = sum(N.m_) or eps; N.u_ = [m/s for m in N.m_]  # normalize matches so Σ_C u = 1
+            dist_ = [np.hypot(*(N.yx - C.yx)) / (N.span + C.span) for C in _C_]  # FCM usually have distance? Or use the feature space distance: TT[1]/(TT[0]+TT[1]) @wTT?
+            s = sum(N.m_) or eps; N.u_ = [m/s*dist for m,dist in zip(N.m_, dist_)]  # normalize matches so Σ_C u = 1
             dU += sum(abs(u-_u) for _u,u in zip(N._u_,N.u_)) if cnt else sum(N.u_)
```
boris-kz, line 604: so we need to form links from base_comp and pack them in C.L_, then raw membership is L.m/( L.m+L.d)?
C.L_s map to arg N_, and reciprocal links will be in N.crim_

boris-kz: Right direction — this brings cluster_P in line with how the rest of CogAlg works (everything is links, not parallel arrays).

**Good:**

1. **Links are already the canonical container** for cross-comparison results elsewhere. The `m_, d_, r_, u_, rN_` parallel arrays in cluster_P are an anomaly. Materializing `C.L_` replaces all of them with one list of Link objects.

2. **`L.m / (L.m + L.d)` is clean** — nonneg (if m≥0), bounded [0,1], dimensionless, no hyperparameter. It's "match share of total comparison energy" and naturally includes d_ in membership without the `val_` detour.

3. **Reciprocal `N.crim_` mirrors the existing `N.rim` pattern.** Traversable from either side, and `proj_C` becomes trivial — just project along `C.L_` the same way `proj_N` projects along `N.rim`.

4. **sum2C reads from L** — derivative updates flow through real objects with their own TT, not index-coordinated arrays. Easier to reason about.

**Things to settle:**

1. **Sign of L.m.** If `L.m` comes from `vt_`, it's already `m_/t_ @ wTT[0] - ave*r` — can be negative. Then `L.m/(L.m+L.d)` misbehaves (denom near zero, sign flips). Two options:
   - Use **raw** `m_/t_ @ wTT[0]` (pre-subtraction) as L.m, since `m_/t_` is nonneg by construction. Keep the `-ave*r` adjustment separate for the keep/prune test.
   - Or `max(L.m, 0)` with eps — ugly but safe.
   
   First is cleaner and matches what the ratio actually means.

2. **u_ still needed on top of L.m/(L.m+L.d).** The ratio is absolute fit quality per link; competition between C's for a given N still needs cross-L normalization over `N.crim_`. So:
   - `L.fit = L.m/(L.m+L.d)` — absolute quality, stored on L
   - `L.u = L.fit / Σ(L'.fit for L' in N.crim_)` — share/competition, computed per N each iter
   
   Both live on L, no parallel arrays.

3. **r (redundancy/cost) stays on C, not L.** It's a C property (cost of representing this centroid), not pairwise. Unchanged.

4. **Iteration updates L in place.** Don't reallocate |N|×|C| links per iter — `base_comp` rewrites L.m, L.d, L.dTT on the existing object. Convergence test compares prior `L.u` (or `L.fit`) to current. Old `_m_`/`_u_` buffers become a single `L._u` field.

5. **sum2C signature changes.** Currently `(N_, _C, i, root)` — the `i` is the index into per-N parallel arrays. New form: `sum2C(C, root)` where C reads its own `L_` and each L knows its N. Cleaner, no indexing.

**One thing to double-check against the rest of the codebase.** `Link` (CL?) is a specific class with its own fields (Nt, Lt, etc. from what I've seen). Are C→N membership links the same type, or do they need a lighter variant? If CL is heavy, a minimal `CmL` (membership link) with just `{_N, _C, m, d, dTT, u, fit}` might be worth defining. Otherwise reuse CL.

Want me to sketch the rewritten cluster_P with L-based structure?

boris-kz: `_C_ = __C_; cnt = 0; r = root.r+1`
`for N in N_: N.crim_ = []`
`L__ = [[comp_N(C,N,r,N.c, full=0) for C in _C_] for N in N_]  # [N][C], once`
`for N,L_ in zip(N_,L__): N.crim_ = L_`
`_u__ = [[0]*len(_C_) for _ in N_]`

while True:

    dU = 0
    for N,L_,_u_ in zip(N_,L__,_u__):
        # refresh m,d on existing L's instead of reallocating:
        for L in L_: base_comp(L._N, L.N, L)  # or whatever in-place form exists
        fit_ = [L.m/(L.m+L.d+eps) for L in L_]
        s = sum(fit_) or eps
        u_ = [f/s for f in fit_]  # per-N membership, Σ_C u = 1
        dU += sum(abs(u-_u) for u,_u in zip(u_,_u_))
        for L,u in zip(L_,u_): L.u = u
    C_ = [sum2C(C,root) for C in _C_]  # sum2C reads C.L_ = column of L__
    ...

frame_2D_alg/agg_recursion.py
kwcckw: > Minimal fix now: use val_ not vt_[0] for u_ basis, so pairwise d_ affects membership.

Okay, so to use d_ in the membership, method 1 should be more suitable right?

I can see your code below is using that ` [L.m/(L.m+L.d+eps) for L in L_]`.

frame_2D_alg/agg_recursion.py
kwcckw: Okay, then we only need to sum this `C_` into H when there's new C_ from G_ here: `[C for N in (G_ if G_ else _N_) for C in N.Ct.N_]`.

frame_2D_alg/agg_recursion.py
kwcckw: On this part:

    L__ = [[comp_N(C,N,r,N.c, full=0) for C in C] for N in N_] # [N][C], once

We comp between N and C above:

And then we have base_comp here again because C could be updated in sum2C below?

     for L in L_: base_comp(L._N, L.N, L)  # or whatever in-place form exists
     
So we shouldn't create new C object in the sum2C?

kwcckw:
> > > This is all fully fuzzy, we have links between all Cs and all Ns, but that
> > > u weights are different
> > 
> > 
> > Okay, so the moment we have proj_C, all Ns are not confirmed/included yet right?
> > Also for this fully fuzzy scheme, we will get N with multiple Cs （overlapping between Cs)?
> 
> You don't seem to understand either cluster_P or FCM, ask Claude

Okay thanks, i was confused it with the k-means. There will be overlapping between nodes since it's just a soft membership assignment, including the existing cluster_P too.

frame_2D_alg/agg_recursion.py
boris-kz: we won't get it here, it's in sub+

frame_2D_alg/agg_recursion.py
boris-kz: so that means m will be in 0:inf instead of -1:1, without r. Not sure that's generally better

boris-kz: Two issues:

**Range:** Not really 0:inf. With `m_ ≥ 0`, `m_/t_ ∈ [0,1]`, so `(m_/t_ @ w)/ave` is bounded by `sum(w)/ave` — a ratio centered at 1 (=average), not -1:1 centered at 0. If `m_` can be signed, then it's `(-sum(w)/ave, +sum(w)/ave)`.

Ratio-to-expected is a standard scale (likelihood ratio, observed/expected, lift in stats/IR). But it's *multiplicative* deviation, not additive — downstream accumulation semantics change: you'd want to combine via product/log, not sum. Mixing ratio-m with subtractive-m in the same pipeline will silently corrupt aggregation.

**r placement looks wrong:** `/ave*r` parses as `(/ave)*r`, so r *amplifies* m in the fdiv branch, while in the subtractive branch `- ave*r` it *penalizes* (higher redundancy → lower m). Opposite signs. Probably meant `/(ave*r)`.

Also worth deciding: is the intent "m relative to noise floor" (then subtraction is right) or "m as multiple of noise floor" (then ratio, and you should take `log` to recover an additive, signed scale near 0)?

frame_2D_alg/agg_recursion.py
boris-kz: It's mostly claude, not mine

frame_2D_alg/agg_recursion.py
boris-kz: base_comp would replace comp_N, with some other ops

kwcckw: Okay, and we also need a different sum2C now since we only need the membership information? 

The proj_C (for member spread information from claude above) should be needed before we sum Ns based on their membership into C?

frame_2D_alg/agg_recursion.py
kwcckw: > so that means m will be in 0:inf instead of -1:1, without r. Not sure that's generally better
> Range: Not really 0:inf. With m_ ≥ 0, m_/t_ ∈ [0,1], so (m_/t_ @ w)/ave is bounded by sum(w)/ave

I think it's still fine if it's bounded by sum(w)/ave, since both w and ave are constant for all Cs here.

frame_2D_alg/agg_recursion.py
kwcckw: Okay, then we not really need this C_, it should be removed.

frame_2D_alg/agg_recursion.py
boris-kz: anyway, it's a mess, should be

    m_,d_ = TT; ad_ = np.abs(d_); t_ = eps_(m_+ad_)  # ~ max comparand
    m = m_/t_ @ wTT[0]; d = ad_/t_ @ wTT[1]  # norm by co-derived
    if fdiv: m/= ave*r; d/= avd*r  # in 0-inf for summation
    else:    m-= ave*r; d-= avd*r  # in -1:1 without r
    return m, d

frame_2D_alg/agg_recursion.py
boris-kz: this is my draft:
def cluster_FCM(__C_, N_, _c, root):  # FCM-style parallel centroid refine, replace cluster_P

    _C_ = __C_; cnt = 0  # r = root.r+1
    _u__= [[0]*len(_C_) for _ in N_]
    while True:
        u__,dU = [], 0  # for N in N_: N.u_ = []
        for C,_u_ in zip(_C_,_u__):
            C_,u_ = [],[]
            for N,_u in zip(N_,_u_):
                TT,rn = base_comp(C,N)
                u_ += [vt_(TT, r=1, fdiv=1)[0]]  # dratf
            s = sum(u_)
            u_ = [f/s for f in u_]  # per-N membership, Σ_C u = 1
            dU += sum(abs(u-_u) for u,_u in zip(u_,_u_))
            C_ += [sum2C(C,root)]
        cnt += 1
        if dU > ave * (root.r+wC) * len(N_):  # memberships change  # for N in N_: N._u_ = N.u_
            _C_ = C_
        else: break  # converged
    out_ = []
    for i, C in enumerate(C_):
        if C.m > ave*(wC+C.r):  # final pruning
            C.N_ = [n for n in C.N_ if n.u_[i] * C.m > ave * n.r_[i]]  # membership-weighted match > redundancy
            if C.N_: out_ += [C]
    for N in N_:  # compact per-N vectors to surviving Cs
        for v_ in (N.m_,N.d_,N.r_,N.u_):
            v_[:] = [v for v,c in zip(v_,out_) if c]
        N.rN_ = [c for c,keep in zip(N.rN_,out_) if keep]
    C_ = [c for c in out_ if c]
    dCt = sum2F(set(__C_)-set(C_), CF())  # compression
    return C_

kwcckw: So u_ is local here, and we will need to parse it into sum2C for when we form C from N's membership?

boris-kz: right. claude's edits:

Diff (showing only changed body; `__C_/N_/_c/root` sig unchanged):

```python
def cluster_FCM(__C_, N_, _c, root):

    _C_ = __C_; cnt = 0
    u__ = [[1/len(_C_)]*len(_C_) for _ in N_]  # [N][C], init uniform
    while True:
        new_u__ = [[0]*len(_C_) for _ in N_]
        for ci, C in enumerate(_C_):
            for ni, N in enumerate(N_):
                TT,rn = base_comp(C, N)
                new_u__[ni][ci] = vt_(TT, r=1, fdiv=1)[0]
        for u_ in new_u__:  # normalize per N: Σ_C u = 1
            s = sum(u_) or 1
            for ci in range(len(u_)): u_[ci] /= s
        dU = sum(abs(u-_u) for r,_r in zip(new_u__,u__) for u,_u in zip(r,_r))
        C_ = [sum2C(C, root) for C in _C_]  # see note
        u__ = new_u__; cnt += 1
        if dU <= ave * (root.r+wC) * len(N_): break
        _C_ = C_
    out_, keep = [], []
    for ci, C in enumerate(C_):
        if C.m > ave*(wC+C.r):
            C.N_ = [N_[ni] for ni in range(len(N_))
                    if u__[ni][ci] * C.m > ave * N_[ni].r_[ci]]
            if C.N_: out_ += [C]; keep += [ci]
    C_ = out_
    for N in N_:  # compact per-N vectors to surviving Cs
        for v_ in (N.m_, N.d_, N.r_):
            v_[:] = [v_[ci] for ci in keep]
        N.rN_ = [N.rN_[ci] for ci in keep]
    dCt = sum2F(set(__C_)-set(C_), CF())
    return C_
```

**Note / open issue:** `sum2C(C, root)` takes no weights. For FCM refit, each N must contribute to C proportional to `u__[ni][ci]`. Either sum2C has to accept a weight list, or it still reads weights off N somehow. Without passing `[u__[ni][ci] for ni in range(len(N_))]` into sum2C, the refit is unweighted and this isn't actually FCM — it'll behave like repeated hard re-summing. Flag for your review; I didn't want to touch sum2C's signature without asking.

frame_2D_alg/agg_recursion.py
boris-kz: C needs all the params for base_comp, probably more for possible downstream comp_C_
proj_C is an optional extension, I didn't add it yet

frame_2D_alg/agg_recursion.py
kwcckw: > Note / open issue: sum2C(C, root) takes no weights.

So we either need to parse u_, so we need to pack u_ into N? But before that we need to update the sum2C first. With u_, it replaces m and we only need them in forming C?

boris-kz: yes, please edit sum2C:

def cluster_P(__C_, N_, _c, root):  # FCM-style parallel centroid refine, replace cluster_P

    cnt = 0  # r = root.r+1
    _u__= [[1]*len(__C_) for _ in N_]
    while True:
        u__,dU = [], 0
        C_ = [sum2C(N_,root.Ct, u_) for u_ in _u__]  # same N_ * updated membership u_s
        for N,_u_ in zip(N_,_u__):
            u_ = []
            for C in C_:
                TT,rn = base_comp(C,N); u_ += [vt_(TT, r=1, fdiv=1)[0]]
            s = sum(u_); u_ = [f/s for f in u_]  # per-N membership, Σ_C u = 1
            dU += sum(abs(u-_u) for u,_u in zip(u_,_u_))
            u__ += [u_]
        cnt += 1
        if dU > ave * (root.r+wC) * len(N_):  # memberships change
            _C_ = C_; _u__ = u__
        else: break  # converged
    out_ = []
    for i, C in enumerate(C_):
        if C.m > ave*(wC+C.r):  # final pruning
            C.N_ = [n for n,u_ in zip(N_,u__) if u_[i] * C.m > ave * C.r]  # not sure
            if C.N_: out_ += [C]
    dCt = sum2F(set(__C_)-set(out_), CF())  # compression
    oF = CoF.get(); oF.N_=out_; oF.rTT=dCt.dTT; oF.r=dCt.r; oF.c=_c-dCt.c
    return out_

kwcckw: Alright, I will go through them again later.

PR #114, opened 2026-04-14
4/14/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -573,8 +574,11 @@ def cluster_P(__C_, N_, _c, root):  # FCM-style parallel centroid refine, replac
     out_ = []
     for i, C in enumerate(C_):
         if C.m > ave*(wC+C.r):  # final pruning
-            C.N_ = [n for n,u_ in zip(N_,u__) if u_[i] * C.m > ave * C.r]  # not sure
-            if C.N_: out_ += [C]
+            N_,u_ = [], []
+            for n,_u_ in zip(C.N_,u__):  # not sure
+                if _u_[i] * C.m > ave * C.r and len(N_)<2:
+                    N_ += [n]; u_ += [_u_[i]]  # pack both n and their soft membership weight of current C
+            if N_: out_ += [sum2C(N_,C, u_,root=root)]
```
kwcckw, line 581: When we prune Ns, i think we should sum using their pruned u_ again in the sum2C.

frame_2D_alg/agg_recursion.py
```diff
@@ -552,13 +552,14 @@ def cluster_C(Ft, E_, r,_c):  # form centroids by clustering exemplar surround v
...
-def cluster_P(__C_, N_, _c, root):  # FCM-style parallel centroid refine, replace cluster_P
+def cluster_P(__C_, N_, _c, root): # FCM-style parallel centroid refine, replace cluster_P
 
     cnt = 0  # r = root.r+1
     _u__= [[1]*len(__C_) for _ in N_]
     while True:
         u__,dU = [], 0
-        C_ = [sum2C(N_,root.Ct, u_) for u_ in _u__]  # same N_ * updated membership u_s
+        # we don't need _C here? Or we can include their value when computing cc in sum2C?
+        C_ = [sum2C(N_,None, u_,root) for u_ in _u__]  # same N_ * updated membership u_s
```
kwcckw, line 562: So we actually don;t need `_C` in sum2C here? Or it will be needed later for the projection?

frame_2D_alg/agg_recursion.py
```diff
@@ -166,7 +166,7 @@ def add_typ_(oF):  # record oF vals for weighting
 wM,wD,wi, wG,wI,wa, wL,wS,wA = 10, 10, 20, 20, 5, 20, 2, 1, 1  # dTT weights = reversed relative ave, update from wTT_ after feedback
 wT = np.array([wM,wD,wi, wG,wI,wa, wL,wS,wA]); wTT = np.array([wT,wT*avd])  # default for comp_N_?
 iCC, iCN_, iCC_, iCN, iCF, iPN, iGE, iCLN, iCLC, iCLP, iFH, iVE, iTE, iFB = range(len(onF_))  # pre-call nF indices
-Fc_ = [5,10,20,10,8,5, 5,10,20,30,30,10,15,10]  # fixed complexity weights from AST, maps to onF_
+Fc_ = [6.38, 5.56, 9.24, 2.29, 1.9, 10.82, 13.08, 6.94, 1.77, 6.73, 6.81, 5.77, 1.62, 1.84] # fixed complexity weights from AST, maps to onF_
```
boris-kz, line 169: we can round them up it's very crude anyway. So these are multiples of vt_ complexity?
Fc_ = [6, 6, 9, 2, 2, 11, 13, 7, 2, 7, 7, 6, 2, 2]  # AST complexity / vt_

frame_2D_alg/agg_recursion.py
boris-kz: right, no _C

frame_2D_alg/agg_recursion.py
boris-kz: right

boris-kz: for C,_u_ in zip(C_,u__):
            N_,u_ = [],[]
            for N, u in zip(C.N_,_u_):
                if u * N.m > ave * N.r: N_+= [N]; u_+= [u]
            if N_: out_ += [sum2C(N_, u_,root=root, final=1)]  # assign C.u_, N.rN_

frame_2D_alg/agg_recursion.py
kwcckw: Yes, it's using the method in meta_code.py and it's using vt_ as the base complexity. So i guess we can round them here and the updated values:

    [9.0, 8.0, 13.0, 3.0, 3.0, 15.0, 18.0, 4.0, 3.0, 9.0, 9.0, 8.0, 2.0, 2.0]

Your updated code is actually more efficient and with lower base cost.

frame_2D_alg/agg_recursion.py
kwcckw: Thanks, u_ is per N_, but it's also per C.N_ here since C packs all Ns before the pruning, so your version is better.

frame_2D_alg/agg_recursion.py
```diff
@@ -665,13 +669,13 @@ def sum_H(N_, Ft):
         elif N.N_: H = [list(N.N_)]
     Ft.H = [sum2F(H[0], CF())] if H else []
 
-def sum2C(N_, _C, _i=None, root=None):  # fuzzy sum + base attrs for centroids
+def sum2C(N_, _C, u_=None, root=None):  # fuzzy sum + base attrs for centroids
```
boris-kz, line 672: def sum2C(N_, _C, u_=None, root=None):  # fuzzy sum + base attrs for centroids

    uc_ = []  #  N/C contribution = c-scaled u_
    for i, N in enumerate(N_):
        if u_ is None:   # cluster_C, fdiv m_, proj C survival?
            i = N._C_.index(_C); m,o = N._m_[i],N._o_[i]; uc_ += [N.c * (m/(ave*o))]  # rational overlap?
        else: uc_ += [N.c * u_[i]]  # N/C contribution
    U = sum(uc_)
    for N, uc in zip(N_,uc_):
        rc = uc/U; TT+= N.dTT*rc; kern+= N.kern*rc; span+= N.span*rc; yx+= N.yx*rc; R+= N.r*rc
    m,d = vt_(TT,R, TTw(root.root,1))  # wTTC
    C = CN(typ=3, Nt= CF(N_=N_),dTT=TT,m=m,d=d,c=U,r=R, yx=yx, kern=kern,span=span, root=root)
    # set param correlation ww / nF:
    for TT, cTT, Fw in zip(wTT_, C.wTT_, Fw_):
        cTT[:] = cent_TT(TT,R) * Fw  # TT += cTT * rc?
    return C

kwcckw: So on this part:

      for TT, cTT, Fw in zip(wTT_, C.wTT_, Fw_):
          cTT[:] = cent_TT(TT,R) * Fw  # TT += cTT * rc?

All wn, wN, wc and wC should be replaced by each weight in Fw_, and added with process cost  Fc_? For wTT_, it should be replaced by FwTT_, so C.wTT_ should be renamed to F.wTT_ now?

boris-kz: Right, but FTT is global, we will get it from CoF anyway, C.wTT is local

boris-kz: so here it's just 

    # set param correlation weights:
    C.wTT = cent_TT(TT,R)

?

kwcckw: > Right, but FTT is global, we will get it from CoF anyway, C.wTT is local

For example, in comp_N:

    angl = [A, np.sign(TT[1] @ wTTn[1])]

`wTTn` should be FwTT_[1] now? Right now we didn't update any of the CoF.wTT, so is CoF wTT is still relevant?

We have global wTT, oF.wTT and CN/CF.wTT, how do we really decide when to use each?

boris-kz: local distribution has priority, so we just need *Fw?
with some renaming:

    wTT = cent_TT(TT, R) * Fw_[cltC if _C else cltP]  # set param correlation weights
    m,d = vt_(TT,R, wTT)
    C = CN(typ=3, Nt= CF(N_=N_),dTT=TT,m=m,d=d,c=U,r=R, yx=yx, kern=kern,span=span, root=root, wTT=wTT)

boris-kz: rename and reorder:

cmpN_, cmpC_, cmpN, cmpF, exem, cltN, cltC, cltP, xcmp, frmH, vctE, trcE, fbac, prjN = range(len(onF_))  # pre-call nF indices
Fc_ = [6, 9, 2, 2, 13, 7, 2, 6, 7, 7, 6, 2, 2, 11]  # AST complexity / vt_, maps to onF_

boris-kz: the weights seem to be totally wrong, get_exemplars is twice as expensive as any clustering?

kwcckw: Okay, so we have wTT per CN now, and for TTw:

    def TTw(G, wi=0): return wTT_[wi] * G.wTT_[wi]

It should just prioritize G.wTT and use FwTT_ when G.wTT is empty?

boris-kz: yes, that's probably better

boris-kz: def TTw(G, wi=0): return (G.wTT or FTT_[wi]) * Fw_[wi]  # * func weight?

kwcckw: Okay, and this probably is not updated yet, but in feedback, we should update FwTT_ based on the frame's TTs, but for each oF.wTT, they should be computed from TT right after function ended? Right now after this section?

       oF.w += vt_(TT, oF.wc+oF.r)[0] if np.any(TT) else 0 
       
 Also we don't use oF.wTT now?

boris-kz: oF.wTT should feed up to Z, and then become global wTT?

updated:

def TTw(G, wi=None): return (G.wTT or FTT_[wi]) * (1 if wi is None else Fw_[wi])  # * func weight?

please edit the ffeedback

boris-kz: not reviewed that yet:

Reading sum2C and its three call sites. Main issues:

**Signature/call-site mismatches (will raise at runtime):**

1. **Line 521** `C = sum2C(N_, root=Ft)` — missing required `_C`. The outer loop binds `_C` already, so it should be `sum2C(N_, _C, root=Ft)`.

2. **Line 559** `sum2C(N_, u_, root)` — positional misalignment: `u_` binds to `_C`, `root` binds to `u_`. Either make `_C` default to `None` and call `sum2C(N_, u_=u_, root=root)`, or pass `sum2C(N_, None, u_, root)`.

3. **Line 578** `sum2C(N_, u_, root=root, final=1)` — same misalignment, plus `final=1` is not in the signature → TypeError. The comment on line 577 ("assign C.u_, N.rN_") implies sum2C used to do this on `final=1`; that logic is now entirely missing.

**Conceptual:**

4. **`C.N_` is never set on C itself.** You only set `Nt=CF(N_=N_)`. But downstream reads `_C._N_ + _C.N_` (529), `C.N_` (545), `C.N_` (575). N_ is hidden inside the Nt fork. Either set `N_=N_` on the CN as well, or fix readers.

5. **Using `_C` as a fork flag in line 680** (`cltC if _C else cltP`) conflates "which prior centroid" with "which clustering path". It works only because cluster_P happens to pass `_C=None`, but it's brittle. Use `u_ is None` (already the discriminator) or an explicit fpath.

6. **`m_`, `o_` aren't returned/attached.** In the u_ branch you compute nothing analogous to per-N membership storage on C, so `final=1` semantics (assigning `C.u_`, back-pointers from N) are lost. If you intend cluster_P to call sum2C twice (once for refine, once final), the final pass currently produces an identical C — no pruning artifact, no membership record.

7. **Asymmetry of contribution weighting (minor, probably intentional, flagging):** cluster_C path uses `N.c * m/(ave*o)` (raw strength, unnormalized across Cs); cluster_P path uses `N.c * u_[i]` where u_ is already per-N normalized over Cs. Both re-normalize within C by U, so per-C ratios end up comparable, but absolute U (used as `c=U`) is on different scales between the two paths. If you compare C.c across forks anywhere, that's a problem.

8. **Shadowed `i`** in line 674 (`i = N._C_.index(_C)` overwrites the enumerate `i`). Harmless in the u_-None branch but confusing — rename one.

Want me to draft the minimal patch covering 1–4?

kwcckw: > FTT_[wi]) * Fw_[wi]

FTT_ is the prior wTT_ right? 

But it shouldn't be aligned with Fw_? Because we only have 4 elements in FTT_?

boris-kz: yes,   prior wTT_ was just a stab, it should cover all onF_

kwcckw: > oF.wTT should feed up to Z, and then become global wTT?

That's mean each oF.dTT should be summed from all callee's dTT before we use them to compute wTT using cent_TT?

And in feedback, we only need Z.dTT for feedback purpose to compute global wTT.

boris-kz:
Yes that wTT summation should be in sum2F

kwcckw:
> Yes that wTT summation should be in sum2F

For CoF, we only have dTT or rTT? So we accumulate dTT/rTT from callees and compute their wTT? We should recompute wTT instead of accumulate them?

boris-kz:
Right, recompute

PR #115, opened 2026-04-15
4/15/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -463,13 +464,13 @@ def trans_cluster(G):
                 c = nc + lc + bc
                 r = (nr*nc + lr*lc + br*bc) /c  # br includes overlap?
                 tt= (nt*nc + lt*lc + bt*bc) /c  # tentative
-                if val_(tt, r, TTw(Ft.root), (len(N_)-1)*Lw) > 0:
+                if val_(tt, r, TTw(Ft.root,cltN), (len(N_)-1)*Lw) > 0:  # apply Fw_ and Fc_ in every eval_?
```
kwcckw, line 467: So is there any general rules to apply Fc_ and Fw_? Typically TT or mw should be adjusted by weight Fw_, while ave or r should be adjusted by cost fc_, but not all eval are using them?

frame_2D_alg/agg_recursion.py
```diff
@@ -1028,10 +1028,14 @@ def expand_lev(_iy, _ix, elev, T):  # seed tile is pixels in 1st lev, or Fg in h
...
-                    for i, dTT, wTT, ww in zip((0,1,2,3), (F.dTT,F.Ct.dTT,F.TTn,F.TTc), wTT_, (wN,wC,wn,wc)):
-                        wTT_[i] = cent_TT(dTT,2) * ww  # max-scope correlation weights
-                if elev == max_elev:  # fb / top lev
-                    rV, wTT_ = ffeedback(F,C,R)  # update globals, rV is not used?
+                    # below should replace ffeedback? Not really sure how to update Fc_, Fw_ yet
+                    add_typ_(Z)  # if we need to use CoF's TT for feedback, we need to compute Z.typ_ first
+                    dTT_ = [None for _ in range(len(onF_))]
+                    for typ in Z.typ_: dTT_[typ.nF] = getattr(typ, 'rTT', typ.dTT)
+                    for i, (dTT, Fw, Fc) in enumerate(zip(dTT_, Fw_, Fc_)):
+                        if dTT is not None: FTT_[i] = cent_TT(F.dTT,2) * Fw  # max-scope correlation weights
```
kwcckw, line 1036: The feedback now should be using each Z's typ's TT, which mean we need to form Z.typ_ first, and retrieve each respective typ oF's TT and update the respective FTT_ via cent_TT?

Also how do we really update Fc_ and Fw_ here? Or Fc_ and Fw_ are constant here and they should be updated on higher level code?

frame_2D_alg/agg_recursion.py
```diff
@@ -157,7 +158,7 @@ def add_typ_(oF):  # record oF vals for weighting
     for tF in typ_: sum2F(tF.typ_,tF)
     oF.typ_ = [tF for tF in typ_ if tF.w * tF.c > ave]  # add coef
     if oF.typ_: sum2F(oF.typ_,oF)  # tVs replace all Vs?
-    for tF in oF.typ_: CoF._wdt[tF.nF] = tF.w
+    # for tF in oF.typ_: CoF._wdt[tF.nF] = tF.w  # this is no longer relevant? We have CoF._W_ and CoF._C_ now
```
boris-kz, line 161: right

frame_2D_alg/agg_recursion.py
boris-kz: yes, I think all these evals should be function-specific

frame_2D_alg/agg_recursion.py
boris-kz: right. Fc_ is fixed, Fw_[i] = vt(FTT)[0]?

kwcckw: > Fw_[i] = vt(FTT)[0]

But vt_ needs wTT, and wTT will be the same FTT[i]?

frame_2D_alg/agg_recursion.py
kwcckw: Yes, they should be function-specific, that's mean each tt should be scaled with Fw_, r is scaled with Fc_, and mw is scaled with Fw_? But i think there will be additional redundancy if we add weights on all params here?

boris-kz: the weights are fractional, they don't add redundancy

frame_2D_alg/agg_recursion.py
boris-kz: ok it's not vt_().
we will have higher-order wwTT at some point, but right now it's just np.sum(wTT)/18?

frame_2D_alg/agg_recursion.py
boris-kz, line 171: the weights seem to be totally wrong, get_exemplars is twice as expensive as any clustering?

frame_2D_alg/agg_recursion.py
```diff
@@ -68,7 +68,7 @@ def __init__(n, **kwargs):
         # 2= G:  + rim,kern,mang,sub,exe
         # 3= C | Cn: +rN_| m_,d_,r_,o_
         n.dTT, n.TTn, n.TTc = [np.copy(kwargs.get(x,np.zeros((2,9)))) for x in ('dTT','TTn','TTc')]  # +wTTC per Ct.dTT, all m_,d_ [M,D,n, I,G,a, L,S,A]
-        n.wTT_ = kwargs.get('wTT_',[np.ones((2,9)) for _ in range(4)])  # dTT: Nt+Lt+Bt, TTC, TTn, TTc
+        n.wTT = kwargs.get('wTT',np.zeros((2,9)))  # dTT: Nt+Lt+Bt, TTC, TTn, TTc
```
boris-kz, line 71: the default is still np.ones?

        n.dTT = kwargs.get('dTT',np.zeros((2,9)))  # Nt+Lt+Bt, Ct is separate, all m_,d_ [M,D,n, I,G,a, L,S,A]
        n.wTT = kwargs.get('wTT',np.ones((2,9)))

frame_2D_alg/agg_recursion.py
```diff
@@ -140,6 +140,7 @@ def inner(*a, **kw):
                     sum2F(oF.call_,oF)  # represents all nested call_s
             TT = getattr(oF,'rTT', oF.dTT)  # rTT includes cluster compression
             oF.w += vt_(TT, oF.wc+oF.r)[0] if np.any(TT) else 0  # default Fw_[oF.nF], get ave?
+            oF.wTT = cent_TT(getattr(oF,'rTT', oF.dTT),oF.r)
```
boris-kz, line 143: oF.w += np.sum(TT)/18  # norm both by c,r? default Fw_[oF.nF]

frame_2D_alg/agg_recursion.py
boris-kz: we can unpack them as globals, they should be fixed per frame

kwcckw: Do you mean unpack Fw_ as separated vars? But i think it's already quite clear with the current Fw_[index] method.

frame_2D_alg/agg_recursion.py
kwcckw: > np.sum(wTT)/18

Do you mean each Fw_[i] is FTT_[i]/18? But this division by 18 is scaling the value down only? So the weights will become smaller and smaller.

frame_2D_alg/agg_recursion.py
kwcckw: I think you copied a wrong set of values?

From the prior discussion, the actual values should be:

     [9.0, 8.0, 13.0, 3.0, 3.0, 15.0, 18.0, 4.0, 3.0, 9.0, 9.0, 8.0, 2.0, 2.0]

frame_2D_alg/agg_recursion.py
kwcckw: We need it as empty ,else TTw will always select those 1s?

boris-kz: then init None?

frame_2D_alg/agg_recursion.py
boris-kz: ok that's 4 for cltP, which should be the most expensive?
so these values are for single loop, we have eval per loop, but it should be * C_ * N_ size?
same for other functions, the costs scale with data?
do they have to be defined as floats?

frame_2D_alg/agg_recursion.py
boris-kz: this is for np.sum(wTT), so /18 gets the mean
or np.mean(wTT)

frame_2D_alg/agg_recursion.py
```diff
@@ -749,17 +749,17 @@ def add_Nt(G, Nt, merge=0):  # addition to Q2R
 def add_Lt(G, Lt):  # addition to Q2R
 
     L_ = Lt.N_
-    if Lt.m > ave*wn*Lt.r:  # comp typ -1 pre-links
+    if Lt.m * Fw_[cltN] > ave*Fc_[cltN]*Lt.r:  # comp typ -1 pre-links (add_Lt is only for clustering, so cltN?)
```
boris-kz, line 752: no, this for prelink comp, so it's cmpN

frame_2D_alg/agg_recursion.py
```diff
@@ -749,17 +749,17 @@ def add_Nt(G, Nt, merge=0):  # addition to Q2R
...
             for L in pL_: L_ += [comp_N(*L.nt, G.r,L.c,1, L.angl[0], L.span)]
             sum2F(L_,Lt); add2F(G,Lt)
     A = np.sum([l.angl[0] for l in L_], axis=0) if L_ else np.zeros(2)
-    G.angl = np.array([A, np.sign(G.dTT[1] @ wTTN[1])], dtype=object)  # add weighting?
+    G.angl = np.array([A, np.sign(G.dTT[1] @ FTT_[cltN])], dtype=object)  # add weighting?
     G.mang = np.mean([comp_A(G.angl[0], l.angl[0])[0] for l in G.L_])  # Ls only?
 
 # utilities:
-def TTw(G, wi=None): return (G.wTT or FTT_[wi]) * (1 if wi is None else Fw_[wi])  # * func weight?
+def TTw(G, wi=None): return np.where(G.wTT != 0, G.wTT, FTT_[wi]) * (1 if wi is None else Fw_[wi])  # * func weight?
```
boris-kz, line 762: def TTw(G, wi=None): 
    return (FTT_[wi] if G.wTT is None else G.wTT) * (1 if wi is None else Fw_[wi])  # * func weight?

frame_2D_alg/agg_recursion.py
boris-kz: it's needless clutter

frame_2D_alg/agg_recursion.py
kwcckw: Okay right, then i guess np.mean should be good here.

frame_2D_alg/agg_recursion.py
kwcckw: I'm not sure if they should be floats, but those are just static analysis?

So runtime info such as loops, C_ or N_ size won't be recorded? I guess it should be much more heavy during runtime when we have while loops.

frame_2D_alg/agg_recursion.py
kwcckw: Okay, None should be fine here.

frame_2D_alg/agg_recursion.py
kwcckw: Okay, then we can just add 'w' infront of those names?

    wcmpN_, wcmpC_, wcmpN, wcmpF, wexem, wcltN, wcltC, wcltP, wxcmp, wfrmH, wvctE, wtrcE, wfbac, wprjN

frame_2D_alg/agg_recursion.py
boris-kz: it's fine for Fc_, but we need * data_size at eval

frame_2D_alg/agg_recursion.py
boris-kz: yes:

wN_,wC_,wN,wF, wE,wcN,wcC,wcP, wX,wFrm,wVct,wTrc,wBac,wPrj = Fw_

frame_2D_alg/agg_recursion.py
kwcckw: Yes, and we already have that? For the mw eval for cross_comp:

    len(G_)-1)*Lw
    
We need to include Fw_?

    len(G_)-1)*Lw*Fw_[xcmp]

frame_2D_alg/agg_recursion.py
```diff
@@ -556,7 +556,7 @@ def cluster_P(__C_, N_, _c, root):  # FCM-style parallel centroid refine, may ad
     cnt = 0; _u__= [[1]*len(__C_) for _ in N_]  # r = root.r+1?
```
boris-kz, line 556: Initial `_u__` in cluster_P should be seeded with m_,d_ formed in cluster_C?

Yes. Current init `[[1]*len(__C_) for _ in N_]` makes the first `sum2C` collapse every C into the same uniform sum of all N_, wasting an iteration and losing the membership info already computed in cluster_C.

Seed from `n._m_, n._C_` (set at line 537) per N, per C in `oC_`:

```python
_u__ = [[(N._m_[N._C_.index(C)] if C in N._C_ else 0) for C in __C_] for N in N_]
# normalize each row
for u_ in _u__:
    s = sum(u_) or 1; u_[:] = [f/s for f in u_]
```

Two notes:
- `_o_` is per-C overlap, not membership weight — m_ alone is the right seed (or m_ - ave*o_ if you want to net it).
- N_ here is `Ft.N_` (all nodes), so non-members must be 0 (or some small floor) to keep the bias from cluster_C.
- After this, the `dU` convergence check should still work since u__ is normalized the same way.

frame_2D_alg/agg_recursion.py
boris-kz: that len(G_)-1 is a gain, not cost, 
we need something like

- ave * (cX*len(G_))

frame_2D_alg/agg_recursion.py
kwcckw: Okay, looks like we can use m in the initial value. Also i think we can include `m - ave*o` as the additional overlapping cost?

frame_2D_alg/agg_recursion.py
kwcckw: Yes, but cost is Fc_? I thought Fw_ is the weight for gain? Or Fw_ is the gain of cost only?

boris-kz: that's a fixed syntax cost, but in a real call the cost is input-proportional

frame_2D_alg/agg_recursion.py
boris-kz: overlap shouldn't affect membership, only final pruning.
and we need to compute u_ instead of m_ in cluster_C too

frame_2D_alg/agg_recursion.py
boris-kz: def TTw(G, wfunc=None, wTT=None):
    return (wTT if G.wTT is None else G.wTT) * wfunc or 1  # func weight

frame_2D_alg/agg_recursion.py
kwcckw: Okay, thanks, so actually both of them are costs for different purposes. Then yes, we should have something like ` (cX*len(G_))`. Also for val_, for example:

    val_(tt, r, TTw(Ft.root,cltN), (len(N_)-1)*Lw) > 0

So `mw` already have data size and scaled by Lw, do we still need any of the cost there?

frame_2D_alg/agg_recursion.py
kwcckw: > and we need to compute u_ instead of m_ in cluster_C too

Okay, that will be added on top of existing m_? But with that cluster_C will be the same as the cluster_P?

frame_2D_alg/agg_recursion.py
boris-kz: mw is gain, not cost

frame_2D_alg/agg_recursion.py
boris-kz: u_ should replace m_. it's not the same at all, cluster_C is sequential with selective expansion

kwcckw: Okay right, we extend the cluster via rim from `_N_` in cluster_C and it's sequential.

boris-kz: so please edit that

kwcckw: Yes, i will add them according to cluster_P. Also the pruning section in cluster_C will be remained the same?

    if M > Ave*O and val_(dTT, cr+O, TTw(_C,cltC),(len(N_)-1)*Lw) > 0
    
That M will be sum of u_ now (prune will be in else section)? Or we need another separated section to prune N from C again based on their o_?

boris-kz: I think N_ pruning should be very similar to that in cluster_P, we mwy even make it a separate function.

if M > Ave*O and val_(dTT, cr+O, TTw(_C,ttcC),(len(N_)-1)*Lw) > 0:

still looks right for Cs

boris-kz: Current `cluster_P` has three conceptual issues independent of the seeding problem:

1. **Typo/dead call** on the seed line: `C(N._u_[...] if C in N._C_ else 0)` — the `C(...)` wrapper shouldn't be there; treats each C as a callable.
2. **Axis confusion.** `_u__` is `[N][C]` (per-N, per-C row), but `sum2C(N_, None, u_, root)` needs `u_` indexed by N for one specific C. The loop `[sum2C(...) for u_ in _u__]` builds `len(N_)` Cs weighted by that N's own C-vector — wrong count, wrong semantic.
3. **Final pruning** reuses the same axis confusion: zips `C_` with `u__` (which has `len(N_)` rows), then zips `C.N_` against a per-N u-vector.

Seeding from cluster_C: by the time cluster_P is called, cluster_C has already swapped (line 537) so `N._C_`, `N._u_`, `N._o_` hold the last round's per-C memberships. That's the seed source — but transpose it to iterate C-major when summing.

Proposed replacement for `cluster_P`:

```python
def cluster_P(__C_, N_, _c, root):  # FCM-style parallel centroid refine, seeded from cluster_C

    # _u__[ni][ci]: per-N membership per C, seeded from n._u_/n._C_, row-normalized
    _u__ = []
    for N in N_:
        u_ = [N._u_[N._C_.index(C)] if C in N._C_ else 0 for C in __C_]
        s = sum(u_) or 1; _u__ += [[u/s for u in u_]]
    cnt = 0
    while True:
        u__,dU = [],0
        C_ = [sum2C(N_, None, [u_[ci] for u_ in _u__], root) for ci,_ in enumerate(__C_)]  # C from col ci
        for N,_u_ in zip(N_,_u__):
            u_ = []
            for C in C_:
                TT,rn = base_comp(C,N); u_ += [vt_(TT, r=1, fdiv=1)[0]]
            s = sum(u_) or 1; u_ = [f/s for f in u_]
            dU += sum(abs(u-_u) for u,_u in zip(u_,_u_))
            u__+= [u_]
        cnt += 1
        if dU*wcP > ave * (root.r+ccP) * len(N_): _u__ = u__
        else: break
    out_ = []
    for ci,C in enumerate(C_):
        if C.m > ave * C.r:
            sN_,u_ = [],[]
            for N,un_ in zip(N_,u__):
                u = un_[ci]
                if u * N.m > ave * N.r: sN_+= [N]; u_+= [u]
            if sN_:
                out_ += [sum2C(sN_, None, u_, root=root, final=1)]
    dCt = sum2F(list(set(__C_)-set(out_)), CF())  # compression, see note
    oF = CoF.get(); oF.N_=out_; oF.rTT=dCt.dTT; oF.r=dCt.r; oF.c=_c-dCt.c
    return out_
```

Upstream issues I noticed while tracing the seed — flagging, not fixing:

- Line 508: `u = [m/(m+d)]` wraps in a list, so `n.u_` ends up as a list-of-1-lists. The seed lookup `N._u_[idx]` will then return `[value]` not `value`, breaking normalization. Probably should be `u = m/(m+d)`.
- Line 533: `n.m_.pop(i)` — `m_` doesn't exist, should be `_m_`.
- Line 512: `N_` is filtered by threshold but `u_`/`o_` are appended every iter, so `zip(N_,u_,o_)` at line 522 misaligns.
- Line 582 (dCt): `out_` are new objects from `sum2C`, so `set(__C_)-set(out_) == __C_` always — the "compression" measure is really the total, not the pruned fraction. If the intent is pruned-only, you need to track which original Cs failed to survive.

boris-kz: re cluster_C:
Seed contract cluster_P expects: per node in `Ft.N_`, aligned lists `n._C_`, `n._u_`, `n._o_` (scalars, one per C that assessed it). Current cluster_C breaks that in four places, plus a dead `_m_` slot shadowing `_u_`.

Proposed diffs (old → new):

**L494** — init seed membership as scalar, drop `_m_`:
```python
# old:
for n in C._N_+C.N_: n._m_+=[C.m*(n.c/C.c)]; n._o_+=[1]; n._C_+=[C]  # not sure
# new:
for n in C._N_+C.N_: n._u_+=[n.c/C.c]; n._o_+=[1]; n._C_+=[C]  # seed u = cost ratio
```

**L508–509** — scalar `u`, don't append yet (wait for pass branch so u_/o_ stay aligned with N_):
```python
# old:
u = [m/ (m+d)]; o = np.sum([_u - u for _u in n._u_ if _u > u])
u_ +=[u]; o_ += [o]  # from all comps
# new:
u = m/(m+d+eps); o = sum(_u-u for _u in n._u_ if _u > u)
```

**L511–516** — append `u_/o_` inside pass branch; update frontier init guard (drop `_m_`):
```python
# old:
if m > 0 and m > Ave * o:
    N_+=[n]; L_+=n.L_; O+=o  # convergence val
        if not hasattr(_n,'_m_'): _n._C_,_n.u_,_n._u_,_n.o_,_n._o_,_n.rN_ = [],[],[],[],[],[]
        N__ += [_n]
# new:
if m > 0 and m > Ave * o:
    N_+=[n]; L_+=n.L_; O+=o; u_+=[u]; o_+=[o]  # aligned with N_
        if not hasattr(_n,'_C_'): _n._C_,_n._u_,_n._o_,_n.u_,_n.o_,_n.rN_ = [],[],[],[],[],[]
        N__ += [_n]
```

**L518** — `_m_` → `_u_`:
```python
# old: if _C in n._C_: i = n._C_.index(_C); dm+=n._m_[i]; do+=n._o_[i]
# new: if _C in n._C_: i = n._C_.index(_C); dm+=n._u_[i]; do+=n._o_[i]
```

**L521** — pass aligned `u_` to sum2C (and retire the `u_=None / N._m_` lookup path):
```python
# old: C = sum2C(N_,_C=None,root=Ft)
# new: C = sum2C(N_, None, u_, root=Ft)
```

**L533** — typo fix:
```python
# old: n.rN_.pop(i); n.m_.pop(i);n.o_.pop(i); break
# new: n.rN_.pop(i); n.u_.pop(i); n.o_.pop(i); break
```

**L546** — uses nonexistent `n.m_`/`n.o_`; after the L536–537 swap, last-round values live in `_u_`/`_o_`:
```python
# old: n.exe = (n.d if n.typ==1 else n.m) + np.sum([m-ave*o for m, o in zip(n.m_, n.o_)]) - ave
# new: n.exe = (n.d if n.typ==1 else n.m) + np.sum([u-ave*o for u,o in zip(n._u_, n._o_)]) - ave
```

Adjacent cleanup (out of strict scope but implied):

- `sum2C`'s `u_ is None` branch at L677–679 becomes dead after this — both callers (cluster_C, cluster_P) pass `u_` explicitly. The `_C` parameter also becomes unused. Can drop the branch and the `_C` arg.
- Semantic note: seeding `_u_` with `n.c/C.c` is simple but unscaled by match quality. Alternative: seed with `C.m/(C.m+C.d+eps) * (n.c/C.c)` so the first cluster_P hand-off reflects both capacity and match, not just capacity.

kwcckw: > I think N_ pruning should be very similar to that in cluster_P, we mwy even make it a separate function.
> 
> if M > Ave*O and val_(dTT, cr+O, TTw(_C,ttcC),(len(N_)-1)*Lw) > 0:
> 
> still looks right for Cs

Okay, so that additional u_ shohld be relevant only in sum2C, thats when we pack eah N to C based on thyr membership value.

PR #116, opened 2026-04-16
4/16/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -491,34 +491,35 @@ def cluster_C(Ft, E_, r,_c):  # form centroids by clustering exemplar surround v
         C.wTT = cent_TT(C.dTT, r) * wcC
         C._N_ = list({n for l in E.rim for n in l.nt if (n is not E and n in Ft.N_)})
         C._L_ = set(E.rim)  # init peer links
-        for n in C._N_+C.N_: n._m_+=[C.m*(n.c/C.c)]; n._o_+=[1]; n._C_+=[C]  # not sure
+        for n in C._N_+C.N_: n._u_+=[n.m*(n.c/C.c)]; n._o_+=[1]; n._C_+=[C]  # seed u based on first m
```
kwcckw, line 494: So i think we still can use m when computing the seed u?

frame_2D_alg/agg_recursion.py
```diff
@@ -491,34 +491,35 @@ def cluster_C(Ft, E_, r,_c):  # form centroids by clustering exemplar surround v
...
                     dtt,_ = base_comp(_C, n); cc+=1  # or comp_N, decay?
                     m,d = vt_(dtt, cr, ttcC, fdiv=1); dTT += dtt; m *= n.c  # rm,olp / C
                     oL_ = set(n.rim) & _C._L_  # replace peer rim overlap with more precise m
                     if oL_: _m,_d = sum_vt(oL_,fm=1, fdiv=1)[:2]; m+=_m; d+=_d
-                    u = [m/ (m+d)]; o = np.sum([_u - u for _u in n._u_ if _u > u])  # higher-m overlap  convert to u_?
+                    # recompute u from m isntead of u=vt_(dtt, r=1, fdiv=1)[0]]  due to oL_?
+                    u = m/ (m+d); o = np.sum([_u - u for _u in n._u_ if _u > u])  # higher-m overlap  convert to u_?
                     u_ +=[u]; o_ += [o]  # from all comps
-                    M += m; D += abs(d)
-                    if m > 0 and m > Ave * o:
```
kwcckw, line 511: So most of the edits here are just changing m to u, but by using u, we not really need `o` in the eval? We can remove all `o` from `Ave * o`  and use `o` only in the pruning function later?

frame_2D_alg/agg_recursion.py
```diff
@@ -527,23 +528,26 @@ def cluster_C(Ft, E_, r,_c):  # form centroids by clustering exemplar surround v
                     else:        C_ += [C]  # reform
                 else:
                     for n in _C._N_+_C.N_:
-                        n.exe = n.m/n.c > 2 * ave
+                        n.exe = n.m/n.c > 2 * ave  # or based on sum of u_ now?
```
kwcckw, line 531: With u as membership, exemplars should be selected on those Ns with more membership, which is just sum of `_u_`?

frame_2D_alg/agg_recursion.py
```diff
@@ -763,12 +772,12 @@ def add_Lt(G, Lt):  # addition to Q2R
...
             sum2F(L_,Lt); add2F(G,Lt)
     A = np.sum([l.angl[0] for l in L_], axis=0) if L_ else np.zeros(2)
-    G.angl = np.array([A, np.sign(G.dTT[1] @ ttcN)], dtype=object)  # add weighting?
+    G.angl = np.array([A, np.sign(G.dTT[1] @ ttcN[1])], dtype=object)  # add weighting?
     G.mang = np.mean([comp_A(G.angl[0], l.angl[0])[0] for l in G.L_])  # Ls only?
 
 # utilities:
-def TTw(G, wfunc=None, wTT=None):
-    return (wTT if G.wTT is None else G.wTT) * wfunc or 1  # func weight
+def TTw(G, wTT=None, wfunc=None):  # we are not parsing wfunc now?
```
kwcckw, line 779: So we are not parsing any of the wfunc here?

frame_2D_alg/agg_recursion.py
boris-kz: yes, but it should be computed with fdiv?

frame_2D_alg/agg_recursion.py
boris-kz: that o still matters for frontier expansion, overlap adds the cost regardless of membership.

frame_2D_alg/agg_recursion.py
boris-kz: yes sorry, that should be outside TTw

boris-kz: Store the fallback on G at init (e.g. G.wTT = wTT or default), then just use G.wTT?
oF wTT is a separate term, it should be applied anyway?

frame_2D_alg/agg_recursion.py
kwcckw: Okay, and right now u is normalized with `m+d`, and we actually need `u / U`?

frame_2D_alg/agg_recursion.py
kwcckw: To make it consistent with fdiv, so we need something like n.m/n.r？

frame_2D_alg/agg_recursion.py
kwcckw: Okay, but what would be the wTT in G.wTT = wTT? That global wTT will be shared and used across all Gs? To be more precise, i think G should be init with their oF.wTT?

boris-kz: ok, that's a comp oF per L and clust oF per G, but it can't be default in def?

frame_2D_alg/agg_recursion.py
boris-kz: we probably need fdiv as default in val_,vt_?

frame_2D_alg/agg_recursion.py
boris-kz: u/U actually replaces o, and m/ (m+d) is redundant, that's how m is defined in vt_.
except that we don't need -/ ave*r in that vt_ call, it should be external?

frame_2D_alg/agg_recursion.py
boris-kz: right

frame_2D_alg/agg_recursion.py
boris-kz: Conceptually sound — oF is the natural context-specific default, the module-level `wTT` is only meaningful as bootstrap before any oF exists.

Implementation: set `CN.wTT = oF.wTT` at creation (inheriting the parent CoF's weights), then centroid formation overwrites with `cent_TT(...) * wc`. That eliminates the None check and `TTw` entirely — call sites just use `G.wTT`.

Two questions:
1. Is oF always available when a CN is created? If some CNs are formed before their containing oF's wTT is computed, you still need a fallback path.
2. Should the centroid reweighting *replace* oF.wTT (current behavior at line 491, 688) or *multiply* it? "Conditional re-weighting" suggests multiply — `C.wTT = oF.wTT * cent_TT(C.dTT, r) * wcC` — which preserves oF context.

frame_2D_alg/agg_recursion.py
boris-kz: Yes, that's a real redundancy. The callers already apply `ave*r` (or `ave`) thresholds on the result:

- `vt_(TT, tile.r)[0] > ave` → effectively `(m - ave*tile.r) > ave`
- `vt_(Link.dTT, Link.r)[0] > ave*r` → `(m - ave*Link.r) > ave*r`, double-counted
- `vt_(ntt+ltt, r)[0] > ave*r` → same

And the `fdiv` branch forces a specific normalization choice inside `vt_`, when some callers (e.g. `sum_vt` → `G.m` in `sum2G`) would want the raw normalized `m,d` to be re-thresholded by the caller's own `r`/weight.

Cleaner split:

```python
def vt_(TT, wTT=wTT):
    m = m_/t_ @ wTT[0]; d = ad_/t_ @ wTT[1]
    return m, d
```

Then callers do `m,d = vt_(TT,wTT); v = m - ave*r` or `m/(ave*r)` as needed. `fdiv` moves to caller too.

Two questions before I diff:

1. Keep `r` arg on `vt_` at all? I'd drop it entirely — the caller has `r`.
2. `sum_vt` returns `(*vt_(TT,R,wTT), TT,C,R)` — want that `m,d` to stay thresholded (current behavior, many call sites rely on it), or also raw?

kwcckw: > we probably need fdiv as default in val_,vt_?

Okay, so we actually already use the same ave*r in the eval, then we not really need it in the function, and the fdiv can be removed too.

frame_2D_alg/agg_recursion.py
kwcckw: > ok, that's a comp oF per L and clust oF per G, but it can't be default in def?

Yes, and i think it can't be default in the function since oF.wTT is not computed yet before the function ended. So it's actually still pointing to the global TTs.

boris-kz: oF.wTT can be set by name:

            oF = CoF(nF=onF_.index(func.__name__), root=_CoF); oF.wTT = FTT_[oF.nF]; oF.wc = Fc_[oF.nF]; _CoF.call_+=[oF]

kwcckw: Yes, so if oF.wTT is FTT, then G.wTT is always point to FTT since oF.wTT only will be recomputed after the function ends.

boris-kz: Except when we init CN() manually?

frame_2D_alg/agg_recursion.py
boris-kz: we still need it in

def sum_vt(N_, fr=0, fm=0, wTT=wTT,fdiv=1):  # basic weighted sum of CN|CF list

    for n in N_:
        rc = n.c / C; TT += (n.rTT if fr else n.dTT)*rc; R += n.r*rc  # * weight
    if fm: 
        m,d = vt_(TT, wTT)
        if fdiv: m/= ave*R; d/= avd*R  # in 0-inf for summation
        else:    m-= ave*R; d-= avd*R  # in -1:1 without r
        return   m,d,TT,C,R
    else: return TT,C,R

frame_2D_alg/agg_recursion.py
kwcckw: Actually i don't see how that can be different? When we init CN manually, we still access the same oF and wTT?

boris-kz: ok we can make wTT default anyway, for clarity

frame_2D_alg/agg_recursion.py
boris-kz, line 1087: simple version:

def ffeedback(frame,c,r):  # adjust filters: all aves *= rV, ultimately differential backprop per ave?

    '''
    future code consolidation, for all onF_:
    compare aligned ops between Z.typ_[i] AST sequences, while prior ops matched?
    cluster matches into higher oF typs
    '''
    rTT_ = [np.ones((2,9)) for _ in range(len(FTT_))]
    _wTT_ = frame.H[0]  # assign per level?
    for lev in frame.H[1:]:
        # sum ratios between consecutive-level TTs, top-down frame expansion levels, not lev-selective or sub-lev recursive
        rtt_ = np.divide(_wTT_, lev.wTT_) 
        rTT_ += rtt_
        _wTT_ = lev.wTT_
    rM = rD = 0
        rm, rd = vt_(rTT,FTT_[i]); rM+=rm; rD+=rd
    return rM+rD, rTT_

frame_2D_alg/agg_recursion.py
```diff
@@ -527,23 +528,26 @@ def cluster_C(Ft, E_, r,_c):  # form centroids by clustering exemplar surround v
...
         if oC_+C_ and mat*dif*olp*wcC > ave*ccC*2:  # if val_(DTT,len(oC_+C_)?
-            oC_+= C_; cc = sum([f.c for f in oC_])
+            oC_+= C_; cc = sum([f.c for f in oC_]); fp=1
             oC_ = cluster_P(oC_, Ft.N_, cc, Ft)  # refine all memberships in parallel by global backprop|EM
             break
-        if Do and Dm/Do > Ave: _C_=C_  # dval vs. dolp: overlap increases with Cs expansion
+        if Do and Du/Do > Ave: _C_=C_  # dval vs. dolp: overlap increases with Cs expansion
         else: oC_ += C_; break  # converged
     if oC_:
+        if not fp:  # skip to prevent pruning again after cluster_P
```
boris-kz, line 546: remove pruning in cluster_P instead?

frame_2D_alg/agg_recursion.py
boris-kz: so with nosrmalized u/=U, we don't need olp, it's something like

def cluster_C(Ft, E_, r,_c):  # form centroids by clustering exemplar surround via rims of new member nodes, within root

    oF = CoF.get(); oF.c += _c; oF.r += r  # revert if no clustering
    C_,_C_ = [],[]  # form root.Ct, may call cross_comp-> cluster_N, incr rc
    for n in Ft.N_: n._C_,n.rN_, n.u_,n._u_ = [],[],[],[]  # add m_,d_,c_ for C.L_?
    for E in E_:
        C = Copy_(E,Ft,init=1,typ=3)  # all rims in root, sequence along eigenvector?
        C.wTT = cent_TT(C.dTT,r) * wcC
        C._N_ = list({n for l in E.rim for n in l.nt if (n is not E and n in Ft.N_)})
        C._L_ = set(E.rim)  # init peer links
        for n in C._N_+C.N_: n._u_+=[n.m*(n.c/C.c)]; n._C_+=[C]  # seed u
        _C_ += [C]
    oC_ = []  # output stable Cs
    while True: # reform C_, add direct in-C_ cross-links for membership?
        C_,cnt, memb,dif,DTT,Du = [],0,0,0,np.zeros((2,9)),0; Ave = ave*(r+ccC); Avd = avd*(r+ccC)
        _Ct_,Ct_,_U,_cc = [],[],0,0 
        for C in _C_:_Ct_ += [C, C.m,C.c,C.r]; _cc+=C.c
        for C,u,_c,r in _C_: u *= (_c/_cc); _U += u; _Ct_ += [[C,u]] 
        for cr, (_C,_U) in enumerate(sorted(_Ct_, key=lambda t: t[1], reverse=True),start=1):
            if _U > Ave:
                L_, N_,N__,u_,U,D,cc, dTT,du = [],[],[],[],0,0,0, np.zeros((2,9)),0  # /C
                    dtt,_ = base_comp(_C,n); cc+=1  # or comp_N, decay?
                    u,d = vt_(dtt,ttcC); dTT+=dtt
                    m = u * n.c  # rm,olp / C, / cr?
                    oL_ = set(n.rim) & _C._L_  # replace peer rim overlap with more precise m
                    if oL_: _m,_d = sum_vt(oL_,fm=1, fdiv=1)[:2]; m+=_m; d+=_d
                    u_+=[u]; U+=u; D+=abs(d)  # from all comps
                    if u > Ave:
                        N_+=[n]; L_+=n.L_  # convergence val
                            N__ += [_n]  # +|-Ls
                        if _C not in n._C_: du+=u  # not in extended _N__
                    else:
                DTT+=dTT; memb+=U; dif+=D; cnt+=cc
                if U > Ave and val_(dTT, cr,_C.wTT*ttcC,(len(N_)-1)*Lw) > 0:
                    C =  sum2C(N_,None, u_, root=Ft)
                    for n,u in zip(N_,u_):
                        n.rN_ += [C]; n.u_+=[u]
                    C._N_ = list(set(N__)-set(N_))  # new frontier
                    C._L_ = set(L_)  # peer links
                    if D<Avd: oC_+= [C]  # output if stable, actually if val_(DTT, fi=0) + D?
                    else:     C_ += [C]  # reform
                else:
                    for n in _C._N_+_C.N_:
                        n.exe = n.m/n.c > 2 * ave  # or based on sum of u_ now?
                        for i, c in enumerate(n.rN_):
                            if c is _C:  # remove _C-mapping u:
                                n.rN_.pop(i); n.u_.pop(i); break
                Du += du
            else: break  # the rest is weaker
        for n in Ft.N_:
            n._C_ = n.rN_; n._u_= n.u_; n.rN_,n.u_ = [],[]  # new n.Ct.N_s, combine with v_ in Ct_?
        if oC_+C_ and memb*dif*wcC > ave*ccC*2:  # if val_(DTT,len(oC_+C_)?
            oC_+= C_; cc = sum([f.c for f in oC_]); fp=1
            break
        if Du > Ave: _C_ = C_;   # dval vs. dolp: overlap increases with Cs expansion
        else: oC_ += C_; break  # converged
    if oC_:
        u__ = [[n._u_[n._C_.index(C)] for n in C.N_] for C in oC_]
        oC_ = prune_C_(oC_, u__, Ft)
        for n in [N for C in oC_ for N in C.N_]:  # exemplar V + sum n match_dev to Cs, m* ||C rvals:
            n.exe = (n.d if n.typ==1 else n.m) + np.sum([u-ave*o for u, o in zip(n.u_, n.o_)]) - ave
        if val_(DTT, r, Ft.root.wTT*ttcC, (len(oC_)-1)*Lw) > 0:
            Ct = sum2F(oC_, Ft.root.Ct)  # ?
            _, r = cross_comp(Ct,r)  # all distant Cs, seq C_ in eigenvector = argmax(root.wTT)?
            sel_TT = (Ct.dTT*Ct.c - Ft.dTT*Ft.c) / eps_(Ct.dTT * Ct.c)
            oF = CoF.get(); oF.N_=oC_; oF.rTT=sel_TT; oF.r=Ct.r; oF.c = Ft.c-Ct.c  # data, select in Nt.N_?
    return oC_,r

frame_2D_alg/agg_recursion.py
kwcckw: Okay, but we need to repack `N._C_` since we didn't update their `_C_` reference in cluster_P, and we need that to get u__.

frame_2D_alg/agg_recursion.py
kwcckw: Actually the problem is we don't have wTT_ per level now? Because we only have single wTT per CN, which mean wTT_ needs to be retrieved from Z's typ_'s wTT?

frame_2D_alg/agg_recursion.py
kwcckw: So i guess `o` actually is packed implicitly in u now, which can be skipped since `u` partly shows their overlapping too.

boris-kz: in u/U, except there is no specifc priority of stronger u s that we had before. That was more accurate, but probably not worth the trouble and not consistent with cluster_P

frame_2D_alg/agg_recursion.py
boris-kz: yes, we need to pack that in levs

kwcckw: Okay, to get that per lev's wTT_, we need to get all calls under each lev, which mean it's in each tile, so we need to pack eacj cross_comp's wTT_ to tile before return them?

frame_2D_alg/agg_recursion.py
kwcckw: I see, thanks for clarifying.

frame_2D_alg/agg_recursion.py
boris-kz: probably, from cross_comp oF.typ_

frame_2D_alg/agg_recursion.py
boris-kz: so that means sum2C should be the same for cluster_C and cluster_P, we don't need `_C` and there is always `u_`?

frame_2D_alg/agg_recursion.py
kwcckw: I think that should be done in CoF?

When oF is cross_comp, before return out, we need to add:

    add_typ_(oF)

And then update each type's wTT to out.wTT_?

frame_2D_alg/agg_recursion.py
kwcckw: > so that means sum2C should be the same for cluster_C and cluster_P, we don't need `_C` and there is always `u_`?

Yes thanks. I missed that in the commit, since u_ should replace `_m_` and `_o_` anyway.

frame_2D_alg/agg_recursion.py
boris-kz: not sure, I left it as is for now

frame_2D_alg/agg_recursion.py
boris-kz: but it's not for each cross_comp, only per frame?

frame_2D_alg/agg_recursion.py
kwcckw: > not sure, I left it as is for now

We probably can return u__ from cluster_P and skip the process to recompute u__.

frame_2D_alg/agg_recursion.py
kwcckw: Each `lev` is in frame.H, and lev is per cross_comp, which is not per frame?

frame_2D_alg/agg_recursion.py
boris-kz: So I pushed that, but u/U is not implemented for all Cs yet, probably a bunch of other bugs.
also please check where we need to add 

    if fdiv: m/= ave*r; d/= avd*r  # in 0-inf for summation
    else:    m-= ave*r; d-= avd*r  # in -1:1 without r

now that it's removed from vt_

frame_2D_alg/agg_recursion.py
boris-kz: ok try that

frame_2D_alg/agg_recursion.py
boris-kz: we also have cross_comp per agg+, but I guess FTT_ should be recomputed there too.
so right, add it in CoF

frame_2D_alg/agg_recursion.py
kwcckw: Alright, i will go through them later.

kwcckw: So this section:

      if fdiv: m/= ave*r; d/= avd*r  # in 0-inf for summation
      else:    m-= ave*r; d-= avd*r  # in -1:1 without r
      
 Should be added externally now?
 
 For example, in cluster_C:
 
     u,d = vt_(dtt,ttcC); dTT+=dtt
     
 We eval this u with `u > Ave`, so we just need to have u/r here?

boris-kz:
We don't need the aves to get u_, only m and d

PR #117, opened 2026-04-17
4/17/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -144,6 +157,12 @@ def inner(*a, **kw):
             _CoF._W_[oF.nF] += Fw_[oF.nF]  # nF is index in onF_
             _CoF._C_[oF.nF] += Fc_[oF.nF]
             CoF._cur.set(_CoF)
+            
+            if oF.nF == 8:  # cross_comp
+                tile = a[0].root  # a is Ft
+                tile.wTT_ = deepcopy(FTT_); add_typ_(oF)
+                for typ in oF.typ_:
+                    tile.wTT_[typ.nF] =  cent_TT(getattr(typ,'rTT', typ.dTT),typ.r)  # compute wTT
```
kwcckw, line 165: Right now wTT_ is computed per tile here and summed into level later, or we should pack oF.rTT\dTT into tile, then sum them and recompute wTT_ from the summed rTT|dTT when we forming lev later?

frame_2D_alg/agg_recursion.py
```diff
@@ -562,24 +570,24 @@ def cluster_P(__C_, N_, _c, root):  # FCM-style parallel centroid refine, may ad
...
             u__+= [u_]
         cnt += 1
         if dU*wcP > ave * (root.r+ccP) * len(N_):  # memberships change
             _u__ = u__
         else: break  # converged
-    out_ = prune_C_(C_, u__, root)
+    out_ = prune_C_(C_, u__, root, ttcP)
     dCt = sum2F(set(__C_)-set(out_), CF())  # compression
-    oF = CoF.get(); oF.N_=out_; oF.rTT=dCt.dTT; oF.r=dCt.r; oF.c=_c-dCt.c
+    oF = CoF.get(); oF.N_=out_; oF.rTT=dCt.dTT; oF.r=dCt.r; oF.c=_c-dCt.c  # looks like it's better to prune it here
```
kwcckw, line 587: So it's actually better to prune within cluster_P since we need the updated Cs to update oF, and it's is causing more trouble if we want to update cluster_P's oF in cluster_C later/

frame_2D_alg/agg_recursion.py
```diff
@@ -506,11 +514,11 @@ def cluster_C(Ft, E_, r,_c):  # form centroids by clustering exemplar surround v
...
                     dtt,_ = base_comp(_C,n); cc+=1  # or comp_N, decay?
                     u,d = vt_(dtt,ttcC); dTT+=dtt
-                    m = u * n.c  # rm,olp / C, / cr?
+                    u *= n.c  # rm,olp / C, / cr?
                     oL_ = set(n.rim) & _C._L_  # replace peer rim overlap with more precise m
-                    if oL_: _m,_d = sum_vt(oL_,fm=1, fdiv=1)[:2]; u+=m; m+=_m; d+=_d
+                    if oL_: _m,_d = sum_vt(oL_,fm=1, fdiv=1, wTT=ttcC)[:2]; u+=_m*n.c; d+=_d
                     u_+=[u]; U+=u; D+=abs(d)  # from all comps, U is not c-weighted?
-                    if u > Ave:
+                    if u/U > Ave:  # only normalize for eval?
```
kwcckw, line 521: For the normalization with U, i think we just need it for eval here?

frame_2D_alg/agg_recursion.py
```diff
@@ -506,11 +514,11 @@ def cluster_C(Ft, E_, r,_c):  # form centroids by clustering exemplar surround v
                 for n in set(_C.N_+_C._N_):  # current + frontier
                     dtt,_ = base_comp(_C,n); cc+=1  # or comp_N, decay?
                     u,d = vt_(dtt,ttcC); dTT+=dtt
-                    m = u * n.c  # rm,olp / C, / cr?
+                    u *= n.c  # rm,olp / C, / cr?
                     oL_ = set(n.rim) & _C._L_  # replace peer rim overlap with more precise m
-                    if oL_: _m,_d = sum_vt(oL_,fm=1, fdiv=1)[:2]; u+=m; m+=_m; d+=_d
+                    if oL_: _m,_d = sum_vt(oL_,fm=1, fdiv=1, wTT=ttcC)[:2]; u+=_m*n.c; d+=_d
```
kwcckw, line 519: I'm not sure why we need the `m` here since it's better to add it into `u`?

frame_2D_alg/agg_recursion.py
boris-kz: cross_comp is root oF anyway, it's really the same as Z, so lev can get it from Z?

frame_2D_alg/agg_recursion.py
boris-kz: that was meant for gating below, m is c-weighted while u is not, but I guess it's the same if weighting is only local

frame_2D_alg/agg_recursion.py
boris-kz: why only for eval, we need it for summation in C too?

frame_2D_alg/agg_recursion.py
```diff
@@ -665,6 +673,8 @@ def sum2F(N_, root, fr=0, merge=0, froot=0):  # -> CF/CN
     else:      root = new
     if getattr(root,'C_',None):
         root.Ct = sum2F(root.C_, CF(root=root))
+    if getattr(N_[0], 'wTT_', None):  # summing tile into lev
```
boris-kz, line 676: this should be in frame_H, sum2F is generic

frame_2D_alg/agg_recursion.py
boris-kz: ok

frame_2D_alg/agg_recursion.py
boris-kz, line 1047: if tile_:  # sparse higher-scope tile, if expanded
            F.H = []; F.N_ = tile_
            F.H += [lev := sum2F(tile_,CF())]  # top lev, same vals as F?
                elev += 1
                if rV > ave:
                    add_typ_(Z)  # should map to FTT_, with cent_TT?
                    FTT_ = lev.wTT_ = np.array([typ.wTT for typ in Z.typ])
?

kwcckw: > FTT_ = lev.wTT_ = np.array([typ.wTT for typ in Z.typ])

Okay, i think we probably need to unpack all Z.typ and match their typ.nF index to the FTT_ here, unless we sort it based on onF_ sequence in add_typ_.

frame_2D_alg/agg_recursion.py
kwcckw: So we actually shouldn't add weight to u and weight only should be added during the eval?

frame_2D_alg/agg_recursion.py
kwcckw: We need raw value for summation, and norm them again after the summation?

frame_2D_alg/agg_recursion.py
kwcckw: Okay, but when we have multiple levs, Z only map to the top lev's wTT_?

boris-kz: We only need to set the top level, they are added sequentially

frame_2D_alg/agg_recursion.py
boris-kz: ok right

frame_2D_alg/agg_recursion.py
boris-kz: yes:

    call_ = flat_(oF)
    for F in call_: typ_[F.nF] += [F]
    for i, F_ in enumerate(typ_):
        if F_: T=sum2F(F_,CF(nF=F_[0].nF)); T.wTT=cent_TT(getattr(T,'rTT',T.dTT),T.r); typ_[i]=T
    oF.typ_ = np.array(typ_)
    if any(typ_): add2F(oF, sum2F([t for t in typ_ if t], CF()))  # refine summed call_?


                    FTT_ = lev.wTT_ = [t.wTT if t else wTT for t in Z.typ_]

boris-kz: if rV > ave:
                    add_typ_(Z)  # maps to FTT_
                    FTT_ = lev.wTT_ = [t.wTT if t else wTT for t in Z.typ_]
                        rV, wTT_ = ffeedback(F,C,R)  # update globals, rV is not used?
            else: break
        else: break

def ffeedback(frame,c,r):  # adjust filters: all aves *= rV, ultimately differential backprop per ave?

    '''
    add code consolidation, for all onF_:
    compare aligned ops between Z.typ_[i] AST sequences, while prior ops matched?
    cluster matches into higher oF typs
    '''
    rTT_ = np.divide(frame.H[0].wTT_, frame.H[0].wTT_)
    _wTT_ = frame.H[1].wTT_
    for lev in frame.H[2:]:  # sum ratios between consecutive-level TTs, top-down frame expansion levels, not lev-selective or sub-lev recursive
        rTT_ += np.divide(_wTT_,lev.wTT_)
        _wTT_ = lev.wTT_
    rM = rD = 0
        rm, rd = vt_(rTT,FTT_[i]); rM+=rm; rD+=rd
    return rM+rD, rTT_

kwcckw: In this line:

     FTT_ = lev.wTT_ = [t.wTT if t else wTT for t in Z.typ_]
 
 If certain typ in Z is empty, we shouldn't update the FTT_ at all and use the prior wTT?

boris-kz: This is a new level, prior wTT doesn't apply

frame_2D_alg/agg_recursion.py
boris-kz: yes, there is a bumch of other issues:
Main issues, ordered by severity:

**1. `N_` vs `u_` misalignment (line 505-524) — correctness bug.**
```
u_+=[u]; U+=u; D+=abs(d)  # u_ gets every candidate
if u/U > Ave:
    N_+=[n]; ...            # N_ only gets accepted ones
```
Then `sum2C(N_, u_, ...)` (523) and `zip(N_, u_)` (524) pair filtered nodes with unfiltered u values. Wrong node ↔ wrong u.

**2. Order-dependent membership test (line 513).** `u/U > Ave` uses running `U`. First node in the set-iteration always passes (u/u=1), last always fails (u/ΣU small). Result depends on `set(_C.N_+_C._N_)` hash order.

**3. Stale `n._C_` across iters (line 547).** Line 539 swaps `_C_ = rN_; rN_=[]` every iter. Centroids added to `oC_` in iter k are in n's rN_ only during iter k, then moved to _C_ at iter-boundary, then overwritten in iter k+1's swap. So `n._C_.index(C)` fails for any C in oC_ from an earlier iter.

**4. Dimensional mismatch (line 504).** `_U = c.m/U` is a fraction summing to 1 across centroids; `Ave = ave*(r+ccC)` is absolute. Either normalize both or neither.

**5. `D<Avd` stability (line 528).** D accumulates `abs(d)` over all candidates including `u/U > Ave` rejects (line 512, outside the `if`). Stability of accepted set is being measured with rejects' noise folded in.

**6. Line 540 threshold is trivially small.** `memb*dif*wcC > ave*ccC*2` — LHS is product of two sums, RHS is a small constant. cluster_P is effectively always called. Is that the intent?

**7. `cc` name collision** — line 500 (sum of centroid costs) vs line 505 (per-C counter in inner loop). Different meanings, same name.

**Minor:**
- Line 553: `cross_comp(Ct, r)` return unpacked as `_, r` — discards G_ return. Deliberate?
- Line 523: `sum2C(..., wTT=ttcC)` — sum2C overwrites wTT internally via `cent_TT` regardless, so the passed arg is dead.
- Line 551: `DTT` measures comparison residuals during expansion, used to eval the output set. Quality signal for oC_ should come from the centroids themselves, not residuals.

Do you want me to focus on fixing #1 and #2 first? Those are the ones that corrupt output rather than just inefficient eval.

boris-kz: if _U > Ave:

                L_, N_,N__,u_,U,D,cc, dTT,du = [],[],[],[],0,0,0, np.zeros((2,9)),0  # /C
                cand_ = []  # (n,u,d) per candidate, filter after U is final
                    dtt,_ = base_comp(_C,n); cc+=1  # or comp_N, decay?
                    u,d = vt_(dtt,ttcC); dTT+=dtt
                    u *= n.c  # rm,olp / C, / cr?
                    oL_ = set(n.rim) & _C._L_  # replace peer rim overlap with more precise m
                    if oL_: _m,_d = sum_vt(oL_,fm=1, fdiv=1, wTT=ttcC)[:2]; u+=_m*n.c; d+=_d
                    cand_+=[(n,u,d)]; U+=u; D+=abs(d)  # from all comps
                for n,u,d in cand_:
                    if u/U > Ave:  # filter with final U, N_/u_ aligned
                        N_+=[n]; u_+=[u]; L_+=n.L_
                            N__ += [_n]  # +|-Ls
                        if _C not in n._C_: du+=u  # not in extended _N__
                    else:
                DTT+=dTT; memb+=U; dif+=D; cnt+=cc

frame_2D_alg/agg_recursion.py
kwcckw: > This is a new level, prior wTT doesn't apply

Okay, but shouldn't be the prior wTT is a better choice if it's empty? That wTT is global and will be the same across all levels?

boris-kz: no, higher levels are higher orders of scope, the lower "globals" are insignificant

frame_2D_alg/agg_recursion.py
kwcckw: > if _U > Ave:
> 
> ```
>             L_, N_,N__,u_,U,D,cc, dTT,du = [],[],[],[],0,0,0, np.zeros((2,9)),0  # /C
>             cand_ = []  # (n,u,d) per candidate, filter after U is final
>             for n in set(_C.N_+_C._N_):  # current + frontier
>                 dtt,_ = base_comp(_C,n); cc+=1  # or comp_N, decay?
>                 u,d = vt_(dtt,ttcC); dTT+=dtt
>                 u *= n.c  # rm,olp / C, / cr?
>                 oL_ = set(n.rim) & _C._L_  # replace peer rim overlap with more precise m
>                 if oL_: _m,_d = sum_vt(oL_,fm=1, fdiv=1, wTT=ttcC)[:2]; u+=_m*n.c; d+=_d
>                 cand_+=[(n,u,d)]; U+=u; D+=abs(d)  # from all comps
>             for n,u,d in cand_:
>                 if u/U > Ave:  # filter with final U, N_/u_ aligned
>                     N_+=[n]; u_+=[u]; L_+=n.L_
>                     for _n in [_n for l in n.rim for _n in l.nt if _n is not n]:
>                         if not hasattr(_n,'_u_'): _n._C_,_n.rN_,_n.u_,_n._u_ = [],[],[],[]
>                         N__ += [_n]  # +|-Ls
>                     if _C not in n._C_: du+=u  # not in extended _N__
>                 else:
>                     if _C in n._C_: i = n._C_.index(_C); du+=n._u_[i]
>             DTT+=dTT; memb+=U; dif+=D; cnt+=cc
> ```

Yes, thanks, this is actually better since we should get all u first before we can compute U, the prior code compute using only the U up to that point.

boris-kz: that u*c is wrong:

Membership quality is per-node fit to C — c-weighting biases the test and double-weights in sum2C.

Two problems stack:

1. **Bias in membership test.** `u *= n.c` then `u/U > Ave` means high-c nodes dominate U, so a good-match low-c node fails the fraction test regardless of fit quality. Membership should ask "does this node fit C?", not "is this node's c-scaled fit a large fraction of total c-scaled fit?"

2. **Double c-weighting in sum2C.** `sum2C` already does `uc_ = N.c * u_[i]` internally. cluster_C passes already-c-weighted u in, so the centroid sums `n.c² * m_n`.

Fix — strip `n.c` from u in cluster_C, let sum2C handle the weighting:

```python
                    u,d = vt_(dtt,ttcC); dTT+=dtt
                    oL_ = set(n.rim) & _C._L_  # peer rim overlap -> more precise m
                    if oL_: _m,_d = sum_vt(oL_,fm=1, fdiv=1, wTT=ttcC)[:2]; u+=_m; d+=_d
                    cand_+=[(n,u,d)]; U+=u; D+=abs(d)
```

Changes from prior: dropped `u *= n.c` (line 509) and the `*n.c` on `u+=_m*n.c` (line 511). `u` is now raw per-node fit; U is unweighted sum; `u/U` tests pure fit fraction; sum2C applies n.c once.

Apply?

boris-kz: remaining substantive issues:

**1. Stale `n._C_` across iters — still a real bug.**
Line 547: `u__ = [[n._u_[n._C_.index(C)] for n in C.N_] for C in oC_]`. Line 539 replaces `n._C_ = n.rN_` every iter, so `n._C_` only holds centroids from the *final* iteration. Any C that was added to `oC_` in an earlier iter (D<Avd branch, line 528) won't be found — `.index(C)` raises `ValueError`. Only survives when the loop runs one iter (which is most cases today because of #3 below, but not safe).

**2. D-based stability (line 528) uses rejected nodes.**
`D += abs(d)` runs inside the first pass over all candidates, before the filter. `D<Avd` tests stability of a set that wasn't selected. Move D accumulation into the second pass's `if u/U > Ave` branch.

**3. `_U > Ave` scale mismatch (line 504).**
`_U = c.m/U` is a fraction (Σ=1 across centroids), `Ave = ave*(r+ccC)` is absolute. With more than a few centroids, no `_U` exceeds `Ave` and the outer loop breaks immediately. Also `c.m` can be negative, breaking the sort.

**4. Line 540 threshold.**
`memb*dif*wcC > ave*ccC*2` — product of two positive sums vs a tiny constant. cluster_P is effectively unconditional. Is that the intent, or should it be something like `dif/memb > some_ratio`?

**5. `du` asymmetry (lines 518, 520).**
Adds use current-iter `u`; drops use prior-iter `_u_[i]`; nodes that stay but change contribute 0. Du undercounts churn, convergence test on line 544 is biased.

**6. `sum2C(..., wTT=ttcC)` — the arg is dead.**
sum2C overwrites its `wTT` param with `cent_TT(TT,R)*wcC` internally. Either drop the kw here or change sum2C to respect it.

**Minor:**
- `cc` is two different variables (line 500: sum of centroid costs; inside loop: comparison count). Rename one.
- Line 544 trailing semicolon on `_C_ = C_;`.

Want me to apply fixes for 1, 2, 3?

kwcckw: > that u*c is wrong:
> Membership quality is per-node fit to C — c-weighting biases the test and double-weights in sum2C.

Okay, i was thinking if we apply it the same to all nodes or members, they should be added equally and shouldn't be a problem?

boris-kz: c-weighting is for different test, it's a process control instead of node inclusion, which should by membership only. That's why I used it on m instead of u, but that's probably wrong too

kwcckw: > remaining substantive issues:
> 
> **1. Stale `n._C_` across iters — still a real bug.** Line 547: `u__ = [[n._u_[n._C_.index(C)] for n in C.N_] for C in oC_]`. Line 539 replaces `n._C_ = n.rN_` every iter, so `n._C_` only holds centroids from the _final_ iteration. Any C that was added to `oC_` in an earlier iter (D<Avd branch, line 528) won't be found — `.index(C)` raises `ValueError`. Only survives when the loop runs one iter (which is most cases today because of #3 below, but not safe).
> 
> **2. D-based stability (line 528) uses rejected nodes.** `D += abs(d)` runs inside the first pass over all candidates, before the filter. `D<Avd` tests stability of a set that wasn't selected. Move D accumulation into the second pass's `if u/U > Ave` branch.
> 
> **3. `_U > Ave` scale mismatch (line 504).** `_U = c.m/U` is a fraction (Σ=1 across centroids), `Ave = ave*(r+ccC)` is absolute. With more than a few centroids, no `_U` exceeds `Ave` and the outer loop breaks immediately. Also `c.m` can be negative, breaking the sort.
> 
> **4. Line 540 threshold.** `memb*dif*wcC > ave*ccC*2` — product of two positive sums vs a tiny constant. cluster_P is effectively unconditional. Is that the intent, or should it be something like `dif/memb > some_ratio`?
> 
> **5. `du` asymmetry (lines 518, 520).** Adds use current-iter `u`; drops use prior-iter `_u_[i]`; nodes that stay but change contribute 0. Du undercounts churn, convergence test on line 544 is biased.
> 
> **6. `sum2C(..., wTT=ttcC)` — the arg is dead.** sum2C overwrites its `wTT` param with `cent_TT(TT,R)*wcC` internally. Either drop the kw here or change sum2C to respect it.
> 
> **Minor:**
> 
>     * `cc` is two different variables (line 500: sum of centroid costs; inside loop: comparison count). Rename one.
> 
>     * Line 544 trailing semicolon on `_C_ = C_;`.
> 
> 
> Want me to apply fixes for 1, 2, 3?

Okay, i will check again when your commit is available later.

kwcckw: > c-weighting is for different test, it's a process control instead of node inclusion, which should by membership only. That's why I used it on m instead of u, but that's probably wrong too

Okay, then we really should exclude it and include it when we need it in the eval only?

boris-kz: yes, I am sure it's still buggy, I haven't checked those suggestions yet

PR #118, opened 2026-04-18
4/18/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -509,24 +509,24 @@ def cluster_C(Ft, E_, r,_c):  # form centroids by clustering exemplar surround v
...
-                            for _n in [_n for l in n.rim for _n in l.nt if _n is not n]:
-                                if not hasattr(_n,'_u_'): _n._C_,_n.rN_,_n.u_,_n._u_ = [],[],[],[]
-                                N__ += [_n]  # +|-Ls
-                            if _C not in n._C_: du+=u  # not in extended _N__
-                        else:
-                            if _C in n._C_: i = n._C_.index(_C); du+=n._u_[i]
+                    cand_ += [(n,u,d)]; U+=u
+                for n,u,d in cand_:
+                    if u/U > Ave:  # filter with final U, N_/u_ aligned
+                        N_+=[n]; u_+=[u]; L_+=n.L_;  D+=abs(d)  # D should be here? Since it should be valid for added n in C only
```
kwcckw, line 515: So i think most of the posted suggestions from LLM are relevant. The D here is eval only when we sum N_ to C, so it should be added only if we add n to N_?

frame_2D_alg/agg_recursion.py
```diff
@@ -543,14 +543,13 @@ def cluster_C(Ft, E_, r,_c):  # form centroids by clustering exemplar surround v
             oC_ = cluster_P(oC_, Ft.N_, cc, Ft)  # refine all memberships in parallel by global backprop|EM
             break
         if Du > Ave: _C_ = C_;   # dval vs. dolp: overlap increases with Cs expansion
-        else: oC_ += C_; break  # converged
+        else: oC_ += C_; u__ += [[n._u_[n._C_.index(C)] for C in C_ for n in C.N_]]; break  # converged
     if oC_:
-        u__ = [[n._u_[n._C_.index(C)] for n in C.N_] for C in oC_]
```
kwcckw, line 548: Sorry, oC_ are added across all while loops, which mean their u__ should be added along with each append to oC_, because we only able to retrieve the last iteration's u__ here, which is not sufficient.

frame_2D_alg/agg_recursion.py
```diff
@@ -543,14 +543,13 @@ def cluster_C(Ft, E_, r,_c):  # form centroids by clustering exemplar surround v
...
-        if not fP: oC_ = prune_C_(oC_, u__, Ft, ttcC)
+        if not fP: oC_ = prune_C_(oC_, u__, Ft)
         for n in [N for C in oC_ for N in C.N_]:  # exemplar V + sum n match_dev to Cs, m* ||C rvals:
             n.exe = (n.d if n.typ==1 else n.m) + np.sum(n.u_) - ave
         if val_(DTT, r, Ft.root.wTT*ttcC, (len(oC_)-1)*Lw) > 0:
-            Ct = sum2F(oC_, Ft.root.Ct)  # ?
+            Ct = sum2F(oC_, Ft.root.Ct)  # Ct should be the same as Ft.root.Ct? Or inherit from Ft.root.Ct?
```
kwcckw, line 552: So this Ct should replace Ft.root.Ct? But we didn't reassign it back to Ft.root.Ct?

frame_2D_alg/agg_recursion.py
```diff
@@ -683,12 +680,12 @@ def sum_H(N_, Ft):
...
     Ft.H = [sum2F(H[0], CF())] if H else []
 
-def sum2C(N_, u_, root=None, wTT=wTT, final=0, fP=0):  # fuzzy sum + base attrs for centroids
+def sum2C(N_, u_, root=None, final=0, fP=0):  # fuzzy sum + base attrs for centroids
 
     L_, uc_ = [],[]  # N/C contribution = c-scaled u_
     for i, N in enumerate(N_):
         if final: L_ += [CN(typ=1, w=[N.c * u_[i]])]  # add u,c,r,TT,span,angl for proj_C?
-        else:     uc_ += [N.c * u_[i]]  # N/C contribution
+        uc_ += [N.c * u_[i]]  # N/C contribution  (this should be default even in final?)
```
kwcckw, line 688: if N_:  # assign C.u_, N.rN_
          out_ += [sum2C(N_, u_,root=root, final=1)]

We need this ` uc_ += [N.c * u_[i]]` when final == 1? Because we still need the ` U = sum(uc_)` and  `c=U` below?

    C = CN(typ=3, Nt= CF(N_=N_),dTT=TT,m=m,d=d,c=U,r=R, yx=yx, kern=kern,span=span, root=root, wTT=wTT, L_=L_)

frame_2D_alg/agg_recursion.py
```diff
@@ -683,12 +680,12 @@ def sum_H(N_, Ft):
         elif N.N_: H = [list(N.N_)]
     Ft.H = [sum2F(H[0], CF())] if H else []
 
-def sum2C(N_, u_, root=None, wTT=wTT, final=0, fP=0):  # fuzzy sum + base attrs for centroids
```
kwcckw, line 686: Sorry, this wTT is irrelevant since we will recompute their wTT below anyway.

frame_2D_alg/agg_recursion.py
```diff
@@ -1084,4 +1081,4 @@ def trace_func(module_dict, module_name=None):
     # frame = agg_frame(0, image=imread('./images/toucan.jpg'), iY=Y, iX=X)
     frame = frame_H(image=imread('./images/toucan.jpg'), iY=Y//2 -31, iX=X//2 -31, Ly=64,Lx=64, Y=Y, X=X, rV=1)
     # search frames ( tiles inside image, at this size it should be 4K, or 256K panorama, won't actually work on toucan
-    add_typ_(Z)
\ No newline at end of file
+    # add_typ_(Z)  we already have this in ffeedback?
```
boris-kz, line 1084: right. 
Claude discussion on comparing my scheme with EM, related to feedback:
https://claude.ai/share/5dda0f18-4a32-4888-ac5e-842e4ed7ffd9

frame_2D_alg/agg_recursion.py
boris-kz: we may need that D in C-to-N links, for the future proj_C

kwcckw: Okay, so we need another D that's exclusively for projection and another D to eval D<avd? Since we shouldn't use the difference if we didn't include the N in C?

boris-kz: for projection D should be signed, not abs, so it's a differen mechanism

frame_2D_alg/agg_recursion.py
boris-kz: right

frame_2D_alg/agg_recursion.py
boris-kz: Ft.root.Ct is reformed inside sum2F anyway?

kwcckw: Yes, but from this section:

    new = CF(N_=N_, c=C, r=R, wc=wC, **{a: TT})
    sum_H(N_, new)
    if root.c: add2F(root,new)
    else:      root = new
    
We didn't reassign this reformed Ct as Ft.root.Ct?

frame_2D_alg/agg_recursion.py
kwcckw: Okay, then we really only should sum abs(d) here, while the D for projection should be summed separately without the abs operation.

frame_2D_alg/agg_recursion.py
boris-kz: that root is Ft.root.Ct

frame_2D_alg/agg_recursion.py
kwcckw: > projection is a feedback mechanism

Is this applicable to all projections? So projection computes the expected outcome, and that expectation will be used to compared with the actual outputs, so that it can be used to adjust the future process again?

boris-kz: initially it's used to evaluate the comparison operation, we may not have the actual output if that fails

frame_2D_alg/agg_recursion.py
kwcckw: The froot and fmerge are 0 here, and Ct.c is empty, so `root = new` is true and we are returning new CF as Ct here, but this Ct is not updated back to Ft.root.Ct? So we need to reassign it?

    Ft.root.Ct = Ct?

boris-kz: if root.c: add2F(root,new)
else:      root = new

root is Ft.root.Ct, new is Ct

kwcckw: > root is Ft.root.Ct, new is Ct

Yes, but root.c is empty for Ft.root.Ct before any accumulation, so only the else section will be executed, which mean we don't use the Ft.root.Ct at all.

boris-kz: else:  Ft.root.Ct = Ct

frame_2D_alg/agg_recursion.py
kwcckw: > initially it's used to evaluate the comparison operation, we may not have the actual output if that fails

Okay, that's in comp_N, but proj_N and proj_focus are different in expand_lev?

boris-kz: it's the same, only on a tile level

kwcckw: Okay right thanks. Looks like proj_N is use to eval proj_focus, and then proj_focus update PV__ that uses to eval the next frame_H。

frame_2D_alg/agg_recursion.py
```diff
@@ -509,24 +509,24 @@ def cluster_C(Ft, E_, r,_c):  # form centroids by clustering exemplar surround v
...
-                            N_+=[n]; u_+=[u]; L_+=n.L_
-                            for _n in [_n for l in n.rim for _n in l.nt if _n is not n]:
-                                if not hasattr(_n,'_u_'): _n._C_,_n.rN_,_n.u_,_n._u_ = [],[],[],[]
-                                N__ += [_n]  # +|-Ls
-                            if _C not in n._C_: du+=u  # not in extended _N__
-                        else:
-                            if _C in n._C_: i = n._C_.index(_C); du+=n._u_[i]
+                    cand_ += [(n,u,d)]; U+=u
+                for n,u,d in cand_:
+                    if u/U > Ave:  # filter with final U, N_/u_ aligned
```
boris-kz, line 514: this /U is actually wrong, there is no overlap between Ns in C.N_
we need overlap between C.N_s, but that can be computed after loop Cs are formed

kwcckw: > this /U is actually wrong, there is no overlap between Ns in C.N_

Yes, but /U is meant for normalization?

> we need overlap between C.N_s, but that can be computed after loop Cs are formed

But overlap should be exists between Cs only, so we can check overlapping Cs with N._C_?

boris-kz: I am reverting a bunch of changes, most of that claude slop is not relevant
partial revert:

    while True: # reform C_, add direct in-C_ cross-links for membership?
        C_,cnt,mat,dif,DTT,Du = [],0,0,0,np.zeros((2,9)),0; Ave = ave*(r+ccC); Avd = avd*(r+ccC)
        _Ct_ = [[C,0,0,0] for C in _C_]
        for cr, (_C,_U,rc) in enumerate(sorted(_Ct_, key=lambda t: t[1]*t[2], reverse=True),start=1):  # rc weighting?
            if _U > Ave:
                L_, N_,N__,u_,M,D,cc, dTT,du = [],[],[],[],0,0,0, np.zeros((2,9)),0  # /C
                Ct_ = []  # candidate Nts
                    m,d = vt_(dtt,ttcC); dTT+=dtt; c=n.c
                    oL_ = set(n.rim) & _C._L_  # peer rim overlap -> more precise m
                    if oL_: _m,_d,_,_c,_ = sum_vt(oL_,fm=1, fdiv=1, wTT=ttcC); m+=_m; d+=_d; c+=_c
                    Ct_ += [(n,m,d,c)]
                for n,m,d,c in Ct_:
                    if m*c > Ave:
                        N_+= [n]; u_+=[m]; L_+=n.L_; M+=m; D+=abs(d)*c; cc+=c
                            N__ += [_n]  # +|-Ls
                        if _C not in n._C_: du+=m  # not in extended _N__
                    else:
                DTT+=dTT; mat+=M; dif+=D; cnt+=cc
                if M*cc > Ave and val_(dTT, cr,_C.wTT*ttcC,(len(N_)-1)*Lw) > 0:
                    C = sum2C(N_,u_, root=Ft)
                    for n,u in zip(N_,u_):
                        n.rN_ += [C]; n.u_+=[u]
                    C._N_ = list(set(N__)-set(N_))  # new frontier
                    C._L_ = set(L_)  # peer links
                    if D<Avd: oC_+= [C]; u__ += [[n._u_[n._C_.index(C)] for n in C.N_]]  # output if stable, or if val_(DTT,fi=0) + D?
                    else:     C_ += [C]  # reform

kwcckw: Okay, i can see only `U` is removed, the others should be remained the same? With the norm, u could be > 1 when we have added `_m` from oL_?

boris-kz:
I actually don't think we need that oL_?
These partial edits, I haven't finished

kwcckw:
> I actually don't think we need that oL_?
These partial edits, I haven't finished

Okay, so that " peer rim overlap " shouldn't be valid with the usage of u now, i guess we can remove them.

PR #120, opened 2026-04-20
4/20/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -489,14 +489,14 @@ def cluster_C(Ft, E_,_r,_c):  # form centroids by clustering exemplar surround v
     _C_ = []
     for E in E_:  # along eigenvector?
         C = Copy_(E, Ft, init=1,typ=3)
-        C.N_,C.L_,C.m_,C.d_ = [],[],[],[]  # aligned
+        C.N_,C.L_,C.m_,C.d_ = [E],[],[E.m],[E.d]  # aligned
```
kwcckw, line 492: We need to init C.m_ here? Else we can't start the first eval below:

    if sum(_C.m_) *_C.c > Ave*_C.r:

frame_2D_alg/agg_recursion.py
```diff
@@ -540,11 +541,11 @@ def cluster_C(Ft, E_,_r,_c):  # form centroids by clustering exemplar surround v
 def cluster_P(__C_, N_, _c, root):  # FCM-style parallel centroid refine, may add proj_C
 
     cnt = 0  # r = root.r+1?:
-    _u__ = [[(N._m_[N._C_.index(C)] if C in N._C_ else 0) for N in N_] for C in __C_]  # N memberships per C
+    _u__ = [[(N.m_[N._c_.index(C)] if C in N._c_ else 0) for N in N_] for C in __C_]  # N memberships per C
     for u_ in _u__: s = sum(u_) or 1; u_[:] = [f/s for f in u_]
     while True:
         u__, dU = [], 0
-        C_ = [sum2C(N_, u_, root) for u_ in _u__]  # same N_ * updated membership u_
+        C_ = [sum2C(N_, u_, d_=[0 for _ in u_], root=root) for u_ in _u__]  # same N_ * updated membership u_
```
kwcckw, line 548: For sum2C in cluster_P, do we still need their d_ here?

Else we need to return it from vt_ below:

    u_ += [vt_(TT, ttcP)[0]]

frame_2D_alg/agg_recursion.py
```diff
@@ -566,9 +567,22 @@ def prune_C_(C_, root):
...
+        for n in C.N_: n.c_ = []; n.m_ = []; n.d_ = []
+    n_= []
+    for C in out_:
+        for n, m, d in zip(C.N_, C.m_, C.d_): 
+            n.c_ += [C]; n.m_ += [m]; n.d_ += [d]
+            n_ += [n]
+
+    for n in n_:
+        n.o = sum(n.m_)  # overlap is sum of m? or len(n.C_)?
+        n.o_ = sum(n.m_)/n.o
```
kwcckw, line 585: I'm not sure if we should use m here, or o could be simpler and use the length of n.c_?

frame_2D_alg/agg_recursion.py
boris-kz: that eval should be for prior vals:

`if sum(_C._m_) *_C.c > Ave*_C.r:`

frame_2D_alg/agg_recursion.py
boris-kz: yes, we will need it for proj_C

frame_2D_alg/agg_recursion.py
boris-kz: yes, length of n.c_ is good to start with, and we only need as multiplier to Ave in evals.
so o doesn't even need to be stored separately?

frame_2D_alg/agg_recursion.py
```diff
@@ -673,6 +687,7 @@ def sum2C(N_, u_, root=None, final=0, fP=0):  # fuzzy sum + base attrs for centr
     wTT = cent_TT(TT, R) * (wcP if fP else wcC)  # set param correlation weights
     m,d = vt_(TT,wTT)
     C = CN(typ=3, Nt= CF(N_=N_),dTT=TT,m=m,d=d,c=U,r=R, yx=yx, kern=kern,span=span, root=root, wTT=wTT, L_=L_)
+    C.m_ = u_; C.d_ = d_ 
     return C
```
boris-kz, line 692: so the globals should be updated with rV after ffeedback, not before?

                        rV, FTT_ = ffeedback(F)  # update globals:
                        global ave,avd, Lw,intw,distw, Fw_,FTT_
                        (ave,avd, Lw,intw,distw), Fw_,FTT_ = np.array([ave,avd, Lw,intw,distw])/rV, np.array(Fw_)/rV, np.array(FTT_)/rV

kwcckw: Yes, thanks. Right now that ` (ave,avd, Lw,intw,distw), Fw_,FTT_ = np.array([ave,avd, Lw,intw,distw])/rV, np.array(Fw_)/rV, np.array(FTT_)/rV` is outside of while loop, so globals are only updated with it in the nested frame_H.

frame_2D_alg/agg_recursion.py
kwcckw: If we need length of `n.c_` only, i guess it could be retrieved from len(n.c_), so we not really need to pack it.

frame_2D_alg/agg_recursion.py
kwcckw: Okay,so we need to add d_ the same way:

    _d__ = [[(N.d_[N._c_.index(C)] if C in N._c_ else 0) for N in N_] for C in __C_]

And we need to pack d below:

    m,d = vt_(TT, ttcP)
    u_ += [m]
    d_ += [d]

frame_2D_alg/agg_recursion.py
kwcckw: But we still need to init `_m_`? Unless it is default to true in the first iteration.

boris-kz: right, the first loop should be default

frame_2D_alg/agg_recursion.py
boris-kz: we don't need to use index, just pass i in sum2C, use as in my draft

frame_2D_alg/agg_recursion.py
boris-kz: you don't need to repeat what I said :)

frame_2D_alg/agg_recursion.py
kwcckw: Or we not even need i? Since it will be per `N._m_[i] or N._d_[i]` of the loop?

frame_2D_alg/agg_recursion.py
kwcckw: Yea, i just agree with your question above.

frame_2D_alg/agg_recursion.py
boris-kz: sum2C is per C, not per N

kwcckw: Okay right, thanks. We just need to parse `i` in order to get the current C's m and d.

frame_2D_alg/agg_recursion.py
```diff
@@ -540,11 +541,11 @@ def cluster_C(Ft, E_,_r,_c):  # form centroids by clustering exemplar surround v
 def cluster_P(__C_, N_, _c, root):  # FCM-style parallel centroid refine, may add proj_C
 
     cnt = 0  # r = root.r+1?:
-    _u__ = [[(N._m_[N._C_.index(C)] if C in N._C_ else 0) for N in N_] for C in __C_]  # N memberships per C
+    _u__ = [[(N.m_[N._c_.index(C)] if C in N._c_ else 0) for N in N_] for C in __C_]  # N memberships per C
```
boris-kz, line 544: so with the dense scheme:

    _u__ = [[N._m_ for N in N_] for C in __C_]  # N memberships per C

frame_2D_alg/agg_recursion.py
```diff
@@ -489,14 +489,14 @@ def cluster_C(Ft, E_,_r,_c):  # form centroids by clustering exemplar surround v
...
+        C.N_,C.L_,C.m_,C.d_ = [E],[],[E.m],[E.d]  # aligned
         C._N_ = list({n for l in E.rim for n in l.nt if (n is not E and n in Ft.N_)})  # frontier
         _C_ += [C]
     oC_=[]; fP = 0
     while True: # reform C_
-        C_, Ct_,cnt,mat,dif,rdn,DTT,Up = [],[],0,0,0,0,np.zeros((2,9)),0; Ave = ave*(r+ccC); Avd = avd*(r+ccC)
+        C_,cnt,mat,dif,rdn,DTT,Up = [],0,0,0,0,np.zeros((2,9)),0; Ave = ave*(_r+ccC); Avd = avd*(_r+ccC)
         c_= [[] for _ in _C_]
-        for _n in set([_n for _C in _C_ for _n in _C.N_]):
+        for _n in set([_n for _C in _C_ for _n in _C.N_+_C._N_]):
```
boris-kz, line 499: this `+_C._N_ assumes that _C._N_` has no external Cs?

frame_2D_alg/agg_recursion.py
```diff
@@ -525,6 +525,7 @@ def cluster_C(Ft, E_,_r,_c):  # form centroids by clustering exemplar surround v
             break
         if Up > Avd*rdn: _C_ = C_  # add n_ overlap in C_?
         else:   oC_+= C_; break  # converged
+    r = _r
```
boris-kz, line 528: we don't need this, r is loop-local?

frame_2D_alg/agg_recursion.py
kwcckw: Yes, thanks, this is actually better. Since we `_m_` and `_d_` mapped to `_C_`, we can retrieve d_ the same way here.

frame_2D_alg/agg_recursion.py
kwcckw: We are gonna eval both "frontier + current" again:

So their c_, m_ and d_ will be reset anyway? Or this reset should be conditional when we form new C only, under the section of this eval:

    if M > Ave*r and val_(dTT, r,_C.wTT*ttcC, (L-1)*Lw) > 0:

frame_2D_alg/agg_recursion.py
kwcckw: But we return this r? This r maybe updated from cross_comp:

    _,r = cross_comp(Ct,r)

frame_2D_alg/agg_recursion.py
boris-kz: It's probably fine for a sequential algo, but if the frontier Ns already have their own c_s, they should be merged in `_C_`

frame_2D_alg/agg_recursion.py
boris-kz: it should still be last loop only, that's where we get final C_:

                r = _r+R/cc  # loop-local, not ave?
                if M > Ave*r and val_(dTT, r,_C.wTT*ttcC, (len(n_)-1)*Lw) > 0:
                    for _n in [_n for l in n.rim for _n in l.nt if _n is not n]: N__+= [_n]  # +|-Ls
                    C = sum2C(n_,m_,d_, i, root=Ft)
                    if D<Avd: oC_+= [C]  # output if stable, or if val_(DTT,fi=0)+D?
                    else:     C_ += [C]  # reform
                DTT+=dTT; mat+=M; dif+=D; cnt+=cc; rdn+=R; Up+=up
        r = _r+ rdn/cnt

frame_2D_alg/agg_recursion.py
kwcckw: >  but if the frontier Ns already have their own c_s, they should be merged in _C_

Okay, `_C._N_` should be already have `_m_` and `_d_`, but if we remove it here, we would need to init their c_, m_ and d_ for the first iteration, since we init _N_ with E above, but we didn't init E.c_, E.m_ and E.d_:

    C._N_ = list({n for l in E.rim for n in l.nt if (n is not E and n in Ft.N_)})

boris-kz: they are empty for all nodes:

`for n in Ft.N_: n._c_,n.c_,n.m_,n._m_,n.d_,n._d_ = [],[],[],[],[],[]`

kwcckw: Yes, that's the problem because we need to replace them later:

    for n,m,d in zip(C.N_,C.m_,C.d_): n.c_[i]=C; n.m_[i]=m; n.d_[i]=d  # mapping-C vals

Without the init, we are getting error since there's nothing to replace.

boris-kz: they are replaced in the loop:

        c_= [[] for _ in _C_]
        for _n in set([_n for _C in _C_ for _n in _C.N_+_C._N_]):  # merge frontier C_ += C._N_'c__?
            _n.c_=copy(c_); _n.m_=copy(c_); _n.d_=copy(c_)  # fill /C index

boris-kz: So I pushed, haven't fully reviewed sum2C yet

kwcckw: Alright, i will check again later.

PR #122, opened 2026-04-22
4/22/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -485,17 +485,16 @@ def trans_cluster(G):
 def cluster_C(Ft, E_,_r,_c):  # form centroids by clustering exemplar surround via rims of new member nodes, within root
 
     N_= Ft.N_; oF = CoF.get(); oF.c += _c; oF.r+=_r  # revert if 0 clusters?
-    for n in N_: n.c_,n.m_,n.d_ = [],[],[]
     c_ = [[] for _ in E_]
-    for n in N_: n._c_=copy(c_); n._m_=copy(c_); n._d_=copy(c_)
+    for n in N_: n.c_=deepcopy(c_); n.m_=deepcopy(c_); n.d_=deepcopy(c_); n._c_=deepcopy(c_); n._m_=deepcopy(c_); n._d_=deepcopy(c_)
```
kwcckw, line 489: So we need deepcopy here, else those nested lists will be referencing the same list.

frame_2D_alg/agg_recursion.py
```diff
@@ -510,30 +509,31 @@ def cluster_C(Ft, E_,_r,_c):  # form centroids by clustering exemplar surround v
             r = _r+R/T  # loop-local, not ave?
             if M > Ave*r and val_(dTT, r,_C.wTT*ttcC, (len(n_)-1)*Lw) > 0:  # else: Up+= sum(_C._m_)+ sum([abs(d) for d in _C._d_])?
                 for n in n_: N__ += [_n for l in n.rim for _n in l.nt if _n is not n]  # +|-Ls
-                C = sum2C(n_,m_,d_, i, root=Ft)
+                C = sum2C(n_,m_,d_, i, final=2, root=Ft)
```
kwcckw, line 512: Looks like we can't use `i` here since `i` is per C from `for i,_C in enumerate(_C_)` , while `m_` and `d_` is added per C's core + frontier `for n in _C.N_+_C._N_`, so each `m_[i]`, `d_[i]` should be corresponding to `n_[i]` instead.

frame_2D_alg/agg_recursion.py
```diff
@@ -543,8 +543,9 @@ def cluster_P(_C_, _c, root):  # FCM-style parallel centroid refine, may add pro
 
     cnt = 0  # r = root.r+1?:
     N_ = list(set([N for C in _C_ for N in C.N_]))  # fully fuzzy: all Ns are in all Cs
-    _u__ = [N._m_ for N in N_]  # N/C vals, aligned with _C_
-    _d__ = [N._d_ for N in N_]
+    # below is not align with _C_, but aligned with N_ instead
+    _u__ = [[(0 if v == [] else v) for v in N.m_] for N in N_]  # N/C vals, aligned with _C_
```
kwcckw, line 547: I checked again and looks like `_u__` and `_d__` here is per `N_` instead of aligned with `_C_`?

Then `N.m_` is aligned with `_C_`, but we form `C_` based on `N_`?

    C_ = [sum2C(N_,u_,d_,i, root) for i,(u_,d_) in enumerate(zip(_u__,_d__))]

We may get more N_ compare to C_ (not all C has same set of Ns), which will become a problem.

frame_2D_alg/agg_recursion.py
```diff
@@ -564,7 +565,7 @@ def cluster_P(_C_, _c, root):  # FCM-style parallel centroid refine, may add pro
     for i, C in enumerate(C_):
         if C.m > ave * C.r:  # final pruning, C vals are competitive
             N_,m_,d_ = [],[],[]
-            for N,m,d in zip(C.N_,C._m_,C._d_):
+            for N,m,d in zip(C.N_,C.m_,C.d_):
```
kwcckw, line 568: It should be m_ and d_ here since we assign m_ and d_ to C in sum2C.

frame_2D_alg/agg_recursion.py
boris-kz: they are not nested

frame_2D_alg/agg_recursion.py
boris-kz: `C_ maps to _u__,_d__,` it's the sequence of same N_ with different weights

frame_2D_alg/agg_recursion.py
```diff
@@ -485,17 +485,16 @@ def trans_cluster(G):
 def cluster_C(Ft, E_,_r,_c):  # form centroids by clustering exemplar surround via rims of new member nodes, within root
 
     N_= Ft.N_; oF = CoF.get(); oF.c += _c; oF.r+=_r  # revert if 0 clusters?
-    for n in N_: n.c_,n.m_,n.d_ = [],[],[]
```
boris-kz, line 488: why did you remove that init?

frame_2D_alg/agg_recursion.py
boris-kz: thanks

frame_2D_alg/agg_recursion.py
kwcckw: Sorry, this shouldn't be relevant, they will be replaced anyway.

frame_2D_alg/agg_recursion.py
kwcckw: The problem is each C may have different N_?

C's N_ is added from core + rim : ` _C.N_+_C._N_`, which may not be the same for each C.

When we call this line:

    C_ = [sum2C(N_,u_,d_,i, root) for i,(u_,d_) in enumerate(zip(_u__,_d__))] 

Each `u_` and `d_` is `N.m_` and `N.d_`, which they have the size of `_C_`, but `i` is per `_u__` (from all Ns), so this `i` will not be matched with the size of `u_` and `d_`.


Also we init N.m_, N.d_ with c_, so N.m_ and N.d_ should be aligned with `_C_`, but we may not form C_ when this eval is false:

    if M > Ave*r and val_(dTT, r,_C.wTT*ttcC, (len(n_)-1)*Lw) > 0:

Which mean their values will remained as empty list and we need to replace it with 0?

frame_2D_alg/agg_recursion.py
kwcckw: They should be init the same way using copy(c_)?

boris-kz: that's only for priors: `_c_,_m_,_d_`

frame_2D_alg/agg_recursion.py
boris-kz: ok right, that's from cluster_C

kwcckw: Yes, and if we need the new Cs to be per N_, we can init it same way, but we can't reuse the same N.m_, N.d_ since they are pointing the Cs from cluster_C. I guess we can just fill them with the same N.m and N.d instead?

frame_2D_alg/agg_recursion.py
kwcckw: We need it for current m_, d_ and c_?

Because we will be assigning them in sum2C:

          C = CN(typ=3, Nt= CF(N_=N_),dTT=TT,m=m,d=d,c=M,r=R, yx=yx, kern=kern,span=span, root=root, wTT=wTT, L_=L_)
          C.m_=m_; C.d_=d_
          return C

boris-kz: right, thanks

frame_2D_alg/agg_recursion.py
```diff
@@ -510,30 +509,31 @@ def cluster_C(Ft, E_,_r,_c):  # form centroids by clustering exemplar surround v
...
-                C = sum2C(n_,m_,d_, i, root=Ft)
+                C = sum2C(n_,m_,d_, i, final=2, root=Ft)
                 C._N_ = list(set(N__)- set(n_))  # new frontier
                 if D<Avd: out_+=[C]; C_+=[None]  # output if stable, keep slot for alignment
                 else:     C_ += [C]  # reform
                 DTT+=dTT; mat+=M; dif+=D; cnt+=T; rdn+=R; Up+=up
         r = _r+ rdn/(cnt or eps)
-        if out_+C_ and mat*dif*wcC > ave*(r+ccC+2):  # if val_(DTT,len(oC_+C_)?
-            out_+= C_; T = sum([f.c for f in out_])
+        Cfil_ = [C for C in C_ if C is not None]  # filled Cs
```
boris-kz, line 518: thanks,

        vC_ = [C for C in C_ if C is not None]  # valid
        if  (mat+dif)* ((len(out_+vC_)-1)*Lw + wcC) > ave*(r+ccC+2):
            out_ += vC_; T = sum([f.c for f in out_])
            out_ = cluster_P(out_, T, Ft)  # refine all memberships in parallel by global backprop
            break
        if Up*((len(out_)-1)*Lw) > Avd*r:  # do next loop

frame_2D_alg/agg_recursion.py
```diff
@@ -510,30 +509,31 @@ def cluster_C(Ft, E_,_r,_c):  # form centroids by clustering exemplar surround v
...
-        iter += 1
+        else:   out_+= Cfil_; break  # converged
+
     if out_:
         for n in [N for C in out_ for N in C.N_]:  # exemplar V + sum n match_dev to Cs, m* ||C rvals:
             n.exe = (n.d if n.typ==1 else n.m) + np.sum(n.m_) - ave
         if val_(DTT, r, Ft.root.wTT*ttcC, (len(out_)-1)*Lw) > 0:
             Ct = sum2F(out_,Ft.root.Ct)
-            if not Ft.root.Ct: Ft.root.Ct = Ct
+            if not Ft.root.Ct: Ft.root.Ct = Ct; Ct.root = Ft.root
```
boris-kz, line 536: thanks

frame_2D_alg/agg_recursion.py
boris-kz: How about this:

    N_ = list(set([N for C in _C_ for N in C.N_]))  # all Ns are in all Cs
    _c_ = [0 for _ in _C_]; _u__,_d__ = [],[]
    for N in N_:
        m_= copy(_c_); d_= copy(_c_)
        for c,m,d in zip(N.c_,N.m_,N.d_): i= _C_.index(c); m_[i] =m; d_[i] =d
        _u__+=[m_]; _d__+=[d_]

frame_2D_alg/agg_recursion.py
boris-kz: I rechecked `cluster_C`. It is still not safe. The biggest problems are not minor style issues; a few paths will either crash or prevent centroid expansion from actually iterating. References below are to the uploaded `agg_recursion.py`. 

## Main issues

### 1. `Up` continuation test is using the wrong count

Current:

```python
if Up*((len(out_)-1)*Lw) > Avd*r:
```

This is wrong because `out_` is often empty on the first pass. Then `(len(out_)-1)*Lw` is negative, so the loop exits even when `Up` is large and the active centroids are still changing.

Use active+finished count, or at least active count:

```python
K = len(out_) + len(V_)
if Up * (max(K-1, 1)*Lw + wcC) > Avd*r:
```

Without this, `cluster_C` usually becomes one-pass expansion, not iterative centroid reform.

---

### 2. `None` placeholders will crash the reset path

You preserve index alignment with:

```python
C_ += [None]
```

But later:

```python
for _n in set([_n for _C in _C_ for _n in _C.N_+_C._N_]):
```

If `_C_` contains `None`, this crashes. Minimal fix:

```python
for _n in {n for C in _C_ if C is not None for n in C.N_+C._N_}:
```

This matters because stable centroids are moved to `out_` while unstable centroids continue.

---

### 3. Membership arrays should not use `[]` placeholders

Current:

```python
c_ = [[] for _ in E_]
for n in N_:
    n.c_=copy(c_); n.m_=copy(c_); n.d_=copy(c_)
```

This causes two problems:

```python
np.sum(n.m_)
```

can fail if `n.m_` contains mixed floats and empty lists, and `cluster_P` later does `_C_.index(c)`, where `c` may be `[]`.

Use `None` for centroid slots and `0.0` for numeric slots:

```python
c0 = [None for _ in E_]
v0 = [0.0 for _ in E_]

for n in N_:
    n.c_ = copy(c0); n.m_ = copy(v0); n.d_ = copy(v0)
    n._c_ = copy(c0); n._m_ = copy(v0); n._d_ = copy(v0)
```

Then replace:

```python
if _C in n._c_:
```

with:

```python
if n._c_[i] is _C:
```

That preserves index semantics and avoids false matches.

---

### 4. New frontier is no longer constrained to `Ft.N_`

Initial frontier is constrained:

```python
C._N_= list({n for l in E.rim for n in l.nt if (n is not E and n in N_)})
```

But later frontier update is not:

```python
C._N_ = list(set(N__)- set(n_))
```

That can pull nodes from outside the current fork/root. It contradicts the function comment: “within root”.

Add an `Nset` once:

```python
Nset = set(N_)
```

Then update frontier as:

```python
C._N_ = list((set(N__) & Nset) - set(n_))
```

---

### 5. `M` and `D` are totals, but thresholds are not scaled by `T`

You accumulate:

```python
T += c
M += m*c
D += abs(d)*c
```

But compare:

```python
if M > Ave*r ...
if D < Avd:
```

This mixes total `M,D` with unscaled thresholds. For larger clusters, `D < Avd` becomes almost impossible, while `M > Ave*r` becomes increasingly permissive or inconsistent depending on `m`.

Probably closer:

```python
if M > Ave * T * r ...
...
if D < Avd * T:
```

or compare normalized means:

```python
mC = M / (T or eps)
dC = D / (T or eps)

if mC > Ave*r ...
...
if dC < Avd:
```

The second version is cleaner if `m,d` from `vt_` are intended as per-node values.

---

### 6. `DTT` only represents the last loop, not all `out_`

`DTT` is reset every `while` iteration:

```python
C_, cnt,mat,dif,rdn,DTT,Up = [],0,0,0,0,np.zeros((2,9)),0
```

But after the loop you evaluate all `out_`:

```python
if val_(DTT,r, Ft.root.wTT*ttcC, (len(out_)-1)*Lw) > 0:
```

If some centroids were stabilized in earlier iterations, they are in `out_` but not in the final-loop `DTT`.

Safer: recompute from `out_` before final evaluation:

```python
DTT, Cc, r = sum_vt(out_, wTT=ttcC)
```

Then use that `DTT,r`.

---

### 7. `Copy_(E, Ft, init=1, typ=3)` mutates `E.root` too early

`Copy_` with `init=1` sets `N.root = C`. In `cluster_C`, this happens while the centroid is still only a tentative seed. If the centroid is later rejected, the exemplar root remains corrupted.

For centroid seeds, avoid `Copy_(..., init=1)` unless you remove that side effect. Either make a centroid-specific copy path, or restore `E.root` immediately. Conceptually, candidate centroid formation should not alter node ownership until final acceptance.

---

## The handoff to `cluster_P` is currently broken

This affects `cluster_C` because it calls:

```python
out_ = cluster_P(out_, T, Ft)
```

The main bug is here:

```python
C_ = [sum2C(N_,u_,d_,i, root) for i,(u_,d_) in enumerate(zip(_u__,_d__))]
```

`_u__` is shaped as `N x C`, but this loop treats it as if it were `C x N`. With more nodes than centroids, it can directly hit `IndexError` in `sum2C`.

The centroid recompute should iterate over centroid index:

```python
C_ = []
for i in range(len(_C_)):
    u_ = [u[i] for u in _u__]
    d_ = [d[i] for d in _d__]
    C_ += [sum2C(N_, u_, d_, i, root, final=2, fP=1)]
```

Also guard sparse membership import:

```python
for c,m,d in zip(N.c_,N.m_,N.d_):
    if c in _C_:
        i = _C_.index(c)
        m_[i] = m
        d_[i] = d
```

And this can fail when no centroids are dropped:

```python
dCt = sum2F(list(set(_C_)-set(out_)),CF())
```

because `sum2F([])` is invalid. Use:

```python
drop_ = list(set(_C_) - set(out_))
dCt = sum2F(drop_, CF()) if drop_ else CF()
```

Also make the call robust:

```python
out_ = cluster_P(out_, T, Ft) or []
```

---

## Minimal patch priority

I would patch in this order:

1. Replace `[]` membership placeholders with `None` / `0.0`.
2. Fix the `Up` gate to use `len(out_)+len(V_)`, not `len(out_)`.
3. Filter `None` in the reset comprehension.
4. Constrain renewed frontier to `Ft.N_`.
5. Fix `cluster_P`’s `N x C` vs. `C x N` transpose bug.
6. Recompute final `DTT,r` from `out_` before final `val_`.
7. Avoid `Copy_(init=1)` root mutation for tentative centroids.

After those, the remaining question is conceptual: whether signed `m` should be used directly as centroid mass in `sum2C`. I think negative `m` should not contribute negative mass to a centroid; either filter those nodes before `sum2C`, or use a non-negative membership such as `max(m, 0)` / relative match.

frame_2D_alg/agg_recursion.py
kwcckw: So using this approach, we are creating new `C_` that's aligned with `N_`, which may have different size with `_C_`.

Also after the init of `_u__` and `_d__`, we need to init N_'s m_,d_ and c_ with `c_ = [0 for _ in N_]`? Because their m_, d_ and c_ will be reassigned in the subsequent sum2C:

     C_ = [sum2C(N_,u_,d_,i, root) for i,(u_,d_) in enumerate(zip(_u__,_d__))]

frame_2D_alg/agg_recursion.py
kwcckw: This eval `if Up*((len(out_)-1)*Lw) > Avd*r:` shouldn't be using the stable out_? Because we only do the next loop when they are not stable, and we can eval with vC_ instead?

boris-kz: right, should be len(V_)

frame_2D_alg/agg_recursion.py
boris-kz: `_C_` is an arg, it's not aligned with N_ 

out_ = cluster_P(out_, T, Ft)  # refine all memberships in parallel by global backprop

kwcckw: Yes, but the problem i mentioned above is another problem where we need to reinit m_, d_ and c_ in each N, else they are still pointing to the prior `_C_` size. Also could you paste the latest full code of cluster_P again?

boris-kz: I have that above

def cluster_P(_C_, _c, root):  # FCM-style parallel centroid refine, may add proj_C

    cnt = 0  # r = root.r+1?:
    N_= list(set([N for C in _C_ for N in C.N_]))  # all Ns are in all Cs
    Q = [0 for _ in _C_]; _u__,_d__ = [],[]  # all replaced
    for N in N_:
        m_= copy(Q); d_= copy(Q)  # N/C vals
        for c,m,d in zip(N.c_,N.m_,N.d_): i =_C_.index(c); m_[i] = m; d_[i] = d  # aligned with _C_
        _u__ += [m_]; _d__ += [d_]  # aligned with N_
    while True:
        u__,d__, dU,dD = [],[],0,0  # fixed-length C_ and N_
        C_ = [sum2C(N_,u_,d_,i, root,wO=wcP) for i,(u_,d_) in enumerate(zip(_u__,_d__))]  # same N_ * updated membership
        for N,_u_,_d_ in zip(N_,_u__,_d__):
            u_,d_ = [],[]
            for C in C_:
                TT,_ = base_comp(C,N); u,d = vt_(TT, ttcP); u_+=[u]; d_+=[d]
            s = sum(u_); u_ = [f/s for f in u_]  # per-N membership, Σ_u_=1: norm for cross-C redundancy, or for totals only?
            dU += sum(abs(u-_u) for u,_u in zip(u_,_u_))  # normalize?
            dD += sum(abs(d-_d) for d,_d in zip(d_,_d_))
            u__+=[u_]; d__+=[d_]
        cnt += 1
        if (dU+dD) * wcP > ave * (root.r+ccP) * len(N_):  # memberships change
            _u__= u__; _d__= d__
        else: break  # converged
    out_ = []
    for i, C in enumerate(C_):
            N_,m_,d_ = [],[],[]
            for N,m,d in zip(C.N_,C.m_,C.d_):
                if m * N.c > ave * N.r: N_+=[N]; m_+=[m]; d_+=[d]
            if N_: out_ += [sum2C(N_, m_, d_, i, root=root, final=1)]
    if out_:
        return out_

kwcckw: Okay, and i think we need to reinit m_, d_ and c_?

    for N in N_:
        m_= copy(Q); d_= copy(Q)  # N/C vals
        for c,m,d in zip(N.c_,N.m_,N.d_): i =_C_.index(c); m_[i] = m; d_[i] = d  # aligned with _C_
        N.m_ = copy(Q);  N.d_ = copy(Q); N.c_ = copy(Q)
        _u__ += [m_]; _d__ += [d_]  # aligned with N_
        
Because we will remap it in sum2C:

      C = CN(typ=3, Nt= CF(N_=N_),dTT=TT,m=m,d=d,c=M,r=R, yx=yx, kern=kern,span=span, root=root, wTT=wTT, L_=L_)
      C.m_=m_; C.d_=d_
      return C
      
The original N.m_, N.d_ and N.c_ is based on `_C_`, and we need to reinit it so that the size matches the size of N_.

boris-kz: I have that above:

m_= copy(Q); d_= copy(Q)  # N/C vals
for c,m,d in zip(N.c_,N.m_,N.d_): i =_C_.index(c); m_[i] = m; d_[i] = d  # aligned with _C_

N.c_ is redundant to C_ here

kwcckw: Your code actually works differently:

    for c,m,d in zip(N.c_,N.m_,N.d_): i =C.index(c); m_[i] = m; d_[i] = d # aligned with C
    
Because this is using existing N.m_, which is not the same size as N_? Now we are packing ` _u__ += [m_]; _d__ += [d_] ` per `N_` instead of per `_C_`?

boris-kz: ok just need to add:

        N.m_= m_; N.d_= d_

kwcckw: Okay, but we probably need to copy it because C.m_ and C.d_ will be assigned with the same m_ and d_. And we still need to init c_ since we will be assigning it in sum2C although it's redundant to C_.

Also this assignment should be valid for cluster_P only since we have the same number of C_ and N_ here.

boris-kz: so here we get values from N.m_,N.d_, but `i =_C_.index(c)`, so their new placement maps to _C_

boris-kz: it is copied:

m_= copy(Q); d_= copy(Q)

frame_2D_alg/agg_recursion.py
```diff
@@ -666,8 +667,8 @@ def sum2C(N_, m_,d_, i, root=None, final=0, fP=0):  # fuzzy sum + base attrs for
 
     L_, mc_ = [],[]  # N/C contribution = c-scaled u_
     for j, N in enumerate(N_):
-        if final: L_ += [CN(typ=1, w=[N.c * m_[j]])]  # add u,c,r,TT,span,angl for proj_C?
-        mc_ += [N.c * m_[i]]  # N/C contribution  (this should be default even in final?)
+        if final == 1: L_ += [CN(typ=1, w=[N.c * m_[j]])]  # add u,c,r,TT,span,angl for proj_C?
```
boris-kz, line 670: This L forming is not complete, we need to flesh it out

frame_2D_alg/agg_recursion.py
kwcckw: > so here we get values from N.m_,N.d_, but `i =_C_.index(c)`, so their new placement maps to _C_

Sorry, i was confused earlier. So it's still having size incompatibility issue. 

    m_= copy(Q); d_= copy(Q)

Input m_ and d_ have the same size as C_, but we are creating new Cs per N_, which has a size mismatch.

N.c_, N.m_ and N.d_ should be init with size of N_ instead of copying Q.

To clarify, we are creating new Cs per `N_`, and both `C.m_` and `N.m_` should have the size of `N_` instead of `_C_`?

From this line ` for c,m,d in zip(N.c_,N.m_,N.d_): i =_C_.index(c); m_[i] = m; d_[i] = d`, `m_` and `d_` are having the same size with `_C_` rather than `N_`.

boris-kz: It should not be size N_.
You are still confused

kwcckw: Okay, then this line is wrong?

    C_ = [sum2C(N_,u_,d_,i, root) for i,(u_,d_) in enumerate(zip(_u__,_d__))]

That `_u__` is per N_ and this is creating new C per N_.

boris-kz: They don't have to align, just use * sum(m_) instead of element-wise

PR #123, opened 2026-04-23
4/23/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -595,6 +605,13 @@ def sum2C(N_, m_,d_, i, root=None, final=0, wO=wcC):  # fuzzy sum + base attrs f
     C = CN(typ=3, Nt= CF(N_=N_),dTT=TT,m=m,d=d,c=M,r=R, yx=yx, kern=kern,span=span, root=root, wTT=wTT, L_=L_)
     C.m_=m_; C.d_=d_
     for n,m,d in zip(C.N_,C.m_,C.d_): n.c_[i] = C; n.m_[i] = m; n.d_[i] = d  # mapping-C vals
+
+    if final==1:  # this L could be from comp instead?
```
kwcckw, line 609: > This L forming is not complete, we need to flesh it out

So this L should be formed by comparing C with the N instead? With comp, we can get all the actual params.

frame_2D_alg/agg_recursion.py
```diff
@@ -546,12 +546,24 @@ def cluster_P(_C_, _c, root):  # FCM-style parallel centroid refine, may add pro
 
     cnt = 0  # r = root.r+1?:
     N_= list(set([N for C in _C_ for N in C.N_]))  # all Ns are in all Cs
-    Q = [0 for _ in _C_]; _u__,_d__ = [],[]  # all replaced
-    for N in N_:
+    # per N_ (the existing method, each N is a seed)
+    Q = [0 for _ in N_]; _u__,_d__ = [],[]  # all replaced
```
kwcckw, line 550: > They don't have to align, just use * sum(m_) instead of element-wise

Actually the prior version of cluster_P is meant to have each N in N_ as seed, so the number of C_ and N_ should be the same. By using the sum(m_), it looks better and makes more sense to me, something like the code below (my line 550 to 554).

If the new C_ is per input `_C_` (my code 558 - 564 below), it looks wrong to me since we are reinit the same number of C_, which each of their C>N_ will be different.

boris-kz: It's the same len C_, each has the same len N_, only m_ and d_ are different

kwcckw: Okay thanks, so if it;s the same len C_, we need to retrieve each N's membership of C, so this line is wrong:

    C_ = [sum2C(N_,u_,d_,i, root,wO=wcP) for i,(u_,d_) in enumerate(zip(_u__,_d__))]  # same N_ * updated membership
    
It should be something like:

        C_ = []
        for i in range(len(Q)):  # Q is per C_, i is C's index
            u_ = [_u__[j][i] for j in range(len(N_))]
            d_ = [_d__[j][i] for j in range(len(N_))]
            C_.append(sum2C(N_, u_, d_,i, root=root, wO=wcP, final=2))

frame_2D_alg/agg_recursion.py
boris-kz: It's already compared in cluster_ functions, forming m_ and d_, we just need assign more attrs to the Ls

frame_2D_alg/agg_recursion.py
boris-kz: for N in N_:
        m_= copy(Q); d_= copy(Q)
        for c,m,d in zip(N.c_,N.m_,N.d_):  # N/C vals
            i = _C_.index(c); m_[i] = m; d_[i] = d  # align with _C_
        N.m_ = m_; N.d_ = d_; _u__ += [m_]; _d__ += [d_]  # aligned with N_

Everything is aligned already

frame_2D_alg/agg_recursion.py
```diff
@@ -546,12 +546,24 @@ def cluster_P(_C_, _c, root):  # FCM-style parallel centroid refine, may add pro
...
+        
+    '''
+    # per C_ (the number of N_ and C_ are different)
+    Q = [0 for _ in _C_]
+    _u__ = [list(np.zeros(len(N_))) for _ in _C_]; _d__ = [list(np.zeros(len(N_))) for _ in _C_]
+    for N in N_:
         for c,m,d in zip(N.c_,N.m_,N.d_):  # N/C vals
-            i = _C_.index(c); m_[i] = m; d_[i] = d  # align with _C_
-        N.m_ = m_; N.d_ = d_; _u__ += [m_]; _d__ += [d_]  # aligned with N_
+            if c is not None:
```
boris-kz, line 562: if c: i = _C_.index(c); m_[i] = m; d_[i] = d  # align with _C_
?

frame_2D_alg/agg_recursion.py
```diff
@@ -595,6 +605,13 @@ def sum2C(N_, m_,d_, i, root=None, final=0, wO=wcC):  # fuzzy sum + base attrs f
     C = CN(typ=3, Nt= CF(N_=N_),dTT=TT,m=m,d=d,c=M,r=R, yx=yx, kern=kern,span=span, root=root, wTT=wTT, L_=L_)
     C.m_=m_; C.d_=d_
     for n,m,d in zip(C.N_,C.m_,C.d_): n.c_[i] = C; n.m_[i] = m; n.d_[i] = d  # mapping-C vals
+
+    if final==1:  # this L could be from comp instead?
+        for j, N in enumerate(N_):
+            dTT, _ = base_comp(C, N)  # but C isn't constructed yet — order issue
```
boris-kz, line 611: so base_comp was already done, L.c = N.c, L.r=N.r, we just need to compute L span,angl?

frame_2D_alg/agg_recursion.py
kwcckw: > ```
> for N in N_:
>     m_= copy(Q); d_= copy(Q)
>     for c,m,d in zip(N.c_,N.m_,N.d_):  # N/C vals
>         i = _C_.index(c); m_[i] = m; d_[i] = d  # align with _C_
>     N.m_ = m_; N.d_ = d_; _u__ += [m_]; _d__ += [d_]  # aligned with N_
> ```
> 
> Everything is aligned already

But this part is still creating Cs that's same size with N_:

    C_ = [sum2C(N_,u_,d_,i, root,wO=wcP) for i,(u_,d_) in enumerate(zip(_u__,_d__))]  # same N_ * updated membership
    
 Each N.m_ is aligned to Q, but not N_ (there will be error when there's more Ns compare to C since their size is different). This problem is already highlighted by your prior posted LLM:
 
 
image

frame_2D_alg/agg_recursion.py
kwcckw: Yes, we just need to skip None here.

frame_2D_alg/agg_recursion.py
kwcckw: > It's already compared in cluster_ functions, forming m_ and d_, we just need assign more attrs to the Ls

Okay, for span and angl, we can just use dy_dx?

    dy_dx = C.yx - N.yx
    span = np.hypot(*dy_dx)
    angl = np.array([dy_dx, np.sign(N.dTT[1] @ ttcN[1])], dtype=object)

frame_2D_alg/agg_recursion.py
boris-kz: ok so these arrays are not nested, we need reciprocal asign instead? 
dropping these u_s for consitency:

    n_ = [0 for _ in N_]; Cm_,Cd_ = copy(n_),copy(n_)  # N_/C, align with N_
    c_ = [0 for _ in _C_]  # C_/N, align with _C_
    for j, N in enumerate(N_):
        m_,d_ = copy(c_),copy(c_)  # C_/N vals
        for c,m,d in zip(N.c_,N.m_,N.d_): 
            i =_C_.index(c); m_[i] = m; d_[i] = d
            Cm_[j] = [m]; Cd_[j] = [d]  # per C, aligned with N_
        N.m_, N.d_ = m_, d_  # reciprocal per N, aligned with C_

frame_2D_alg/agg_recursion.py
boris-kz: right

frame_2D_alg/agg_recursion.py
kwcckw: But if we form C with Cm_ and Cd_, it's still creating N_'s size C_?

I think we need to retrieve per C's m_ and d_  from the prior ` _u__` and `_d__`:

    C_ = []
    for i in range(len(Q)):  # Q is per C_, i is C's index
        u_ = [_u__[j][i] for j in range(len(N_))]
        d_ = [_d__[j][i] for j in range(len(N_))]
        C_.append(sum2C(N_, u_, d_,i, root=root, wO=wcP, final=2))

boris-kz: ok so it's a  rectangle matrix for each, but the original values are still retrieved from

        for c,m,d in zip(N.c_,N.m_,N.d_):
            i =_C_.index(c); m_[i] = m; d_[i] = d

?

frame_2D_alg/agg_recursion.py
boris-kz, line 512: we need more specific evals, like

`if M*(_C.w+ wC_*len(n_)+wC_*len(n_)) > Ave*(r+_C.r+ cC_*len(n_)):  `

frame_2D_alg/agg_recursion.py
boris-kz: def cluster_P(_C_, _c, root):  # FCM-style parallel centroid refine, may add proj_C

    cnt = 0  # r = root.r+1?:
    N_ = list(set([N for C in _C_ for N in C.N_]))  # all Ns are in all Cs
    md__ = np.zeros((len(N_), len(_C_), 2))
    for j, N in enumerate(N_):  # row
        for c,m,d in zip(N.c_,N.m_,N.d_):  # col
            if c in _C_:  # else pruned
                i = _C_.index(c); md__[j,i] = m,d

boris-kz: Claude: 
Diff for `cluster_P`:

```python
def cluster_P(_C_, _c, root):  # FCM-style parallel centroid refine, may add proj_C

    cnt = 0  # r = root.r+1?:
    N_ = list(set([N for C in _C_ for N in C.N_]))  # all Ns are in all Cs
    _md__ = np.zeros((len(N_), len(_C_), 2))  # [m,d] per N,C
    for j, N in enumerate(N_):
        for c,m,d in zip(N.c_,N.m_,N.d_):
            if c in _C_: i = _C_.index(c); _md__[j,i] = m,d
    while True:
        md__ = np.zeros_like(_md__)
        C_ = [sum2C(N_, _md__[:,i,0], _md__[:,i,1], i, root) for i in range(len(_C_))]
        for j,N in enumerate(N_):
            for i,C in enumerate(C_):
                TT,_ = base_comp(C,N); md__[j,i] = vt_(TT, ttcP)
        s_ = md__[:,:,0].sum(axis=1, keepdims=True); md__[:,:,0] /= eps_(s_)  # row-norm u
        dU = np.abs(md__[:,:,0]-_md__[:,:,0]).sum()
        dD = np.abs(md__[:,:,1]-_md__[:,:,1]).sum()
        cnt += 1
        if (dU+dD) * wcP > ave * (root.r+ccP) * len(N_): _md__ = md__
        else: break  # converged
    out_ = []
    for i, C in enumerate(C_):
            N_,m_,d_ = [],[],[]
            for N,m,d in zip(C.N_, md__[:,i,0], md__[:,i,1]):
                if m * N.c > ave * N.r: N_+=[N]; m_+=[m]; d_+=[d]
            if N_: out_ += [sum2C(N_, m_, d_, i, root=root, final=1)]
    if out_:
        return out_
```

Two questions:
1. `sum2C` reads `m_[j if final else i]` — passing a 1D slice `md__[:,i,0]` to it works as a sequence, indexed by `j`. OK?
2. Final pruning iterates `C.N_` paired with `md__[:,i,*]` — assumes `C.N_ == N_` order (it does, since `sum2C(N_,...)` stores `N_` as-is). Confirm intended.

boris-kz: edited:

def cluster_P(_C_, _c, root):  # FCM-style parallel centroid refine, may add proj_C

    cnt = 0  # r = root.r+1?:
    N_ = list(set([N for C in _C_ for N in C.N_]))  # all Ns are in all Cs
    Ln, Lc = len(N_),len(_C_); L=Lc*Ln  # constant
    _md__ = np.zeros((Ln, Lc, 2))  # [m,d] per N,C
    for j, N in enumerate(N_):
        for c,m,d in zip(N.c_,N.m_,N.d_):
            if c in _C_: i = _C_.index(c); _md__[j,i] = m,d
    while True:
        md__ = np.zeros_like(_md__)
        for j,N in enumerate(N_):
            for i,C in enumerate(_C_):
                TT,_ = base_comp(C,N); md__[j,i] = vt_(TT, ttcP)  # use rc?
        C_ = [sum2C(N_, md__[:,i,0], md__[:,i,1], i, root) for i in range(Lc)]
        M = md__[:,:,0].sum()  # abs value
        dM = np.abs(md__[:,:,0] -_md__[:,:,0]).sum()
        dD = np.abs(md__[:,:,1] -_md__[:,:,1]).sum()  # update signal only
        cnt += 1
        if M * (dM+dD) * (wcP*L) > ave * (root.r+ ccP*L):
            _C_= C_; _md__ = md__
        else: break  # converged or weak
    out_ = []
    for i, C in enumerate(C_):
            N_,m_,d_ = [],[],[]
            for N, m,d in zip(C.N_, md__[:,i,0], md__[:,i,1]):
            if N_:
                out_ += [sum2C(N_,m_,d_,i, root=root, final=1)]
    if out_:
        return out_

kwcckw: I think we can skip `if c in _C_` and just `if c`? All Cs must be in `_C_` if it's not a None.

Also each N.m_, d_ and c_ should be init with size of `_C_`? So we need to include this after the assignment:

    for j, N in enumerate(N_):
        for c,m,d in zip(N.c_,N.m_,N.d_):
            if c in _C_: i = _C_.index(c); _md__[j,i] = m,d
        N.m_ = _md__[j][:,0].copy()
        N.d_ = _md__[j][:,1].copy()


Then in this line:

    C_ = [sum2C(N_, _md__[:,i,0], _md__[:,i,1], i, root) for i in range(len(_C_))]

This `_md__[:,i,0],` (m_) is aligned with N_, which is the membership of each N within C, so in sum2C, we should use `j` since it is aligned now:

    for j, (N,m,d) in enumerate(zip(N_,m_,d_)):
        mc_ += [N.c * m_[j]]  # N/C else C/N contribution

boris-kz: >Also each N.m_, d_ and c_ should be init with size of _C_?

not in initial Ns. And Ns don't change during refining, so we only need to assign final c_,m_,d_ 

>mc_ += [N.c * m_[j]]  # N/C else C/N contribution

that weighted contribution is probably for pruning only, we already have that in sum2C and:

            for N, m,d in zip(C.N_, md__[:,i,0], md__[:,i,1]):

boris-kz: Most of sum2C can be replaced by sum2F?

kwcckw: > not in initial Ns. And Ns don't change during refining, so we only need to assign final c_,m_,d_

But those initial Ns may still carry the m_, d_ and c_ init from prior cluster_C? We need to replace that?

> that weighted contribution is probably for pruning only, we already have that in sum2C and

I think the weighted contribution is the core section since we need that to compute each N's contributed params based on their soft membership?

> Most of sum2C can be replaced by sum2F?

But sum2F has many redundant sections, such as merge, froot, sum_H, which is not relevant to Cs. I think it's still better to separate them for clarity. Also this part is unique for sum2C since it computes based on membership:

    M = sum(mc_)
    for N, mc in zip(N_,mc_):
        rc = mc/M; TT+= N.dTT*rc; kern+= N.kern*rc; span+= N.span*rc; yx+= N.yx*rc; R+= N.r*rc

boris-kz: N m_, d_, c_ should be only from final loop of cluster_C 

> I think the weighted contribution is the core section since we need that to compute each N's contributed params based on their soft membership?

basic weighting is by N.m_[i] per C, not sure about combining that with N.c

boris-kz: Current code does both: `mc_ += [N.c * m_[j if final else i]]`, so contribution = `N.c * m`. That mixes two different things:

- `N.c` = N's own mass (how much it counts globally, independent of any C).
- `m_[i]` = N's fit to *this* C.

Combining them weights a node by "intrinsic importance × membership in C", which is reasonable if you want strong-and-fitting nodes to dominate. But it also means a high-`c` node with mediocre `m` outweighs a low-`c` node with excellent fit. That's a bias toward massive nodes, not toward cluster coherence.

For a pure centroid (prototype of fit), use `m_[i]` alone — FCM convention.
For a mass-weighted prototype (accounts for node significance), keep `N.c * m_[i]`.

In CogAlg terms: centroid is "shared quantity among members". If `N.c` already reflects N's summed match from prior comparisons, then using it double-counts — that evidence is what produced high `m` in the first place. So I'd lean toward `m_[i]` only here, and let `N.c` enter separately via `C.c` (total mass) for downstream eval.

Which interpretation did you intend?

kwcckw: > N m_, d_, c_ should be only from final loop of cluster_C

Yes, but actually they may still carry empty None value if that E doesn't form C at all, that's when the 

    if M > Ave*r and val_(dTT, r,_C.wTT*ttcC, (len(n_)-1)*Lw) > 0`

or the current eval 

    if M*(_C.w+ wC_*len(n_)+wC_*len(n_)) > Ave*(r+_C.r+ cC_*len(n_)): 

 is false.

> basic weighting is by N.m_[i] per C, not sure about combining that with N.c

Yes, but existing sum2F uses `rc_ = c_/C`, which is different with the N.m_[i] weighting.

>For a pure centroid (prototype of fit), use m_[i] alone — FCM convention.
> For a mass-weighted prototype (accounts for node significance), keep N.c * m_[i].

Okay, but i would prefer the mass-weighted method since it is consistent with the sum2F where their values are adjusted by c too. Also we are not really use c in the eval, so it's better to apply them here with m_[i].

boris-kz: >Yes, but existing sum2F uses rc_ = c_/C, which is different with the N.m_[i] weighting.

 we can add a fork with rc_ = m_/M, it should work the same 

also this padding N c_,m_,d_ with None for all Cs is not used now, it was just for cluster_P? 
sorry I wasted so much time on it

kwcckw: > we can add a fork with rc_ = m_/M, it should work the same

Alright, then we can try to merge them.

> also this padding N c_,m_,d_ with None for all Cs is not used now, it was just for cluster_P?
sorry I wasted so much time on it

Okay, i was thinking that's for aligning the m_, d_ and c_ when we call sum2C with i in cluster_C, so if the padding is removed, i think that `i` may not be relevant now.

boris-kz: So I removed the padding, please go though sum2C / F

kwcckw: Alright, i will go through them later.

PR #124, opened 2026-04-24
4/24/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -647,24 +653,45 @@ def add2F(F, n, fr=0, merge=0):  # unpack for batching in sum2F
     if hasattr(F,'H') and getattr(n,'H',None): add_H(F.H, n.H, F)
     if hasattr(n,'C_') and hasattr(n,'C_'): F.C_ = getattr(F,'C_',[]) + n.C_  # same for L_?
 
-def sum2F(N_, root, fr=0, merge=0, froot=0):  # -> CF/CN
+def sum2F(N_, root, fr=0, merge=0, froot=0, m_=None, d_=None, ci=None, final=0, wO=wcC):
 
     a = 'rTT' if fr else 'dTT'
-    c_ = np.array([n.c for n in N_], dtype=float); C = c_.sum(); rc_ = c_/C
+    fC = m_ is not None
```
kwcckw, line 659: To combine both sum2F and sum2C, only the section below is the same for both:

    a = 'rTT' if fr else 'dTT'
    fC = m_ is not None
    w_ = np.array(m_ if fC else [n.c for n in N_]); C = w_.sum(); rc_ = w_/C  # weights
    wC = rc_ @ np.array([getattr(n,'wc',1) for n in N_])
    if hasattr(N_[0], 'kern'):
        kern = rc_ @ np.stack([n.kern for n in N_])
        span = rc_ @ np.array([n.span for n in N_])
        yx   = rc_ @ np.stack([n.yx for n in N_])
  

Else they work very differently. I think it's better we pack this as a function and call them separately from sum2F or sum2C?

frame_2D_alg/agg_recursion.py
```diff
@@ -502,13 +502,14 @@ def cluster_C(Ft, E_,_r,_c):  # form centroids by clustering exemplar surround v
                 dtt,_ = base_comp(_C,n)  # or comp_N, decay?
                 m,d = vt_(dtt,ttcC); dTT+=dtt
                 n_+=[n]; m_+=[m]; d_+=[d]; c=n.c; T+=c; M+=m*c; D+=abs(d)*c; R+=n.r*c  # scale totals only?
+                n.m_ += [m]; n.d_ += [d]; n.c_ += [None]  # to align position
```
kwcckw, line 505: So we need this to align their m_, d_ and c_, so that we can assign them in sum2C later:

frame_2D_alg/agg_recursion.py
```diff
@@ -566,28 +567,33 @@ def cluster_P(_C_, _c, root):  # FCM-style parallel centroid refine, may add pro
...
-                out_ += [sum2C(N_,m_,d_,i, root=root, final=1)]
+                out_ += [sum2F(N_, root, m_=m_, d_=d_, ci=i, final=1)]
     if out_:
         dCt = sum2F(list(set(_C_)-set(out_)),CF())  # compress
         oF = CoF.get(); oF.N_=C_; oF.rTT=dCt.dTT; oF.r=dCt.r; oF.c=_c-dCt.c
         return out_
 
 def sum2C(N_, m_,d_, i, root=None, final=0, wO=wcC):  # fuzzy sum + base attrs for centroids
 
-    for N in N_: N.w = N.c*N.m_[i]  # combined weight of N in C?
```
kwcckw, line 577: This should be `N.c*m_` instead of `N.c*N.m_`? In cluster_P, n.m_ is from prior loop, which is wrong.

frame_2D_alg/agg_recursion.py
```diff
@@ -488,7 +488,7 @@ def cluster_C(Ft, E_,_r,_c):  # form centroids by clustering exemplar surround v
     for n in N_: n.c_,n.m_,n.d_, n._c_,n._m_,n._d_ = [],[],[], [],[],[]
     _C_ = []
     for i,E in enumerate(E_):
-        C = Copy_(E, Ft,init=1,typ=3)
+        C = Copy_(E, Ft,init=1,typ=3); C.w = C.m * C.c
```
boris-kz, line 491: That's a possible overloading, but default w = np.mean(wTT)

frame_2D_alg/agg_recursion.py
boris-kz: that makes no sense to me, just enumerate m_?

frame_2D_alg/agg_recursion.py
boris-kz: yes, but it's probably wrong anyway, it should be just local w or N.w =  m_[i]

frame_2D_alg/agg_recursion.py
boris-kz: if you really have to align, it's  n.c_ += [_C]

frame_2D_alg/agg_recursion.py
boris-kz: that shared part is really sum_vt, just vectorized?

frame_2D_alg/agg_recursion.py
boris-kz: something like 

def sum2C(N_, m_,d_,i, root=None, final=0, wO=wcC):  # fuzzy sum + base attrs for centroids

    TT= np.zeros((2,9)); kern= np.zeros(4); yx= np.zeros(2); T=R=span= 0
    for N, m in zip(N_,m_):
        w = m/sum(N.m_); TT+= N.dTT*w; kern+= N.kern*w; span+= N.span*w; yx+= N.yx*w; R+= N.r*w; T+=N.c  # not weighted?
    wTT = cent_TT(TT,R)* wO  # set param correlation weights
    m,d = vt_(TT,wTT)
    C = CF(typ=3, N_=N_,dTT=TT,m=m,d=d,c=T,r=R, root=root, wTT=wTT); C.m_=m_; C.d_=d_; C.yx=yx; C.kern=kern; C.span=span
    for n,m,d in zip(N_,m_,d_):
        n.c_[i] = C; n.m_[i] = m; n.d_[i] = d  # mapping-C vals
    if final:
        C.L_ = []
        for N in N_:
            dy_dx = C.yx-N.yx; span = np.hypot(*dy_dx); angl = np.array([dy_dx, np.sign(N.dTT[1] @ ttcN[1])], dtype=object)
            C.L_ += [CN(typ=1,nt=[C,N],dTT=N.dTT,c=N.c,r=N.r,m=N.m,d=N.d,span=span,angl=angl)]  # for proj_C?
    return C

boris-kz: Key issues:

**1. `n.c_[i] = C` indexing is inconsistent between callers**

In `cluster_C`, `n.c_` is append-built (starts empty, grows as N encounters Cs). `i` is position in `_C_`, not position in `n.c_`. If N first joins at C index 3, `n.c_[3] = C` IndexErrors because `n.c_` is still empty. List-index assignment ≠ append.

In `cluster_P`, every N is in every C, so index by `_C_` position is consistent — but your cluster_P call site no longer passes `i`:
```python
out_ += [sum2C(N_,m_,d_, root=root, final=1)]  # missing i arg
```

Two different semantics for `n.c_` in the two clusterers, one function can't serve both cleanly.

**2. `w = m/sum(N.m_)` — zero division + wrong scope**

- First iteration: `N.m_` was reset to `[]` in cluster_C entry → sum=0 → inf.
- During the iteration, `N.m_` has only entries from *earlier* Cs in this pass, not this C or later ones. So `w` depends on iteration order of `_C_`.
- Scope is N-local (normalize N's contribution across all Cs it belongs to), not C-local (FCM-style: normalize N's weight within this C). If you want fuzzy centroid, standard is `w = m / sum(m_)` where `m_` is the membership list for *this* C — use the local `m_` you're building, not `N.m_`.

**3. Weight-sum not 1 → `yx, kern, span` are unscaled**

With N-local w, Σ w over N_ in this C is arbitrary. `yx`, `kern`, `span` accumulate with w but are never divided by Σw, so they're not means. Same for `R`. Only `T += N.c` is unweighted (flagged by your own comment) — that one might actually be intentional (mass is count, not fractional), but then it's inconsistent with the weighted rest.

**4. Minor: `span` inner shadow**

`if final:` overwrites outer `span` (already stored on C). Harmless but easy to misread.

---

Core question: do you want N's contribution to C weighted by *its membership in this C* (C-local, standard FCM) or by *its exclusivity to this C* (N-local)? Current code mixes both — the `w = m/sum(N.m_)` is N-local but everything else (C.m, C.d from `vt_(TT)`, T as raw sum) reads C-local.

boris-kz: def sum2C(N_, m_,d_, root=None, final=0, wO=wcC):  # fuzzy sum + base attrs for centroids

    TT= np.zeros((2,9)); kern= np.zeros(4); yx= np.zeros(2); T=R=span= 0
    for N, m in zip(N_,m_):
        w = m/sum(N.m_); TT+= N.dTT*w; kern+= N.kern*w; span+= N.span*w; yx+= N.yx*w; R+= N.r*w; T+=N.c  # not weighted?
    wTT = cent_TT(TT,R)* wO  # set param correlation weights
    m,d = vt_(TT,wTT)
    C = CF(typ=3, N_=N_,dTT=TT,m=m,d=d,c=T,r=R, root=root, wTT=wTT); C.m_=m_; C.d_=d_; C.yx=yx; C.kern=kern; C.span=span
    for n,m,d in zip(N_,m_,d_):
        n.c_ += [C]; n.m_ = [m]; n.d_ = [d]  # mapping-C vals

frame_2D_alg/agg_recursion.py
kwcckw: That `wTT` is global wTT? But then all of them have the same w?

frame_2D_alg/agg_recursion.py
kwcckw: I see, thanks.

frame_2D_alg/agg_recursion.py
kwcckw: This is for the assignment in sum2C, because now n.m_ is init empty? Or we can just append them now since we only know the number of Cs after we form them?

frame_2D_alg/agg_recursion.py
kwcckw: Okay, yes, thanks. So we not really need the new func too, and we just need to add new flag to sum from m_ instead of c_.

frame_2D_alg/agg_recursion.py
boris-kz: no, G.wTT

frame_2D_alg/agg_recursion.py
boris-kz: I think append. Your version won't work anyway

frame_2D_alg/agg_recursion.py
kwcckw: But most of the time G.wTT is init from global's wTT too?

    n.wTT = kwargs.get('wTT',wTT)
    
We not really update it?

boris-kz: It can be updated by cent_TT

frame_2D_alg/agg_recursion.py
kwcckw: Yes, so we it should be just append since each C should be added only once to each of the member N..

frame_2D_alg/agg_recursion.py
kwcckw: Okay right, that's only updated in sum2C, but it should be fine here for initialization. For non C, their wTT is remained the same?

boris-kz: probably

frame_2D_alg/agg_recursion.py
boris-kz: so does it make sense to vectorize sum_vt and call it from sum2C, sum2F?

kwcckw: They all shared some sections but not all. If we need to vectorize it and call it form sum2C and sum2F, we may increase the complexity of sum_vt since it requires additional if checks for fm and those additional params such as wc, kern, psan, etc. I would prefer pack the vectorized code as a separated function.

frame_2D_alg/agg_recursion.py
kwcckw: > Key issues:
> 
> **1. `n.c_[i] = C` indexing is inconsistent between callers**
> 
> In `cluster_C`, `n.c_` is append-built (starts empty, grows as N encounters Cs). `i` is position in `_C_`, not position in `n.c_`. If N first joins at C index 3, `n.c_[3] = C` IndexErrors because `n.c_` is still empty. List-index assignment ≠ append.
> 
> In `cluster_P`, every N is in every C, so index by `_C_` position is consistent — but your cluster_P call site no longer passes `i`:
> 
> ```python
> out_ += [sum2C(N_,m_,d_, root=root, final=1)]  # missing i arg
> ```
> 
> Two different semantics for `n.c_` in the two clusterers, one function can't serve both cleanly.
> 
> **2. `w = m/sum(N.m_)` — zero division + wrong scope**
> 
>     * First iteration: `N.m_` was reset to `[]` in cluster_C entry → sum=0 → inf.
> 
>     * During the iteration, `N.m_` has only entries from _earlier_ Cs in this pass, not this C or later ones. So `w` depends on iteration order of `_C_`.
> 
>     * Scope is N-local (normalize N's contribution across all Cs it belongs to), not C-local (FCM-style: normalize N's weight within this C). If you want fuzzy centroid, standard is `w = m / sum(m_)` where `m_` is the membership list for _this_ C — use the local `m_` you're building, not `N.m_`.
> 
> 
> **3. Weight-sum not 1 → `yx, kern, span` are unscaled**
> 
> With N-local w, Σ w over N_ in this C is arbitrary. `yx`, `kern`, `span` accumulate with w but are never divided by Σw, so they're not means. Same for `R`. Only `T += N.c` is unweighted (flagged by your own comment) — that one might actually be intentional (mass is count, not fractional), but then it's inconsistent with the weighted rest.
> 
> **4. Minor: `span` inner shadow**
> 
> `if final:` overwrites outer `span` (already stored on C). Harmless but easy to misread.
> 
> Core question: do you want N's contribution to C weighted by _its membership in this C_ (C-local, standard FCM) or by _its exclusivity to this C_ (N-local)? Current code mixes both — the `w = m/sum(N.m_)` is N-local but everything else (C.m, C.d from `vt_(TT)`, T as raw sum) reads C-local.

So those issues shouldn't be relevant now. For 3, `rc_ @ ` should be already scaled the values?

frame_2D_alg/agg_recursion.py
boris-kz: ok, but is that vectorization really equivalent?

frame_2D_alg/agg_recursion.py
boris-kz: yes,

    M = sum(m_) or eps  # C-normalizer
    for N, m in zip(N_,m_):
        w = m/M; TT+= N.dTT*w; kern+= N.kern*w; span+= N.span*w; yx+= N.yx*w; R+= N.r*w; T+=N.c  # not weighted

frame_2D_alg/agg_recursion.py
kwcckw: Yes, it's the same * rc in both processes?

I tried to run both with the same N_, and getting the same results. Also LLM stated the same:

image

frame_2D_alg/agg_recursion.py
kwcckw: Okay, the difference is really just m/M and c/C now.

frame_2D_alg/agg_recursion.py
```diff
@@ -566,28 +567,33 @@ def cluster_P(_C_, _c, root):  # FCM-style parallel centroid refine, may add pro
...
         rc = m/M; TT+= N.dTT*rc; kern+= N.kern*rc; span+= N.span*rc; yx+= N.yx*rc; R+= N.r*rc
     wTT = cent_TT(TT,R) * wO  # set param correlation weights
     m,d = vt_(TT,wTT)
-    C = CN(typ=3, Nt= CF(N_=N_),dTT=TT,m=m,d=d,c=M,r=R, yx=yx, kern=kern,span=span, root=root, wTT=wTT, L_=L_)
-    C.m_=m_; C.d_=d_
+    C = CN(typ=3, Nt= CF(N_=N_),dTT=TT,m=m,d=d,c=M,r=R, yx=yx, kern=kern,span=span, root=root, wTT=wTT)
+    C.m_=m_; C.d_=d_; C.w=W; C.c_=[]
     for n,m,d in zip(C.N_,C.m_,C.d_): n.c_[i] = C; n.m_[i] = m; n.d_[i] = d  # mapping-C vals
+    if final==1:
+        L_ = []  # draft:
```
boris-kz, line 591: So now we need to add these C-to-N links to compute

    V = (m-ave*n) + (d-avd*n)

in proj_focus?

frame_2D_alg/agg_recursion.py
boris-kz: in the new scheme the Cs are CFs, so we can use C.N_ directly, without C.Nt. 
do you see any problem with that?

frame_2D_alg/agg_recursion.py
kwcckw: Do you mean adding projection in Cs now?

So the projection should be similar to proj_focus? The existing proj_focus estimates neighbors with decays computed via distance. For C, it will be within members N_, estimates the membership of members, but it's based on their m_ and probably d_?

frame_2D_alg/agg_recursion.py
kwcckw: It might have some problems when we have subsequent cross_comp of Ct. But why C.Nt.N_ is an issue?

boris-kz: C is a single fork, it shouldn't have Nt, Bt, Ct at this point. That Nt is just confusing

frame_2D_alg/agg_recursion.py
boris-kz: These variables used in proj_focus are currently computed from tile.Nt.Lt, 
we need to add tile.Ct.Lt to that?

frame_2D_alg/agg_recursion.py
kwcckw: Okay, we can try that. At this point i think mostly is just compatibility issue, i need to go through them first to see if there's a bigger issue.

frame_2D_alg/agg_recursion.py
kwcckw: Okay, yes, so we need only add tile.Ct in cluster_N, and for Ct.Lt, it will be added the same from Ct.N_'s L_ (which is C.L_) into tile.Ct.Lt?

    if C_: sum2F(C_, Ft.root.Ct,froot=0)  # Ct.r includes overlap?

boris-kz: right

kwcckw: Alright, I will check again much later

PR #125, opened 2026-04-25
4/25/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -286,7 +286,7 @@ def comp_C_(C_,_r, _C_=[], fall=1, fC=0):  # simplified for centroids, trans-N_s
             tc += c; tr += r*c
             comp_N(_C,C,_r, c ,A=dy_dx,span=dist,L_=L_,N_=N_,acc=acc)  # simplified for typ=3
     for N in list(set(N_)):
-        if N.rim: sum2F(N.rim, N.Rt)
+        if N.rim: N.Rt = CF(); sum2F(N.rim, N.Rt)  # not sure
```
kwcckw, line 289: So when we convert C to CF, most of the issues are from Rt, so i guess we can init them here? Or we can just remain their rim as list instead of creating Rt?

frame_2D_alg/agg_recursion.py
```diff
@@ -591,11 +594,17 @@ def sum2C(N_, m_,d_, root=None, final=0, wO=wcC):  # fuzzy sum + base attrs for
...
+    box = np.array([np.inf, np.inf, -np.inf, -np.inf])
     for N, m in zip(N_,m_):
         w = m/M; TT+= N.dTT*w; kern+= N.kern*w; span+= N.span*w; yx+= N.yx*w; R+= N.r*w; T+=N.c  # not weighted
+        box = extend_box(box, N.box)
     wTT = cent_TT(TT,R)* wO  # set param correlation weights
     m,d = vt_(TT,wTT)
-    C = CF(typ=3, N_=N_,dTT=TT,m=m,d=d,c=T,r=R, root=root, wTT=wTT); C.m_=m_; C.d_=d_; C.yx=yx; C.kern=kern; C.span=span
+    C = CF(typ=3, N_=N_,dTT=TT,m=m,d=d,c=T,r=R, root=root, wTT=wTT); 
+    C.m_=m_; C.d_=d_; C.c_ = []; C.yx=yx; C.kern=kern; C.span=span
+    C.angl = np.array([np.zeros(2), 1],dtype=object)
```
kwcckw, line 605: For C.angl, we can use the same way as in add_Lt?

    A = np.sum([N.angl[0] for N in N__], axis=0) 
    C.angl = np.array([A, np.sign(C.dTT[1] @ ttcN[1])], dtype=object)

frame_2D_alg/agg_recursion.py
```diff
@@ -911,7 +925,7 @@ def proj_focus(PV__, y,x, tile):  # radial accum of projected focus value in PV_
     m,d,n = tile.m, tile.d, tile.c  # add r?
     V = (m-ave*n) + (d-avd*n)
     dy,dx = tile.angl[0]; a = dy / (dx or eps)  # average link_ orientation, projection
-    # include C/N links?
+    # include C/N links? (include Nt and Ct's Lt to compute decays?)
```
kwcckw, line 928: So when we have Nt and Ct's Lt now, they will be used in computing decay? That will be based on the ratio of their distance and total span across all Links' Nt?

frame_2D_alg/agg_recursion.py
```diff
@@ -739,7 +748,7 @@ def add_Nt(G, Nt, merge=0):  # addition to Q2R
         if hasattr(N,'m_'):
             if not hasattr(G, 'c_'): G.c_, G.m_, G.d_ = [], [], []
             G.r += N.r; G.c_ += N.c_; G.m_ += N.m_; G.d_ += N.d_
-        G.C_ += N.rN_  # Ct||Nt;  A,a = G.angl[0],N.angl[0]; A[:]= (A*C+a*c)/C # vect only, if in Nt?
+        # G.C_ += N.rN_  # Ct||Nt;  A,a = G.angl[0],N.angl[0]; A[:]= (A*C+a*c)/C # vect only, if in Nt?  (rN not relevant now?)
```
boris-kz, line 751: It's N.c_ now, add it to CN?

frame_2D_alg/agg_recursion.py
```diff
@@ -270,7 +270,7 @@ def comp_C_(C_,_r, _C_=[], fall=1, fC=0):  # simplified for centroids, trans-N_s
                 if val_(L.dTT,_r+cC_, ttC_,fi=0) < 0:
                     C0,C1 = L.nt
                     if C0 is C1 or C1 in merg_ or C0 in merg_: continue  # not merged
-                    add2F(C0,C1,merge=2); add_Nt(C0, C1.Nt)
+                    add2F(C0,C1,merge=2)  # add_Nt(C0, C1.Nt)  # without Nt, no need to add_Nt?
```
boris-kz, line 273: right

frame_2D_alg/agg_recursion.py
boris-kz: how about adding it in sum2F?

frame_2D_alg/agg_recursion.py
```diff
@@ -304,7 +304,7 @@ def comp_H(_Nt,Nt, Link):  # tentative pre-comp
 
     TT= base_comp(_N,N)[0] if full else comp_derT(_N.dTT[1],N.dTT[1]); m,d = vt_(TT, ttN_)
     L = CN(typ=1, nt=[_N,N], dTT=TT,m=m,d=d,c=_c,r=r, root=rL, exe=1)
-    if N.typ > 0 and m * wN_ > ave*cN_:
+    if N.typ > 0 and N.typ != 3 and m * wN_ > ave*cN_:
```
boris-kz, line 307: We only have comp_N for Cs if they call cross_comp, so we need to add a set of CN attrs to them there?
We had this scheme before, do you remember why it was removed?

frame_2D_alg/agg_recursion.py
boris-kz: I think decay can still be computed from lateral links only, but then also applied to these vertical links

frame_2D_alg/agg_recursion.py
boris-kz: right

frame_2D_alg/agg_recursion.py
```diff
@@ -436,9 +436,9 @@ def trans_cluster(G):
             L.Nt,L.Bt,L.Ct = CF(),CF(),CF()
             # merge roots
     G_= []  # add prelink pL_,pN_? include merged Cs, in feature space for Cs
-    if _N_ and val_(Ft.dTT, r+ccN, Ft.root.wTT*ttcN, mw=(len(_N_)-1)*Lw *wcN) > 0:  #| fL?
+    if _N_ and val_(Ft.dTT, r+ccN, wTT*ttcN, mw=(len(_N_)-1)*Lw *wcN) > 0:  #| fL?
```
boris-kz, line 439: these separate Fw_[i] and Fc_[i] * same data_scope are pretty awkward.
I think we can replace them with ave * (Fw/Fc)?

boris-kz: Claude:
The current pattern `m*wF > ave*cF` is just `m > ave*(cF/wF)`, so define one per-fork ave:

```python
aF_ = ave * np.array(Fc_) / np.array(Fw_)  # default = ave when Fw_=Fc_
```

Then `m*wF > ave*cF` → `m > aF`, and feedback updates Fw_ instead of touching two parallel arrays.

**One thing to decide first** — current `val_` usage of cF is inconsistent:

- `val_(Ft.dTT, r+ccN, ...)` — additive (cF as extra rdn)
- `val_(TT, r*cX, ttX, ...)` — multiplicative (in `cross_comp`)
- `val_(L.dTT,_r+cC_, ttC_,fi=0)` — additive
- `val_(TT, r+ccC, ...)` — additive

If aF replaces ave inside val_, the rdn term naturally becomes `aF*r` and that ambiguity disappears — `cF` is no longer added to `r`, it's folded into the multiplier:

```python
val_(TT, r, ..., aF=aF)  # m - aF*r, no separate cF term
```

That seems strictly cleaner, but it changes one thing: currently `r+cF` makes call overhead a *constant* contribution independent of how deep r already is. With `aF*r`, fork cost scales with recursion depth. Probably fine — deeper recursion *should* face proportionally higher per-fork bar — but worth confirming you want that.

**Question:** in `m*wF > ave*cF` checks (the threshold gates, no `r`), do you want pure `m > aF`, or `m > aF*(1+r)` so call cost still applies at depth 0 with rdn=1?

boris-kz: **Header (lines 62–66):**
```diff
-Fc_ = [9,8,13,3,3,15,18,4,3,9,9,8,2,2]  # AST complexity / vt_, maps to onF_, *= data_size: not implemented yet
-Fw_ = copy(Fc_)  # ave gain summed from calls, init neutral to Fc_
-wN_,wC_,wN,wF, wE,wcN,wcC,wcP, wX,wFrm,wVct,wTrc,wBac,wPrj = Fw_
-cN_,cC_,cN,cF, cE,ccN,ccC,ccP, cX,cFrm,cVct,cTrc,cBac,cPrj = Fc_
+Fc_ = np.array([9,8,13,3,3,15,18,4,3,9,9,8,2,2])  # AST complexity / vt_, maps to onF_
+Fw_ = Fc_.copy()  # ave gain from calls, init = Fc_, updated by ffeedback
+aF_ = Fc_ / Fw_  # per-fork rdn = cost/gain, slots into ave*(r+aF)
+aN_,aC_,aN,aF, aE,acN,acC,acP, aX,aFrm,aVct,aTrc,aBac,aPrj = aF_
```

**Call sites — the clean ones:**
```diff
-        if M * ((len(L_)-1)*Lw) * wE > ave * cE:                         # 195
+        if M * ((len(L_)-1)*Lw) > ave * aE:
-                if val_(TT, r*cX, ttX, (len(G_)-1)*Lw,1,TTd, rd/(r+rd)) > 0:   # 200
+                if val_(TT, r+aX, ttX, (len(G_)-1)*Lw,1,TTd, rd/(r+rd)) > 0:
-        if abs(vt_(eTT,ttN_)[0]) > ave*(cN+r+(_N.r+N.r)/2):               # 226
+        if abs(vt_(eTT,ttN_)[0]) > ave*(aN+r+(_N.r+N.r)/2):
-            if abs(m)*wN < ave*cN:                                        # 239
+            if abs(m) < ave*aN:
-                if val_(L.dTT,_r+cC_, ttC_,fi=0) < 0:                     # 271
+                if val_(L.dTT,_r+aC_, ttC_,fi=0) < 0:
-    if N.typ > 0 and m * wN_ > ave*cN_:                                   # 308
+    if N.typ > 0 and m > ave*aN_:
-        if m + vt_(htt,ttN_)[0] * wN_ > ave*cN_:                          # 310
+        if (m + vt_(htt,ttN_)[0]) > ave*aN_:
-            if (rL.m+m)/2 * ((len(Np_)-1)*Lw) * wF > ave * cF:            # 345
+            if (rL.m+m)/2 * ((len(Np_)-1)*Lw) > ave * aF:
-        if N.Rt.m * N.c * wE > ave* (_r+rdn+cE+oV):                       # 413
+        if N.Rt.m * N.c > ave* (_r+rdn+aE+oV):
-                    if tL.m*wcN> ave*ccN:                                 # 434
+                    if tL.m > ave*acN:
-    if _N_ and val_(Ft.dTT, r+ccN, Ft.root.wTT*ttcN, mw=(len(_N_)-1)*Lw *wcN) > 0:   # 440
+    if _N_ and val_(Ft.dTT, r+acN, Ft.root.wTT*ttcN, mw=(len(_N_)-1)*Lw) > 0:
-        C_, cnt,mat,dif,rdn,DTT,Up = [],0,0,0,0,np.zeros((2,9)),0; Ave = ave*(_r+ccC); Avd = avd*(_r+ccC)   # 498
+        C_, cnt,mat,dif,rdn,DTT,Up = [],0,0,0,0,np.zeros((2,9)),0; Ave = ave*(_r+acC); Avd = avd*(_r+acC)
-        if lm*ld* ((len(L_)-1)*Lw) > ave*avd* (lr+1+ccN):                 # 699
+        if lm*ld* ((len(L_)-1)*Lw) > ave*avd* (lr+1+acN):
-            V = lm - ave*(lr+1+ccN); c = G.Lt.c                           # 700
+            V = lm - ave*(lr+1+acN); c = G.Lt.c
-                G_,r= cluster_C(G.Nt,E_,r,c) if (mdecay(L_)-decay)*V > ave*ccC else cluster_N(G.Nt, E_,r,c)   # 703
+                G_,r= cluster_C(G.Nt,E_,r,c) if (mdecay(L_)-decay)*V > ave*acC else cluster_N(G.Nt, E_,r,c)
-                if G_ and val_(G.Nt.dTT, r+cX, G.wTT*ttX, mw=(len(G_)-1)*Lw) > 0:   # 704
+                if G_ and val_(G.Nt.dTT, r+aX, G.wTT*ttX, mw=(len(G_)-1)*Lw) > 0:
-        if bd > avd*br*cX and N.typ!=1: cross_comp(Bt, br,'Bt')           # 708
+        if bd > avd*br*aX and N.typ!=1: cross_comp(Bt, br,'Bt')
-    if Lt.m * wN > ave*(cN+Lt.r):                                         # 751
+    if Lt.m > ave*(aN+Lt.r):
-            if T and val_(T.dTT,T.r+cFrm+elev, T.wTT*ttFrm, mw=(len(T.N_)-1)*Lw) > 0:   # 996
+            if T and val_(T.dTT,T.r+aFrm+elev, T.wTT*ttFrm, mw=(len(T.N_)-1)*Lw) > 0:
```

**Three sites I left alone — flagging:**

1. **Line 508** `M*(_C.w + wC_*len(n_) + wC_*len(n_)) > Ave*(r+_C.r + cC_*len(n_))` — `wC_` and `cC_` both multiply `len(n_)`, but `Ave` already absorbed `_r+acC`, so `Ave/wC_` doesn't simplify. Cleanest fix: drop wC_ on value side, swap `cC_→aC_`:
   ```
   M*(_C.w + 2*len(n_)) > Ave*(r+_C.r + aC_*len(n_))
   ```
   But this changes the relative weight of `_C.w` vs the count term. Want me to do it?

2. **Lines 518, 522, 559** — same pattern (`wcP*L` vs `ccP*L`, `wcC*len(C_)` vs `ccC*len(C_)`). Same call: drop the `w*` from value side, replace `c*` with `a*` on cost. OK to apply uniformly?

3. **Line 591** `wO=wcC` in `sum2C` — different use entirely, `wO` scales `cent_TT` output, not a threshold. With gain folded in, the analog is `wO=1/acC`, but probably cleaner to drop `wO` and let global feedback handle scaling. Your call.

4. **Line 753** `sum_vt(pL_,fm=1,wTT=wTT)[0]* ave*wN > ave*(cN * np.mean([L.r for L in pL_]))` — has a stray `* ave` on the value side that looks buggy independent of this refactor. Leave as `[0] > ave*(aN * mean(L.r))`?

**Wiring note (out of scope but related):** `_W_,_C_` accumulators in `CoF.traced` are never read back into `Fw_`. After this refactor, `aF_` should be recomputed at the end of `ffeedback` once you wire that update — otherwise the consolidation is cosmetic.

frame_2D_alg/agg_recursion.py
kwcckw: We init C with Copy_ and sum2C, so sum2C should be merged to sum2F now?

Those params should be added when we init C (typ == 3), so we can just add N.Rt = CF(), but we probably need to add N.rim = N.Rt.N_.

        C.yx = np.zeros(2); C.kern = np.zeros(4)
        C.angl = np.array([np.zeros(2), 1],dtype=object)
        C.rim = []; C.compared = set(); C.fin = 0; C.box = copy(N.box)
        C.sub = N.sub; C.Rt = CF(); C.rim = C.Rt.N_

frame_2D_alg/agg_recursion.py
kwcckw: That's already added in line 750 above? 

    G.r += N.r; G.c_ += N.c_; G.m_ += N.m_; G.d_ += N.d_

frame_2D_alg/agg_recursion.py
kwcckw: Sorry, i can't remember that now. But this sub+ should be still relevant to Cs? Those Cn attrs should be added when we init C .

frame_2D_alg/agg_recursion.py
kwcckw: > I think we can replace them with ave * (Fw/Fc)?

So in short, we only need the ratio instead of two separated Fw and Fc?

frame_2D_alg/agg_recursion.py
boris-kz: That only when Cs are cross-compared, not the base_comp in cluster_C

frame_2D_alg/agg_recursion.py
boris-kz: then it should be G.C_ += N.c_

frame_2D_alg/agg_recursion.py
boris-kz: I think so

frame_2D_alg/agg_recursion.py
kwcckw: Okay, so they are actually applied implicitly here? But what is really mean by this include C/N links?

frame_2D_alg/agg_recursion.py
kwcckw: Okay right thanks, so this should be added to G.Ct.N_ (G.C_).

frame_2D_alg/agg_recursion.py
kwcckw: Yes, those compatibility issues are not really from the base_comp in cluster. So i guess we need to add those params to Ct.N_ right before the cross_comp.

frame_2D_alg/agg_recursion.py
boris-kz: Not really:
You're right, I conflated them. mw (data_size) is a value-side gain term, parallel to Fw. Fc is a cost-side rdn term, parallel to r. They live on opposite sides of the inequality:

```
m * mw * Fw  >  ave * (r + Fc * data_size)
   └─gain─┘         └────cost────┘
```

Reducing to `aF = Fc/Fw` only works if Fw and Fc *always* appear paired in the same ratio. The moment Fw needs to combine with other gain terms (mw, Lw, intw...) or Fc with other cost terms (r, olp, _C.r...), they have to stay separate — they're not commensurable, just like you can't fold mw into r.

So the original parallel-arrays structure is conceptually right; the awkwardness was just the naming/duplication, not the duality. Two cleaner reorganizations that preserve the duality:

**Option A: per-fork pair object**
```python
class Fwc: w, c  # one per fork
F_ = [Fwc(w=9,c=9), ...]
```
Access as `F_[nF].w`, `F_[nF].c`. Eliminates the parallel-array bookkeeping.

**Option B: keep arrays, drop the unpacked aliases**
Just use `Fw_[onF_.index('comp_N')]` directly — verbose but no shadow names. Probably worse.

**The deeper question your point raises:** does `mw` belong in `val_` as a value-side multiplier *with* an Fw companion? Currently val_ takes mw but no Fw — it's hardcoded `rv*mw`. If Fw is genuinely a per-fork gain multiplier on value, val_ should be:
```
rv * mw * Fw - ave * (r + Fc * mw_or_data)
```

Otherwise Fw lives only in the threshold sites outside val_, which is asymmetric with Fc (which gets added into r everywhere).

Want to think about whether Fw should propagate into val_ on the value side?

frame_2D_alg/agg_recursion.py
boris-kz, line 187: I think we need to unpack val_:

```
def cross_comp(Ft, rr):  # core function mediating recursive rng+ and der+ cross-comp and clustering

    N_,G_ = Ft.N_,[]; fC = N_[0].typ==3; root=Ft.root  # rc=rdn+olp, comp N_|B_|C_:
    # comp_ s update root oF:
    if L_: # Lm_, no +|- Ft.Lt?
        L= len(L_)-1  # val=m+d /clust, m/comp
            if G_:
                root.Nt = sum2F(G_,CF(root=root),froot=1); L=len(G_)-1  # or G.N_ = spliced C_ if higher value?
                if vt_(TT,ttX)[0]* (wX*L) > ave* (r+cX*L):  # val_ without borrw?

```

frame_2D_alg/agg_recursion.py
boris-kz: something like that, pack it in F2N? and then N2G to add other forks in cross_comp?

frame_2D_alg/agg_recursion.py
boris-kz: **1. The template is now clear:**
```
value * wF * mw  >  cost * (r + cF * mw)
```
where `value` is `m`, `d`, `m+d`, etc. and `cost` is `ave`, `avd`, or `ave+avd` matching which value terms participate. wF and cF stay on opposite sides, both scaled by data_size mw.

**2. `Lw` is absorbed into wE/wX.** Original had `(len(L_)-1)*Lw * wE`, now `wE*L`. Same role, one weight instead of two. Cleaner — Lw was a redundant global constant.

**3. The m vs m+d split is meaningful and should be preserved:**
- First eval (clustering gate): `sum(vt_)` with `ave+avd` — any predictable signal (match or contrast) justifies clustering, since contrast forms boundaries
- Second eval (recursion gate): `vt_[0]` with `ave` — only positive match justifies going deeper, contrast doesn't recurse

The original `val_(..., cr=rd/(r+rd))` borrow mechanism was conflating these — borrowing d into m. The unpacked form drops it cleanly because m+d is now explicit when you want both, otherwise just m.

**One question this raises:** if every eval follows this template, val_ as a function is just obscuring it. The non-trivial part of val_ was `rv = m_/t_ @ wTT[0]` (normalization by total magnitude) — that's what `vt_` already does. So val_ adds nothing; it can disappear, and call sites use `vt_` + the template directly. Worth doing across the file?

**One asymmetry to flag:** in your sketch the second eval uses `r+cX*L` where L is `len(G_)-1` (post-clustering count), while r came from the comp/cluster cycle that produced those G_. That's correct — cost of the *next* recursion step scales with what's being fed in. Just noting that L switches meaning between the two evals (L_ vs G_), which is the right behavior but easy to miss.

boris-kz: Yes, and the simplification exposes that two distinct mechanisms were collapsed into one mixing parameter:

**Root borrow is subtractive (selectivity):** clustering value = `m - root.m` (or `root.Lt.m` for link-mediated). The cluster only justifies itself by exceeding what the parent already represents. This is loss reduction, not lending.

**Boundary borrow is additive (complementation):** core+contour from the README — high-contrast boundary stabilizes the core cluster, so `m + boundary.m * decay` (decay because boundary support attenuates with distance/composition gap).

Current `val_` mixes both as `rv*(1-cr) + _rv*cr`, which can't represent opposite signs. Recomputing `_rv` from `_TT` adds nothing — root and boundary already cache m from when they were formed under the same wTT.

Cleaner form:
```
selective_m = m - root.m          # selectivity
complemented_m = m + B.m * decay  # core+contour
```
Then the eval template is just:
```
selected_or_complemented_m * wF * mw > cost * (r + cF * mw)
```

**One thing to decide:** should the two combine? A cluster could be both selective vs root *and* complemented by boundary:
```
(m - root.m) + B.m * decay
```
This is coherent — both modify "what this cluster contributes beyond what's already represented", just from opposite directions. Selectivity removes redundancy with parent; complementation adds support from contrast that wouldn't be in the cluster itself.

Or are these meant to apply in different phases (selectivity in cross_comp recursion gate, complementation in cluster_N gate)? That would keep them separate and match the m vs m+d split in your sketch.

kwcckw: > I think we need to unpack val_:

Okay, so unpacking val_ should be clearer to show the general formula:

    value * weight > base cost * (r + function cost)
    
> One thing to decide: should the two combine?

So the current borrow in val_ can't differentiate whether it is subtractive or additive, which i think is better to separate the eval logic as suggested above:

      selective_m = m - root.m          # selectivity
      complemented_m = m + B.m * decay  # core+contour

Logically i think both also can exist at the same time since they are two different thing from root and core+contour, but we may need only either one of them during the eval, and mostly not both at the same time, since it's rare to eval both root +core+contour.

frame_2D_alg/agg_recursion.py
kwcckw: So using ratio is wrong, because we need to apply weight and cost differently to different val.

frame_2D_alg/agg_recursion.py
boris-kz: yes, that root, contour borrow should be specifc add-ons, and we don't need dTT for that

frame_2D_alg/agg_recursion.py
```diff
@@ -739,7 +748,7 @@ def add_Nt(G, Nt, merge=0):  # addition to Q2R
         if hasattr(N,'m_'):
             if not hasattr(G, 'c_'): G.c_, G.m_, G.d_ = [], [], []
             G.r += N.r; G.c_ += N.c_; G.m_ += N.m_; G.d_ += N.d_
```
boris-kz, line 750: G.r += N.r is probably redundant

frame_2D_alg/agg_recursion.py
boris-kz: in proj_N:

    for L in N.L_+ N.B_: proj_TT(L, cos_d, dist, L.r+r, iTT, wTT, dec=dec)  # accum TT internally
    for L in N.Ct.Lt.N_: proj_TT(L, cos_d, dist, L.r + r, iTT, wTT, dec=dec)  # C-to-N links?
    for L in N.rim:  proj_TT(L, cos_d, dist, L.r+r, eTT, wTT, dec=dec)

frame_2D_alg/agg_recursion.py
kwcckw: Okay, so we can remove val_ now, and all eval should be follow this now? 

    value * weight > base cost * (r + function cost)

frame_2D_alg/agg_recursion.py
kwcckw: This G.r accumulation should be done separately ?

frame_2D_alg/agg_recursion.py
boris-kz: yes, as I did in cross_comp, please do that in other functions

frame_2D_alg/agg_recursion.py
boris-kz: So please add those

frame_2D_alg/agg_recursion.py
kwcckw: Alright , I will check and update later.

frame_2D_alg/agg_recursion.py
kwcckw: I see, thanks for claryfing again.

frame_2D_alg/agg_recursion.py
kwcckw: Sorry, I overlooked this, I will check abd add later.

frame_2D_alg/agg_recursion.py
boris-kz: It's already done in sum2G?

kwcckw: Yes, we already have that in sum2G:

    G.typ=typ; r=G.r

So it's should be done separately in sum2G instead of add_Nt.

PR #127, opened 2026-04-27
4/27/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -148,6 +148,7 @@ def inner(*a, **kw):
             oF = CoF(nF=onF_.index(func.__name__), root=_CoF); oF.wTT = FTT_[oF.nF]; oF.wc = Fc_[oF.nF]; _CoF.call_+=[oF]
             CoF._cur.set(oF); out = func(*a, **kw)
             if oF.call_:  # complete at this point
+                # right now we only use Lw here, change it to woF?
                 if (len(flat_(oF))-1)*Lw > ave*(_CoF.wc+oF.r):
```
kwcckw, line 152: So right now we only need Lw here, we can rename it as woF or LoF now?

frame_2D_alg/agg_recursion.py
```diff
@@ -548,7 +548,7 @@ def cluster_P(_C_, _c, root):  # FCM-style parallel centroid refine, may add pro
         md__ = np.zeros_like(_md__)
         for j,N in enumerate(N_):
             for i,C in enumerate(_C_):
-                TT,_ = base_comp(C,N); md__[j,i] = vt_(TT, ttcP)  # use rc?
+                TT = comp_derT(C.dTT[1],N.dTT[1]); md__[j,i] = vt_(TT, ttcP)  # use rc? (kern is empty before final, use comp_derT?)
```
kwcckw, line 551: So kern is empty before final assignment, i think we can just comp_derT here to get TT?

boris-kz: Kern is not positional, we still need it

kwcckw: Okay, so we need to include kern exclusively in sum2C now? Something like this:

              if i: TT+=n.dTT*w; R+=n.r*w; n_ += n.N_ if merge>1 else [n] if merge else []; (kern:=kern+n.kern*w) if np.any(m_) else None
              else: TT = n.dTT*w; R=n.r*w; n_=[]; (kern:=n.kern*w) if np.any(m_) else None
          if not merge: n_ = N_
          F = CF(N_=n_, dTT=TT, c=C, r=R);
          if np.any(m_): F.typ=3; F.m_,F.d_=m_,d_; F.m,F.d=sum(m_),sum(d_); F.kern = kern
          else:  F.m, F.d = vt_(TT)
          root = add2F(root,F) if root!=None else F
          if froot:
              for n in N_: n.root=root
          sum_H(N_, root)
          if getattr(root,'C_',None):
          return root

frame_2D_alg/agg_recursion.py
boris-kz: that woF is typ-specific? By analogy with wc, it should be ww?
also these are CoF-only, we don't need them in CF?
they should be called Fw and Fc?

boris-kz: On naming: yes, `Fc`/`Fw` reads cleanly — matches the `Fc_`/`Fw_` globals they're sourced from. Currently `f.wc` shadows the global `Fw_` semantics (gain) vs `Fc_` (cost) confusingly. Rename:

- `f.wc` → `f.Fc` (per-instance AST cost, from `Fc_[nF]`)
- `f.w`  → `f.Fw` (per-instance gain, accumulated)

Both live on `CoF` only, removed from `CF`.

On `Fr`: the question is what "alternative oFs on the same data" means structurally. Two cases I can see:

**1. Sequential alternatives within one cycle** — e.g. `cluster_N` vs `cluster_C` vs `cluster_P` are gated by different conditions on the same `Ft`, so at most one runs per call. No redundancy because no overlap in execution. `Fr` doesn't apply.

**2. Alternative paths producing overlapping outputs** — e.g. `cross_comp` recurses, and within it `comp_N_` may produce Ls that feed `cluster_N`, whose Gs feed another `cross_comp`. These aren't alternatives, they're a pipeline. Still no redundancy.

The only place I see actual fork-level redundancy is when the same data is processed by structurally similar oFs whose outputs are then both used downstream — which doesn't really happen in current code. `cluster_N` and `cluster_C` are gated as either/or in `sum2G` via `mdecay`.

So `Fr` only makes sense if you plan to run alternative clustering oFs in parallel and combine their outputs (analogous to data `r` for overlapping clusters). Right now that's not the architecture. I'd skip it until there's a real use.

One genuine question: should `Fc` accumulate as sum (total cost down the call tree) or mean (typical cost per fork)? The eval in `traced`:
```
if (len(flat_(oF))-1)*Lw > ave*(_CoF.wc + oF.r):
```
treats `_CoF.wc` as parent's accumulated cost — wants sum. But `add2F` blends by mean. Which is intended?

boris-kz: class CoF(CF):  # oF/ code fork, N_,dTT: data scope, w = vt_(wTT)[0]?

    name = "func"
    _W_,_C_ = np.zeros(len(onF_)),np.zeros(len(onF_))  # sum called oF weights and costs, global Fw_ is not summed
    def __init__(f, **kw):
        super().__init__(**kw)
        f.typ_ = kw.get('typ_', [])  # flattened and nested with tFs
        f.fw,f.fc,f.fr = [kw.get(x,0) for x in ('fw','fc','fr')]  # fr if multiple oFs on the same data?
    @staticmethod
    @staticmethod
    def traced(func):
        @wraps(func)
        def inner(*a, **kw):
            _CoF = CoF._cur.get()
            oF = CoF(nF=onF_.index(func.__name__),root=_CoF); oF.wTT = FTT_[oF.nF]; oF.fw = Fw_[oF.nF]; oF.fc = Fc_[oF.nF]; _CoF.call_+=[oF]
                cT = flat_(oF); L=(len(cT)-1)  # flatten call tree
                if oF.fw*L > ave*(oF.fc*L): sum2F(cT,oF)  # eval summarize, fc*fr if multiple oFs on the same data?
            oF.wTT= cent_TT(getattr(oF,'rTT',oF.dTT), oF.r)  # rTT covers cluster compression
            oF.w += np.mean(wTT)  # / fc*fr?
            _CoF._W_[oF.nF] += Fw_[oF.nF]  # nF is index in onF_
            _CoF._C_[oF.nF] += Fc_[oF.nF]
            CoF._cur.set(_CoF)
            return out
        inner.wrapped = True
        return inner

frame_2D_alg/agg_recursion.py
```diff
@@ -222,7 +223,7 @@ def proj_V(_N,N, dist, dy_dx, dec):  # _N x N induction
         Dec = dec or decay ** ((dist/((_N.span+N.span)/2)))
         iTT = (_N.dTT + N.dTT) * Dec
         eTT = (_N.Rt.dTT + N.Rt.dTT) * Dec
-        if abs(vt_(eTT,ttN_)[0]) > ave*(cN+r+(_N.r+N.r)/2):  # spec N links
+        if abs(vt_(eTT,ttN_)[0]) * wN > ave*(cN+r+(_N.r+N.r)/2):  # spec N links
```
boris-kz, line 226: wPrj, cPrj

frame_2D_alg/agg_recursion.py
```diff
@@ -340,7 +341,7 @@ def comp_F(_F, F, ir=0, rL=None):
         if  _N_ and N_:
             if f: Np_ = [[_n,n] for _n,n in zip(_N_,N_) if _n and n]  # same forks
             else: Np_ = list(product(_N_,N_))  # pairs
-            if (rL.m+m)/2 * ((len(Np_)-1)*Lw) * wF > ave * cF:  # * cF +r?
+            if (rL.m+m)/2 * (len(Np_)-1) * wF > ave * (r + cF*(len(Np_)-1)):  # * cF +r?
```
boris-kz, line 344: L = len(Np_)-1
            if np.mean(rL.m,m)* (wF*L) > ave* (r+cF*L):

frame_2D_alg/agg_recursion.py
```diff
@@ -468,9 +469,8 @@ def trans_cluster(G):
         if G_:
             for G in G_: trans_cluster(G)  # splice trans_links, merge L.nt.roots
             C = sum([g[1] for g in Gt_]); TT=np.zeros((2,9)); R=0; r+=1  # + wC?
-            for tt,c,r in Gt_: rc=c/C; TT+=tt*rc; R+=r*rc
-            # r is no longer relevant? So only R is used
-            if sum(vt_(TT, Ft.root.wTT*ttcN))*wcN*(len(G_)-1) > (ave+avd)*(R+ccN*(len(G_)-1)):  # reform root,Nt, no other forks yet:
+            for tt,c,gr in Gt_: rc=c/C; TT+=tt*rc; R+=gr*rc
```
boris-kz, line 472: L = len(G_)-1
            if sum(vt_(TT, Ft.root.wTT*ttcN))*(wcN*L) > (ave+avd)*(r+R+ccN*L):  # reform root,Nt, no other forks yet:

frame_2D_alg/agg_recursion.py
```diff
@@ -504,7 +504,7 @@ def cluster_C(Ft, E_,_r,_c):  # form centroids by clustering exemplar surround v
                 if _C in n._c_: k=n._c_.index(_C); up += abs(n._m_[k]-m) + abs(n._d_[k]-d)
                 else:           up += m+abs(d)
             r = _r+ R/T  # loop-local, not ave?
-            if M*(_C.w+ wC_*len(n_)+wC_*len(n_)) > Ave*(r+_C.r+ cC_*len(n_)):  # else: Up+= sum(_C._m_)+ sum([abs(d) for d in _C._d_])?
+            if M*(_C.m+ wC_*len(n_)+wC_*len(n_)) > Ave*(r+_C.r+ cC_*len(n_)):  # else: Up+= sum(_C._m_)+ sum([abs(d) for d in _C._d_])?
```
boris-kz, line 507: thanks

frame_2D_alg/agg_recursion.py
boris-kz: yes

frame_2D_alg/agg_recursion.py
```diff
@@ -830,7 +838,7 @@ def PP2N(PP):
                         PPd_ = [B.root for B in N.B_]; sum2F(PPd_,N.Bt)
                         N.Bt.N_ = PPd_; [setattr(B,'root',N.Bt) for B in PPd_]
                 L = len(PPm_)-1
-                if sum(vt_(sum_vt(N_,wTT=ttVct))) * (wVct*L) > (ave+avd) * (3+cVct*L):
+                if sum(vt_(sum_vt(N_,wTT=ttVct)[0])) * (wVct*L) > (ave+avd) * (3+cVct*L):
```
boris-kz, line 841: thanks

frame_2D_alg/agg_recursion.py
kwcckw: > that woF is typ-specific? By analogy with wc, it should be ww?

Yes, it will be more precise to have each fw and fc for each typ. I guess we will have 2 sets of fw and fc for code and function now. 

Also we have `_W_` and `_C_`, accumulated from each typ's call, so we are not using them now?

frame_2D_alg/agg_recursion.py
kwcckw: Okay right thanks, i always mistaken it by looking at the caller function name instead of the calling function name.

frame_2D_alg/agg_recursion.py
boris-kz: code and function? do you mean local and global?

we are using
FTT_ = lev.wTT_ = np.array([t.wTT if not isinstance(t,list) else wTT for t in Z.typ_])
instead of` _W_ and _C_, that's missing _C_`?

boris-kz: base CF should not have wc and wr?

kwcckw: > code and function? do you mean local and global?

Sorry i was confused earlier, that shouldn't be relevant. 

> we are using
FTT_ = lev.wTT_ = np.array([t.wTT if not isinstance(t,list) else wTT for t in Z.typ_])
instead of _W_ and _C_, that's missing _C_?

Yes, so cost should be updated too? But we are not really using `_W_` now? `_W_` should be used to update Fw_ and `_C_` should be used to update Fc_?

> base CF should not have wc and wr?

I think yes, because we only need them in CoF now.

boris-kz: >But we are not really using _W_ now? _W_ should be used to update Fw_ and _C_ should be used to update Fc_?

We don't need them now:

Yes — once `fw`/`fc` propagate through `sum2F`/`add2F`, `add_typ_(Z)` produces exactly the per-`nF` aggregates that `_W_`/`_C_` were trying to maintain, and with realized values (`F.fw` accumulated from subtrees) instead of static `Fw_[nF]`/`Fc_[nF]` constants.

So `Z.typ_[i].fw` ≡ what `_W_[i]` should have been, and superior because it's data-dependent.

Difference is timing: `_W_/_C_` update eagerly per `traced` call; `Z.typ_` only after `add_typ_(Z)` runs (currently once per `frame_H` elev cycle). If you don't need real-time mid-call totals, drop `_W_`/`_C_` entirely.

The eager updates also have the class-attribute-persistence bug (state leaks across `frame_H` runs). Removing them fixes that too.

kwcckw: I see, so `_W_` and`_C_` are redundant to fw and fc, and fw and fc is summed from deeper callees.

For CoF fw and fc, right now it's init from globals:

    oF.fw = Fw_[oF.nF]; oF.fc = Fc_[oF.nF]

Do we need to add additional data's fw and fc into each oF?

boris-kz: that's add2F per typ?

def add2F(F, n, fr=0, merge=0):  # unpack for batching in sum2F

    if F.c:
        if isinstance(F,CoF) and isinstance(n,CoF): F.fw += n.fw; F.fc += n.fc  # sum subtree gain/cost, extensive: no weighting
    else:
        if isinstance(F,CoF) and isinstance(n,CoF): F.fw=n.fw; F.fc=n.fc
    if hasattr(n,'C_') and hasattr(n,'C_'): F.C_ = getattr(F,'C_',[]) + n.C_  # same for L_?
    return F

boris-kz: then in frame_H:

                    for i, tF in enumerate(Z.typ_):
?

kwcckw: > that's add2F per typ?

Yes, but the deepest typ's fw and fc are init from globals, so basically it's just accumulating globals?

> then in frame_H:
> 
> ```
>             if rV > ave:
>                 add_typ_(Z)  # maps to FTT_
>                 for i, tF in enumerate(Z.typ_): 
>                     if tF: Fw_[i] = tF.fw; Fc_[i] = tF.fc; FTT_[i] = lev.wTT_[i] = tF.wTT
> ```
> 
> ?

We need to assign Fc_ Fw_ and FTT after the feedback?

The current existing frame_H has /rV, we still need that too?

              if rV > ave:
                    add_typ_(Z)  # maps to FTT_
                    FTT_ = lev.wTT_ = np.array([t.wTT if not isinstance(t,list) else wTT for t in Z.typ_])
                        rV, FTT_ = ffeedback(F)
                        (ave,avd, Lw), Fw_,FTT_ = np.array([ave,avd, Lw])/rV, np.array(Fw_)/rV, np.array(FTT_)/rV

frame_2D_alg/agg_recursion.py
boris-kz: sorry, we still need to accumulate them:

    fO = isinstance(N_[0],CoF)
        if i:
            TT+=n.dTT*w; R+=n.r*w; n_ += n.N_ if merge>1 else [n] if merge else []
            if m_: box=extend_box(box,n.box); kern+=n.kern*w; span+=n.span*w; yx+=n.yx*w
            elif fO: Fw += n.Fw
        else:
            TT = n.dTT*w; R=n.r*w; n_=[]
            if m_: box=copy(n.box); kern=n.kern*w; span=n.span*w; yx=n.yx*w
            elif fO: Fw = n.Fw
    if not merge: n_ = N_
    if m_: F.typ=3; F.m_,F.d_=m_,d_, F.m,F.d=sum(m_),sum(d_); F.box=box; F.kern=kern; F.span=span; F.yx=yx
    else:  
        F.m, F.d = vt_(TT)
        if fO: F.fw = Fw
    root = add2F(root,F) if root!=None else F
    if froot:
        for n in N_: n.root=root
    sum_H(N_, root)
    if getattr(root,'C_',None):
    return root

frame_2D_alg/agg_recursion.py
```diff
@@ -813,14 +819,16 @@ def PP2N(PP):
...
+            PP.rN_ = []  # init, do we really still need rN_?
             for dP in P_:
                 for P in dP.nt: PP.rN_ += [P.root]  # PPm
         return PP
     blob_ = tile.N_; G_,TT,C,R = [],np.zeros((2,9)),0,0
     for blob in blob_:
         if not blob.sign and blob.G > aveB:
             edge = slice_edge(blob, rV)
-            if edge.G * ((len(edge.P_)-1)*Lw) > ave * sum([P.latT[4] for P in edge.P_]):
+            # below is eval for comp_slice, but we don't have w comp_slice, use wVct first?
```
boris-kz, line 830: I guess so, I also added aveCS instead of ave

frame_2D_alg/agg_recursion.py
kwcckw: > elif fO: Fw += n.Fw

So this shouldn't be scaled by w? But then all of them are accumulating the default n.fw, which is from the global?

frame_2D_alg/agg_recursion.py
boris-kz: ok, so F.fw should be accumulating m+d per call?

>We need to assign Fc_ Fw_ and FTT after the feedback?

yes, after the feedback, but not sure we need to update Fc_, it's really static?

frame_2D_alg/agg_recursion.py
kwcckw: Alright, so for those cross module functions, we actually should include them in the onF since they will be called too?

frame_2D_alg/agg_recursion.py
boris-kz: not sure about w, I think it's m+d? actually just m, d is borrowing

frame_2D_alg/agg_recursion.py
boris-kz: right, please add that

frame_2D_alg/agg_recursion.py
kwcckw: > ok, so F.fw should be accumulating m+d per call?
> not sure about w, I think it's m+d? actually just m, d is borrowing

I think probably just m? If we need both, maybe m + abs(d)?

> yes, after the feedback, but not sure we need to update Fc_, it's really static?

But why cost is static here? It should be adjusted by feedback the same way as weight?

boris-kz: cost per call only scales with data, F.c

kwcckw: Okay right, so for code's cost, it should be scaled in higher level?

boris-kz: It's fixed per call, so we just need to count calls: len oF.N_

PR #128, opened 2026-04-28
4/28/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -741,9 +740,9 @@ def add_Lt(G, Lt,wTT):  # addition to Q2R
     G.mang = np.mean([comp_A(G.angl[0], l.angl[0])[0] for l in G.L_])  # Ls only?
 
 # utilities:
-def F2N(F, fG=0):  # extend attrs, or fG only?
-    F.yx=np.zeros(2); F.angl=np.zeros(2); F.kern=np.zeros(4); F.span=0; F.box=np.array([np.inf,np.inf,-np.inf,-np.inf])
-    if fG: F.rim=[]; F.Nt,F.Bt,F.Ct = CF(),CF(),CF(); F.compared = set(); F.fin = 0
+def F2N(F):  # extend attrs, or fG only?
```
kwcckw, line 743: So this F2N should be adding additional params such as Nt, compared, fin , etc only? Since those positional and kern will be added by default now.

frame_2D_alg/agg_recursion.py
```diff
@@ -828,7 +827,7 @@ def PP2N(PP):
     for blob in blob_:
         if not blob.sign and blob.G > aveB:
             edge = slice_edge(blob, rV)
-            if edge.G * ((len(edge.P_)-1)*wVct) > aveCS * sum([P.latT[4] for P in edge.P_]):
+            if edge.G * ((len(edge.P_)-1)*wCS) > cCS * sum([P.latT[4] for P in edge.P_]):
```
kwcckw, line 830: At this point we only need weight and cost for comp_slice, but not frame_blob_root since we don't need to eval in order to run frame_blob_root?

frame_2D_alg/agg_recursion.py
```diff
@@ -546,8 +546,7 @@ def cluster_C(Ft, E_,_r,_c):  # form centroids by clustering exemplar surround v
         for n in [N for C in out_ for N in C.N_]:  # exemplar V + sum n match_dev to Cs, m* ||C rvals:
             n.exe = (n.d if n.typ==1 else n.m) + np.sum(n.m_) - ave
         if vt_(DTT,Ft.root.wTT*ttcC)[0]*wcC*(len(out_)-1) > ave*(r+ccC*(len(out_)-1)):
-            Ct = sum2F(out_,F2N(Ft.root.Ct))
-            if not Ft.root.Ct: Ft.root.Ct = Ct; Ct.root = Ft.root
+            Ct = sum2F([F2N(C) for C in out_],Ft.root.Ct,merge=1)
```
boris-kz, line 549: thanks

frame_2D_alg/agg_recursion.py
```diff
@@ -610,23 +609,23 @@ def cent_TT(dTT, r):  # EM-like weight attr matches | diffs by their match to th
...
 
     c_ = np.array([n.c for n in N_], dtype=float); C = c_.sum(); w_ = c_/C
-    fO = isinstance(N_[0],CoF)
+    fO = isinstance(N_[0],CoF); fC = np.any(m_)
     for i, (n,w) in enumerate(zip(N_,w_)):
         if i:
-            TT+=n.dTT*w; R+=n.r*w; n_ += n.N_ if merge>1 else [n] if merge else []
-            if m_: box=extend_box(box,n.box); kern+=n.kern*w; span+=n.span*w; yx+=n.yx*w
+            TT+=n.dTT*w; R+=n.r*w; n_ += (n.N_ if merge>1 else [n] if merge else [])
+            if fC: box=extend_box(box,n.box); kern+=n.kern*w; span+=n.span*w; yx+=n.yx*w; angl += n.angl[0]
```
boris-kz, line 616: the angl should be empty until we form C.L_?

frame_2D_alg/agg_recursion.py
```diff
@@ -610,23 +609,23 @@ def cent_TT(dTT, r):  # EM-like weight attr matches | diffs by their match to th
...
             elif fO: Fw += n.m
         else:
-            TT = n.dTT*w; R=n.r*w; n_=[]
-            if m_: box=copy(n.box); kern=n.kern*w; span=n.span*w; yx=n.yx*w
+            TT = n.dTT*w; R=n.r*w; n_ = copy(n.N_ if merge>1 else [n] if merge else [])
```
boris-kz, line 619: thanks

frame_2D_alg/agg_recursion.py
boris-kz: right, but do we need to return F?

frame_2D_alg/agg_recursion.py
boris-kz: It's aveB?

frame_2D_alg/meta_code.py
```diff
@@ -25,13 +25,19 @@ def get_ops(node):
         }
         return sum(weights.get(type(n), 0) for n in ast.walk(node))
 
-    with open(path, "r", encoding="utf-8") as f:
-        tree = ast.parse(f.read(), filename=path)
+
     # onF_+ vt_ names and function objects:
-    funcs = {func.name: func for func in ast.walk(tree) if isinstance(func, ast.FunctionDef) and (func.name in onF_ or func.name=='vt_')}
+    funcs = {}
+    for path in paths:
```
boris-kz, line 31: thanks

frame_2D_alg/agg_recursion.py
kwcckw: Probably not, and i think it could F2N_ to convert C_ instead of individual C.

frame_2D_alg/agg_recursion.py
kwcckw: That `aveB` is needed for slice_edge only? So we probably need to add additional weight and cost for slice_edge instead.

frame_2D_alg/agg_recursion.py
kwcckw: So for the base_comp between C and their Ns, their angl's computation should be skipped instead?

boris-kz: it should be skipped in most cases actually, that angl is mostly meaningful in links

kwcckw: Okay, that mA and dA should be default to ((0,0),1) when angl[0] is empty in base_comp?

boris-kz: ((0,0),1) is for angl, mA and dA are scalars, 0 each

frame_2D_alg/agg_recursion.py
boris-kz: right, all downstream functions

frame_2D_alg/agg_recursion.py
kwcckw: Yes, so if N.angl[0] is empty, we can skip the comp_A in base_comp below?

    mA,dA = comp_A(_N.angl[0]*_N.angl[1], N.angl[0]*N.angl[1])
    m_,d_ = comp(_pars,pars, mA,dA)  # M,D,n, I,G,a, L,S,A
    dm_,dd_ = comp_derT(_N.dTT[1], N.dTT[1])

boris-kz: probably

boris-kz: actually just make a default angl None?

kwcckw: > actually just make a default angl None?

Okay, so it could be simpler:

    mA,dA =  comp_A(_N.angl[0]*_N.angl[1], N.angl[0]*N.angl[1]) if (_N.angl and N.angl) else (0,0)

frame_2D_alg/agg_recursion.py
kwcckw: For F2N, C.Nt should be summed from C.N_, while their Lt should be summed from C.L_? I can see Lt is not init here, and we need it for sub-comp later?

boris-kz: yes for C.Nt, but CF already has Lt, although it may not be summed from L_ at this point.

kwcckw: > yes for C.Nt, but CF already has Lt, although it may not be summed from L_ at this point.

That Lt is init as empy list, so i think we need to init it as CF with C.L_ here.

boris-kz: ok that was meant  for cross_comp anyway, so L_ should be empty before F2N?
that means we need to remove Lt and Ct from CF definition?

boris-kz: so we need to add all forks:

def F2N(F):  # extend for cross_comp

    F.N_,F.C_,F.L_,F.B_,F.X_,F.rim = prop_F_('Nt'),prop_F_('Ct'),prop_F_('Lt'),prop_F_('Bt'),prop_F_('Xt'),prop_F_('Rt')
    F.Nt,F.Lt,F.Bt,F.Ct,F.Rt = sum2F(F.N_),CF(typ='Lt'),CF(typ='Bt'),CF(typ='Ct'),CF(typ='Rt')
    F.compared = set(); F.fin = 0

kwcckw: For L_, it should be empty until we fill it in cluster_P? So we may get Lt in those Cs before F2N.

For Ct, i guess we not really need it in CF definition.

boris-kz: too confusing, just run F2N:

            if N_:
                C = sum2F(N_, root, m_,d_); F2N(C)
                C.L_= [CN(typ=1,nt=[N,C],dTT=N.dTT,c=N.c,r=N.r,m=N.m,d=N.d,span=np.hypot(*(dy_dx:=C.yx-N.yx)),angl=np.array([dy_dx,np.sign(N.dTT[1]@ttcN[1])],dtype=object))
                       for N in N_]  # form C-to-N links

frame_2D_alg/agg_recursion.py
boris-kz, line 998: TT,C,R = sum_vt(T_, wTT=ttFrm); R += elev; L = len(T_)-1
            if sum(vt_(TT,ttFrm))*(wFrm*L) <= (ave+avd)*(R+elev+cFrm*L): T_=[]; C=0; R=0

frame_2D_alg/agg_recursion.py
kwcckw: > too confusing, just run F2N:
> 
> ```
>         if N_:
>             C = sum2F(N_, root, m_,d_); F2N(C)
>             C.L_= [CN(typ=1,nt=[N,C],dTT=N.dTT,c=N.c,r=N.r,m=N.m,d=N.d,span=np.hypot(*(dy_dx:=C.yx-N.yx)),angl=np.array([dy_dx,np.sign(N.dTT[1]@ttcN[1])],dtype=object))
>                    for N in N_]  # form C-to-N links
> ```

So C.Lt is empty, but C.L_ is filled here?

boris-kz: not empty, Lt.N_ = L_, it's just not summed

kwcckw: But this is CF? `CF.L_` is not `CF.Lt.N_` since we don't have prop_F_ here, unless we assign it with C.Lt.N_ instead of C.L_.

boris-kz: I added  prop_F_

kwcckw: > I added prop_F_

Alright, let me go through them again later.

frame_2D_alg/agg_recursion.py
```diff
@@ -610,23 +609,23 @@ def cent_TT(dTT, r):  # EM-like weight attr matches | diffs by their match to th
 def sum2F(N_, root=None, m_=[],d_=[], merge=0, froot=0):  # -> CF/CN
```
boris-kz, line 609: I think #6 is correct?

Rechecking `sum2F` — a few conceptual issues, ordered by importance:

**1. `fC = np.any(m_)` is fragile.**
If `m_` is `[0,0,...]` (all zero matches, degenerate but legal), fC is False and centroid mode silently turns off — no `box/kern/span/yx/angl/m_/d_` get attached. Should be `fC = bool(len(m_))` or pass an explicit flag.

**2. `angl` accumulation is broken for typical Gs.**
- Init sets `angl = copy(n.angl[0]) if n.angl else None`. If first N has `n.angl is None`, angl stays None — then on iter 2 `angl += n.angl[0]` errors (`None + array`).
- Final guard `if angl:` raises for any numpy array `angl` of length > 1 (ambiguous truth value). And typical G.angl[0] from `add_Lt` *is* a 2-element np.array.

So fC + Gs-with-array-angls = ValueError on the `if angl:` line. Either centroids never actually have array-typed angl[0] in practice, or this path hasn't been exercised. Which is it?

**3. `F.m, F.d = vt_(TT)` uses module-default wTT**, not `F.wTT` (or the caller's). Everywhere else in the file evals pass explicit weights (ttN_, ttE, ttcN, etc.) — this one slot uses the generic global, which makes summed F.m/F.d incommensurable with eval-side m/d. Intentional?

**4. `sum_H(N_, root)` unconditionally overwrites `root.H`.**
Fine when root is freshly constructed in this call, but if root was passed in with existing H (via `add2F(root,F,merge=2)` branch), the prior H is silently clobbered. Looks like a bug for incremental accumulation.

**5. Side-effect at the tail:**
```python
if getattr(root,'C_',None):
```
Why is C-tree rebuilding part of "sum N_ into F"? It also reassigns `root.Ct`, discarding whatever was there.

**6. `merge=1` vs `merge=0`** produces effectively identical `F.N_`:
- merge=0: `F.N_ = N_` (alias).
- merge=1: `F.N_ = [N_[0], N_[1], …]` (fresh list, same elements).
Only merge>1 is meaningfully different (flatten one level). Is merge=1 just to force a new list, or is there intended semantic distinction?

**Minor:** the loop builds `n_` even when `not merge`, then `n_ = N_` discards it.

Want me to draft a minimal patch for #1 and #2 only? The rest are questions, not edits.

kwcckw: I think yes, so when merge == 1, n_ should be merged with n.N_, while n_ should append n when merge == 0?

boris-kz: merge=0: F.N_ = N_
merge=1: F.N_ = [*n.N_ for n in N_]

PR #129, opened 2026-04-29
4/29/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -490,8 +490,10 @@ def trans_cluster(G):
         sel_TT = (Ft.dTT*Ft.c - Ft.root.Lt.dTT*Ft.root.Lt.c) / eps_(Ft.root.Lt.dTT * Ft.root.Lt.c)
         oF = CoF.get(); oF.N_=copy(G_); oF.rTT=sel_TT; oF.c = Ft.root.c - Ft.c; oF.r=sum([G.r * G.c for G in G_])/oF.c  # sum_vt?
         # combine C_:
-        C_ = [C for N in (G_ if G_ else _N_) for C in N.Ct.N_]
-        if C_: sum2F(C_, root=Ft.root.Ct)  # Ct.r includes overlap?
+        C_ = [C for N in (G_ if G_ else _N_) for C in N.Ct.N_]   
+        if C_:
+            sum2F(C_, root=Ft.root.Ct)  # Ct.r includes overlap?
+            Ft.root.Ct.Lt = sum2F([L for C in C_ for L in C.L_], root=Ft.root.Ct)
```
kwcckw, line 496: If not mistaken, we need to add Lt into Ct?

frame_2D_alg/agg_recursion.py
```diff
@@ -733,14 +734,15 @@ def add_Lt(G, Lt,wTT):  # addition to Q2R
...
             sum2F(L_,Lt); add2F(G,Lt)
     A = np.sum([l.angl[0] for l in L_], axis=0) if L_ else np.zeros(2)
-    G.angl = np.array([A, np.sign(G.dTT[1] @ ttcN[1])], dtype=object)
+    G.angl = [A, np.sign(G.dTT[1] @ ttcN[1])]
     G.mang = np.mean([comp_A(G.angl[0], l.angl[0])[0] for l in G.L_])  # Ls only?
 
 # utilities:
 def F2N(F):  # extend for cross_comp F.N_
-    F.N_, F.L_, F.B_, F.C_, F.X_, F.rim = prop_F_('Nt'),prop_F_('Lt'),prop_F_('Bt'),prop_F_('Ct'),prop_F_('Xt'),prop_F_('Rt')
+    # F.N_, F.L_, F.B_, F.C_, F.X_, F.rim = prop_F_('Nt'),prop_F_('Lt'),prop_F_('Bt'),prop_F_('Ct'),prop_F_('Xt'),prop_F_('Rt')
```
kwcckw, line 742: We can't assign prop_F in runtime, so it's better to just convert their class below?

frame_2D_alg/agg_recursion.py
```diff
@@ -935,11 +938,12 @@ def proj_TT(L, cos_d, dist, r, pTT, wTT, dec=1, fdec=0, frec=0):  # accumulate l
 def proj_N(N, dist, A, r, _c, dec=1):  # arg rc += N.rc+Nw, recursively specify N projection val, add pN if comp_pN?
 
     oF = CoF.get(); oF.c = _c; oF.r = r
-    cos_d = (N.angl[0].dot(A) / ((np.hypot(*N.angl[0]) * dist) or eps)) * N.angl[1]  # internal x external angle alignment
+    if N.angl: cos_d = (N.angl[0].dot(A) / ((np.hypot(*N.angl[0]) * dist) or eps)) * N.angl[1]  # internal x external angle alignment
+    else:      cos_d = 1
```
kwcckw, line 942: We might not have N.angl (when cross_comp of Ct.N_ call comp_N_ via comp_F) here, so we should skip the projection?

frame_2D_alg/agg_recursion.py
```diff
@@ -640,16 +641,16 @@ def add2F(F, n, fr=0, merge=0):  # unpack for batching in sum2F
...
-    if hasattr(n,'C_') and hasattr(n,'C_'): F.C_ = getattr(F,'C_',[]) + n.C_  # same for L_?
+    if hasattr(n,'C_'): F.C_ = getattr(F,'C_',[]) + n.C_  # same for L_?
     return F
 
 def add_H(H,h, root, fN=0):
     # bottom-up:
     for Lev,lev in zip_longest(H, h):
         if lev:
-            if Lev: add2F(Lev,lev, merge=2)
+            if Lev: add2F(Lev,lev, merge=1)
```
boris-kz, line 653: do we ever have merge=2 now?

frame_2D_alg/agg_recursion.py
boris-kz: I don't think we need to sum L_. also if C_ goes through cross_comp, that L_ and Lt will be replaced anyway

frame_2D_alg/agg_recursion.py
```diff
@@ -536,12 +538,12 @@ def cluster_C(Ft, E_,_r,_c):  # form centroids by clustering exemplar surround v
             for n in N_: n._c_ = n.c_; n._m_ = n.m_; n._d_ = n.d_
             for n in set([_n for _C in _C_ for _n in _C.N_ + _C._N_]): n.c_, n.m_, n.d_ = [], [], []
             _C_ = C_
-        else: out_+=C_; break  # converged
+        else: out_+=C_; [F2N(C) for C in out_]; break  # converged
```
boris-kz, line 541: this should be conditional on the eval below

frame_2D_alg/agg_recursion.py
```diff
@@ -607,22 +609,21 @@ def sum2F(N_, root=None, m_=[],d_=[], merge=0, froot=0):  # -> CF/CN
     fO = isinstance(N_[0],CoF); fC = np.any(m_)
     for i, (n,w) in enumerate(zip(N_,w_)):
         if i:
-            TT+=n.dTT*w; R+=n.r*w; n_ += (n.N_ if merge>1 else [n] if merge else [])
-            if fC: box=extend_box(box,n.box); kern+=n.kern*w; span+=n.span*w; yx+=n.yx*w; angl += n.angl[0] if n.angl else None  # mostly for Ls?
+            TT+=n.dTT*w; R+=n.r*w; n_ += (n.N_ if merge else [n])
+            if fC: box=extend_box(box,n.box); kern+=n.kern*w; span+=n.span*w; yx+=n.yx*w; (angl := angl + n.angl[0]) if n.angl else None  # mostly for Ls?
```
boris-kz, line 613: thanks

frame_2D_alg/agg_recursion.py
```diff
@@ -607,22 +609,21 @@ def sum2F(N_, root=None, m_=[],d_=[], merge=0, froot=0):  # -> CF/CN
...
             elif fO: Fw += n.m
         else:  # init with N_[0]
-            TT = n.dTT*w; R=n.r*w; n_ = copy(n.N_ if merge>1 else [n] if merge else [])
-            if fC: box=copy(n.box); kern=n.kern*w; span=n.span*w; yx=n.yx*w; angl = copy(n.angl[0]) if n.angl else None
+            TT = n.dTT*w; R=n.r*w; n_ = copy(n.N_ if merge else [n])
+            if fC: box=copy(n.box); kern=n.kern*w; span=n.span*w; yx=n.yx*w; angl = copy(n.angl[0]) if n.angl else np.zeros(2)
```
boris-kz, line 617: angl = copy(n.angl[0]) if n.angl else None?

frame_2D_alg/agg_recursion.py
boris-kz: that changes the object?

frame_2D_alg/agg_recursion.py
boris-kz: no, this alignment is a secondary factor, projection is valid even without N.angl

frame_2D_alg/agg_recursion.py
kwcckw: > if C_ goes through cross_comp, that L_ and Lt will be replaced anyway

We didn't assign any Lt to root from cross_comp now? So there will be no Lt for the Ct anyway.

frame_2D_alg/agg_recursion.py
kwcckw: It shouldn't change the object, from LLM:
image

frame_2D_alg/agg_recursion.py
kwcckw: I think no since we only need to merge or append n now.

frame_2D_alg/agg_recursion.py
kwcckw: Okay, but we may have converted Cs from the eval below, so we need additional check to see if they are converted before we running F2N again.

frame_2D_alg/agg_recursion.py
kwcckw: Okay, so cos_d should be default to 1 since it's one of the factor.

frame_2D_alg/agg_recursion.py
boris-kz: great, then we don't need F2N?

kwcckw: We still need F2N to assign the attrs since it doesn't add the additional attrs from CN:

      Changing __class__ does not call CN.__init__. The instance keeps only what was in its __dict__ from CF.__init__. Any attrs that CN.__init__ would have added are missing

frame_2D_alg/agg_recursion.py
boris-kz: actually Lt should be assigned to the root, replacing old Lt, if any
It's confusing because we have two ways of forming Lt: +only in sum2G and +- in cross_comp.
tile / frame are not formed by sum2G, so this Lt replacement only happens in sub+?

frame_2D_alg/agg_recursion.py
boris-kz: check where? there is no F2N before 546?

frame_2D_alg/agg_recursion.py
boris-kz: ok, just call CN.__init__?

frame_2D_alg/agg_recursion.py
```diff
@@ -935,11 +938,12 @@ def proj_TT(L, cos_d, dist, r, pTT, wTT, dec=1, fdec=0, frec=0):  # accumulate l
...
 
     oF = CoF.get(); oF.c = _c; oF.r = r
-    cos_d = (N.angl[0].dot(A) / ((np.hypot(*N.angl[0]) * dist) or eps)) * N.angl[1]  # internal x external angle alignment
+    if N.angl: cos_d = (N.angl[0].dot(A) / ((np.hypot(*N.angl[0]) * dist) or eps)) * N.angl[1]  # internal x external angle alignment
+    else:      cos_d = 1
     iTT, eTT = np.zeros((2,9)), np.zeros((2,9))
     wTT = oF.wTT*ttPrj
     for L in N.L_+ N.B_: proj_TT(L, cos_d, dist, L.r+r, iTT, wTT, dec)  # accums TT internally
-    if N.Ct.Lt: [proj_TT(L,cos_d,dist,L.r+r,iTT,wTT,dec) for L in N.Ct.Lt.N_]  # C-to-N links
+    if hasattr(N.Ct,'Lt') and N.Ct.Lt: [proj_TT(L,cos_d,dist,L.r+r,iTT,wTT,dec) for L in N.Ct.Lt.N_]  # C-to-N links
```
boris-kz, line 946: should be isinstance(CN) now?

frame_2D_alg/agg_recursion.py
boris-kz: actually the default should be mean cos, which is 0, so in proj_TT we need (cos_d + dec), not * dec?

frame_2D_alg/agg_recursion.py
boris-kz: you have to specify which LLM you are using, they are not the same

frame_2D_alg/agg_recursion.py
boris-kz: `if fC: box=extend_box(box,n.box); kern+=n.kern*w; span+=n.span*w; yx+=n.yx*w; angl = angl+n.angl[0] if n.angl is not None else None`

frame_2D_alg/agg_recursion.py
```diff
@@ -607,22 +609,21 @@ def sum2F(N_, root=None, m_=[],d_=[], merge=0, froot=0):  # -> CF/CN
...
-            if fC: box=copy(n.box); kern=n.kern*w; span=n.span*w; yx=n.yx*w; angl = copy(n.angl[0]) if n.angl else None
+            TT = n.dTT*w; R=n.r*w; n_ = copy(n.N_ if merge else [n])
+            if fC: box=copy(n.box); kern=n.kern*w; span=n.span*w; yx=n.yx*w; angl = copy(n.angl[0]) if n.angl else np.zeros(2)
             elif fO: Fw = n.m
-    if not merge: n_ = N_
     F = (CoF if fO else CF)(N_=n_, dTT=TT, c=C, r=R)
     if fC:
         F.typ=3; F.m_,F.d_=m_,d_; F.m,F.d=sum(m_),sum(d_); F.box=box;F.kern=kern;F.span=span;F.yx=yx
-        if angl: F.angl=np.array([angl,np.sign(F.dTT[1] @ ttcP[1])],dtype=object)
+        F.angl = [angl,np.sign(F.dTT[1] @ ttcP[1])] if np.any(angl) else None
```
boris-kz, line 622: if angl is not None: F.angl,np.sign(F.dTT[1] @ ttcP[1])

frame_2D_alg/agg_recursion.py
kwcckw: > , so this Lt replacement only happens in sub+?

Yes exactly, the top tile/frame won't get Lt since Lt is only summed in sum2G. So i guess we need to add back the prior section where we pack new L_ into root.Lt?

frame_2D_alg/agg_recursion.py
kwcckw: I'm using opus 4.6 now? I checked with the code below and getting true below:

      n1 = CN()
      n2 = CN()
      n1.root = n2
      n2.__class__ = CF
      n1.root is n2

frame_2D_alg/agg_recursion.py
kwcckw: I see, if we looking at mean of cos, that value should be default to 0. But that ` (cos_d + dec),` may always scale dTT[1] to a larger value compare to `L.dTT[0] * dec`?

frame_2D_alg/agg_recursion.py
kwcckw: We have F2N in cluster_P now? So `out_` actually are those converted Cs.

    out_ = cluster_P(out_,T, Ft)

If we convert it under `if out_:`, we might convert them again.

frame_2D_alg/agg_recursion.py
kwcckw: We may still get Lt from Cs (CF)? Added from cluster_P?

frame_2D_alg/agg_recursion.py
boris-kz: right, I just don't know if we need to preserve old Lt in sub+

frame_2D_alg/agg_recursion.py
boris-kz: just call 
`F.__class__ = CN; F.__init__`?

frame_2D_alg/agg_recursion.py
boris-kz: doesn't matter, it's more accurate?

frame_2D_alg/agg_recursion.py
kwcckw: That could be added in H?

frame_2D_alg/agg_recursion.py
kwcckw: Yes, we can use `__init__`, but not directly since we need to init with the attrs, from claude:

    saved = {k: v for k, v in F.__dict__.items() 
             if k in ('dTT','m','d','c','r','wTT','root','typ')}
    
    F.__class__ = CN
    CN.__init__(F, **saved)  # fills ALL CN attrs with defaults, overrides with save

frame_2D_alg/agg_recursion.py
kwcckw: Okay, i guess we can use cos_d + dec first.

frame_2D_alg/agg_recursion.py
boris-kz: They should only be converted if they go through cross-comp

frame_2D_alg/agg_recursion.py
boris-kz: probably not Lt, make it cL_?

frame_2D_alg/agg_recursion.py
boris-kz: ok that's +ve lev.L_, with the original L_ in H[-1]?

frame_2D_alg/agg_recursion.py
```diff
@@ -571,13 +573,13 @@ def cluster_P(_C_, _c, root):  # FCM-style parallel centroid refine, may add pro
...
     out_ = []
     for i, _C in enumerate(C_):
-        if _C.m > ave * C.r:  # prune, add olp as stronger ms?
+        if _C.m > ave * _C.r:  # prune, add olp as stronger ms?
             N_,m_,d_ = [],[],[]
             for N, m,d in zip(_C.N_, md__[:,i,0], md__[:,i,1]):
                 if m*N.c > ave*N.r: N_+=[N]; m_+=[m]; d_+=[d]
             if N_:
                 C = sum2F(N_, root, m_,d_); F2N(C)
-                C.L_=[CN(typ=1,nt=[N,C],dTT=N.dTT,c=N.c,r=N.r,m=N.m,d=N.d,span=np.hypot(*(dy_dx:=C.yx-N.yx)),angl=np.array([dy_dx,np.sign(N.dTT[1]@ttcN[1])],dtype=object))
```
boris-kz, line 580: so this can be CF.L_, packed in C.H if C goes through cross_comp?

frame_2D_alg/agg_recursion.py
boris-kz: Then that root Ft in cross_comp should be converted to CN, adding H?

frame_2D_alg/agg_recursion.py
boris-kz:
```python
def F2N(F):
    Nt = CopyF(F, root=F); Nt.N_ = F.N_  # Nt inherits F's scalars + N_
    Lt = CF(N_=F.L_, root=F) if F.L_ else CF(root=F)
    Ct = CF(N_=F.C_, root=F) if getattr(F,'C_',None) else CF(root=F)
    F.__class__ = CN
    for k,v in Na_.items(): setattr(F, k, copy(v))
    F.compared = set()
    F.Nt, F.Lt, F.Ct = Nt, Lt, Ct
    for ft in ('Bt','Xt','Rt'): setattr(F, ft, CF(root=F))
    if Lt.N_:
        lev = sum2F(Lt.N_, F); lev.N_=Nt.N_; lev.L_=Lt.N_; F.H += [lev]
```

With Na_ as previously defined:

```python
Na_ = dict(H=[], kern=np.zeros(4), span=1, angl=None, mang=1,
           box=np.array([np.inf,np.inf,-np.inf,-np.inf]),
           yx=np.zeros(2), sub=0, exe=0, fin=0, nt=[], c_=[])
```

One catch worth flagging: `CopyF` recursively copies `F.N_` (with `CopyF` for nested CF, identity for CN). Since you want Nt to *share* the same N_ list as F (so the property sees the same nodes), my `Nt.N_ = F.N_` overwrites CopyF's copy — good. But if any caller later mutates `F.N_` expecting it to also update internal copies, that's not what happens. Probably fine for your use, just noting.

frame_2D_alg/agg_recursion.py
kwcckw: > probably not Lt, make it cL_?

So it should be L_ instead of Lt.N_ now? I think we should stick with L_ instead of cL_ since we might add L_ in the subsequent cross_comp, so L_ should be added to H in the same way.

frame_2D_alg/agg_recursion.py
kwcckw: > Then that root Ft in cross_comp should be converted to CN, adding H?

We already have existing H in CF? 

        f.H  = kw.get('H', [])  # lower CF levs / Nt||Ct
        
> They should only be converted if they go through cross-comp

Right now it's converted in cluster_P to add L_ into Lt:

        C = sum2F(N_, root, m_,d_); F2N(C)
        C.L_=[CN(typ=1,nt=[N,C],dTT=N.dTT,c=N.c,r=N.r,m=N.m,d=N.d,span=np.hypot(*(dy_dx:=C.yx-N.yx)),angl=[dy_dx,np.sign(N.dTT[1]@ttcN[1])])
              for N in N_]  # form C-to-N links
        out_ += [C]
        
With cL_ or L_ version, this F2N should be removed and added later.

frame_2D_alg/agg_recursion.py
kwcckw: Yes, it should be only +ve in sub+, also for consistency, i think we can pack the original Lt into H[-1]?

boris-kz: lev should include F.N_, so it's not just Lt

frame_2D_alg/agg_recursion.py
boris-kz: I moved that H to CN

frame_2D_alg/agg_recursion.py
boris-kz: right, CF only needs L_

frame_2D_alg/agg_recursion.py
boris-kz: def F2N(F):  # convert for cross_comp

    Nt = CopyF(F,root=F); Nt.N_ = F.N_; L_ = F.L_  # replace in cross_comp
    F.__class__ = CN
    for k,v in dict(H=[],kern=np.zeros(4),span=1,angl=None,mang=1,
                    box=np.array([np.inf,np.inf,-np.inf,-np.inf]),
                    yx=np.zeros(2),sub=0,exe=0,fin=0,nt=[],c_=[]).items():
        setattr(F, k, copy(v))
    F.Nt = Nt; F.compared = set()
    if L_: lev = sum2F(L_, F); lev.N_ = Nt.N_; lev.L_ = L_; F.H += [lev]

frame_2D_alg/agg_recursion.py
kwcckw: But this is level for Lt only? F.N_ should be those Ls?

frame_2D_alg/agg_recursion.py
kwcckw: Thanks, I will step through this later.

frame_2D_alg/agg_recursion.py
kwcckw: Okay, and we need to add H in cross_comp now? That's when we reform Ft with G_? But that will be packed in Ft.root.H instead of Ft.H?

frame_2D_alg/agg_recursion.py
boris-kz: it's a compositional level. but we are not using that N_, and it can be recovered from L_ nt, 
so I guess lev=Lt is fine

frame_2D_alg/agg_recursion.py
boris-kz: I think Ft.H is fine, that's the current scheme, and we need F2N anyway

kwcckw: So the existing Ft should be packed in new Ft.H? That's mean the max level is a single level only?

frame_2D_alg/agg_recursion.py
kwcckw: Okay, but this H is in Lt.H right? If we pack lev as CF, those Ns could be retrieved from lev.N_? Then the top L_ will be in Lt.N_?

frame_2D_alg/agg_recursion.py
boris-kz: that's a first level, deeper levels should be added in 

Five sites:

- **`add_Nt`** (line 716): `G.Nt.H += [sum2F(N_+ (G.Nt.H.pop().N_ if G.Nt.H else []), G.Nt)]` — top-level grow during graph combination.
- **`sum_H`** (line 661, called from `sum2F` line 630): `Ft.H = [sum2F(H[0], CF())] if H else []` — initializes Ft.H from N_'s contents.
- **`add_H`** (line 650, called from `add2F` line 646): `add2F(Lev,lev,1)` per zip-pair — bottom-up extend during accumulation.
- **`F2N`** (line 743): `F.H += [lev]` — appends an L-derived level on conversion.
- **`frame_H`** (lines 1007–1009): `F.H = []`, then `add_H(F.H, T.H, ...)` per tile, then `F.H += [lev := sum2F(tile_, CF())]` — assembles per-frame H across elevations.

So writers come in two flavors: ones that **append a new top level** (add_Nt, F2N, frame_H's last line), and ones that **merge level-by-level** (add_H, sum_H from sum2F).

One thing to flag while you're auditing: in `add_Nt` the line `G.Nt.H if merge else G.Nt.H[:-1]` (line 713) pops/excludes the top before passing to `add_H`, then line 716 does its own `G.Nt.H.pop()` again. That's two different pop sites depending on merge — possibly intentional but worth re-deriving.

Need to review those

frame_2D_alg/agg_recursion.py
boris-kz: no, it's a root H

kwcckw: Okay,  I thought we are refering to Lt here. If we pack Lt in root.H, this should be consistent across all H?

boris-kz: yes, please review those

frame_2D_alg/agg_recursion.py
kwcckw: I think other than cross_comp and frame_H, the rest should be just combining levels?

frame_2D_alg/agg_recursion.py
kwcckw: Okay, and the H in frame_H should be a unique case for tiles only。

frame_2D_alg/agg_recursion.py
boris-kz: we add a level every we add new N_, in agg+/sub+
that new N_ pushes the old N_ onto H

PR #130, opened 2026-04-30
4/30/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -206,6 +207,7 @@ def cross_comp(Ft, rr):  # core function mediating recursive rng+ and der+ cross
             E_ = get_exemplars({N for L in L_ for N in L.nt}, r,c)
             G_,r = cluster_N(Ft, E_,r,c)  # -> cluster_C,_P, eval?
             if G_:
+                root.H += [sum2F(L_, root)]  # replaces Nt, but pack from L_
```
kwcckw, line 210: Sorry, i'm still not fully review the H. So if we pack H with Ls, root.Nt will be replaced, and we will no longer able to retrieve the prior root.Nt? The prior Ns can be retrieved via H's N_'s nt only.

frame_2D_alg/agg_recursion.py
```diff
@@ -147,6 +147,7 @@ def inner(*a, **kw):
             CoF._cur.set(oF); out = func(*a, **kw)
             if oF.call_:  # complete at this point
                 cT = flat_(oF); L=(len(cT)-1)  # flatten call tree
+                # sum2F here is packing CoF into CoF.N_?
                 if oF.fw*L > ave*(oF.fc*L): sum2F(cT,oF)  # eval summarize, fc*fr if multiple oFs on the same data?
```
kwcckw, line 151: This line is packing calls into CoF.N_, but we may pack data into CoF.N_ too.

For example, in cluster_N:

     oF.N_=copy(G_)

So oF.N_ should be calls or data here?

frame_2D_alg/agg_recursion.py
```diff
@@ -147,6 +147,7 @@ def inner(*a, **kw):
             CoF._cur.set(oF); out = func(*a, **kw)
             if oF.call_:  # complete at this point
                 cT = flat_(oF); L=(len(cT)-1)  # flatten call tree
+                # sum2F here is packing CoF into CoF.N_?
```
boris-kz, line 150: I think it was `_oF: oF is a new call in oF.call_, which maps to oF.N_`. Not sure how that got changed

frame_2D_alg/agg_recursion.py
```diff
@@ -327,7 +329,7 @@ def comp_H(_Nt,Nt, Link):  # tentative pre-comp
                         r = (i or 1) -1  # Nt,Lt are core, not redundant
             for fb_, nF in zip((L.Nt.fb_,L.Lt.fb_,L.Bt.fb_,L.Ct.fb_), ('Nt','Lt','Bt','Ct')):
                 if fb_:  # += trans-links, python-batched bottom-up
-                    Ft = sum2F(fb_, getattr(L,nF)); add2F(L,Ft)
+                    sum2F(fb_, getattr(L,nF))  # add2F(L,Ft,merge=1) This add2F is redundant now since we already have add2F when root is not empty
```
kwcckw, line 332: Actually most of the add2F is redundant now since we already have add_F within sum2F?

    root = add2F(root,F,1) if (root!=None and not fC) else F

frame_2D_alg/agg_recursion.py
boris-kz: This should be in clustering: sum2G. cross_comp adds L_, sum2G packs it in H.
so F2N should be the first time we form new N_ in root, not before cross_comp

frame_2D_alg/agg_recursion.py
boris-kz: calls contain data, so oF.call_ is redundant?

frame_2D_alg/agg_recursion.py
boris-kz: Sorry, this is after clustering, so it looks right

boris-kz: we just need

            if G_:
                if isinstance(root,CF): F2N(Ft)  # upgrade per 1st sub+|agg+ per root
                root.H += [sum2F(L_,Ft)]  # L_ is new derivatives

frame_2D_alg/agg_recursion.py
boris-kz: probably

frame_2D_alg/agg_recursion.py
kwcckw: Okay, but Ft is the root of the level? That should be Ft.root?

frame_2D_alg/agg_recursion.py
kwcckw: But oF is a call too, which is the same?

frame_2D_alg/agg_recursion.py
kwcckw: Sorry, could you explain further ? Why oF is a new call in the same oF.call_?

frame_2D_alg/agg_recursion.py
boris-kz: no, froot=0 for levels. doesn't matter, cross_comp is per fork, not level

frame_2D_alg/agg_recursion.py
boris-kz: it's a different level of calls

frame_2D_alg/agg_recursion.py
boris-kz: in `_oF`, not oF

kwcckw: Okay, but how `_oF` is related to the `oF.call_` here?

We are packing all nested `oF.call_` after the oF function ends, so it should be relevant to `oF` only? `_oF` should be already packed their calls when their function ends too.

frame_2D_alg/agg_recursion.py
kwcckw: > calls contain data, so oF.call_ is redundant?

`oF.call_` contains deeper nested calls, while data is packed in oF.N_, which we can't access calls from data, why it is redundant here?

frame_2D_alg/agg_recursion.py
kwcckw: Okay right, but we shouldn't parse Ft, else Ft will be returned and packed into root.H.

frame_2D_alg/agg_recursion.py
boris-kz: we don't need to access calls from the data, only data from calls

frame_2D_alg/agg_recursion.py
boris-kz: Ft is the root now, we don't need Ft.root because we convert it to CN:

    N_,G_ = root.N_,[]; fC = N_[0].typ==3  # root is Ft, converted below, rc=rdn+olp, comp N_|B_|C_:
    # comp_ s update root oF:
    if L_: # Lm_, no +|- Ft.Lt?
            G_,r = cluster_N(root, E_,r,c)  # -> cluster_C,_P, eval?
            if G_:
                if isinstance(root,CF): F2N(root)  # upgrade @ 1st sub|agg+
                root.H += [sum2F(L_,root)]  # L_ only: new derivatives
                root.Nt = sum2F(G_,CF(root=root),froot=1); L=len(G_)-1  # or G.N_ = spliced C_ if higher value?
                if vt_(TT,ttX)[0]* (wX*L) > ave* (r+cX*L):  # root borrw| rdn?

frame_2D_alg/agg_recursion.py
```diff
@@ -658,7 +665,7 @@ def sum_H(N_, Ft):
     for N in N_:
         if H: H[0] += N.N_
         elif N.N_: H = [list(N.N_)]
-    Ft.H = [sum2F(H[0], CF())] if H else []
+    Ft.H += [sum2F(H[0], CF())] if H else []  # should be append here? Else it overwrites existing levels
```
boris-kz, line 668: that was for combining H into single lev, in L, to avoid nested H. But it's probably not relevant for Gs

frame_2D_alg/agg_recursion.py
boris-kz: sorry it's oF, but sum2F doesn't pack oF, it summarizes oF. call_ (now N_) in oF vals. refactored:

def flat_(oF, N_=None):  # all nested call_ s

    if N_ is None: N_ = []
    for sub in oF.N_:
        N_ += [sub]
        if sub.N_: flat_(sub, N_)
    return N_

    name = "func"
    def __init__(f, **kw):
        super().__init__(**kw)
        f.typ_ = kw.get('typ_', [])  # call tree flattened and nested with tFs
        f.fw,f.fc,f.fr = [kw.get(x,0) for x in ('fw','fc','fr')]  # call_ gain, cost, rdn: multiple oFs on same data?
    @staticmethod
    @staticmethod
    def traced(func):
        @wraps(func)
        def inner(*a, **kw):
            _CoF = CoF._cur.get()
            oF = CoF(nF=onF_.index(func.__name__),root=_CoF); oF.wTT = FTT_[oF.nF]; oF.fw = Fw_[oF.nF]; oF.fc = Fc_[oF.nF]; _CoF.N_+=[oF]
            if oF.N_:  # complete at this point
                call_ = flat_(oF); L=(len(call_)-1)  # flat call tree
                if oF.fw*L > ave*(oF.fc*L): sum2F(call_,oF)  # eval summarize, fc*fr if multiple oFs on the same data?
            oF.wTT= cent_TT(getattr(oF,'rTT',oF.dTT), oF.r)  # rTT covers cluster compression
            oF.w += np.mean(oF.wTT)
            CoF._cur.set(_CoF)
            return out
        inner.wrapped = True
        return inner

frame_2D_alg/agg_recursion.py
boris-kz: then G.H = G.Nt.H:

def prop_F_(F, attr='N_'):  # factory function to get and update top-composition fork.N_|H

    def get(N): return getattr(getattr(N,F), attr)
    def set(N, new_val): setattr(getattr(N,F), attr, new_val)
    return property(get,set)

class CN(CBase):  # full node / graph attrs:

    name = "node"
    N_,C_,L_,B_,X_,rim,H = prop_F_('Nt'),prop_F_('Ct'),prop_F_('Lt'),prop_F_('Bt'),prop_F_('Xt'),prop_F_('Rt'),prop_F_('Nt','H')  # ext|int -defined Ns,Ls

frame_2D_alg/agg_recursion.py
```diff
@@ -611,7 +613,9 @@ def sum2F(N_, root=None, m_=[],d_=[], merge=0, froot=0):  # -> CF/CN
     for i, (n,w) in enumerate(zip(N_,w_)):
         if i:
             TT+=n.dTT*w; R+=n.r*w; n_ += (n.N_ if merge else [n])
-            if fC: box=extend_box(box,n.box); kern+=n.kern*w; span+=n.span*w; yx+=n.yx*w; angl += n.angl[0] if n.angl is not None else None  # mostly Ls?
```
boris-kz, line 614: what's wrong here?

frame_2D_alg/agg_recursion.py
kwcckw: Okay, so we actually can remain the Nt.H, and use N.H when needed.

frame_2D_alg/agg_recursion.py
kwcckw: > we don't need to access calls from the data, only data from calls

But we may pack data into oF.N_  from those clustering functions , and then packing deeper calls into the same oF.N_ via this line `if oF.fw*L > ave*(oF.fc*L): sum2F(call_,oF)`,  this is causing oF.N_ has a combination of data and CoF.

frame_2D_alg/agg_recursion.py
kwcckw: I see, then we should include a flag for combining purpose.

frame_2D_alg/agg_recursion.py
kwcckw: There's a syntax error because angle can't sum with None when n.angl is None, so a better way is just skip it entirely when n.angl is None.

frame_2D_alg/agg_recursion.py
boris-kz: why do we need data in oF.N_?

frame_2D_alg/agg_recursion.py
boris-kz: ok thanks

frame_2D_alg/agg_recursion.py
boris-kz: actually, we don't need L.H, comp_H or subcomp should directly add to L vals, L_ is single level:

Both consistent with the model:

**1. L_-derived fold** matches "G.L_ becomes G.H[-1]" — links from this round are the new packed level. Replacing N.N_ → L_ in sum_H:

```python
def sum_H(L_, Ft):  # fold this round's L_ as new top level in Ft.H
    if L_: Ft.H += [sum2F(L_, CF())]
```

Called once per cross_comp round, not from every sum2F. Move the call out of sum2F (line 630) and into the cross_comp / cluster_N finalization where L_ is known.

**2. L has no H** — a link is a single comparison, no hierarchy of its own. comp_N's sub-comp currently builds dH-like structure via `L.Nt.fb_`, `L.Lt.fb_`, etc. (lines 318–323), then sum2Fs them into `L.Nt`, `L.Lt`, etc. (lines 324–326), and `add2F(L,Ft)` rolls each up into L's combined dTT/m/d/c/r.

If L has no H, then `[Nt] + Nt.H` in comp_H (line 309) shouldn't apply to L either — comp_H is only meaningful for N-N comparison where both sides have multi-level histories. For L-L comparison (when does that happen? cluster_C? comp_dP_-equivalent?), only the flat dTT compares.

So the cleanup is:

- Drop `n.H = kw.get('H', [])` from CN init? No — N still has H. Just don't populate it on Ls.
- In comp_N, the H block (`if N.typ > 0 and m*wF > ave*(r+cF)`) already gates on `N.typ > 0`, which excludes L (typ=1)... wait, typ=1 *is* > 0. Check: typ=0 PP, typ=1 L, typ=2 G, typ=3 C. So `N.typ > 0` includes Ls. That's probably wrong if Ls shouldn't recurse into H.

Should the gate be `N.typ > 1` (G and C only), or `N.typ != 1` (everything except L)?

Also: with L having no H, `comp_H` in comp_N (line 307) walking `[_Nt] + _Nt.H` — when `_N` is an L, `_N.Nt.H` is empty, so the loop just compares the top dTTs. That works. But then the whole `dH` accumulation reduces to one entry, which is the same info as `comp_derT(_N.dTT[1], N.dTT[1])` — already computed at the top of comp_N. So for Ls, comp_H is redundant.

Want me to draft the edits, or do you want to think through the typ gate first?

frame_2D_alg/agg_recursion.py
```diff
@@ -354,10 +356,10 @@ def comp_F(_F, F, ir=0, rL=None):
             else: Np_ = list(product(_N_,N_))  # pairs
             L = len(Np_)-1
             if np.mean([rL.m,m])* (wF*L) > ave* (r+cF*L):
-                if f: L_= [comp_F(*Np, r,rL=dF) for Np in Np_]; TT,C,R = sum_vt(L_, wTT=ttF)
+                if f: L_= [L for Np in Np_ for L in comp_F(*Np, r,rL=dF).N_]; TT,C,R = sum_vt(L_, wTT=ttF)
                 else: L_,TT,C,R,_,_,_= comp_N_(Np_,r,nF,rL)
                 if L_:
-                    add2F(dF,CF(N_=L_,dTT=TT,c=C,r=R)); add2F(rL,dF); sum2F(L_,CoF.get())
+                    add2F(dF,CF(N_=L_,dTT=TT,c=C,r=R),merge=1); sum2F(L_,CoF.get())  # do we really need this "add2F(rL,dF)" here? This is packing dF to rL.N_, and rL.N_ should be dF.N_ only?
```
boris-kz, line 362: dF should be summed in rL, I think we need that merge=2 to skip dF.N_

frame_2D_alg/agg_recursion.py
```diff
@@ -620,16 +624,19 @@ def sum2F(N_, root=None, m_=[],d_=[], merge=0, froot=0):  # -> CF/CN
...
         F.m, F.d = vt_(TT)
         if fO: F.fw = Fw
     root = add2F(root,F,1) if (root!=None and not fC) else F
     if froot:
         for n in N_: n.root=root
-    sum_H(N_, root)
-    if getattr(root,'C_',None):
-        root.Ct = sum2F(root.C_, CF(root=root))
+     
+    if root and not isinstance(root, CoF):  # not valid for CoF?
```
boris-kz, line 635: this should be in 631?

frame_2D_alg/agg_recursion.py
```diff
@@ -620,16 +624,19 @@ def sum2F(N_, root=None, m_=[],d_=[], merge=0, froot=0):  # -> CF/CN
     F = (CoF if fO else CF)(N_=n_, dTT=TT, c=C, r=R)
     if fC:
         F.typ=3; F.m_,F.d_=m_,d_; F.m,F.d=sum(m_),sum(d_); F.box=box;F.kern=kern;F.span=span;F.yx=yx
-        if angl is not None: F.angl,np.sign(F.dTT[1] @ ttcP[1])
+        if angl is not None: F.angl = [angl, np.sign(F.dTT[1] @ ttcP[1])]
     else:
         F.m, F.d = vt_(TT)
         if fO: F.fw = Fw
     root = add2F(root,F,1) if (root!=None and not fC) else F
```
boris-kz, line 631: why if not fC?

frame_2D_alg/agg_recursion.py
boris-kz: if root: 
        add2F(root, F,2)
        if getattr(root,'C_',None): root.Ct = sum2F(root.C_, CF(root=root))
    if froot:
        for n in N_: n.root=root
    return root

frame_2D_alg/agg_recursion.py
kwcckw: We included that previously to access data from calls? It's just now we have flatten calls that packed in the same N_,

kwcckw:
Sorry, I'm travelling today and not in my desk now, will check again when I back to my desk later.

frame_2D_alg/agg_recursion.py
boris-kz: ok, I will keep call_ for now

frame_2D_alg/agg_recursion.py
boris-kz: remaining L.N_, L.H uses, we probably need to remove them, maybe make links a separate class again?
 
Three live sites, all via the `prop_F_` indirection (`L.N_` → `L.Nt.N_`, `L.H` → `L.Nt.H`):

**1. `comp_H` inside `comp_N`, line 310** — `[_Nt]+_Nt.H`
When `_N` is an L, `_N.Nt.H` is empty, so the loop collapses to one iter on the top dTT. That's redundant with the `comp_derT(_N.dTT[1], N.dTT[1])` already computed at line 317. The block is gated by `N.typ > 0`, which still includes L (typ=1) — should be `N.typ > 1` if Ls are meant to skip H entirely.

**2. `proj_TT`, line 942** — `for lev in L.Nt.N_`
Treats `L.Nt.N_` as a sequence of levels for projection refinement. This is the main place "link.H semantics" actually persists, just stored under `Nt.N_` instead of `H`.

**3. `trans_cluster` (cluster_N), lines 445–450** — walks `L.Nt.N_, L.Bt.N_, L.Ct.N_`
Reads sub-links/sub-nodes accumulated in L's forks during comp_N's sub-comp, then clears them.

**Populating sites** (all in `comp_N`, lines 322–332): `L.Nt.fb_ += …` and `getattr(L,tnF).fb_ += dFt.N_`, then `sum2F(fb_, getattr(L,nF))` writes those into `L.Nt.N_ / L.Lt.N_ / L.Bt.N_ / L.Ct.N_`.

So if you want to fully drop H/N_ on links, the four touch-points are: comp_H gate, comp_N sub-comp population, proj_TT level walk, trans_cluster merge.

Want me to draft the typ-gate fix for comp_H first, or work backward from proj_TT to figure out what should replace the level walk?

frame_2D_alg/agg_recursion.py
kwcckw: We still need dF.N_ for merging purpose in trans_cluster? So dF.N_ are those Ls from trans_comp, we shouldn't skip them?

frame_2D_alg/agg_recursion.py
kwcckw: For Cs, we should append instead of merging Ns when we cluster Ns into C? Else C.N_is Ns' N_ instead of N_.

frame_2D_alg/agg_recursion.py
kwcckw: This is the eval for sum_H, but why sum_H is removed now?

frame_2D_alg/agg_recursion.py
kwcckw: Okay, so L_ is a flatten level now, to pack those additional params, I guess we need to init L as a new class, probably a lesser CN class as suggested above, but without H and Fts?

frame_2D_alg/agg_recursion.py
boris-kz: Sum_H is only good for L, which are not supposed to have H now

PR #131, opened 2026-05-01
5/1/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -633,7 +634,7 @@ def sum2F(N_, root=None, m_=[],d_=[], merge=0, froot=0):  # -> CF/CN
         if getattr(root,'C_',None): root.Ct = sum2F(root.C_, CF(root=root))
     if froot:
         for n in N_: n.root=root
-    return root
+    return F if fF else root  # we may need to return F? I guess we need another flag to return F or root?
```
kwcckw, line 637: So we may need to return F here? In some sum2F, we need to return F and root is only good for froot assignment?

frame_2D_alg/agg_recursion.py
```diff
@@ -621,7 +622,7 @@ def sum2F(N_, root=None, m_=[],d_=[], merge=0, froot=0):  # -> CF/CN
             TT = n.dTT*w; R=n.r*w; n_ = copy(n.N_ if merge else [n])
             if fC: box=copy(n.box); kern=n.kern*w; span=n.span*w; yx=n.yx*w; angl = copy(n.angl[0]) if n.angl is not None else None
             elif fO: Fw = n.m
-    F = (CoF if fO else CF)(N_=n_, dTT=TT, c=C, r=R)
+    F = (CoF if fO else CF)(dTT=TT, c=C, r=R); setattr(F, 'call__' if fO else 'N_', n_)  # call__ to pack flatten call__
```
kwcckw, line 625: So for flatten calls' CF, i guess we can pack it as call__? Then data will be packed separately in oF.N_.

kwcckw:
> actually, we don't need L.H, comp_H or subcomp should directly add to L vals, L_ is single level:

So for new L class, we can skip those Fts and H, but we need to include additional attrs such as trans to pack trans dFts?

frame_2D_alg/agg_recursion.py
```diff
@@ -110,6 +110,7 @@ def __init__(f, **kw):
         super().__init__()
         f.N_ = kw.get('N_',[])  # flat top lev, calls in oF, all sub-forks added conditionally
         f.L_ = kw.get('L_',[])  # +-Ls in levs or cLs in C
+        f.H = kw.get('H',[])    # lower CF levs / Nt||Ct
```
boris-kz, line 113: why do you need this?

frame_2D_alg/agg_recursion.py
boris-kz: why do you need to flatten it here?

frame_2D_alg/agg_recursion.py
```diff
@@ -739,12 +740,13 @@ def add_Lt(G, Lt,wTT):  # addition to Q2R
 def F2N(F):  # convert for cross_comp
 
     Nt = CopyF(F, root=F); Nt.N_ = F.N_; L_ = F.L_  # replace in cross_comp
-    F.__class__ = CN
-    for k,v in dict(H=[],kern=np.zeros(4),span=1,angl=None,mang=1,box=np.array([np.inf,np.inf,-np.inf,-np.inf]),yx=np.zeros(2),sub=0,exe=0,fin=0,nt=[],c_=[]).items():
-        setattr(F,k, copy(v))
-    if L_: F.H += [sum2F(L_,F)]  # level Ls only
+    F.__class__ = CN; F.Nt = Nt;
+    # H is in Nt.H now, we can skip it
```
boris-kz, line 744: we shouldn't have H before effective clustering, which is when F2N is called

frame_2D_alg/agg_recursion.py
boris-kz: use froot instead of fF?

frame_2D_alg/agg_recursion.py
boris-kz, line 107: so links can be CF by default, just add span and angl, and N_=nt?
then convert to CN if they are cross__comped or extended with trans-Ls?

boris-kz: Concept is sound — most Ls are terminal, paying full CN cost is waste. But the "replacing nt with N_" part has a real conflict.

**The conflict:** endpoints and sub-comp output currently occupy separate slots (`L.nt` and `L.Nt.N_/Lt.N_/Bt.N_/Ct.N_`). proj_TT line 942 walks `L.Nt.N_` *as levels*, populated by comp_N's sub-comp block (lines 322–332). If you collapse to a single `L.N_`:

- Endpoints (`[_N, N]`, ordered, exactly 2) and sub-comp products (variable-length, unordered) end up in the same list. Position-2-is-endpoints is fragile.
- Cleaner: keep `nt` for endpoints, use `N_` for sub-comp / trans-Ls. Then "replacing nt with N_" is wrong — they're different slots.

**Promotion triggers, two candidates:**

(a) L gets cross-compared itself → gains rim, compared → CN needed.
(b) comp_N sub-comp block fires → currently splays output across `L.Nt/Lt/Bt/Ct.fb_` → needs forks → CN.

If you also flatten sub-comp into a single `L.N_` (drop the per-fork partitioning, which is the "flat dTT for L-L" position from before), then (b) no longer needs CN, only (a) does. That's a much cleaner promotion line.

But flattening sub-comp drops the four-way fb_ structure that comp_N currently produces. Are you OK losing that? It's the same question as before re: comp_H redundancy on Ls.

**Other things that break:**

- `base_comp` reads `len(N.N_)` as the L attr. Today: `L.N_` → `L.Nt.N_` → empty (0). Tomorrow: `L.N_` → endpoints (2). The pars change. Need to either keep len(N_)=0 for fresh Ls, or accept the shift.
- All `L.nt[0]/[1]` sites (40+) stay as `L.nt` — no rename needed if endpoints stay in nt.
- `L.kern, L.span, L.angl, L.mang` are already on CN; if L becomes CF, those move to a CL subclass. Worth defining `class CL(CF)` rather than tacking onto bare CF (keeps non-link CFs lean).
- `prop_F_` properties on CN (`N_, L_, rim, H` → forks) don't exist on CF, so `L.rim` etc. would fail until promotion. Either CL provides empty defaults, or promotion is mandatory before any rim-touching code.

So the design question reduces to: **is sub-comp output on Ls flat (single L.N_) or partitioned (four forks)?** That decides whether (b) promotes or not, which decides whether most Ls ever get promoted at all.

boris-kz: sub-comp should trigger promotion
we can add nt to CF at link init, along with  kern,span,angl?

Claude
Clean enough. Couple of things to nail down:

**Promotion semantics.** Sub-comp triggers promotion *during* comp_N — but the sub-comp block (lines 319–332) reads `L.Nt.fb_`, calls `getattr(L,tnF)` for the four forks, then does `sum2F([L.Nt,L.Lt,L.Bt,L.Ct], CoF.get())`. None of that exists on CF. So promotion has to happen *before* line 322, not after. Sequence:

```
TT = base_comp(...); L = CL(...)  # CF subclass with nt,kern,span,angl
if N.typ > 1 and m*wF > ave*(r+cF):  # gate raised to typ>1, drops Ls
    dH,htt,C,R = comp_H(...)
    if m + ...:
        F2N(L)  # promote here, before fork access
        # ... existing sub-comp block runs on now-CN L
```

Note the gate raise: `N.typ > 0` → `N.typ > 1` excludes Ls from comp_H entirely, which we agreed is correct since L-L comp is flat dTT only. Then the inner sub-comp block only runs for Gs/Cs anyway, and promotion of *the link being formed* is a separate question — this L is between two N typ>1 things, so it's a meta-link being elevated because its endpoints have rich H.

Wait, that's confused. Re-read: the gate is on `N.typ`, but the *link* L is what's being potentially promoted. Endpoint typ controls whether *their* H is worth comparing. Link promotion is a consequence of comp_H finding enough match. So the trigger is correct — promotion fires when endpoints have comparable H *and* their match clears the sub-comp threshold.

**(a) cross-comp trigger** — separate from sub-comp. When does an L get cross-compared? In `comp_link_`-style flows, or when L is treated as a node in the next round. Both paths need to call `F2N(L)` at entry. cross_comp itself doesn't currently distinguish — it just iterates `N_`. So either:

- cross_comp's first step: `for N in N_: if isinstance(N,CF) and not isinstance(N,CN): F2N(N)`
- or promote at the boundary where L enters cross_comp's input.

Latter is cleaner — promotion stays explicit at the trigger sites.

**CL fields.** Adding `nt, kern, span, angl, mang` to a CF subclass is fine. `mang=1` for fresh Ls, populated only in CN.add_Lt. Don't add `rim, compared, fin, exe, sub` — those are CN-only. Default `H=[]` on CL would let comp_H walk safely without isinstance, but if the gate is raised to typ>1 you don't need that.

**`base_comp` and `len(N.N_)`.** L starts CF with no N_, so `len(L.N_)=0` — stays consistent with current behavior. After promotion, `L.N_` resolves to `L.Nt.N_` via prop_F_ on CN, which sub-comp populates. So len shifts from 0 to non-zero exactly when promotion happens. Fine.

**Promotion idempotence.** F2N currently sets `F.__class__ = CN` and adds attrs. Calling it twice on the same object would double-init — CN attrs would reset. Add an early-return: `if isinstance(F,CN): return F`.

**Ordering quirk.** F2N stashes a Nt copy: `Nt = CopyF(F, root=F); Nt.N_ = F.N_`. For a fresh L with empty `F.N_`, this gives `L.Nt.N_ = []`, which is what comp_N's sub-comp block then writes into via `L.Nt.fb_ += [...]` and `sum2F(fb_, L.Nt)`. Works.

So the changes: (1) define CL with nt/kern/span/angl/mang on top of CF, (2) `comp_N` creates CL not CN at line 318, (3) gate at 319 changes to `N.typ > 1`, (4) `F2N(L)` called before line 322 inside the sub-comp branch, (5) F2N idempotence guard, (6) cross_comp entry promotes any CL it receives.

Anything I'm missing — does L ever get rim before being cross-compared? If yes, promotion has to happen earlier than (6).

boris-kz: we can have CF as a base class for C0F and CL,
then CL is base class for CN,
then F2N forks for F = CF|CL?

Workable. Let me sketch the hierarchy and check it against actual usage:

```
CF       — pure aggregate: N_, dTT, m,d,c,r, root, nF, fb_, wTT
├── CoF  — code fork: + typ_, fw, fc, fr; ContextVar machinery
└── CL   — link base: + nt, kern, span, angl, mang
    └── CN — full node: + rim, compared, H (via prop_F_), Lt/Bt/Ct/Xt/Rt forks, yx, box, sub, exe, fin, etc.
```

**Things that fall out cleanly:**

- `prop_F_` properties (`N_, L_, rim, H`) currently live on CN. They'd stay on CN since CL doesn't have Lt/Bt/etc. forks. Fresh CL has `nt` but no `rim` — code that touches `L.rim` only fires after promotion to CN, which is what you want.
- `comp_N` line 318: `L = CL(typ=1, nt=[_N,N], dTT=TT, ...)`. Cheap.
- `F2N(CL)` upgrades to CN — adds the fork structure, initializes rim/compared/H. The existing F2N already does most of this.

**Things to watch:**

1. **CN inherits CL's `nt, kern, span, angl, mang`.** Currently CN.__init__ sets `kern, span, angl, mang, nt` directly. That overlap is fine if CL.__init__ runs first via super(). But the kwargs flow through both — so CL needs to consume only its own kwargs and pass the rest, or set defaults that CN can override. Cleanest: CL.__init__ sets all CL-fields from kwargs, CN.__init__ calls super and adds CN-only fields. Then nt/kern/span/angl/mang are unified, no duplicate code paths.

2. **F2N idempotence.** After hierarchy change, F2N must distinguish CL→CN (promotion) from CF→CN (different operation: needs to invent nt, kern, etc.). Two functions or one with isinstance branching. The function name F2N suggests one — but CF→CN is rarer than CL→CN. Maybe `L2N(CL) → CN` and keep F2N for the CF→CN path used by sum2G's `F2N(C)` at line 542 and `F2N(C) for C in out_` at 580. Those are centroids being elevated, semantically different from link promotion.

3. **Existing F2N at line 766.** Reads `F.__class__ = CN`. Works for either CL or CF since CN is subclass of both. The added-attrs dict assumes CF-flavored input (nothing about nt/kern). For CL input, nt/kern/span/angl/mang already exist — F2N shouldn't reinit them. Need an isinstance guard: skip those fields if F is CL.

4. **isinstance checks elsewhere.** Any `isinstance(x, CN)` currently means "full node, has rim/H." Stays correct since CL won't pass. Any `isinstance(x, CF)` currently means "aggregate, not a node." That **breaks** — CN now passes `isinstance(CN, CF)` as True. Audit sites: `sum2F` line 689 (`fO = isinstance(N_[0],CoF)` — fine, CoF specific), `Copy_` line 808, `add2F` line 738 (`isinstance(F,CoF)` — fine), and the CoF wrapper. Most existing checks look CoF-specific. Worth grepping.

5. **CL has no Lt/Bt/Ct.** comp_N's sub-comp block (322–332) accesses `L.Nt.fb_, L.Lt.fb_, L.Bt.fb_, L.Ct.fb_`. Promotion has to happen *before* that block — same conclusion as last turn.

So: CF → {CoF, CL}, CL → CN. Promotion CL→CN is the cheap path (most Ls), CF→CN stays for centroid elevation. F2N becomes two functions or one with branching. Audit `isinstance(x, CF)` sites — that's the main risk.

Want to grep the isinstance sites first, or sketch the CL class?

kwcckw: > so links can be CF by default, just add span and angl, and N_=nt?
> then convert to CN if they are cross__comped or extended with trans-Ls?
> we can have CF as a base class for C0F and CL, then CL is base class for CN, then F2N forks for F = CF|CL?


Okay, but i'm not sure if we need the CL base class in CN? Since most of the CN should be init from CN instead of CL?

frame_2D_alg/agg_recursion.py
kwcckw: > we shouldn't have H before effective clustering, which is when F2N is called

Okay, sorry this is not relevant now, so H should be added only in runtime.

frame_2D_alg/agg_recursion.py
kwcckw: This `n_` is the flatten callees, called from CoF:

    if oF.fw*L > ave*(oF.fc*L): sum2F(call_,oF)
    
If we don't assign it here, it will replaces the data from oF.N_ instead.

frame_2D_alg/agg_recursion.py
kwcckw: But froot is used for n.root assignment? We may return F or root for both cases of froot ==1 and froot == 0 too.

frame_2D_alg/agg_recursion.py
boris-kz: something like this:

class CF(CBase):  # clustering fork: rim, Nt,Ct, Bt,Lt: ext|int- defined nodes, ext|int- defining links, Lt/Ft, 
Ct/lev, Bt/G

    name ="fork"  # add sub-forks by F2N
    def __init__(f, **kw):
        super().__init__()
        f.N_ = kw.get('N_',[])  # flat top lev, calls in oF, all sub-forks added conditionally
        f.L_ = kw.get('L_',[])  # +-Ls in levs or cLs in C
        f.nF = kw.get('nF','Nt')
        f.dTT = kw.get('dTT',np.zeros((2,9))); f.m, f.d, f.c, f.r = [kw.get(x,0) for x in ('m','d','c','r')]  # rdpTT in oF?
        f.wTT = kw.get('wTT',wTT); f.w = kw.get('w',0)  # or np.sum(wTT)
        f.fb_ = kw.get('fb_',[])
        f.typ = kw.get('typ',0)  # blocks sub_comp
        f.root = kw.get('root',None)

class CC(CF):  # typ=3, no other forks, Rt,Nt,Bt,Ct from comp_C F2N

    name = "cent"
    def __init__(c, **kw):
        super().__init__(**kw)
        c.m_ = kw.get('m_',[])  # add _m_,_d_,_c_?
        c.d_ = kw.get('d_',[])
        c.c_ = kw.get('c_',[])  # c.r_ = kw.get('r_',[]), c.o_ = kw.get('o_',[])?
        c.yx = kw.get('yx', np.zeros(2))  # comparable means:
        c.kern = kw.get('kern',np.zeros(4))
        c.span = kw.get('span',1)

class CL(CF):  # typ=1, no other forks, Rt,Nt,Bt,Ct from comp_sub F2N

    name = "link"
    def __init__(l, **kw):
        super().__init__(**kw)
        l.nt = kw.get('nt',[])  # nodet
        l.yx = kw.get('yx', np.zeros(2))  # mean nodet?
        l.kern = kw.get('kern',np.zeros(4))  # I,G,A diffs in links
        l.span = kw.get('span',1)  # distance in nodet or aRad, comp with kern or len(N_)
        l.angl = kw.get('angl',None)  # (dy,dx),dir, sum from L_, rarely?

def prop_F_(F, attr='N_'):  # factory function to get and update top-composition fork.N_|H

    def get(N): return getattr(getattr(N,F), attr)
    def set(N, new_val): setattr(getattr(N,F), attr, new_val)
    return property(get,set)

class CN(CL):  # full node | graph attr set
 
   name = "node"  # ext|int -defined Ns,Ls:
    N_,C_,L_,B_,X_,rim,H = prop_F_('Nt'),prop_F_('Ct'),prop_F_('Lt'),prop_F_('Bt'),prop_F_('Xt'),prop_F_('Rt'),prop_F_('Nt','H')
    def __init__(n, **kw):
        super().__init__(**kw)
        n.root = kw.get('root',None)  # immediate
        n.mang = kw.get('mang',1)  # ave match of angles in L_, =1 in links
        n.box  = kw.get('box',np.array([np.inf, np.inf, -np.inf, -np.inf]))  # y0, x0, yn, xn
        n.c_ = kw.get('c_',[])  # reciprocal root Cs
        n.sub = 0  # composition depth relative to top-composition peers?
        n.compared = set()
        # ftree: list =z([[]])  # indices in all layers(forks, if no fback merge, G.fback_=[] # node fb buffer, n in fb[-1]
    def __bool__(n): return bool(n.c)

kwcckw: Okay, so the base class is just for new class initialization, most of the time we don't need to extend CL into CN unless we cross compare it.

boris-kz: yes, so now we have a class for each main fork: node, link, cent

frame_2D_alg/agg_recursion.py
boris-kz: ok, right

frame_2D_alg/agg_recursion.py
boris-kz: typ tracks construction complexity:

```
CF (bare aggregate)        typ=0
CL (link, +geom)           typ=1
CC (centroid, +arrays)     typ=2
CN (node, +forks/rim/H)    typ=3
```

CC's complexity ≈ CL's (sibling additions to CF, not a chain), but giving it 2 still works as a complexity proxy because m_/d_/c_ arrays add per-member cost beyond CL's fixed-size geom.

**Implications for existing typ checks:**

- `N.typ > 0` → was "exclude PP" (PP=0). Now: excludes only bare CF. If PPs become CL (typ=1) they pass the gate. If PPs stay CF-equivalent (typ=0), still excluded. Need to decide PP's class.
- `N.typ > 1` → was the H-walk gate, intent "G and C only" (G=2, C=3). Now: typ>1 selects CC and CN, i.e. centroid and node. CL excluded. That matches "Ls don't recurse into H" — actually correct under new numbering.
- `N.typ == 1` → was "is link" (L=1). Still correct (CL=1).
- `N.typ == 3` → was "is centroid" (C=3). **Now means CN**, not CC. Breaks centroid identification.
- `N.typ == 2` → was "is G". Now means CC. Breaks G identification — but G doesn't really exist as a separate class anymore, every full node is CN.
- `N.typ == -1` → pre-link (comp_N_ line 248). Still works, untouched.
- `N.typ == 0` → vect_edge's PP2N currently sets typ=0 on a CN. Under new numbering, "typ=0 on CN" is contradictory (CN should be typ=3). Either PP2N changes to make PPs CL/CC/CN with appropriate typ, or typ=0 is overloaded as "PP marker on CN."

**Sites needing audit** (grep `\.typ\b`):

- comp_N line 319: `N.typ > 0` — now lets through CL/CC/CN, blocks bare CF/PP. If PPs are CN-with-typ=0, they're blocked — probably wrong since PPs should walk H. Was originally intended to walk H for everything except links. Now walks H for everything except CL. CC (centroid) walks H — does C have H? CC has no `H` attr until F2N(CC)→CN. Reading `_Nt.H` on a CC-not-promoted would AttributeError unless CC has H=[].
- comp_N line 322: `if abs(N.typ) == 1` — selects L and pre-link L. Still correct.
- sum2G line 698: `G.typ = N.typ + 1 if G.L_ else N.typ` — incremented per sub-level. Under complexity numbering this means "deeper composition = higher complexity". Possibly OK, but this typ inc is independent of class hierarchy — a CN with typ=4,5,... has no class meaning. Was the original intent.
- sum2G line 706: `if N.typ != 1` — exclude L. Still correct.
- cluster_C line 605: `(n.d if n.typ==1 else n.m)` — L vs non-L. Still correct.

**Real conflict:** typ=3 used to mean centroid, now means node. cluster_C line 542 `F2N(C)` — F2N currently sets typ=2 on output (`typ=2 if init<2 else N.typ`). After renumber, F2N for "G" output should be typ=3. And for centroid identification, you need a different mechanism.

**Suggestion:** if typ is purely complexity, identify *role* by `isinstance`:
- `isinstance(N, CC)` for centroid
- `isinstance(N, CL) and not isinstance(N, CN)` for unpromoted link  
- `isinstance(N, CN)` for full node

Then `N.typ` is just complexity, and code reads class for role. Cleaner separation.

Worth the audit, or do you want to keep typ doing both jobs?

frame_2D_alg/agg_recursion.py
boris-kz: how about this:

    if root:
        add2F(root, F,2)
        if getattr(root,'C_',None): root.Ct = sum2F(root.C_, CF(root=root))
    else: root=F
    if froot:
        for n in N_: n.root=root
    return root

boris-kz: with new typs and classes:

    typ = N_[0].typ; cls_= [CF,CL,CC,CN,CoF]
        if i:
            TT+=n.dTT*w; R+=n.r*w; n_ += (n.N_ if merge else [n])
            if typ:  # links and higher
                if n.angl is not None : angl += n.angl[0]  # mostly Ls?
                if typ==3: box=extend_box(box,n.box)  # CN
                if typ==4: Fw += n.m  # oF
        else:  # init with N_[0]
            if typ:
                if typ==3: box = copy(n.box)  # CN
                if typ==4: Fw = n.m  # oF
    F = (cls_[typ])(typ=typ, dTT=TT, c=C, r=R)  # keep typ?
    if typ:
        F.kern=kern;F.span=span;F.yx=yx; F.angl = [angl, np.sign(F.dTT[1] @ ttcP[1])] if angl is not None else None
        if typ==2: F.m_,F.d_=m_,d_; F.m,F.d=sum(m_),sum(d_)
        else:      F.m, F.d = vt_(TT)
        if typ==3: F.box = box
        if typ==4: F.call_=n_; F.fw = Fw
        else:      F.N_=n_
    if root:
        add2F(root,F,1)
        if typ==3 and getattr(root, 'C_', None): root.Ct = sum2F(root.C_, CF(root=root))
    else: root = F
    if froot:
        for n in N_: n.root = root
    return root

frame_2D_alg/agg_recursion.py
kwcckw: Okay, so higher typ means more extended attrs with higher complexity, this is actually cleaner. But the downside is we may need to change all of the higher complexity's type if we need to add any new lower complexity type.

frame_2D_alg/agg_recursion.py
kwcckw: This won't work because we may have need to return F when root is parsed, for example:

    root.H += [sum2F(L_,root)]  

We need to return F as level，and root is parsed to accumulate params?0

boris-kz: Yes. Audit of `sum2F` callers and what they do with the return:

**Use the return:**
- `cluster_C`: `C = sum2F(n_, Ft, m_, d_)` — needs new centroid (typ==2) ✓ handled
- `cluster_P`: `C = sum2F(N_, root, m_, d_)` — same ✓
- `cluster_N`: `Ft.root.Ct.Lt = sum2F([L for C in C_ for L in C.L_], root=Ft.root.Ct)` — wants the new Lt fork F, not the root it was folded into
- `cross_comp`: `root.Nt = sum2F(G_, CF(root=root), froot=1)` — wants new F (the CF passed in is just a parent ref)
- `comb_Ft` via `T.dTT,T.c,T.r = sum_vt([T,Ft],...)` — not sum2F
- `vect_edge`: `tile.Nt = ...` no, that's CF directly. But `sum2F(PPd_, N.Bt)` then `N.Bt.N_ = PPd_` — uses N.Bt which is the root, OK
- `frame_H`: `F.H += [lev := sum2F(tile_, CF())]` — wants new F
- `sum2G` calls `comb_Ft` which doesn't use return... but `sum2F([F2N(C) for C in out_], Ft.root.Ct)` in cluster_C — return ignored

**Ignore the return:**
- `get_exemplars`: `sum2F(E_, CoF.get())` — side effect only
- `comp_N_`, `comp_C_`: `sum2F(L_, CoF.get())` and `sum2F(N.rim, N.Rt)` — side effect

So the pattern is: when root is a **container being populated** (CoF call tree, N.Rt rim sum), caller ignores return. When root is a **parent ref for the new F** (CF(root=root), Ft.root.Ct, etc.), caller wants F.

Cleanest rule: **always return F.** The callers that ignore the return don't care; the callers that use it want F.

```python
if root and typ != 2:
    add2F(root, F, 1)
    if typ==3 and getattr(root,'C_',None):
if froot:
return F
```

This matches the *original* sum2F semantics too — original returned `root` only because `root = add2F(...) if ... else F` rebinds root to F when no add2F happens. The downstream callers like `lev := sum2F(tile_, CF())` work in the original because root was a fresh CF and add2F returns it after folding F in — so they get a CF that *contains* F as its sole N_ entry, not F itself.

Wait — that's a real semantic difference. Let me re-check `frame_H`:

```python
F.H += [lev := sum2F(tile_, CF())]
```

Original: `CF()` is root, `add2F(root, F, 1)` appends F to root.N_ and returns root. So `lev` is the wrapper CF with `lev.N_ = [F]`. If I switch to "always return F", `lev` becomes F directly, and tile_ ends up at lev.N_ instead of lev.N_[0].N_. Different shape.

So "always return F" is **not** a drop-in replacement. The safer rule:

```python
return F if typ==2 else (root or F)
```

Centroid returns the new C (fixes cluster_C/_P). Everything else returns root if root was passed (preserves original wrapper semantics), F if not.

boris-kz: def sum2F(N_, root=None, m_=[],d_=[], merge=0, froot=0):  # -> CF/CL/CC/CN/CoF
 
    typ = N_[0].typ; cls_= [CF,CL,CC,CN,CoF]
        if i:
            TT+=n.dTT*w; R+=n.r*w; n_ += (n.N_ if merge else [n])
            if typ and typ < 4:  # CL/CC/CN
                if n.angl is not None: angl += n.angl[0]
                if typ==3: box=extend_box(box,n.box)
            elif typ==4: Fw += n.m
        else:  # init with N_[0]
            if typ and typ < 4:
                if typ==3: box=copy(n.box)
            elif typ==4: Fw = n.m
    F = cls_[typ](dTT=TT, c=C, r=R)
    if typ:
        if typ < 4:
        if typ==2: F.m_,F.d_=m_,d_; F.m,F.d=sum(m_),sum(d_)
        else:      F.m, F.d = vt_(TT)
        if typ==3: F.box = box; F.N_ = n_
        elif typ==4: F.call_ = n_; F.fw = Fw
        else: F.N_ = n_
    else:
        F.N_ = n_; F.m,F.d = vt_(TT)
    if root and typ != 2:  # don't fold centroid into root
        add2F(root, F, 1)
        if typ==3 and getattr(root,'C_',None):
    if froot:
    return F if typ==2 else (root or F)

kwcckw: This `return F if typ==2 else (root or F)` is still wrong for `root.H += [sum2F(L_,root)]  `?

When we have add2F: `add2F(root, F, 1)` , root is filled, so root will be returned instead of F?

boris-kz: Sorry, I haven't checked that yet.
Right now the bigger issue is refactoring with those new classes, including F2N

kwcckw: Okay, i will add the relevant fix first. For the classes, i think we need to check most of the eval for typ, and also the related extension when we convert CL to CN, or CC to CN.

boris-kz: > This `return F if typ==2 else (root or F)` is still wrong for `root.H += [sum2F(L_,root)] `?
> 
> When we have add2F: `add2F(root, F, 1)` , root is filled, so root will be returned instead of F?

Opus 4.7:

Yes, you're right. `add2F(root,F,1)` mutates root in place — so `root or F` returns `root` (the mutated container), but `root.H += [sum2F(L_,root)]` wants the new lev F, not root itself (which would be self-referential).

Auditing all call sites that use the return value:
- `C = sum2F(n_, Ft, m_, d_)` (cluster_C) — wants F (centroid) ✓ already returns F
- `T = sum2F(F_,CoF())` (add_typ_) — wants F
- `Ct = sum2F([F2N(C) for C in out_], Ft.root.Ct)` (cluster_C) — wants F
- `Ft.root.Ct.Lt = sum2F(L_for_Cs, Ft.root.Ct)` (cluster_N) — wants F
- `root.H += [sum2F(L_,root)]` (cross_comp) — wants F
- `root.Nt = sum2F(G_,CF(root=root),froot=1)` (cross_comp) — wants F
- `lev := sum2F(tile_, CF())` (frame_H) — wants F
- `dCt = sum2F(..., CF())` (cluster_P) — wants F
- `root.Ct = sum2F(root.C_, CF(root=root))` (sum2F itself) — wants F

Container-fill callers that ignore return (`sum2F(N.rim, N.Rt)`, `sum2F(L_, CoF.get())`, etc.) only care about the add2F side effect.

So `return F` always is right:One note: with `return F`, the `sum2F(L_, root)` call in cross_comp still has the side effect of folding L_ into root.N_ via add2F. If that's undesired (root is a graph, its N_ should be sub-graphs not links), the call should be `root.H += [sum2F(L_)]` — drop the root arg. Want me to check that line?

kwcckw: Thanks, so i guess we can return F, and we actually only need to return root here:

    Ct = sum2F([F2N(C) for C in out_], Ft.root.Ct) (cluster_C) — wants F
    
Which can be assigned separately.

boris-kz: That was opus 4.7, are you using it?

kwcckw: Yes, i shifted to 4.7 recently, but looks like the prior 4.6 version thinks deeper and somehow the latest version seems "lazier".

boris-kz: It's faster, not quite sure the replies are better, but easier to guide along

kwcckw: I see, probably that's the reason why it doesn't think deeper. I'm still observing how it's better than the prior 4.6 version.

PR #132, opened 2026-05-02
5/2/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -70,8 +70,8 @@ class CF(CBase):  # clustering fork: rim, Nt,Ct, Bt,Lt: ext|int- defined nodes,
     name ="fork"  # add sub-forks by F2N
     def __init__(f, **kw):
         super().__init__()
-        f.N_ = kw.get('N_',[])  # flat top lev, calls in oF, all sub-forks added conditionally
-        f.L_ = kw.get('L_',[])  # +-Ls in levs or cLs in C
+        if not hasattr(f, 'N_'): f.N_ = kw.get('N_',[])  # flat top lev, calls in oF, all sub-forks added conditionally
+        if not hasattr(f, 'L_'): f.L_ = kw.get('L_',[])  # +-Ls in levs or cLs in C
```
kwcckw, line 74: So there's some compatibility issue when we have CF as base class of CN because this may overwrites the existing N_ init from CN, so we need to skip if the attr is already available.

frame_2D_alg/agg_recursion.py
```diff
@@ -109,10 +109,11 @@ class CN(CL):  # full node | graph fork set
     name = "node"  # ext|int -defined Ns,Ls:
     N_,C_,L_,B_,X_,rim,H = prop_F_('Nt'),prop_F_('Ct'),prop_F_('Lt'),prop_F_('Bt'),prop_F_('Xt'),prop_F_('Rt'),prop_F_('Nt','H')
     def __init__(n, **kw):
+        # Ft needs before init because f.N_ = kw.get('N_',[]) trigger setter on Nt, which is not available yet
+        n.Nt,n.Bt,n.Ct,n.Lt,n.Xt,n.Rt = ((kw.get(f) if f in kw else CF(root=n) for f in ('Nt','Bt','Ct','Lt','Xt','Rt')))  # CN if nest, Ct||Nt
```
kwcckw, line 113: The `super().__init__(**kw)` below runs CF's init, which calls f.N_, and that triggers `prop_F` to get `N_` from Nt. That's causing error if we don't init Nt first since Nt.N_ is unavailable when we trigger N_ from CF init.

frame_2D_alg/agg_recursion.py
```diff
@@ -828,7 +831,7 @@ def PP2N(PP):
                          np.array([dM, dD, dL, dI, dG, dA, dL, dL / 2, 0])])
         y,x,Y,X = box; dy,dx = Y+1-y, X+1-x
         A = [np.array(A), np.sign(dTT[1] @ ttVct[1])]  # append sign
-        PP = CN(typ=0, dTT=dTT,m=m,d=d,c=c,r=1, kern=kern,box=box,yx=yx,angl=A,span=np.hypot(dy/2,dx/2))  # set root in trace_edge
+        PP = CN(typ=3, dTT=dTT,m=m,d=d,c=c,r=1, kern=kern,box=box,yx=yx,angl=A,span=np.hypot(dy/2,dx/2))  # set root in trace_edge
```
kwcckw, line 834: For PP, their typ should be 3 since it will be compared the same way as CN?

boris-kz: I think it's 1, it's only base_comp attrs?

kwcckw: But we may still need to add some attrs such as yx, rim? We may need those in comp_N , called within trace_edge.

boris-kz: 1 has all the attrs you set here, except for box which we don't need.
it doesn't have really matter, but we need to skip sub-comp.

frame_2D_alg/agg_recursion.py
boris-kz: isn't it the other way around, the higher class overwrites the base class?

frame_2D_alg/agg_recursion.py
```diff
@@ -322,10 +323,11 @@ def comp_H(_Nt,Nt, Link):  # tentative pre-comp
         return dH,tt,C, (r* Link.c+R)/ (Link.c+C)  # same norm for tt?
 
     TT= base_comp(_N,N)[0] if full else comp_derT(_N.dTT[1],N.dTT[1]); m,d = vt_(TT, ttN_)
-    L = CN(typ=1, nt=[_N,N], dTT=TT,m=m,d=d,c=_c,r=r, root=rL, exe=1)
-    if N.typ > 0 and m*wF > ave*(r+cF):
+    L = CL(typ=1, nt=[_N,N], dTT=TT,m=m,d=d,c=_c,r=r, root=rL, exe=1)
+    if N.typ and m*wF > ave*(r+cF):  # compares L, C and N only? Skip CF
```
boris-kz, line 327: N only, the rest have no sub-structures, that's why PP should be 1

frame_2D_alg/agg_recursion.py
```diff
@@ -699,7 +701,7 @@ def sum2G(ft_, root=None, init=1, typ=None):
                 cross_comp(G.Nt,r)
     if G.Bt:
         Bt = G.Bt; bd,br,L = Bt.d,Bt.r,len(Bt.N_); rroot = root.root if root.root else 0
-        if N.typ!=1 and bd*(wX*L) > avd*(br+cX*L): cross_comp(Bt, br)  # no ddfork
+        if N.typ!=1 and bd*(wX*L) > avd*(br+cX*L): [F2N(L) for L in Bt.N_]; cross_comp(Bt, br)  # no ddfork
```
boris-kz, line 704: F2N should have forks for typs 0,1,2

frame_2D_alg/agg_recursion.py
boris-kz: Pulled the diff and 5 comment threads. Summary of what's in PR #132 and where I'd push back:

**Type renumbering applied:** `1=L (and PP per your call), 2=C, 3=N`. Mostly consistent but a few sites still mismatched.

**Your anchor comment (r3176549246, "F2N should have forks for typs 0,1,2"):**

The real bug F2N has now: the setattr loop unconditionally writes defaults for `kern, span, angl, yx, nt` — but a CL already has all of those populated from `comp_N`. So `F2N(L)` clobbers L's positional/kernel data with zeros. Same risk for CC. The fix is what you flagged — branch on source typ:

- typ=1 (CL): preserve `kern, span, yx, angl, nt`; add only `H, mang, box, sub, exe, fin, c_, compared` and the fork tree.
- typ=2 (CC): same as CL plus preserve `m_, d_, c_`.
- typ=0 (bare CF): add all of the above.

So roughly:

```python
def F2N(F):
    Nt = CopyF(F, root=F); Nt.N_ = F.N_; L_ = F.L_
    F.__class__ = CN
    Na_ = dict(H=[],mang=1,box=np.array([np.inf,np.inf,-np.inf,-np.inf]),sub=0,exe=0,fin=0,c_=[])
    if F.typ == 0:  # CF -> add all
        Na_.update(kern=np.zeros(4),span=1,angl=None,yx=np.zeros(2),nt=[])
    for k,v in Na_.items(): setattr(F,k, copy(v))
    if L_: F.H += [sum2F(L_,F)]
    F.Nt = Nt; F.compared = set()
    return F
```

**Other issues from the diff:**

1. **`if N.typ and m*wF > ave*(r+cF)`** in comp_N (your comment r3176534662 said "N only"): this still admits CL (typ=1) and CC (typ=2). Should be `if N.typ == 3`. PP at typ=1 is correctly excluded.

2. **PP2N still has `typ=3`** (kwcckw's edit). You said "I think it's 1" in r3176389463. Inconsistent with your final call. Either revert to 1 or update the comment.

3. **`F2N(L)` in comp_N before sub-comp**: L was just made `CL(typ=1)`. F2N promotes it to CN with fork tree — that's fine, sub-comp needs `L.Nt.fb_`. But same clobbering bug as above hits L's freshly-set `kern, span, angl, yx, nt`. Fix #1 above resolves this too.

4. **Side-effect leak in `Ft.root.Ct.Lt = sum2F([...], root=Ft.root.Ct)`** (cluster_N, unchanged in PR): with `return F` always, sum2F still calls `add2F(Ft.root.Ct, F, 1)` — folds the new lev into `Ft.root.Ct.N_` as a side effect, then you assign the same F to `.Lt`. Probably not what you want. The cluster_C site `Ct = sum2F([F2N(C)...]); Ft.root.Ct = Ct` was fixed by dropping root; this one still has the issue.

**Re your reply r3176527094** ("isn't it the other way around, the higher class overwrites the base class?"):

kwcckw's guard is correct, your intuition is reversed for this case. `super().__init__(**kw)` *delegates downward* to CF, so CF.__init__ runs **during** CN construction, in the CN instance's context. When CF does `f.N_ = []`, on a CN instance that triggers `prop_F_('Nt')`'s setter, which writes to `f.Nt.N_`. So CF's init can overwrite what CN already set. Order in current CN:

```python
n.Nt,n.Bt,... = ...   # subclass sets first
super().__init__(**kw)  # then base runs and would overwrite via prop_F_
```

Without the `hasattr` guard in CF, super() would clobber `n.Nt.N_`. The guard correctly skips. The "higher overwrites base" rule applies to method *resolution*, not to the order of writes during chained inits.

frame_2D_alg/agg_recursion.py
```diff
@@ -322,10 +323,11 @@ def comp_H(_Nt,Nt, Link):  # tentative pre-comp
         return dH,tt,C, (r* Link.c+R)/ (Link.c+C)  # same norm for tt?
 
     TT= base_comp(_N,N)[0] if full else comp_derT(_N.dTT[1],N.dTT[1]); m,d = vt_(TT, ttN_)
-    L = CN(typ=1, nt=[_N,N], dTT=TT,m=m,d=d,c=_c,r=r, root=rL, exe=1)
-    if N.typ > 0 and m*wF > ave*(r+cF):
+    L = CL(typ=1, nt=[_N,N], dTT=TT,m=m,d=d,c=_c,r=r, root=rL, exe=1)
```
boris-kz, line 326: we don't need typ and exe?

frame_2D_alg/agg_recursion.py
kwcckw: Yes, it's actually should be the other way around, so in lower base class, we should skip init of N_ and L_ when typ is 3 (called from CN's superinit), else it gets error when it tries to access Nt.N_.

frame_2D_alg/agg_recursion.py
kwcckw: Okay right, so after the conversion via F2N, their typ should be updated right?

frame_2D_alg/agg_recursion.py
kwcckw: We need to convert CC and CL to CN only, and since CC inherit from CL, and CC's attrs is unique to C only, so both of the classes should add the same N's attrs?

frame_2D_alg/agg_recursion.py
kwcckw: Okay right, i'm not sure why we need exe = 1 here, it should be not relevant now.

frame_2D_alg/agg_recursion.py
boris-kz: sorry, we still to eval sub-comp for N.N_ or N.nt, except for PPs. so PP.typ = 0, despite added attrs:

    if N.typ==3 and m*wF > ave*(r+cF):
        dH,htt,C,R = comp_H(_N,N, L); r+=R; m+= vt_(htt,ttN_)[0]  # tentative dH, replace if subcomp?
    if N.typ and m*cN_ > ave*(r+cN_):  # skip PPs 
        F2N(L)  # add tFs for subcomp
        if abs(N.typ)==1:
            for _n,n in product(_N.nt,N.nt): L.Nt.fb_ += [comp_N(_n,n,r,_c)]  # link sub-comp
        else:
            for i,(_Ft,Ft,tnF) in enumerate(zip((_N.Nt,_N.Lt,_N.Bt,_N.Ct),(N.Nt,N.Lt,N.Bt,N.Ct),('Nt','Lt','Bt','Ct'))):
                if _Ft and Ft:  # sub-comp
                    dFt = comp_F(_Ft,Ft,r,L); getattr(L,tnF).fb_ += dFt.N_  # tFt feedback
            if fb_:  # L+= trans-Ls, python-batched bottom-up:
                sum2F(fb_,getattr(L,nF))
        sum2F([L.Nt,L.Lt,L.Bt,L.Ct], CoF.get())

boris-kz: > Okay right, so after the conversion via F2N, their typ should be updated right?

no, I think just class

frame_2D_alg/agg_recursion.py
boris-kz: no, there's an issue with converting existing attrs, so we need to check typ, not class

frame_2D_alg/agg_recursion.py
boris-kz: typ doesn't have to match class:
PP = CL(typ=0, dTT=dTT,m=m,d=d,c=c,r=1, kern=kern,yx=yx,angl=A,span=np.hypot(dy/2,dx/2))

frame_2D_alg/agg_recursion.py
boris-kz: def comp_N(_N,N, r,_c, full=1, A=np.zeros(2),span=None, rL=None, L_=None, N_=None, acc=None):

    def comp_H(_Nt,Nt, Link):  # tentative pre-comp
        dH, tt, C,R = [],np.zeros((2,9)),0,0
        for _lev, lev in zip([_Nt]+_Nt.H, [Nt]+Nt.H):  # should be top-down
            ltt = comp_derT(_lev.dTT[1],lev.dTT[1])
            lc = min(_lev.c,lev.c); lr = (_lev.r+lev.r)/2; m,d = vt_(ltt,ttN_)
            dH += [CF(dTT=ltt,m=m,d=d,c=lc,r=lr,root=Link)]
            tt += ltt*lc; C+=lc; R+=lr*lc

    L = CL(nt=[_N,N], dTT=TT,m=m,d=d,c=_c,r=r, root=rL)
        dH,htt,C,R = comp_H(_N,N, L); r+=R; m+= vt_(htt,ttN_)[0]  # tentative dH, replace if subcomp?
    if N.typ and m*wN_ > ave*(r+cN_):  # skip PPs
        F2N(L)  # add tFs for subcomp
        if N.typ==1:
            for _n,n in product(_N.nt,N.nt): L.Nt.fb_ += [comp_N(_n,n,r,_c)]  # link sub-comp
        else:
            for i,(_Ft,Ft,tnF) in enumerate(zip((_N.Nt,_N.Lt,_N.Bt,_N.Ct),(N.Nt,N.Lt,N.Bt,N.Ct),('Nt','Lt','Bt','Ct'))):
                if _Ft and Ft:  # sub-comp
                    dFt = comp_F(_Ft,Ft,r,L); getattr(L,tnF).fb_ += dFt.N_  # tFt feedback
                    r+= (i or 1) -1  # Nt,Lt are core, not redundant
            if fb_:  # L+= trans-Ls, python-batched bottom-up:
                sum2F(fb_,getattr(L,nF))
        sum2F([L.Nt,L.Lt,L.Bt,L.Ct], CoF.get())
    if full:
        angl = [A, np.sign(TT[1] @ ttN_[1])]
    for n, _n in (_N,N),(N,_N):
    if L_ is not None:
        N_+= [_N,N]  # for sum_vt(rim)
        elif L.typ==1 and L.d > avd: acc[3]+=L.dTT*L.c; acc[4]+=L.c; acc[5]+=L.r*L.c  # no pLs?
    return L

frame_2D_alg/agg_recursion.py
kwcckw: Yes, but both typs should add the same N's attrs?

frame_2D_alg/agg_recursion.py
kwcckw: > no, I think just class

Okay, i guess that's better since we still can differentiate them with typ.

frame_2D_alg/agg_recursion.py
boris-kz: the question is what to do with existing attrs

kwcckw: The existing attrs should remain the same? Why we need to modify the existing attrs?  

If we look at CC and CL: 


      class CL(CF):  # typ=1, add kern+positionals for base comp, Rt,Nt,Bt,Ct from comp_sub F2N
          name = "link"
          def __init__(l, **kw):
              super().__init__(**kw)
              l.nt = kw.get('nt',[])  # nodet
              l.yx = kw.get('yx', np.zeros(2))  # mean nodet? comp box is not meaningful?
              l.kern = kw.get('kern',np.zeros(4))  # I,G,A diffs in links
              l.span = kw.get('span',1)  # distance in nodet or aRad, comp with kern or len(N_)
              l.angl = kw.get('angl',None)  # (dy,dx),dir, sum from L_, rarely?
              l.typ  = kw.get('typ',1)  # blocks sub_comp
      
      class CC(CL):  # typ=2, adds arrays per N_
          name = "cent"
          def __init__(c, **kw):
              super().__init__(**kw)
              c.m_ = kw.get('m_',[])  # add _m_,_d_,_c_?
              c.d_ = kw.get('d_',[])
              c.c_ = kw.get('c_',[])  # c.r_ = kw.get('r_',[]), c.o_ = kw.get('o_',[])?
              c.typ = kw.get('typ',2)
              
  CC should have all the existing CL attrs, and `m_, d_, c_` is only unique to C. So both should add the same additional CN attrs?
  
      class CN(CL):  # full node | graph fork set
        name = "node"  # ext|int -defined Ns,Ls:
        N_,C_,L_,B_,X_,rim,H = prop_F_('Nt'),prop_F_('Ct'),prop_F_('Lt'),prop_F_('Bt'),prop_F_('Xt'),prop_F_('Rt'),prop_F_('Nt','H')
        def __init__(n, **kw):
            # Ft needs before init because f.N_ = kw.get('N_',[]) trigger setter on Nt, which is not available yet
            super().__init__(**kw)
            n.c_ = kw.get('c_',[])  # reciprocal root Cs
            n.mang = kw.get('mang',1)  # ave match of angles in L_, =1 in links
            n.box  = kw.get('box',np.array([np.inf, np.inf, -np.inf, -np.inf]))  # y0, x0, yn, xn
            n.sub = 0  # composition depth relative to top-composition peers?
            n.compared = set()
            # ftree: list =z([[]])  # indices in all layers(forks, if no fback merge, G.fback_=[] # node fb buffer, n in fb[-1]
        def __bool__(n): return bool(n.c)

boris-kz: their values should be preserved, not reset to default

kwcckw: We only need to preserve the overlapping attrs' values, and the only overlapping attrs should be just N_ and L_ between CF and CN? All the other attrs should be init to default when we convert their class. For N_ and L_, we can check if they are available and assign N_ and L_ to Nt.N_ and Lt.N_ when we convert them.

boris-kz: If we convert from CL, it will have valid yx,kern, etc. They should not be replaced with defaults

boris-kz: def F2N(F):  # convert for cross_comp

    Nt = CopyF(F, root=F); Nt.N_=F.N_; L_=F.L_  # replace in cross_comp
    F.__class__ = CN; F.Nt = Nt
    Na_ = dict(H=[], mang=1, box=np.array([np.inf,np.inf,-np.inf,-np.inf]), sub=0, exe=0, fin=0, c_=[], compared = set())
    if F.typ==0:  # CF | PP, no overlap for Cs
        Na_.update(kern=np.zeros(4), span=1, angl=None, yx=np.zeros(2))
    for k,v in Na_.items(): setattr(F, k, copy(v))
    if L_: F.H += [sum2F(L_, F)]
    return F

kwcckw: > If we convert from CL, it will have valid yx,kern, etc. They should not be replaced with defaults

Yes, they will not be replaced since those attrs are not in CN's __init__:

      class CN(CL):  # full node | graph fork set
          name = "node"  # ext|int -defined Ns,Ls:
          N_,C_,L_,B_,X_,rim,H = prop_F_('Nt'),prop_F_('Ct'),prop_F_('Lt'),prop_F_('Bt'),prop_F_('Xt'),prop_F_('Rt'),prop_F_('Nt','H')
          def __init__(n, **kw):
              # Ft needs before init because f.N_ = kw.get('N_',[]) trigger setter on Nt, which is not available yet
              super().__init__(**kw)
              n.c_ = kw.get('c_',[])  # reciprocal root Cs
              n.mang = kw.get('mang',1)  # ave match of angles in L_, =1 in links
              n.box  = kw.get('box',np.array([np.inf, np.inf, -np.inf, -np.inf]))  # y0, x0, yn, xn
              n.sub = 0  # composition depth relative to top-composition peers?
              n.compared = set()
              # ftree: list =z([[]])  # indices in all layers(forks, if no fback merge, G.fback_=[] # node fb buffer, n in fb[-1]
          def __bool__(n): return bool(n.c)

boris-kz: ok, then if we are upgrading CF, we need to add CL attrs + CN attrs, but for CL it's just CN attrs?

kwcckw: > ok, then if we are upgrading CF, we need to add CL attrs + CN attrs, but for CL it's just CN attrs?

Yes, but we are not really converting CF to CN now? I can see we only need to convert CC or CL to CN.

boris-kz: I think there should be an option to compare Bt in sub-comp too?

kwcckw: That's in comp_F?

But we only need to compare Ft's dTT:

    ddTT = comp_derT(_F.dTT[1],F.dTT[1]); r=(_F.r+F.r)/2; m,d = vt_(ddTT,ttF); r+=ir
    
    
 Or do you mean the sub+ section in sum2G?
 
       if G.Bt:
              if N.typ!=1 and bd*(wX*L) > avd*(br+cX*L): [F2N(L) for L in Bt.N_]; cross_comp(Bt, br)  # no ddfork
              if rroot: Bt.brrw = Bt.m * (rroot.m * (decay * (rroot.span/G.span)))  # external lend only, subtract from root?
              
That's comparing Bt.N_, which is CLs?

boris-kz: sorry, it's not relevant. so yes, CF is currently not promoted.
I would keep it just in case we need to

boris-kz: in cross_comp:

            if G_:
                if isinstance(root,CF): F2N(root)  # upgrade @ 1st sub|agg+


Tracing call sites of `cross_comp`:

1. **`cross_comp(root.Nt, r)`** in cross_comp itself (agg+ recursion) — `root.Nt` is a CF (sub-fork container). **CF.**

2. **`cross_comp(Ct, r)`** in cluster_C, where `Ct = sum2F([F2N(C) for C in out_])` — sum2F returns the new F it builds. `out_` elements are CCs that got F2N'd to CNs. typ matches the input N_[0].typ which is now 3, so sum2F returns a CN. **CN.**

3. **`cross_comp(Bt, br)`** in sum2G — `G.Bt` is a CF sub-fork container. **CF.**

4. **`cross_comp(G.Nt, r)`** in sum2G — `G.Nt` is a CF. **CF.**

5. **`cross_comp(lev, rr=elev)`** in frame_H — `lev := sum2F(tile_)`, no root passed, so sum2F returns F. typ comes from tile_[0].typ which is 3 (CNs), so sum2F builds and returns a CN. **CN.**

6. **`cross_comp(T.Nt, T.r)`** in vect_edge's `base_tile` — T.Nt is a CF. **CF.**

So most call sites pass a **CF sub-fork container**, not a CN. The promotion in cross_comp (`if isinstance(root,CF): F2N(root)`) fires routinely on these.

So yes, typ==0 branch in F2N is live and needed. My "drop it" was wrong.

Looking once more at the cross_comp body:

```python
if G_:
    if isinstance(root,CF): F2N(root)  # upgrade @ 1st sub|agg+
    root.H+= [sum2F(L_,root,froot=1)]
    root.Nt = sum2F(G_,froot=1); ...
```

After F2N, root is now a CN with fresh `kern, span, angl, yx, nt` defaults. But these CFs (Nt, Bt sub-fork containers) had no positional data to preserve anyway — they're aggregates. So defaults are correct.

Restore the typ==0 branch.

kwcckw: > if isinstance(root,CF): F2N(root)

Okay, so we do need to convert CF to CN, but it should be just CN's attrs since those CL's attrs are redundant to root.

boris-kz: the root is CF, it doesn't have CL attrs

kwcckw: Yes, but the conversion doesn't require to convert it from CL -> CN, since we can just use F2N to include all CN's attrs.

boris-kz: you just said that only the attrs in `__init__` are included? so we need both sets of attrs here

kwcckw: Yes, that's when we init new class. But when we convert classes, we still need to manually init attrs since the converted class's attrs will not be included.

boris-kz: That's why we need two sets of attrs in Na_

kwcckw: Okay thanks. So those CL's __init__ attrs are needed in converted CN, then 2 sets of Na is required. I will go through them again later.

PR #133, opened 2026-05-03
5/3/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -93,12 +93,12 @@ def __init__(l, **kw):
 
 class CC(CL):  # typ=2, adds arrays per N_
     name = "cent"
-    def __init__(c, **kw):
+    def __init__(ce, **kw):
```
kwcckw, line 96: So we need to rename c here, else it conflicts with the c attrs when we init CC:

    TypeError: __init__() got multiple values for argument 'c'

frame_2D_alg/agg_recursion.py
```diff
@@ -112,7 +112,7 @@ def __init__(n, **kw):
         n.Nt,n.Bt,n.Ct,n.Lt,n.Xt,n.Rt = ((kw.get(f) if f in kw else CF(root=n) for f in ('Nt','Bt','Ct','Lt','Xt','Rt')))  # CN if nest, Ct||Nt
         super().__init__(**kw)
         n.H  = kw.get('H', [])  # lower CF levs / Nt||Ct
-        n.c_ = kw.get('c_',[])  # reciprocal root Cs
+        n.C_ = kw.get('C_',[])  # reciprocal root Cs
```
kwcckw, line 115: The reciprocal C_ should be C_ instead of c_? That c_ is the root Cs added cluster_C or cluster_P?

frame_2D_alg/agg_recursion.py
```diff
@@ -707,12 +707,12 @@ def sum2G(ft_, root=None, init=1, typ=None):
 def comb_Ft(Nt, Lt, Bt, Ct, root,wTT):  # from sum2G, default Nt
 
     G = CN(Nt=Nt,Lt=Lt,Bt=Bt,Ct=Ct, root=root); Nt.root=G; Lt.root=G; Bt.root=G; Ct.root=G
-    T = CopyF(Nt)  # temporary accumulator
+    T = Nt  # temporary accumulator
```
kwcckw, line 710: Sorry, we should remain the CopyF of (Nt) here, but add2F below should have merge =2 to prevent packing the same Nt.N_ into N_ again.

frame_2D_alg/agg_recursion.py
```diff
@@ -835,7 +835,7 @@ def PP2N(PP):
         for B in B_: m_ += B.verT[0]; d_ += B.verT[1];
         ad_ = np.abs(d_); t_ = m_ + ad_  # ~ max comparand
         m = m_/eps_(t_) @ w_t[0] - ave*2; d = ad_/eps_(t_) @ w_t[1] - avd*2
-        PP.Bt = CF(N_=B_, m=m, d=d, r=2, root=PP,nF='Bt')
+        PP.Bt = CF(N_=B_, m=m, d=d, r=2, root=PP,nF='Bt'); PP.B_ = PP.Bt.N_; PP.rim = []; PP.box = box; PP.sub=0
```
kwcckw, line 838: We actually needs some additional CN's attrs in PPs, that's when we comp PP in trace_edge. So we should init them as CN, but with typ == 0? Or init them as CF and run the conversion later?

Right now i'm just manually adding those additional attrs.

frame_2D_alg/agg_recursion.py
boris-kz: The roots are reciprocal, C_ is internal

kwcckw: I see, thanks for clarifying. I was confused earlier. So during the conversion:

    Na_ = dict(H=[], mang=1, box=np.array([np.inf,np.inf,-np.inf,-np.inf]), sub=0, exe=0, fin=0, c_=[], compared = set())
    
We should skip `c_` if F already have `c_` , that's when we convert CC to CN.

boris-kz: right. we can rename CN.c_ to `root_`, that's more suggestive? and copy along with Nt in F2N

frame_2D_alg/agg_recursion.py
boris-kz: that should be selective, in 860?

G_,TT,C, R = trace_edge([F2N(n) for n in N_], G_,TT,c,3,tile)

frame_2D_alg/agg_recursion.py
boris-kz: yes, just make it n

frame_2D_alg/agg_recursion.py
boris-kz: actually that c_ should not be in Cs, only in C.N_. 
so we add m_ and d_ to both C and N, but they map to correspondingly C.N_ or N.root_

frame_2D_alg/agg_recursion.py
```diff
@@ -212,7 +212,7 @@ def cross_comp(root, rr):  # core function mediating recursive rng+ and der+ cro
             E_ = get_exemplars({N for L in L_ for N in L.nt}, r,c)
             G_,r = cluster_N(root, E_,r,c)  # -> cluster_C,_P, eval?
             if G_:
-                if isinstance(root,CF): F2N(root)  # upgrade @ 1st sub|agg+
+                if not hasattr(root,'H'): F2N(root)  # upgrade @ 1st sub|agg+
```
boris-kz, line 215: if not root.typ?

frame_2D_alg/agg_recursion.py
```diff
@@ -681,7 +681,7 @@ def sum2G(ft_, root=None, init=1, typ=None):
     for ft, nF in zip_longest(ft_,('Nt','Lt','Bt')):
         if ft: n_,_,tt,c,r = ft; Ft_+= [CF(N_=n_,nF=nF,dTT=tt,m=(vt:=vt_(tt,ttTrc if typ == 2 else ttcN))[0],d=vt[1],c=c,r=r)]
         else:  Ft_ += [CF()]
-    C_ = [c for N in ft_[0][0] for c in N.Ct.N_]
+    C_ = [c for N in ft_[0][0] if hasattr(N, 'Ct') for c in N.Ct.N_]
```
boris-kz, line 684: Ct.N_ must be CNs?

C_ = [c for N in ft_[0][0] for c in N.C_]

frame_2D_alg/agg_recursion.py
kwcckw: Yes, so we should remove c_ from CC , since we only use c_ from n now.

frame_2D_alg/agg_recursion.py
kwcckw: Yes, this is better, thanks.

frame_2D_alg/agg_recursion.py
kwcckw: Yes, not root.typ should work the same. We can't use isinstance because `isinstance(root,CF)` returns true for all inherited classes.

frame_2D_alg/agg_recursion.py
kwcckw: This is actually for PP, where they do not have Ct, so this shouldn't be relevant now.

frame_2D_alg/agg_recursion.py
boris-kz: I will change that in cluster_C

kwcckw: We only init c_ from n in cluster_C, so there's no changes needed from those Cs?

boris-kz: yes, I will just rename c_ to root_

boris-kz: also replace N.root_ in cluster_P:

    for i, _C in enumerate(C_):
            N_,m_,d_ = [],[],[]
            if N_:
                C = sum2F(N_,root, m_,d_)
                for N in N_:  
                    N.root_ += [C]  # form C-to-N links:
                    C.L_= [CN(typ=1,nt=[N,C], dTT=N.dTT,c=N.c,r=N.r,m=N.m,d=N.d, span=np.hypot(*(dy_dx:=C.yx-N.yx)), angl=[dy_dx,np.sign(N.dTT[1]@ttcN[1])])]
                out_ += [C]

boris-kz: so these root_-mapping m_,d_ in C.N_ CNs may conflict with prior promoted CC m_,d_ s, they map to CC.N_?

boris-kz: Yes, real conflict. Tracing it:

CC defines `m_,d_` indexed by `C.N_` (per-member). On regular CN, `m_,d_` are indexed by `N.root_` (per-centroid-it-belongs-to). Different lengths, different semantics.

F2N preserves the CC's original `m_,d_` (only resets `root_=[]`). When that promoted CC re-enters `cluster_C` as a member of a higher-level centroid, this line silently clobbers them:

```python
for n in N_: n.root_,n.m_,n.d_,n._root_,n._m_,n._d_ = [],[],[],[],[],[]
```

So the per-N_ arrays the CC carried from its construction in `sum2F`/`cluster_P` are lost, and in between they're briefly interpretable under the wrong indexing if anything reads them.

Also `add_Nt` does:
```python
if hasattr(N,'m_'):
    G.C_ += N.c_; G.m_ += N.m_; G.d_ += N.d_  # Ct || Nt
```
This concatenates regardless of which semantic the source `m_,d_` carry — a CC member would inject N_-indexed values into a root_-indexed accumulator.

Minimal fix in your style: keep `m_,d_` on CN as the per-root_ membership (it's the more universal use, hit by `cluster_C`, `cluster_P`, `add_Nt`), and rename CC's per-N_ arrays. Following your `_` convention something like `cm_,cd_` or `mN_,dN_` on CC. Then:

- `sum2F` typ==2 branch sets `F.cm_,F.cd_ = m_,d_`
- `cluster_P`'s `zip(N.c_,N.m_,N.d_)` stays correct (N is member, not centroid)
- F2N-promoted CC keeps `cm_,cd_`, and its `m_,d_` start empty as for any CN

Want me to show the diffs for the rename?

kwcckw: > so these root_-mapping m_,d_ in C.N_ CNs may conflict with prior promoted CC m_,d_ s, they map to CC.N_?

This is about Cs  when we run cross_comp with Ct, and those Cs may rerun cluster_C and reinit m_ and d_?

boris-kz: yes, it's very unlikely

frame_2D_alg/agg_recursion.py
boris-kz, line 675: def sum2G(ft_, wTT, root=None, init=1):  # core clustering function

    if not init:
        N_,_,ntt,nc,nr = ft_[0]; N_+=root.N_; ntt+=root.Nt.dTT; nc+=root.Nt.c; nr+=root.Nt.r; ft_[0] = N_,_,ntt,nc,nr
    Ft_ = []
        if ft: n_,_,tt,c,r = ft; Ft_+= [CF(N_=n_,nF=nF,dTT=tt,m=(vt:=vt_(tt,wTT))[0],d=vt[1],c=c,r=r)]
        else:  Ft_ += [CF()]
    C_ = [c for N in ft_[0][0] for c in N.C_]  # splice centroids
    G = comb_Ft(*Ft_, root, wTT=wTT)
    if G.Lt:  # sub+
        if Vn > 0:
            c = G.Lt.c; E_ = get_exemplars({N for L in L_ for N in L.nt}, r,c)
            if Vn *(mdecay(L_)-decay) > (ave+avd)*(lr+1+ccC*L):
                r+=1; G_,r = cluster_C(G.Nt,E_,r,c)  # higher V, low decay
            else:     G_,r = cluster_N(G.Nt,E_,r,c)  # updates G
            L= (len(G_)-1) **2  # full cross_comp?
            if G_ and vt_(G.Nt.dTT,G.wTT*ttX)[0]*(wX*L) > ave*(r+cX*L) > 0:
                cross_comp(G.Nt,r)
    if G.Bt:
        if N.typ!=1 and bd*(wX*L) > avd*(br+cX*L): [F2N(L) for L in Bt.N_]; cross_comp(Bt, br)  # no ddfork
        if rroot: Bt.brrw = Bt.m * (rroot.m * (decay * (rroot.span/G.span)))  # external lend only, subtract from root?
    return G

frame_2D_alg/agg_recursion.py
kwcckw: Okay, then we probably can have M_ and D_ to map into C.N_, while when C becomes node, it' using the same m_ and d_?

frame_2D_alg/agg_recursion.py
kwcckw: So input typ is removed, but i think we can identify if sum2G is called from trace_edge by checking N's typ, and use ttTrc or ttcN in different cases.

boris-kz: the typ is always 3, ttTrc or ttcN is arg wTT

frame_2D_alg/agg_recursion.py
boris-kz: I wouldn't worry about it at this point

frame_2D_alg/agg_recursion.py
kwcckw: Thanks, i overlooked that input wTT and was thinking that's the global wTT instead.

frame_2D_alg/agg_recursion.py
```diff
@@ -93,12 +93,12 @@ def __init__(l, **kw):
 
 class CC(CL):  # typ=2, adds arrays per N_
```
boris-kz, line 94: So the next step is code modification, based on CoF usage records. We already have run-time evals of Fw_, Fc_, FTT_, now we need to use them to re-arrange, cluster/split, or extend the functions?

boris-kz: Got it — so the evals already exist (the `if … > ave*(r+cF*L)` gates), and each gated sequence is implicitly a block. The question is meta: when does a gate's stats tell you to split it (finer eval) or merge it with an adjacent one (coarser)?

This maps directly onto your own clustering logic applied to gates instead of data. Each gate i has, over its call history:
- pass-rate `pi` (gate true / total calls)
- conditional gain when passed: mean `m` of the body
- cost paid when failing: `Fc_` of the gate evaluation itself
- behavioral signature `FTT_[i]` of the data it gates on

**Split criterion** — gate is too coarse:
- High variance in body's `m` across calls that all passed → the gate isn't discriminating well, the body's value depends on something the gate doesn't see. Operationally: cluster the passing calls by their `dTT`. If `cent_TT` on those passes returns a clearly bimodal weight profile (one mode dominates a subset), split into two gates with different thresholds or different conditioning attrs.
- Equivalently: if `FTT_[i]` of *passes* differs strongly from `FTT_[i]` of *fails*, the gate is fine; if passes themselves split into sub-clusters with very different downstream `m`, split.

**Merge criterion** — adjacent gates are redundant:
- Two sequential gates A then B where `P(B passes | A passed) ≈ 1` or `≈ 0`: B is determined by A, merge into one gate (drop B, or AND them into a single test). Operationally: high `comp_F(typ_A, typ_B).m`.
- Two gates on the same body / same data with overlapping signatures → same as your trans-link case: cluster gates whose `FTT_` match above ave.

**The unifying picture:** gates *are* nodes, gate-pairs *are* links, and you already have `comp_F` / `cluster_C` to do this. Specifically:
- Treat each gate's accumulated stats as a CoF (which `add_typ_` already does per function — just extend to per-gate).
- `comp_F` gates pairwise → `m` high → merge candidate.
- `cent_TT` on a single gate's per-call `dTT`s → if invdev profile is far from uniform, that gate's input distribution is multimodal → split candidate.

So you don't need new criteria, just new instrumentation: per-gate (not just per-function) accumulation of pass/fail `dTT`. Then split = `cent_TT` returns non-uniform weights on passes; merge = `comp_F` between gates returns high `m`.

**Question:** is each gate already getting its own CoF in `call_` (since the body is usually a `traced` function call, the gate's stats are effectively the body's stats minus failure cost), or do you want explicit gate CoFs separate from body CoFs? The latter is what'd let you measure `P(pass)` directly.

kwcckw: Okay, but to do that, we need to the function to be modular first so that we can re-arrange, insert or remove them? 

Or we need to breakdown some functions to smaller parts first?

boris-kz: the modules are evaled blocks, see above.
for each eval we get a value deviation, which is the value of that eval op.
so ideally we each eval should be oF too, storing that value.
if that oF.w > ave * oF.cost: make the eval terms more detailed, or split the evaled block,
else: simplify the eval or merge adjacent evaled blocks?

boris-kz: Yes, that rule is the right shape but needs one refinement: `w > ave*cost` only tells you the gate is *paying its way overall*, not whether splitting would help. For split-vs-leave-alone you also need *within-pass variance*:

- high w, high variance of body's `m` across passes → split the body (each sub-block gets its own gate, finer eval)
- high w, low variance → gate is well-tuned, leave alone
- low w, high pass-rate (correlated with predecessor gate) → merge with predecessor
- low w, low pass-rate → simplify or drop

So it's a 2D criterion: gate's own deviation, plus dispersion of body outcomes among passes. Both are derivable from per-gate `dTT` history if we record passes and fails separately — variance falls out of `cent_TT` returning non-uniform weights on the passes' `dTT`.

On "one gate per oF": two ways to get there.

**(a) Refactor**: every gated body becomes its own `traced` function. Clean, makes call_ tree exactly mirror gate tree, but invasive — lots of small functions, especially in `cross_comp`, `comp_N_`, `cluster_N`.

**(b) Inline gate marker**: a `gate(V, cost, dTT=…)` call (or `with gate(…)`) that pushes a sub-CoF at each `if` site, recording deviation as `w` regardless of pass/fail. Failure still creates an oF, so you finally see the cost of gates that mostly fail — currently invisible because no body oF is created. Minimal change, ~one line per gate.

(b) seems strictly better for measurement. (a) might still be desirable later as the *output* of the split decision — i.e. once a gate is flagged for splitting, you factor its body out.

**Questions:**

1. Going with (b) — does the gate marker also need to record the *body's* dTT separately from the gate's predicate dTT? Right now the body's traced calls would nest under the gate-oF naturally, so body stats are accessible as `gate_oF.call_`. I think that's enough.

2. For the merge criterion (gate-pair correlation), do you want this measured automatically via `comp_F` between adjacent gate oFs in `call_`, or is that a manual review step once we have the stats?

3. The split/merge decision itself — runtime self-modification, or offline review where I propose changes and you apply them?

kwcckw: I see, I think oF per eval should be better since we can merge those oFs later. But I'm not sure how we can split an existing single eval into multiple evals?

boris-kz: any complex operation: evaluated block, can be split in smaller evaluated blocks.
we just need AST to tell us if block complexity justifies the projected gain of skipping it.  
check (b) Inline gate marker, above

boris-kz: Not in standard AST tooling, no. AST is structural — it doesn't carry runtime stats. What you're describing is closer to **profiling instrumentation** + **counter/probe insertion**, which is a different lineage.

The closest established things:

- **`sys.settrace` / `sys.setprofile`** — Python's built-in hooks fire on every function call/line/return. cProfile uses this. Gives you per-line counts and timing for free, no source changes. But it's all-or-nothing per line; doesn't carry your `dTT` payload.

- **Branch coverage tools** (`coverage.py --branch`) — record which branches taken/not-taken via `sys.settrace`. Stats are pass/fail counts only, no value payload.

- **Manual probes** in performance work — `with timer("name"):` blocks, `@profile` decorators (line_profiler), `nvtx.range_push/pop` in CUDA code. These are the direct ancestors of what you'd write. Common pattern, but ad-hoc per project, not standardized.

- **Aspect-Oriented Programming** (Java/AspectJ, Python `aspectlib`) — declares "around these joinpoints, do X". Originally for logging/auth; same mechanism works for stats. Heavy and fell out of fashion.

- **eBPF / DTrace** — kernel-level dynamic probing. Runtime instrumentation without source changes. Conceptually closest to what you want — probe insertion at named points, payload accumulation — but at OS/syscall level, not Python expressions.

- **JIT profile-guided optimization** (PyPy, V8, JVM) — runtime collects branch frequencies and call-site type distributions, uses them to inline / specialize / deoptimize. This *is* split/merge driven by runtime stats, fully automated, and the closest match to your intent. But it's hardcoded into the runtime; the heuristics are not user-exposed and not tied to any domain-meaningful "value" — only to execution frequency.

So your scheme is in a real lineage (PGO-style adaptive specialization), but the specific combination — gate-as-cluster, value-deviation-driven split/merge, payload is your own `dTT` — is your own. The `gate(V, cost, dTT)` marker pattern itself is conventional probe instrumentation; nothing exotic, nothing AST-level.

If you wanted to get fancier later: a decorator could AST-walk a function, find every `if` whose test is of the form `... > ave*(...)`, and auto-wrap them. That would remove the manual marker. But for ~60 gates across the file, manual is simpler and you'd want to review each anyway.

Want me to sketch the marker?

boris-kz: Minimal version, dropped into the existing `CoF` machinery so it doesn't introduce a new data path:

```python
class CoF(CF):
    ...
    @staticmethod
    @contextmanager
    def gate(V, cost, dTT=None, name=''):  # context manager around an `if V > ave*cost:` body
        _CoF = CoF._cur.get()
        g = CoF(nF=name, root=_CoF); g.fc = cost; g.w = V - ave*cost  # value deviation = gate's own w
        _CoF.call_ += [g]; CoF._cur.set(g)
        try: yield g
        finally: CoF._cur.set(_CoF)
```

Use:

```python
with CoF.gate(sum(vt_(TT,ttE)), cE*L+r, dTT=TT, name='cross_comp:exemplars'):
    G_,r = cluster_N(root, E_,r,c)
    ...
```

What this gives you:

- Every gate becomes a `CoF` in `call_` with `w = V - ave*cost`, regardless of whether the body fires (failed gates still register, just with body cost = 0 since nothing nests under them).
- Body's traced calls nest under the gate oF naturally — so `gate.call_` is exactly the body's call subtree, `flat_(gate)` gives the full subtree cost, and `gate.w / sum(c.fc for c in flat_(gate))` is the per-call efficiency you'd cluster on.
- `gate.dTT` carries the predicate's signature so `cent_TT(gate.dTT, gate.r)` is meaningful for split-detection (multimodal predicate inputs → split).
- Adjacent gates in the same `call_` can be `comp_F`'d for merge candidates.

Three things to decide:

**1. Failed gates** — should they still create the oF (current sketch: yes, body just doesn't execute, so `gate.call_=[]`)? This is what lets you measure pass-rate. Or do you want failures invisible? I'd keep them visible.

**2. Where does `dTT` come from at the gate?** For gates of the form `vt_(TT, wTT)*wF*L > ave*(r+cF*L)`, the natural `dTT` is `TT` itself — what's being evaluated. For gates without an explicit TT (e.g. `if Up * (wcC*len(C_)) > Avd * ...`), there's no signature to record beyond `w`. The marker should accept `dTT=None` and just skip the multimodality check for those.

**3. Marker placement** — every gate, or only the ones whose body is non-trivial? I'd say every gate that calls a `traced` function in its body. The pure-arithmetic ones (e.g. inside `comp_N_`'s loop, deciding whether to call `comp_N` per pair) are too hot — wrapping them in a context manager per pair would dominate runtime. For those, accumulate counts manually into a CoF created once per `comp_N_` call.

Want me to mark up which existing gates I'd wrap vs leave, before any actual edits?

kwcckw: > any complex operation: evaluated block, can be split in smaller evaluated blocks. we just need AST to tell us if block complexity justifies the projected gain of skipping it.
 
That projected gain should dbe computed in runtime using data?

Also the oF per gate probably can be packed in per function oF?

boris-kz: see above. I will check tomorrow

PR #134, opened 2026-05-04
5/4/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -208,13 +262,15 @@ def cross_comp(root, rr):  # core function mediating recursive rng+ and der+ cro
     if L_: # Lm_, no +|- Ft.Lt?
         root.L_ = L_; L= len(L_)-1  # val=m+d /clust, m/comp
         if sum(vt_(TT,ttE)) * (wE*L) > (ave+avd) * (r+cE*L):
+        # if gate(sum(vt_(TT,ttE))*(wE*L), r+cE*L, dTT=TT, name='cross_comp:get_exemplars'):
```
kwcckw, line 265: So this is drafted from the prior discussion using claude. Basically each eval block is a gate, init with CoF. 

During the init, their params will be assigned using the input val, cost and TT:

        g.root = _CoF
        g.fc = cost
        g.w  = V - ave * cost

Then during the feedback, we should eval those gates to merge or split them using the 2D criterion:

      - high w, high variance of body's m across passes → split the body (each sub-block gets its own gate, finer eval)
      - high w, low variance → gate is well-tuned, leave alone
      - low w, high pass-rate (correlated with predecessor gate) → merge with predecessor
      - low w, low pass-rate → simplify or drop

Right now the new gate reference after the merging or split is mainly done by using register and handler in the `gate` class above, and the merging and split is done with `gate_eval` above.

In term of the variance computation, the one in gate_eval should be just a place holder. In actual case, variance should be computed with ds?

frame_2D_alg/agg_recursion.py
```diff
@@ -128,7 +128,7 @@ def flat_(oF, call_=None):  # all nested call_ s
     if call_ is None: call_ = []
```
kwcckw, line 128: > any complex operation: evaluated block, can be split in smaller evaluated blocks.
we just need AST to tell us if block complexity justifies the projected gain of skipping it.

AST is static and that should be done once to identify those complex blocks before we split it manually to smaller blocks?

frame_2D_alg/agg_recursion.py
boris-kz: yes, variance is probably np.sum( goF.dTT[1] ): dTT is data of gate_oF?
these gate_oFs are actually higher-order than current oFs, their gain and cost are the derivatives of corresponding-oF' gain and cost?
g.w is currently g.fw, not sure there's a difference

frame_2D_alg/agg_recursion.py
boris-kz: right, split or merge

frame_2D_alg/agg_recursion.py
```diff
@@ -830,7 +886,7 @@ def PP2N(PP):
         y,x,Y,X = box; dy,dx = Y+1-y, X+1-x
         A = [np.array(A), np.sign(dTT[1] @ ttVct[1])]  # append sign
         PP = CL(typ=0, dTT=dTT,m=m,d=d,c=c,r=1, kern=kern,yx=yx,angl=A,span=np.hypot(dy/2,dx/2))  # set root in trace_edge
-        m_, d_ = np.zeros(6), np.zeros(6)
+        m_, d_ = np.zeros(6), np.zeros(6); PP.B_ = B_  # B_ exclusively for PPd
```
boris-kz, line 889: thanks

frame_2D_alg/agg_recursion.py
boris-kz, line 896: for dGs in Bt, rN_ is current root_, only the roots are core Gs instead of Cs.
that's if multiple Gs can share the same boundary Gs, it makes sense but probably doesn't work yet

frame_2D_alg/agg_recursion.py
```diff
@@ -128,7 +128,7 @@ def flat_(oF, call_=None):  # all nested call_ s
     if call_ is None: call_ = []
     for sub in oF.call_:
         call_ += [sub]
-        if sub.N_: flat_(sub, call_)
+        if sub.call_: flat_(sub, call_)  # should be call_ here
```
boris-kz, line 131: right

frame_2D_alg/agg_recursion.py
```diff
@@ -171,6 +172,59 @@ def add_typ_(oF):  # record oF vals for weighting, mapped to global FTT_
     oF.typ_ = typ_
     if any(typ_): add2F(oF,sum2F([t for t in typ_ if t],CoF()))  # refine summed call_?
 
+# may merge into CoF if needed
```
boris-kz, line 175: why do you need a new class?

kwcckw: Sorry, this is init separately so that it does not mess up the original CoF. We actually can merge them into the CoF later since it's actually inherited from CoF too.

frame_2D_alg/agg_recursion.py
kwcckw: Okay, so we actually should have this first to split or merge the existing code before proceed with the gated CoF?

frame_2D_alg/agg_recursion.py
kwcckw: > yes, variance is probably np.sum( goF.dTT[1] ): dTT is data of gate_oF?

Yes, dTT is the data of gate oF, where we parse this dTT during the gate eval:

    if gate(sum(vt_(TT,ttE))*(wE*L), r+cE*L, dTT=TT, name='cross_comp:get_exemplars'):
    
> these gate_oFs are actually higher-order than current oFs, their gain and cost are the derivatives of corresponding-oF' gain and cost?

I think this is true for most of the cases since gain and cost are computed from the nested function in current function (for example, TT,c, r are computed from comp in cross_comp). So if they are higher -order, we shouldn't pack them per oF.gate_?

frame_2D_alg/agg_recursion.py
kwcckw: Yes, sorry, i forgot to remove the comment, we actually still need this rN_ in trace_edge to get node-mediated `_N_`, .

frame_2D_alg/agg_recursion.py
boris-kz: it's to split or merge gated blocks depending on gating gain, we already have inital blocks

frame_2D_alg/agg_recursion.py
```diff
@@ -171,6 +172,59 @@ def add_typ_(oF):  # record oF vals for weighting, mapped to global FTT_
...
+def gate_eval():  # split/merge/drop gates per 2D criterion
+
+    for oF in [oF for oF in flat_(Z) if oF.gate_]:  # only functions with gates
+        for i, g_ in enumerate([gate.reg[gN.nF] for gN in oF.gate_]):
+            if not g_: continue
+            name = g_[0].nF
+            G = sum2F(g_)                    # sum gates: G.w, G.dTT, G.c, G.r
+            P_ = [g for g in g_ if g.w > 0]  # passes
+            F_ = [g for g in g_ if g.w <=0]  # fails
+            pr = len(P_) / (len(g_) or eps)  # pass-rate
```
boris-kz, line 205: we only need to evaluate the passes, same as calls, there is no data from the fails

frame_2D_alg/agg_recursion.py
kwcckw: > it's to split or merge gated blocks depending on gating gain, we already have inital blocks

Okay, so this gating gain is not static, which mean AST will be part of the split or merge workflow? It should be merged with the 2D criterion above?

boris-kz: AST is the fixed cost part, including fails, the gain is dynamic, from passes only

frame_2D_alg/agg_recursion.py
boris-kz: it should be per oF call, so each goF-gated block is an oF, we have to turn them into functions?

kwcckw: I'm not sure if that's a good idea because we will be creating a lot of sub functions for each gated block? Then when we split them further, we need to create new functions again?

frame_2D_alg/agg_recursion.py
kwcckw: But that AST may not be relevant after we split or merge those gates? So it's only useful in the first pass?

boris-kz: AST can be tested for each new block

frame_2D_alg/agg_recursion.py
boris-kz: split or merge. these gates are already per oF, not much difference, they just have to be folded in or aligned with oFs

boris-kz: The derivative relationship:

- `oF.dTT` = data signature flowing through the function
- `goF.dTT` = predicate value: `vt_(TT, wTT)*w*L` vs. `ave*(r+c*L)` — i.e. a *function of* `oF.dTT` (and weights, cost, etc.)

So `goF` is one derivation level up from `oF` on the same data. Same relation as your link/node hierarchy: links are derivatives of nodes, gates are derivatives of functions.

Concretely: gate's predicate is computed from the data the function is processing — `TT` from `comp_N_`'s accumulator, `m` from a comp result, etc. `goF.w` (value deviation) is what survives the comparison `predicate - threshold`. That's exactly `m`/`d` from a comparison, lifted to function level.

**Containment options:**

1. **`oF.gate_` list** — sibling of `call_`, holds the goFs. Cleanest separation: `call_` = sub-functions invoked, `gate_` = gates over their invocation. Splits the call tree by derivation order.

2. **`goF` inside `oF.call_`** — gates nest in call_ alongside sub-function oFs. Uniform traversal, but mixes derivation orders in one list. Existing `flat_` walks both indistinguishably.

3. **`goF.call_` contains body's sub-functions** — gate is the parent, body's traced calls nest under the gate, gate sits in parent function's `call_`. This is what `gate` context manager does in my earlier sketch.

(3) is what I had. But your point reframes it: if goF is a derivative of oF, the natural containment is *gate inside the function whose data it gates on*, not *function inside the gate*. So (1) is more faithful:

```
oF (function)
├── call_ : [sub-oFs invoked when gates pass]
└── gate_ : [goFs measuring decisions on this oF's data]
```

Each `goF` references the sub-oF(s) it gates (so you can still recover "which sub-call did this gate enable"), but structurally the gate belongs to the gating function, not to the gated body.

This also makes the split/merge logic cleaner:
- compare goFs *within one oF.gate_* → merge candidates (sequential gates on same data)
- cluster goFs *across oF.gate_* by signature → find gates that are doing the same job in different functions
- `cent_TT` on a single goF's accumulated predicate `dTT` → split candidate

And `Fw_`/`FTT_` extend naturally: each oF has `gate_w_`, `gate_TT_` arrays parallel to its gates.

Does this match? The reframe is: gate is a derivative, so it's *owned by* the function whose data it derives from, not by its body.

frame_2D_alg/agg_recursion.py
boris-kz: This is folding goF into in CoF:

Concrete sketch — minimal additions to current CoF, plus the gate marker. Showing only the diff:

```python
from contextlib import contextmanager

class CoF(CF):
    name = "func"
    def __init__(f, **kw):
        super().__init__(**kw)
        f.call_ = kw.get('call_',[])  # sub-oFs (functions invoked)
        f.gate_ = kw.get('gate_',[])  # sub-goFs (gates on this oF's data, derivative)
        f.typ_  = kw.get('typ_', [])
        f.fw,f.fc,f.fr = [kw.get(x,0) for x in ('fw','fc','fr')]
    @staticmethod
    @staticmethod
    def traced(func):  # unchanged except: traversal includes gate_
        @wraps(func)
        def inner(*a, **kw):
            _CoF = CoF._cur.get()
            oF = CoF(nF=onF_.index(func.__name__),root=_CoF)
            oF.wTT = FTT_[oF.nF]; oF.fw = Fw_[oF.nF]; oF.fc = Fc_[oF.nF]
            _CoF.call_ += [oF]
            if oF.call_ or oF.gate_:
                if oF.fw*L > ave*(oF.fc*L):
                    sum2F(tree,oF)
            oF.wTT = cent_TT(getattr(oF,'rTT',oF.dTT), oF.r)
            oF.w += np.mean(oF.wTT)
            CoF._cur.set(_CoF)
            return out
        inner.wrapped = True
        return inner
    @staticmethod
    @contextmanager
    def gate(V, cost, dTT=None, name=''):  # derivative oF over current oF's data
        _CoF = CoF._cur.get()
        g = CoF(nF=name, root=_CoF); g.fc = cost; g.w = V - ave*cost
        _CoF.gate_ += [g]; CoF._cur.set(g)
        try: yield g
        finally: CoF._cur.set(_CoF)
```

And `flat_` extended once:

```python
def flat_(oF, out_=None, fd=2):  # fd: 0=call_, 1=gate_, 2=both
    if out_ is None: out_ = []
    subs = (oF.call_ if fd!=1 else []) + (oF.gate_ if fd!=0 else [])
    for sub in subs:
        out_ += [sub]
        if sub.call_ or sub.gate_: flat_(sub, out_, fd)
    return out_
```

That's it. Everything else (`add_typ_`, `sum2F`, `cent_TT`, `comp_F`) keeps working — they operate on `dTT`/`w`/`c`/`r`, none of which changed.

**Use site, e.g. in `cross_comp`:**

```python
if L_:
    root.L_ = L_; L = len(L_)-1
    with CoF.gate(sum(vt_(TT,ttE)) * (wE*L), (ave+avd)*(r+cE*L), dTT=TT, name='cc:exemplars'):
            G_,r = cluster_N(root, E_,r,c)
            ...
```

The `if` stays — the marker just records the gate; it doesn't replace the test. (Could be merged later via a helper that both records and tests, but keeping them separate makes the marker non-invasive.)

**Things this leaves open:**

1. **`onF_` extension for gate names.** Currently `nF=onF_.index(func.__name__)` is an integer. Gate names are strings (`'cc:exemplars'`). Either gate names also get appended to `onF_` and indexed, or `nF` becomes `int | str` and indexers handle both. The latter is less invasive but `Fw_[i]` indexing breaks. I'd add gates to `onF_` lazily — first time a name is seen, append, get index.

2. **`add_typ_`** currently bins `call_` only via `flat_`. With `fd=2` default it'll bin gates too into `typ_[i]`. Functions and gates share the namespace. Is that what you want? Or do you want separate `typ_` and `gtyp_`?

3. **`sum2F` over `flat_(oF)`** — when both calls and gates are summed together, the resulting summary mixes derivation orders. Maybe sum them separately and store both: `oF.dTT` from `call_`, `oF.gdTT` from `gate_`. Then comp_F between two oFs compares both.

I'd hold on (2) and (3) until you've placed a few markers and seen what the data looks like. The CoF skeleton above is enough to start instrumenting — answer to those falls out from what's actually useful in the stats.

boris-kz: Fair, you're right — alignment is positional and recoverable. If each gate records `len(parent.call_)` at creation, the calls "under" gate i are `parent.call_[gate_i.span : gate_{i+1}.span]` (or to end of call_). No `_cur` push, no `with`.

Revised gate:

```python
@staticmethod
def gate(V, cost, dTT=None, name=''):
    _CoF = CoF._cur.get()
    g = CoF(nF=name, root=_CoF); g.fc = cost; g.w = V - ave*cost
    g.span = len(_CoF.call_)  # index into parent.call_ where this gate's body starts
    _CoF.gate_ += [g]
    return g.w > 0  # plain bool
```

Use:

```python
if CoF.gate(sum(vt_(TT,ttE))*(wE*L), (ave+avd)*(r+cE*L), TT, 'cc:exemplars'):
    E_ = get_exemplars(...)
    G_,r = cluster_N(...)
```

One line per gate, identical to the original `if` shape, gate's CoF always created (pass and fail recorded), body's calls land in `parent.call_` aligned by index. To recover which sub-calls ran under gate g:

```python
def gate_calls(parent, i):
    g = parent.gate_[i]
    end = parent.gate_[i+1].span if i+1 < len(parent.gate_) else len(parent.call_)
    return parent.call_[g.span:end]
```

Cleaner than `with`, and the gate-as-derivative framing holds: gate sits in `gate_`, sibling to `call_`, with `span` linking them.

One edge case: if a gate fails, no calls land under it, so `g.span == next_gate.span` (zero-width slice). That's correct — empty body span means failed gate. Free pass-rate signal: `g.span < next.span` ⟺ pass.

Match what you wanted?

frame_2D_alg/agg_recursion.py
kwcckw: > Each goF references the sub-oF(s) it gates (so you can still recover "which sub-call did this gate enable"), but structurally the gate belongs to the gating function, not to the gated body.

Yes, i think if each goF should reference the oF it gates, but that's requires us to manually insert that information? I don't see how we can do that automatically yet. Also if gated section should be a function for referencing purpose, else goF can't reference them if it's in unpacked form.

frame_2D_alg/agg_recursion.py
kwcckw: Okay, so this is evaluating using AST in the runtime, and they should be eval the same as the code in meta_code? Where vt_ served as the base unit, but the eval is per block of code rather than the function?

frame_2D_alg/agg_recursion.py
kwcckw: In this method using `gate`, how do we deal with merge or split if we don't have any gate reference or handle?

And gate needs a different `__bool__` based on `V * w > ave * cost` since the existing one is for CoF?

boris-kz: gate_ is aligned with call_, so flat_ and typ_ handle both

frame_2D_alg/agg_recursion.py
boris-kz: right

frame_2D_alg/agg_recursion.py
boris-kz: the data is inserted in the same way as in oF, we just need gw,gc per gate_:

            if oF.call_:
                tree,gtree = flat_(oF); L=len(tree)-1; Lg=len(gtree)-1  # flatten call_, gate_ 
                if oF.fw*L > ave*(oF.fc*L): sum2F(tree,oF)
                if oF.gw*L > ave*(oF.gc*Lg): sum2F(gtree,oF)

frame_2D_alg/agg_recursion.py
kwcckw: > gate_ is aligned with call_

But not every call has gate? For example, comp functions in cross_comp doesn't have gate and we only get gate in the clustering functions?

boris-kz: all oFs are gated

frame_2D_alg/agg_recursion.py
boris-kz: Diffs:

```python
class CoF(CF):
    def __init__(f, **kw):
        super().__init__(**kw)
        f.call_ = kw.get('call_',[])
        f.gate_ = kw.get('gate_',[])
        f.typ_  = kw.get('typ_',[])
        f.span  = kw.get('span',0)
        f.fw,f.fc,f.fr = [kw.get(x,0) for x in ('fw','fc','fr')]
        f.gw,f.gc      = [kw.get(x,0) for x in ('gw','gc')]
```

`flat_` returns two lists:

```python
def flat_(oF, ftree=None, gtree=None):
    if ftree is None: ftree, gtree = [], []
    for sub in oF.call_:
        ftree += [sub]
        if sub.call_ or sub.gate_: flat_(sub, ftree, gtree)
    for g in oF.gate_:
        gtree += [g]
    return ftree, gtree
```

`traced` wrapper:

```python
if oF.call_ or oF.gate_:
    tree, gtree = flat_(oF); L=len(tree)-1; Lg=len(gtree)-1
    if oF.fw*L  > ave*(oF.fc*L):   sum2F(tree, oF)
    if oF.gw*Lg > ave*(oF.gc*Lg):  sum2F(gtree, oF)
```

`gate` records into parent's gw/gc as it goes:

```python
@staticmethod
def gate(V, cost, dTT=None, name=''):
    _CoF = CoF._cur.get()
    g = CoF(nF=name, root=_CoF); g.fc=cost; g.w = V - ave*cost; g.span = len(_CoF.call_)
    _CoF.gate_ += [g]; _CoF.gw += g.w; _CoF.gc += cost
    return g.w > 0
```

**Open question — globals:** `oF.fw = Fw_[nF]` reads a *running total per function name*. For gates, the parallel is `Gw_[gate_name]` running totals per gate name, and a goF's `gw` would init from there. Without that, `oF.gw` is just this-invocation's sum (not comparable to `oF.fw` which is cumulative across all past calls). Two options:

(a) **Per-name globals `Gw_`/`Gc_`** — gates get the same running-totals treatment as functions. Requires gate names to index into something stable; lazy-append to `onF_` on first sight, share namespace with functions.

(b) **Per-invocation only** — `oF.gw`/`oF.gc` are just this-call's gate sums; the running aggregation happens implicitly when `sum2F(gtree, oF)` updates `oF.dTT` and that gets summed up to `Z` via parent eval.

(a) is symmetric with how functions work and gives directly comparable signals. (b) is simpler but the eval threshold for gate-summary becomes harder to calibrate. I'd go (a). Which?

frame_2D_alg/agg_recursion.py
kwcckw: > all oFs are gated

So gate is not exclusively for eval only? For example, cross_comp calls comp function by default, the comp function's gate will be referring to the higher gate of cross_comp?

     if T: cross_comp(T.Nt,T.r)

frame_2D_alg/agg_recursion.py
kwcckw: > the data is inserted in the same way as in oF, we just need gw,gc per gate_:
> 

Okay, and from the gate func, oF is accumulated with next level gate's w, so when we sum the flatten gates ` if oF.gw*Lg > ave*(oF.gc*Lg):  sum2F(gtree, oF),` it will be accumulated again? Then this `_CoF.gw += g.w` should be removed and oF.gw should be summed from the flatten gates only?

frame_2D_alg/agg_recursion.py
boris-kz: right

frame_2D_alg/agg_recursion.py
boris-kz: we just need oF.gF, then it works the same:

Got it — each oF carries an embedded gF (its gate-side aggregate), no separate gate_ list, no flat_ split. Sketch:

```python
class CoF(CF):
    name = "func"
    def __init__(f, **kw):
        super().__init__(**kw)
        f.call_ = kw.get('call_',[])
        f.typ_  = kw.get('typ_',[])
        f.fw,f.fc,f.fr = [kw.get(x,0) for x in ('fw','fc','fr')]
        f.gF = kw.get('gF', None)  # embedded gate aggregate; None for goFs themselves
    @staticmethod
    @staticmethod
    def traced(func):
        @wraps(func)
        def inner(*a, **kw):
            _CoF = CoF._cur.get()
            oF = CoF(nF=onF_.index(func.__name__), root=_CoF)
            oF.wTT = FTT_[oF.nF]; oF.fw = Fw_[oF.nF]; oF.fc = Fc_[oF.nF]; _CoF.call_ += [oF]
            gF = CoF(nF=oF.nF, root=oF)  # gate aggregate, same nF index, separate global tables
            gF.wTT = GTT_[gF.nF]; gF.fw = Gw_[gF.nF]; gF.fc = Gc_[gF.nF]
            oF.gF = gF
            if oF.call_:
                tree = flat_(oF); L = len(tree)-1
                if oF.fw*L     > ave*(oF.fc*L): sum2F(tree, oF)
                if gF.fw*L     > ave*(gF.fc*L): sum2F([s.gF for s in tree if s.gF], gF)
            oF.wTT = cent_TT(getattr(oF,'rTT',oF.dTT), oF.r); oF.w += np.mean(oF.wTT)
            gF.wTT = cent_TT(getattr(gF,'rTT',gF.dTT), gF.r); gF.w += np.mean(gF.wTT)
            CoF._cur.set(_CoF)
            return out
        inner.wrapped = True
        return inner
    @staticmethod
    def gate(V, cost, dTT=None):
        _CoF = CoF._cur.get(); gF = _CoF.gF
        gF.fc += cost; gF.fw += V - ave*cost
        if dTT is not None: gF.dTT = gF.dTT + dTT  # accumulate predicate signature
        return V > ave*cost
```

And new globals parallel to F-side:

```python
Gw_ = copy(Fc_)  # init = cost, like Fw_
Gc_ = copy(Fc_)  # could be cheaper than Fc_ since gates are arithmetic, but 1:1 indexing is simpler
GTT_= [deepcopy(wTT) for _ in range(len(onF_))]
```

Notes on the design:
- gates within a function are no longer individually addressable — they all dump into `oF.gF`. You see the function's *aggregate* gate value, not per-gate. That's much simpler; loses the per-gate split/merge granularity but matches what's actually evaluable per function call (multiple gates per function become one gF datum per call).
- `gF.nF == oF.nF` — gate aggregate is indexed by the function it gates, not by its own name. So `Gw_[i]` is "total gate value contribution of function i". No new namespace.
- gate body's nested `traced` calls still go into `oF.call_` as before; the `_cur` is `oF`, not `gF`. gF only collects the gate predicate stats, not the body's calls.
- Per-gate `name` string is dropped — no longer needed since gates aren't individually tracked. If you want it back later for in-function diagnostics, gates could append a thin record to `gF.call_`, but that recovers the complexity we just removed.

**One thing to confirm:** when multiple gates fire in one function call, their `dTT`s have different meanings (different predicates on different data). Summing them into one `gF.dTT` averages incompatible signatures. Two options:

1. Accept the averaging — `gF.dTT` is "this function's overall gate-data signature", coarse but consistent with the per-call aggregation.
2. Skip `dTT` accumulation entirely — `gF` only carries `fw`/`fc` scalars, no signature, and gate-side `cent_TT` doesn't apply. Cleaner if signatures are incompatible.

I'd go (2) unless you want gate-data clustering across functions. Which?

kwcckw: This is having gF per oF, that's mean gate oF is no longer per eval, but it is per function? 

In this case, the only difference between gF and oF is their params are init differently using gTT, gw_, etc rather than wTT , fw_, etc?

With this structure, how do we deal with the eval within the function? It will be calling each oF.gF for all evals?

boris-kz: we need to treat evaled block as a function, probably can be lambda

kwcckw: Okay, so each evaled block must be a function because we may have multiple blocks within a caller function, and we need the block function to trace them for eval purpose, else we can't differentiate them since we only have a single gF per oF.

boris-kz: probably, I will check tomorrow

kwcckw: Alright, I will try to include the runtime AST tester first.

PR #135, opened 2026-05-05
5/5/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -160,11 +165,16 @@ def inner(*a, **kw):
                 if oF.fw*L > ave*(oF.fc*L):
                     sum2F(tree, oF); oF.wTT = cent_TT(getattr(oF,'rTT',oF.dTT), oF.r); oF.w += np.mean(oF.wTT)
                 tree = flat_(gF); L=len(tree)-1
-                if gF.fw*L > ave*(gF.fc*L): sum2F(tree, gF); gF.w += np.mean(gF.wTT)  # no cent_TT?
+                if gF.fw*L > ave*(gF.fc*L): 
+                    sum2F(tree, gF); gF.w += np.mean(gF.wTT)  # no cent_TT?
+                    for gf in tree:  # add gated functions' cost, computed from ast
+                        fc = get_wc(os.path.basename(__file__),gf.name)
+                        gF.fc += fc
```
kwcckw: I added functionality to get gated's functions wc and add to gF.fc in runtime, but this is actually still static because it's retrieved from the code via file path. 

With current per oF's gF, we not really need to explicitly split or merge gF?

boris-kz: Split and merge will be for code blocks, headed by oF. gF provides the data that determines it

frame_2D_alg/agg_recursion.py
```diff
@@ -151,7 +156,7 @@ def traced(func):
         def inner(*a, **kw):
             _CoF = CoF._cur.get()
             oF = CoF(nF=onF_.index(func.__name__), root=_CoF)  # gF auto-created
-            gF = oF.gF
+            gF = oF.gF; gF.name = func.__name__  # add name to retrieve cost via runtime ast
```
boris-kz: the name is oF.nF

frame_2D_alg/agg_recursion.py
boris-kz: so yes, the costs are static, we just need to add Gc_, similar to Fc_
that G is confusing, lets use E instead, for eval

boris-kz: eval cost,gain,distr, ||oF_, ecost = eval complexity, egain = evaled block w - default block w, 
eeval = egain - ecost * ave?
Ec_ = [3,3,4,2,1,5,5,4,1,4,4,4,2,2,1.1]  # placeholders
Ew_ = copy(Ec_)  # ave gain/eval, init = cost
ETT_= [deepcopy(wTT) for _ in range(16)]  # for more precise eeval?

boris-kz: Comment vs. implementation mismatch.** Your comment says:
> gain = evaled_block_w - default_block_w

i.e. value of *having the eval* = `(block run with eval) - (block run without eval)`. That's the savings from filtering out low-value calls, net of eval cost.

But `gate()` currently computes:
```python
g.w = V - ave*cost
```
which is just the predicate deviation. When the gate blocks a low-V call, this records a *negative* w — but per the comment, blocking a low-V call should be a *positive* gain (eval saved you the body cost). Sign is inverted on the failure branch.

If the comment is the intended semantics, gate() needs:
```python
# pass: body ran, gain = predicate value - eval cost
# fail: body skipped, gain = saved body cost - eval cost  (= ave_body_cost - cost)
g.w = (V - ave*cost) if V > ave*cost else (Fc_[parent_nF] - cost)
```
or some variant. The exact formula depends on whether you want to credit the eval for blocking (saved cost) or only for passing (gain delivered). Worth deciding before placing markers.

gate calls supposed to use these E-side costs:
```python
if CoF.gate(V, ecN_):  # eval cost from Ec_

Comment (boris-kz) [reply frame_2D_alg/agg_recursion.py]: Per your comment `gain = evaled_block_w - default_block_w`:

- pass: with eval we pay `Ec + body_cost` to realize `V`; without eval we'd pay `body_cost` to realize same `V`. Gain = `-Ec` (eval pure overhead when "run" is the right answer).
- fail: with eval we pay `Ec`, skip body. Without eval we'd pay `body_cost`, get `V` (low). Gain = `body_cost - Ec - V` (saved body cost, lost predicted V, paid eval cost).

```python
@staticmethod
def gate(V, cost, dTT=None):  # gain = evaled_block_w - default_block_w
    _CoF = CoF._cur.get(); gF = _CoF.gF
    Ec = Ec_[gF.nF]; T = ave * cost  # eval cost; threshold
    if V > T: g_w = -Ec                # pass: eval was overhead, body would've run anyway
    else:     g_w = cost - V - Ec      # fail: saved body cost, paid eval cost, lost predicted V
    g = CoF(nF=gF.nF, root=_CoF, fo=0); g.w = g_w; g.fc = Ec; g.span = len(_CoF.call_)
    gF.call_ += [g]; gF.fw += g_w; gF.fc += Ec
    return V > T
```

**Two assumptions worth flagging:**

1. **`V` and `cost` in same units.** Existing evals are `if V > ave*cost`, so `V` and `ave*cost` are commensurable, but `V - cost` (in the fail branch) assumes `V` and `cost` directly comparable, no `ave` scaling. Mixed: I'm using `cost` raw on fail branch (saved body cost = body cost, no ave) and `cost*ave` only as threshold. If you'd rather the savings be normalized too (`ave*cost - V - Ec` on fail), one-char change.

2. **`Ec_[gF.nF]` is per-containing-function, not per-gate-site.** So multiple gates within one function share the same Ec value. Coarse but matches the embedded-gF design. To go per-gate, gates need their own indices (extension of `onF_` or separate registry), and `Ec_` becomes longer.

Aggregate behavior over many calls: `gF.fw` becomes positive when the eval is doing real filtering work (fails on calls with low V relative to body cost). Stays near `-N*Ec` when eval is mostly overhead (high pass-rate, no information). That's the signal you want for split/merge: high `gF.fw` → consider splitting (eval is paying off, finer granularity might pay more); low or negative `gF.fw` → simplify or merge.

kwcckw: > Split and merge will be for code blocks, headed by oF. gF provides the data that determines it

Actually i'm still not quite clear on how we can actually split them. When we merge gF, we are combining multiple gFs into a single block (new gF?)? And when we split it, it's actually splitting a block into smaller functions? And there should be a smallest possible unit where we can't split it further? For example, if the block only has a comp function now, we are no able to split it further?

frame_2D_alg/agg_recursion.py
kwcckw: That's the index of name in onF_?

And if the function is not in onF_, then we can't retrieve the function name for ast computation.

boris-kz: put blocks in onF_ too

frame_2D_alg/agg_recursion.py
boris-kz: we can split functions. for example comp_C_ has multiple evaled forks

kwcckw: > we can split functions. for example comp_C_ has multiple evaled forks

Okay, so actually each evaled forks should be packed in each new function now? That's actually simpler since we always split function into each of the smaller function, instead of just few lines of code.

frame_2D_alg/agg_recursion.py
kwcckw: > put blocks in onF_ too

Okay, then this requires us to place all functions in onF_ instead of the few selected as in the current code.

boris-kz: most functions don't need to be evaluated individually, and some functions have individually evaluated blocks.
I think the gain/cost of adding a function def is different from that of adding eval?

Splitting and merging is similar to connectivity clustering over the code

frame_2D_alg/agg_recursion.py
boris-kz: why does it have to be a function?
I think the gain/cost of adding a function def is different from that of adding eval

frame_2D_alg/agg_recursion.py
boris-kz: Yes — exactly the same machinery, applied to code as input domain. The mapping:

Data clustering | Code clustering
-- | --
node (PP, G) | function/block (oF)
link (comp result) | gate (gF) — derivative of code data
pairwise comp | comp_F between oFs/gFs (already exists)
connectivity link | sequential adjacency in call_, or shared data flow
match → cluster | high m between two oFs → merge candidate
within-cluster variance → split | cent_TT non-uniform on one oF's call-history → multimodal input → split candidate
boundary (high-contrast) | gate that consistently fires with high gF.w — natural function boundary, don't merge across
centroid cluster | canonical eval pattern that most gates fit; refactor to share
typ_ summarization | already a primitive cluster step (functions binned by nF, summarized)


<p>The domain-specific differences:</p>
<ol>
<li>
<p><strong>Topology is discrete.</strong> Functions don't live in Euclidean space; "proximity" is call-graph adjacency or data-dependency. So the connectivity step doesn't need geometric distance — it uses the call tree directly.</p>
</li>
<li>
<p><strong>Far fewer nodes.</strong> Dozens of functions, not millions of pixels. The clustering loop converges in one pass; no need for sub-clustering hierarchies of code (yet).</p>
</li>
<li>
<p><strong>Output is structural change, not labels.</strong> Data clustering produces named groups. Code clustering has to actually emit modified code — split a function, merge two, factor a common sub-block. That's the part that doesn't have a direct analog and needs separate machinery (AST manipulation, source emission).</p>
</li>
<li>
<p><strong>Two derivation orders coexist.</strong> Function and its gates (oF and gF) are at different derivation orders on the same data. Code clustering operates on both: cluster oFs by their <code>dTT</code>, cluster gFs by their <code>dTT</code>, look at their relationship. This is the dual hierarchy you already have for data (nodes / links) showing up at the meta level.</p>
</li>
</ol>
<p>So your read is right: split/merge of evals is connectivity clustering over code, with <code>comp_F</code> as the comp primitive and the call tree as the connectivity substrate. The README claim that the system should "cross-comp and cluster code elements" is exactly this — and the recent CoF work has been building the substrate to do it. The remaining gap is just the structural-output step (3).</p>
<p>Was there something specific you wanted to settle on this front, or is this the framing-check before continuing to use-site instrumentation?</p></body></html>

frame_2D_alg/agg_recursion.py
boris-kz: Chat: A useful split is:

```python
Cdef   # static cost of adding / keeping / specializing a code unit
Ccall  # runtime call overhead, usually small
Cbody  # runtime cost of the function or block body
Ceval  # runtime cost of evaluating a gate condition
Wbody  # gain if the body runs
Weval  # marginal gain of having evaluated run vs skip
```

So:

```python
function def value != function execution value != eval gate value
```

`function def` value is amortized code value:

```python
Wdef = sum(expected_call_value_) - Cdef
```

It is about whether the code unit is worth having as a reusable operation or specialization.

`eval gate` value is selection value:

```python
Weval = value(selected_policy) - value(default_policy) - Ceval
```

It is only positive if the gate changes decisions often enough to compensate for the cost of evaluating the condition.

In the current `CoF`, `traced()` auto-creates `gF` for every traced function and maps it to `Ec_/Ew_/ETT_`, while `gate()` then uses the current function’s `gF.nF`. That forces:

```python
function id == eval id
```

but that is not valid here. `onF_` mixes whole functions such as `cross_comp`, `cluster_N`, `vect_edge`, `trace_edge`, `frame_H`, plus imported `comp_slice` / `slice_edge`, while actual selective units are often internal blocks such as `cross_comp:get_exemplars` and `cross_comp:cross_comp`. 

I would make `CoF.traced` create only a function-scope node:

```python
# function accounting only
oF = CoF(nF=f_id, root=_CoF)
```

and remove automatic per-function `gF` creation from `CoF.__init__` / `traced`.

Then make gates explicitly block-indexed:

```python
evalB_ = [
    'cross_comp:get_exemplars',
    'cross_comp:agg+',
    'comp_N_:proj_N',
    'comp_N:comp_H',
    'comp_N:sub_comp',
    'comp_F:sub_comp',
    'sum2G:cluster_N',
    'sum2G:cluster_C',
    'sum2G:Bt_cross_comp',
    'vect_edge:slice_edge',
    'vect_edge:comp_slice',
    'vect_edge:trace_edge',
    'trace_edge:wrap_G',
    'frame_H:accept_tile',
    'frame_H:cross_comp',
    'frame_H:expand_lev',
]
```

Then `Ec_/Ew_/ETT_` become eval-block arrays, not function arrays.

For gate accounting, define value relative to the default policy. In your current logic, default seems to be “run body unless gate proves it should be skipped”. Then:

```python
Vbody = gain - ave * body_cost

if gate_passes:
    Weval = -Ceval              # ran anyway, gate was overhead
else:
    Weval = -Vbody - Ceval      # saved negative body value, paid eval
```

If default is “skip body unless gate proves it should run”, then:

```python
if gate_passes:
    Weval = Vbody - Ceval       # gained body value, paid eval
else:
    Weval = -Ceval              # skipped anyway, gate was overhead
```

So `gate()` needs either a `default` flag or separate gate types:

```python
def gate(nB, gain, body_cost, dTT=None, default_run=True):
    ...
```

The important part is that `body_cost` belongs to the block being selected, while `Ceval` belongs to the gate itself. Neither is the cost of adding the enclosing function definition.

So the structure should be:

```python
Defc_, Defw_     # static code definition / specialization value
Fc_, Fw_, FTT_   # function-scope execution/accounting value
Bc_, Bw_, BTT_   # selectable block body value
Ec_, Ew_, ETT_   # eval-gate value
```

Most functions only need `Fc_/Fw_` scope accounting. Only blocks that are optional, recursive, expensive, or data-dependent need `Bc_/Ec_`.

This also fixes the conceptual mismatch where a function with several gated blocks currently has only one `gF`, while a mandatory helper function gets an eval entry even though evaluating whether to run it is meaningless.


Yes. They should be accounted separately.

In current `CoF.traced`, every wrapped function call creates both an `oF` and an embedded `gF`, then appends `oF` to the parent call tree and `gF` to the parent gate tree. That makes gate evaluation implicitly one-to-one with function calls, which is too coarse. `CoF.gate` is already a different mechanism: it scores the value of running an eval that may skip a block, using eval cost `Ec_` separately from body cost/gain. Those two should not share the same unit of valuation. 

I’d separate three things:

1. **Function definition / function boundary**
   This is structural. Its value is amortized: reuse, cleaner credit assignment, and ability to assign separate weights to a recurring code pattern. Its cost is call overhead, trace overhead, and possible fragmentation of credit assignment. It should not be re-evaluated per data instance unless the whole function is optional.

2. **Function call / body execution**
   This is normal `CoF` accounting: body gain minus body cost, including child calls. Most helper functions should be transparent: either folded into the parent `oF`, or traced only for statistics, not individually gated.

3. **Eval / gate block**
   This is a control decision. Its gain is not the block’s output value; it is the value of changing the default decision. With default “run body”, roughly:

   ```text
   pass: V_gate = - eval_cost
   fail: V_gate = saved_body_cost - lost_body_gain - eval_cost
   ```

   That is very different from:

   ```text
   V_func = body_gain - body_cost
   ```

So yes: adding a `def` and adding an `eval` have different gain/cost semantics.

The practical change I’d make in `CoF` is: don’t auto-create a `gF` for every traced function. Create gate records only when an actual block eval happens. A function can then have zero gates, one gate, or many gates.

Conceptually:

```python
CoF.traced(func):
    # create oF only: execution / attribution record
    ...

CoF.gate(block_id, gain, body_cost, eval_cost=None, dTT=None):
    # create gF lazily and append one gate record for this block
    ...
```

And `Ec_`, `Ew_`, `ETT_` should probably be keyed by block identity, not just `nF`:

```python
Ekey = (oF.nF, block_id)
```

That lets you represent things like:

```text
cross_comp:get_exemplars
cross_comp:agg_recursion
comp_N_:proj_N
sum2G:cluster_C_vs_cluster_N
```

while leaving most functions without individual eval accounting.

So the revised model is:

```text
F weight = value of executing / keeping a function as a reusable process unit
E weight = value of evaluating a condition before an optional block
```

The current `Fc_/Fw_/FTT_` can stay function-level. `Ec_/Ew_/ETT_` should become block-level. The explicit gate sites should be the conditional blocks already present in places like `cross_comp`, not every function wrapper.

frame_2D_alg/agg_recursion.py
boris-kz: But the comparison part here is actually evaluation: what happens if the next operation in sequence | fork tree is called or not called? that has to be evaluated on the level of whole function or block?

Claude: this is a different comparison primitive than `comp_F`. Two distinct kinds:

**Similarity comp** (`comp_F`): pairwise, "how alike are these two ops in what they do?" Output = m,d between existing things. This is what data comp does.

**Counterfactual comp** (gate): per-decision, "what happens if next op runs vs doesn't?" Output = gain, the deviation between two regimes (with/without). The current `gate.w = evaled_w - default_w` formula IS this comparison.

For data clustering you mostly use similarity; for code clustering you need both. Similarity tells you which ops are doing the same job (merge into one parametric form). Counterfactual tells you which ops are pulling their weight (drop, simplify, or specialize).

**Granularity:** the comparison resolves at whatever level you have a gate. Currently:

- **per-function** (embedded `oF.gF` aggregate): "is this whole function paying off?" Coarsest. One gain estimate per function call.
- **per-gate-site** (each `gate()` creates a sub-CoF): "is this specific eval-point paying off?" Finer. One gain per `if`.
- **per-sub-call** (sub.dTT inside parent.call_): "is this specific call to next-op paying off?" Finest. Already free from additive structure: `sub.dTT` = marginal contribution of that sub-call to parent. `parent.dTT - sub.dTT` = parent without sub.

The third one is interesting and underused. It's a gain measure without needing a gate at all — any sub-oF in `call_` already carries its marginal contribution as its `dTT`. To ask "should we keep calling X here", look at X's accumulated `dTT` weighted vs cost: `vt_(X.dTT) > ave * X.fc`. That's the same shape as the gate eval but without instrumenting an `if`.

So you have three nested levels of counterfactual comparison, all using the same arithmetic:

1. function-level: should this function exist at all? (aggregate over all its calls)
2. gate-level: should this gate exist? (per-eval, gain formula)
3. call-level: should this sub-call happen here? (free, from sub.dTT)

For sequence-level "what if step B is skipped between A and C", level 3 already gives it: B's contribution to its parent is `B.dTT`. To estimate downstream knock-on (skipping B might change C's input), you'd need to track which sub-calls of C consumed B's output — i.e. a data-dependency edge between sub-oFs in the same call_. That's not currently recorded.

**So:** the comparison is per-decision, and the unit IS the function/block-with-its-own-gate. Whole-function level is the *coarsest* useful resolution; gate-level is finer; sub-call-level is finest and already implicit. The split/merge logic decides between these resolutions per measured variance — same as in data clustering, you go finer where there's signal, coarser where there isn't.

Was your question about which level to operate at, or about whether the counterfactual primitive is fundamentally different from comp_F?

frame_2D_alg/agg_recursion.py
kwcckw: > most functions don't need to be evaluated individually, and some functions have individually evaluated blocks.
I think the gain/cost of adding a function def is different from that of adding eval?

> Splitting and merging is similar to connectivity clustering over the code

So basically we still need the gate oF per eval, and since Ec_, Ew_ and ETT_ is per block, those should be packed per gate oF.


For the additional cost/weight for each of the ops below, they should be mapped to Fc_, Fw_, Ec_ and Ew_? Which mean we need to add more cost and wieght?
      
      Cdef   # static cost of adding / keeping / specializing a code unit
      Ccall  # runtime call overhead, usually small
      Cbody  # runtime cost of the function or block body
      Ceval  # runtime cost of evaluating a gate condition
      Wbody  # gain if the body runs
      Weval  # marginal gain of having evaluated run vs skip

boris-kz: yes it's all per oF, but not per terminal function

Ec_, Ew_ and ETT_ are globals, parallel to Fc_, Fw_, FTT_, and Z.typ_ 

this is all redundant:

  Cdef   # static cost of adding / keeping / specializing a code unit
  Ccall  # runtime call overhead, usually small
  Cbody  # runtime cost of the function or block body
  Ceval  # runtime cost of evaluating a gate condition
  Wbody  # gain if the body runs
  Weval  # marginal gain of having evaluated run vs skip

kwcckw: I see, so Ec_, Ew_ and ETT_ are primarily used for adding eval, while  Fc_, Fw_, FTT_ are used for adding func def. 

For goF, their weight of cost should be init of Ew_ and Ec_ since they will be used as gate to eval block pf code.

> Splitting and merging is similar to connectivity clustering over the code

This will be done across goFs only, and it should be all under the same oF (similar with sub+)? Since we can't merge or split across root function?

boris-kz: Fc_, Fw_, FTT_ are for call evals, Ec_, Ew_, ETT_ are for call eval evals, they can change the code structure. Merge or split will be in ffeedback, to change the process for new inputs

kwcckw: > Merge or split will be in ffeedback, to change the process for new inputs

Okay, so we might have code clustering or split in the ffeedback, and is the diagrams below correctly illustrated the split and merge process? From GPT 5.5:

Merge:
Image

Split:
Image

boris-kz: yes, roughly

boris-kz: Sub-call granularity is already there for free — every traced function call inside a body lands in `parent.call_` as a sub-oF with its own `dTT`. That sub.dTT IS the marginal contribution. Conditional comparison on that level: `vt_(sub.dTT) > ave * sub.fc` answers "was this sub-call worth it?" without any new wrapping.

So three already-free levels:

1. **Whole function** — aggregate gF over many calls, "should this function exist?"
2. **Gate site** — per-`if` gain via `gate()`, "should this branch exist?"
3. **Sub-call within body** — sub.dTT directly, "should this sub-call happen here?"

Below that, no — sub-statement contributions need wrapping. There's no measurement substrate without a named scope. Three options for going finer:

**(a) Statement marker.** `with oF.step('name'):` around the statements you care about. Manual, but precise. Same shape as gate(), different semantics (no predicate, just contribution recording). Works for things like the accumulator updates inside `comp_N_`'s loop.

**(b) AST auto-wrap.** `traced` parses function source, wraps each `Stmt` node in a synthetic CoF. Automatic but invasive — generated wrappers obscure source, fragile against control flow (loops, exceptions, `return`).

**(c) Operator overloading on dTT.** Override `+=` on the dTT arrays so each mutation records (line, contribution) into a side log. Only catches explicit dTT updates, not arbitrary logic. Cute but narrow.

Practically the floor is sub-call — that's the natural unit because functions are already the unit of *named operation* in the code. Going below sub-call into sub-statement gives diminishing returns: at that granularity, the contributions are usually too small to discriminate from noise, and the semantic unit blurs (a single `np.subtract` line has no meaningful "operation identity" separate from its surrounding context).

So I'd say: yes, you've already got 3 levels of granularity for conditional comparison without any new wrapping. Going finer requires markers (a) and probably isn't worth it until you've exhausted the signal at the existing 3 levels.

The harder question is the inverse — going *coarser*. Above whole-function: clusters of functions that should be merged. That's where similarity comp (`comp_F` between oFs) and conditional gain (does keeping both pay off vs. one combined) start interacting. That's the meta-level where split/merge decisions actually live, and it's where the existing machinery (typ_, comp_F, cent_TT) starts mattering for code rather than just data.

boris-kz: <html><head></head><body><p> they're the dual mechanism you already have for data, just applied to code:</p>
<ul>
<li><strong>similarity clustering</strong> (<code>comp_F</code> over oFs/gFs) = <code>cluster_N</code> analog: finds groups of code units doing related work</li>
<li><strong>conditional/performance</strong> (gF.fw, sub.dTT contribution) = <code>cluster_C</code> analog: ranks units by gain, picks representatives, decides keep/drop</li>
</ul>
<p>The decision space is 2D:</p>

  | high gain | low gain
-- | -- | --
high sim | merge into parametric form (both pay off, factor common) | drop the duplicate (one does the work, other replicates)
low sim | keep both separate (distinct valuable work) | drop or specialize


<p>This maps onto your existing pipeline directly. Similarity clusters define candidate groups; within-group performance evaluation refines (which member is representative, which is redundant). Outside-group performance defines the boundary (high-gain-low-sim functions = type boundaries, don't merge across).</p>
<p>There's one subtlety from the dual oF/gF structure:</p>
<ul>
<li><code>comp_F</code> over <strong>oFs</strong> → similar in <em>what they process</em> (data signature). Merge candidates by data-side similarity.</li>
<li><code>comp_F</code> over <strong>gFs</strong> → similar in <em>what they decide</em> (predicate signature). Merge candidates by control-flow similarity.</li>
</ul>
<p>Two functions can process similar data but filter very differently (or vice versa). Both signals matter and they don't always agree. That's just the dual hierarchy showing up at the meta level — exactly as nodes vs. links carry distinct information about a graph.</p>
<p>Concrete combined pipeline for code:</p>
<ol>
<li><code>comp_F</code> over <code>typ_</code> (or all oFs) → pairwise m,d. Existing.</li>
<li><code>cluster_N</code> over those → similarity groups. Existing, applied to code.</li>
<li>Within each group, sort by <code>gF.fw / gF.fc</code> (gain rate) → identify representative.</li>
<li><code>cluster_C</code> to find canonical eval prototypes (centroid of similar gates) → refactor candidates.</li>
<li>Boundary check: high-d pairs across cluster boundaries = stable function-type splits, don't merge.</li>
<li>Within-function: <code>cent_TT</code> on call-history dTT → if multimodal, split candidate (same as today's data-side splitting).</li>
</ol>
<p>The whole thing reuses the existing clustering functions — they just operate on CoFs (code) instead of CNs (data). Same arithmetic, same flow. The only new piece is step 6's <em>output</em> — translating "split here" into actual code modification.</p>
<p>So combining isn't a separate machinery, it's the existing two-stage pipeline applied to code with both similarity and gain as inputs. Which is what you'd want if you take seriously that code is just another input domain — same algorithm, different substrate.</p>
<p>Want me to sketch what step 1+2 would look like operating on <code>typ_</code>?</p></body></html>

kwcckw: > Practically the floor is sub-call
> So three already-free levels:

    Whole function — aggregate gF over many calls, "should this function exist?"
    Gate site — per-if gain via gate(), "should this branch exist?"
    Sub-call within body — sub.dTT directly, "should this sub-call happen here?"

Yes, i would think the same since these 3 levels should be already sufficient to eval at different levels.

> The harder question is the inverse — going coarser. Above whole-function: clusters of functions that should be merged. That's where similarity comp (comp_F between oFs) and conditional gain (does keeping both pay off vs. one combined) start interacting. That's the meta-level where split/merge decisions actually live, and it's where the existing machinery (typ_, comp_F, cent_TT) starts mattering for code rather than just data.

So i guess we should start to draft on the merge and split ops first?

When we split block, we should be adding new goF to oF, that's information should be packed in oF. In the next run, we might not able to add new eval in the code during the runtime? That's mean we need to check each line or block of code to see if it's aligned with the goF in oF, and use the goF for eval purpose?

kwcckw: > The whole thing reuses the existing clustering functions — they just operate on CoFs (code) instead of CNs (data). Same arithmetic, same flow. The only new piece is step 6's output — translating "split here" into actual code modification

Okay, so we actually can reuse the same comp and clustering functions. But initially we can try with a simpler version of comp and cluster function so that tit won't messed up the existing code? 

Also another problem is data comp or cluster requires eval from golF, but that doesn't apply to the comp.or clustering of CoFs?

boris-kz: Not sure we can separate gF from oF: the latter collects the data for the former
I had claude to draft it and added it to meta_code, not fully reviewed

kwcckw: Okay, but I think we at least can pack gFs in oF? But this may be a problem if we are able to merge or split across different functions, then gFs should be per Z.

boris-kz: gFs have always been packed in oFs

kwcckw: Okay, but do we ever split or merge across oFs? Or that is strictly within functions of a single oF only?

PR #136, opened 2026-05-06
5/6/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -7,50 +7,56 @@
 # pairwise similarity over function-type summaries:
 def comp_typ_(typ_):
     L_ = []
-    for _T,T in combinations([t for t in typ_ if t], 2):
-        dF = comp_F(_T,T)  # already gives m,d via comp_derT(dTT)
+    for _T,T in combinations(typ_, 2):
+        dF = comp_F(_T,T)  # already gives m,d via comp_derT(dTT) (most of the oF.dTT is empty, and we only fill rTT?)
```
kwcckw, line 11: By looking at the draft here, there are several issues:

1. Most of the CoF.dTT is empty and we only pack rTT, so we might need to compare with additional attrs, or to make sure all CoF.dTT is filled?

frame_2D_alg/meta_code.py
```diff
@@ -7,50 +7,56 @@
...
+            for T1,T2,dF in L_:
+                _T = T2 if T1 is T else (T1 if T2 is T else None)
+                if _T and _T not in Clust and dF.m > ave*dF.r:
+                    Clust.add(_T); front += [_T]
+        seen |= Clust; Clust_ += [list(Clust)]
+    return Clust_
 
 # nF similarity → "alike", gain → "useful"
-def gain_rate(T): return Ew_[T.nF] / (Ec_[T.nF] or eps)
+def gain_rate(T): return Ew_[T.nF] / (Ec_[T.nF] or eps)  # use only global?
```
kwcckw, line 31: 2. The gain rate should not be solely based on global Ew_ and Ec_? It should includes runtime T.fw and T.fc?

frame_2D_alg/meta_code.py
```diff
@@ -7,50 +7,56 @@
...
+    clust_ = cluster_O(typ_)
+    for clust in clust_:
+        ranked = sorted(clust, key=gain_rate, reverse=True)
         rep = ranked[0]; rest = ranked[1:]
         if len(rest):
             high = [t for t in rest if gain_rate(t) > ave]
             low  = [t for t in rest if gain_rate(t) <= ave]
-            if high: out += [('merge_factor', rep.nF, [t.nF for t in high])]  # both pay off
-            if low:  out += [('drop_dup', rep.nF, [t.nF for t in low])]       # rep does the work
+            if high: out += [('merge_factor', rep.nF, [t.nF for t in high])]  # both pay off      (similar and high gain, merge them?)
```
kwcckw, line 44: 3. So this section is still following the prior 2D criterion?

When it's similar and high gain, we merge those Ts, and leave them when Ts are similar but have low again?

For those leftover Ts (non clustered Ts), if their gain is lower than ave, we should split them?

boris-kz: Similarity and gain actually work in the opposite direction, split value = difference + gain: they both increase the value of resolution

frame_2D_alg/meta_code.py
boris-kz: Use rTT or dTT

frame_2D_alg/meta_code.py
boris-kz: Ideally a weighted product of global and local, depending on decay rate?

kwcckw: But what would be the decay rate here? It should decay based on m or d?

frame_2D_alg/meta_code.py
kwcckw: Then gain is needed for split only? But i thought gain is the amplification factor for both split and merge?

frame_2D_alg/meta_code.py
kwcckw: Yes, but most dTT are just zeros and there;s nothing to compare?

Wee only add rTT in cluster N, C and P. All the other functions' dTT are empty.

frame_2D_alg/meta_code.py
boris-kz: Sorry, that's dumb comment

kwcckw: So we need gain for both merge and split, which mean we need to apply gain for similarity (m) and difference (d)?

High similarity = merge
high difference = split
low on both = leave it?

boris-kz: Just read what you wrote. How can one factor: gain, work in the opposite directions at the same time?

kwcckw: Okay, sorry, my prior understanding was flawed because if gain is high, we may get merge or split at the same time.

So we should apply gain only after we identify if Ts are similar or internally different. 

For similar Ts, high gain = merge, low gain = drop,
If T is internally different, high gain = split, low gain = leave it or ignore?

boris-kz: you are not making any sense

frame_2D_alg/agg_recursion.py
```diff
@@ -177,7 +177,7 @@ def gate(gain,cost, dTT=None):  # gain = evaled_block_w - default_block_w
         if dTT is not None: g.dTT = dTT
         gF.call_+=[g]; gF.fw += gw; gF.fc += Ec
         return gain>T
-    def __bool__(f): return bool(f.N_)
+    def __bool__(f): return bool(f.N_)  # this bool is wrong now? N_ in CoF is mostly empty
```
boris-kz, line 180: should be call_

frame_2D_alg/agg_recursion.py
```diff
@@ -373,15 +373,17 @@ def comp_H(_Nt,Nt, Link):  # tentative pre-comp
 
 def comp_F(_F, F, ir=0, rL=None):
 
-    ddTT = comp_derT(_F.dTT[1],F.dTT[1]); r=(_F.r+F.r)/2; m,d = vt_(ddTT,ttF); r+=ir
+    ddTT = comp_derT(getattr(_F, 'rTT',_F.dTT)[1],getattr(F, 'rTT',F.dTT)[1]); r=(_F.r+F.r)/2; m,d = vt_(ddTT,ttF); r+=ir
```
boris-kz, line 376: rTT is an evaluand, not a comparand

frame_2D_alg/agg_recursion.py
```diff
@@ -373,15 +373,17 @@ def comp_H(_Nt,Nt, Link):  # tentative pre-comp
 
 def comp_F(_F, F, ir=0, rL=None):
```
boris-kz, line 374: you can't take these LLM suggestions literally. 
it makes no sense to compare data between the calls, the calls themselves do the comparison

kwcckw: > rTT is an evaluand, not a comparand

Okay ,sorry, then everything here is actually no longer relevant, that's why it's not making sense at all.

To compare calls, we need call-specific params, but we can't comp and eval with just rTT or dTT since it's from data, so we need to include params that summed from callees from sum2F, `fw, m and d`, which is better in representing the oF?

    elif fO: F.call_ = n_; F.fw = Fw; F.m, F.d = vt_(TT)

boris-kz: That whole discussion was about comparing code, not data.
it's a completely different process

kwcckw: Okay, if it's just strictly code, i think the most relevant ones are those AST derived params of function, and we can compute and pack them in oF during the oF initialization, and then later use them for the comparison and clustering here.

boris-kz: not oF, it will be mapping Fc_

kwcckw: That Fc_ mapping is the total cost using vt_ as a single unit, i think we need more params for comparison purpose? Such as call count, loop count, function count, gate count, depth score, etc?

boris-kz: yes, that's just nested Fc_

kwcckw: Yes, so we need to decide what kind of ast computed feature that needs to be included in the nested Fc_.

At this point, i think it could be an array of features, so that we can compare it the same way as dTT?

Some suggested features, not sure if we need to normalized those count?
    
    op_cost                 # weighted op count / vt_ base
    
    # --- loops ---
    for_count               # number of for-loops
    while_count             # number of while-loops
    loop_count              # for + while (convenience)
    comprehension_count     # list / set / dict / generator comps
    
    # --- conditionals ---
    if_count                # if-statements (not ternary)
    ternary_count           # inline if-expressions  (a if c else b)
    branch_count            # total branches (if + elif + else bodies)
    bool_op_count           # and / or nodes
    compare_count           # comparison expressions (<, ==, >, …)
    
    # --- calls ---
    call_count              # all function / method calls
    gated_call_count        # calls that appear directly inside an if-test
    recursive_call_count    # calls to the function itself (self-recursion)
    inner_func_count        # nested def statements
    
    # --- data ---
    int_literal_count       # integer numeric literals
    float_literal_count     # float numeric literals
    numeric_count           # int + float (convenience)
    string_literal_count    # string constants
    list_literal_count      # explicit [] literals
    dict_literal_count      # explicit {} dict literals
    
    # --- names / scope ---
    unique_names            # distinct local variable names
    aug_assign_count        # += -= *= …
    global_count            # global statements
    nonlocal_count          # nonlocal statements
    
    # --- attributes / subscripts ---
    attr_access_count       # obj.attr reads/writes
    subscript_count         # obj[key] accesses
    
    # --- structural ---
    max_nesting_depth       # deepest indentation level (if/for/while/with/try)
    return_count            # return statements
    assert_count            # assert statements
    exception_handler_count # except clauses
    lambda_count            # lambda expressions
    yield_count             # yield / yield from
    arg_count               # number of function parameters
    
    
Also just count may not be sufficient, we may need to explore some additional features, the possible features:

| Feature class | What it discriminates | Description |
|---|---|---|
| Counts | size / volume | Raw totals of ops, calls, branches — how much a function does |
| Ratios | character / style | Normalized relationships between counts — how a function does it, independent of size |
| Entropy | generalist vs specialist | Diversity of op types — whether a function is one focused thing or many mixed things |
| N-grams | execution pattern | Ordered sequences of op types — the recurring structural idiom a function follows |
| Shape | where cost is concentrated | Distribution of ops across the function body — front-loaded setup vs back-loaded compute |
| Lifetimes | stateful vs pipeline | How long variables survive — accumulated mutable state vs short-lived pass-through data |
| Fan-in/out | coordinator vs leaf | How many other tracked functions this calls or is called by — orchestrator vs worker |

boris-kz: We don't need detailed costs now, it's a generic problem, there will be tools for that.
Most basic merge / split is: `_Q,Q` ops comparison, where Q is function or block. Comparison is just binary matching for aligned ops, and to merge the Qs we need to add if fork for each miss. Or remove in splitting each Q that has if forks. The cost of if is fixed: Ec_[i], while the gain is removing the overhead of having a separate Q: Fc_[i], amortized over usage = oF.w. That must be a very common problem in code optimization?

boris-kz: Yes, this is a well-known problem with a few distinct names depending on the community:

- **Anti-unification** — the formal operation: given two terms, find the most specific generalization that subsumes both, with parameters/branches at the differing positions. Classical in inductive logic programming and program synthesis.
- **Function/procedure merging** — LLVM has `MergeFunctions` (identical) and `MergeSimilarFunctions` (near-identical, dispatched via thunks). Compiler PGO decides merge-vs-specialize based on call counts, exactly your `oF.w` amortization.
- **Code clone refactoring** — the software-engineering side; "extract method" and "introduce parameter" are the merge direction, "inline + specialize" is the split.

So yes, the cost model you described (fixed branch cost vs amortized per-Q overhead, weighted by usage) is the standard PGO cost model. Nothing new there.

Two things worth flagging before you implement:

**1. Predicate source for the `if`.** When you merge Q1 and Q2 with a miss at position i, you need a runtime test that distinguishes which branch to take. The natural source in your code is the input's `typ` field (CF/CL/CC/CN already dispatch this way). Without a clean discriminator, the merged Q must accept the union of input types and the predicate becomes synthetic (a flag passed in by the caller), which leaks complexity outward and partly cancels the merge gain.

**2. Binary aligned matching loses parameterization opportunities.** `a + b` vs `a - b` is a full miss at op level → forces `if`. But the operands match; only the operator differs. Anti-unification would lift just the operator to a parameter (or sub-`if`), keeping more shared structure. With strict op-level binary match, you over-branch. Worth deciding whether at a miss you (a) always insert `if` around the whole op, or (b) recurse into operands and only branch at the deepest divergence.

**The deeper symmetry:** this is the *code dual* of your data fork structure. Clustering merges data instances under a shared parent with a `typ` attribute selecting variants; merging Qs creates a shared procedure with an `if` predicate selecting variants. Same operation, different substrate. Which means the merge/split eval should reuse `vt_`-style match/cost arithmetic, not be a separate mechanism.

Question: for the predicate source — are you assuming the discriminator is always already present (input `typ`), or do you want meta_code to also propose synthetic discriminators when no natural one exists?

boris-kz: So we don't need to do this optimization ourselves, we just need to keep the data that's needed for it:

`Ec_, gF.w*c, Fc_, oF.w*c `

boris-kz: For splitting:

keep the (oF, gF) pairing and just nest it. Each gate's sub-CoF becomes a full oF itself (`fo=1`, with its own embedded gF), and the body block executes with `_cur` set to that sub-oF. Then everything inside the body — sub-traced calls and nested gates — accumulates into it via the same mechanism that already populates the top-level oF/gF tree.

Concretely:

In `gate()`, create `g` with `fo=1` so it owns its own gF. Wrap the body so `_cur` swaps to `g` during execution. Cleanest is context-manager form, since it matches Python's `if`-block syntax that gates are meant to replace:

```python
with gate(gain, cost) as g:
    if g:
        # body — sub-traced calls land in g.call_, sub-gates in g.gF.call_
```

`__enter__` swaps `_cur` to `g` on pass / no-op on fail; `__exit__` restores. Now per branch position you get the same data shape as per-function: `g.fc, g.fw, g.c, g.gF.fc, g.gF.fw` — recursive, no new abstraction.

For split: walk oF's gF.call_; each entry is now a (sub-oF, sub-gF) pair carrying its branch's amortized cost vs gain. Dead branch = `g.fw * g.c << g.fc * g.c`. Same arithmetic as the merge eval, just one level deeper in the tree.

The merge/split optimizer then operates on a uniform recursive (oF, gF) structure where the only distinction between "function" and "branch within function" is depth — which is the right invariant given that Q is "function or block" in your framing.

kwcckw: > So we don't need to do this optimization ourselves, we just need to keep the data that's needed for it:
> 
> `Ec_, gF.w*c, Fc_, oF.w*c `

Okay, so Ec_, Fc_ should be updated from code level feedback. Then w and c is from data, where w is the mean of wTT, and c is the call count?

boris-kz: w and c is from oF.dTT.  fw and fc is from oF.wTT, I think we need to combine them

draft:

```python
def merge(_Q,Q):  # combine aligned ops, if-fork per miss, recurse on matched

    M = CoF(nF=Q.nF, root=Q.root)
    for _f,f in zip(_Q.call_, Q.call_):
        if _f.nF == f.nF:  # match
            sub = merge(_f,f) if _f.call_ else CoF(nF=f.nF, root=M, c=_f.c+f.c, fw=_f.fw+f.fw, fc=_f.fc)
            sub.root = M; M.call_ += [sub]
        else:  # miss: gate selects between alternative bodies
            g = CoF(nF=_f.nF, root=M, fo=0); g.call_=[_f,f]; g.fc=Ec_[_f.nF]; g.c=_f.c+f.c
            _f.root=g; f.root=g; M.call_+=[g]; M.gF.call_+=[g]
    M.c=_Q.c+Q.c; M.fw=_Q.fw+Q.fw; M.fc=Fc_[Q.nF]
    return M

def split(Q):  # invert merge at most cost-unamortized gate, return specialized Qs

    g = max(Q.gF.call_, key=lambda g: max(b.fc*b.c - b.fw*b.c for b in g.call_))
    Qs = []
    for b in g.call_:
        S = CoF(nF=Q.nF, root=Q.root); S.call_=[b if f is g else f for f in Q.call_]
        S.gF.call_=[x for x in Q.gF.call_ if x is not g]
        S.c=b.c; S.fw=sum(f.fw for f in S.call_); S.fc=Fc_[Q.nF]
        Qs += [S]
    return Qs
```

Notes on the choices:

**merge:** matched positions recurse if they have nested `call_` (deeper structure may still differ → deeper gates), or just sum counts/gain at leaves. Missed positions create a gate `g` with `fo=0` (it's a dispatch container, the alternatives `_f, f` carry their own gFs). Gate's `c` = combined call frequency (gate fires every time merged Q runs); each branch retains its source `.c` so split can later read per-branch usage.

**split:** picks the gate where one branch most fails to amortize its body cost (`fc*c > fw*c`). Returns one specialized Q per branch — each keeps everything except `g`, replacing it with the chosen branch. Symmetric inverse of merge in arity (1→2 vs 2→1). Optimizer can call recursively if Q has more imbalanced gates.

**Open: zip truncates on length mismatch.** If `_Q` and `Q` have different `len(call_)`, the tail is silently dropped. Since you specified strict positional alignment with terminate-on-miss earlier, you'd typically only merge Qs that already passed an alignment check, so equal length is expected. If not, the loop should bail or pad with miss-gates — flag if you want the latter.

**Open: gate predicate.** `g.call_=[_f,f]` records the two alternative bodies but no actual runtime predicate. The discriminator (input `typ`, flag, etc.) isn't synthesized here — that's the unsolved question from earlier. As-is, `g` is a structural placeholder; making it executable needs the predicate source decided.

kwcckw: Okay, and for the merge part, this section:

     for _f,f in zip(_Q.call_, Q.call_):

Why we are checking the calls in the same position only? It should be all to all?

boris-kz: should be for _f,f in zip(_Q, Q):  # each Q is ops in func or block

these Q are not accidental, each is a logical sequence

kwcckw: Okay, so the sequence matters here, but we may missed out some calls if one length is shorter than another, as pointed by the LLM? 

We probably can have all to all, but the sequence similarity adds
 extra weight to the params?

boris-kz: every miss is a fork?
edited:

Continuing in that style:

```python
def merge(_Q,Q, root=None):  # combine aligned ops, if-fork per miss, recurse on matched

    mrg = CoF(nF=Q.nF, root=root or Q.root)
    for _f,f in zip(_Q.call_, Q.call_):  # call_: op sequence in func or block
        if _f.nF == f.nF:  # match
            mrg.call_ += [merge(_f,f,mrg) if isinstance(_f,CoF) else f]  # if nested oF
        else:  # miss: gate selects alternative bodies
            g = CoF(nF=_f.nF, root=mrg, fo=0); g.call_=[_f,f]; g.fc=Ec_[_f.nF]; g.c=_f.c+f.c
            _f.root=g; f.root=g; mrg.call_+=[g]; mrg.gF.call_+=[g]
    mrg.c=_Q.c+Q.c; mrg.fw=_Q.fw+Q.fw; mrg.fc=Fc_[Q.nF]
    return mrg

def split(Q, root=None):  # invert merge at most cost-unamortized gate

    g = max(Q.gF.call_, key=lambda g: max(b.fc*b.c - b.fw*b.c for b in g.call_))
    Qs = []
    for b in g.call_:
        S = CoF(nF=Q.nF, root=root or Q.root); S.call_=[b if f is g else f for f in Q.call_]
        S.gF.call_=[x for x in Q.gF.call_ if x is not g]
        S.c=b.c; S.fw=sum(f.fw for f in S.call_ if isinstance(f,CoF)); S.fc=Fc_[Q.nF]
        Qs += [S]
    return Qs
```

Two consequences of the leaf-vs-CoF distinction you introduced:

- Leaf ops have no `.c, .fw, .fc` to combine, so matched leaves just pass through (your `else f`). That's clean but means the merged `mrg` only inherits *aggregate* counts from `_Q.c+Q.c` at the top, not from per-leaf accumulation — fine if leaves are identical-by-construction (same nF means same op), since there's nothing to combine.
- In `split`, `sum(f.fw for f in S.call_)` needs the `isinstance(f,CoF)` guard, since matched leaves aren't CoFs and don't have `.fw`.

If `call_` actually mixes leaf and CoF entries (matched leaves stay leaves, matched nested ops become merged CoFs, misses become gate CoFs), the guard pattern propagates anywhere we sum across `call_`. Worth deciding now whether leaves should be wrapped in trivial CoFs at trace time so downstream code stays uniform — costs one allocation per op but removes the type check everywhere.

boris-kz: The placement question maps cleanly onto a timescale distinction:

- **ffeedback** = intra-run, between elevs in one `frame_H`. Adjusts filters (rV, FTT_) — same code, scaled thresholds.
- **merge/split** = inter-run, between full `frame_H` invocations. Adjusts code structure — different ops next time.

So the natural architecture is meta_code as outer loop, agg_recursion as per-run engine:

```
meta_code.__main__:
    while ...:
        frame_H(...)         # populates Z.typ_
        ffeedback updates aves/wTT  (already inside frame_H, intra-run)
        merge/split on Z.typ_  (inter-run, code-level)
        apply changes
        Z.typ_ = []  # reset for next run
```

agg_recursion's `__main__` stays as the single-shot test entry; meta_code's `__main__` owns the self-modifying loop. Imports flow upward only (meta_code imports frame_H, ffeedback, Z, CoF, the typ accumulator). No circular dep.

**The unresolved piece is "apply changes".** Collecting `mrg_, spl_` lists tells you what the optimizer *thinks* should happen, but doesn't actually alter what `frame_H` runs next. Three options:

1. **Source rewrite via AST** — meta_code edits agg_recursion.py, restart. Heavy, risky, but real code modification, which is the README's stated goal.
2. **Parallel dispatch layer** — agg_recursion's `traced` decorator consults a "merged Q registry" maintained by meta_code; when calling `comp_N`, dispatch to merged version if one exists. No source change, fully runtime, but requires every traced call to go through registry lookup.
3. **Offline only** — meta_code prints/serializes refactor candidates, human reviews and edits manually. Simplest, breaks autonomy, but probably the right starting point given how easy it is to get the merge/split logic subtly wrong.

Without picking one, the data collection is open-loop — useful for analysis but not actually self-modifying. Which direction were you thinking? My guess is (3) for now, (2) eventually, (1) only if (2) hits a wall.

Also: `combinations(Z.typ_, 2)` will try to merge unrelated function types (e.g., `comp_N` with `cluster_C`). Pre-filter by structural similarity (op_seq prefix match length above some threshold) or this generates mostly junk merges where every position misses → degenerate gate-soup.

kwcckw: > every miss is a fork?

Yes, but without zip_longest, we are skipping the entire match or miss checking, so we need to use zip_longest in order to pack the longer Q?

kwcckw: > meta_code.__main__:
    while ...:
        frame_H(...)         # populates Z.typ_
        ffeedback updates aves/wTT  (already inside frame_H, intra-run)
        merge/split on Z.typ_  (inter-run, code-level)
        apply changes
        Z.typ_ = []  # reset for next run

With this structure, Z is actually updated outside  frame_H, and we need to parse  updated Z back into frame_H so that it uses the merged or split oF?

PR #137, opened 2026-05-07
5/7/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -29,12 +29,18 @@ def split(Q, root=None):  # invert merge at most cost-unamortized gate
...
+    from agg_recursion import frame_H, imread, trace_func, add_typ_
+    
+    trace_func(vars(agg_recursion))  # add oF tracing
     Y,X = imread('./images/toucan.jpg').shape
     frame_H(image=imread('./images/toucan.jpg'), iY=Y//2-31, iX=X//2-31, Ly=64,Lx=64, Y=Y,X=X, rV=1)
+    add_typ_(Z)   # each call_ in Z.typ_ is the flatten calls of same typ
     mrg_,spl_ = [],[]
-    # add in ffeedback?:
-    for F,_F in combinations(Z.typ_,2): mrg_ += [merge(F,_F)]
+    # add in ffeedback?:  (I think it should be here, ffeedback is per frame_H, and we may call frame_H at higher elevation)
```
kwcckw, line 41: I think this merge and split should be done outside of ffeedback, since ffeedback is one of the call and they maybe merged or split?

frame_2D_alg/meta_code.py
```diff
@@ -29,12 +29,18 @@ def split(Q, root=None):  # invert merge at most cost-unamortized gate
...
+    import agg_recursion
+    from agg_recursion import frame_H, imread, trace_func, add_typ_
+    
+    trace_func(vars(agg_recursion))  # add oF tracing
     Y,X = imread('./images/toucan.jpg').shape
     frame_H(image=imread('./images/toucan.jpg'), iY=Y//2-31, iX=X//2-31, Ly=64,Lx=64, Y=Y,X=X, rV=1)
+    add_typ_(Z)   # each call_ in Z.typ_ is the flatten calls of same typ
     mrg_,spl_ = [],[]
-    # add in ffeedback?:
-    for F,_F in combinations(Z.typ_,2): mrg_ += [merge(F,_F)]
```
kwcckw, line 37: We should merge calls?  But using Z.typ_ here is merging typs, where all calls in typ.call_ are the same typ since they have the same nF? It's just their calling location or sequence will be different. 

So we need to unpack their call_ and merge their calls again?

frame_2D_alg/meta_code.py
```diff
@@ -29,12 +29,18 @@ def split(Q, root=None):  # invert merge at most cost-unamortized gate
...
+    # add in ffeedback?:  (I think it should be here, ffeedback is per frame_H, and we may call frame_H at higher elevation)
+    call_ = [call for typ in Z.typ_ if isinstance(typ, CoF) for call in typ.call_]  # typ.call_ are the oFs with the same typ, unpack to match with others
+    for F,_F in combinations(call_,2): mrg_ += [merge(F,_F)]
```
kwcckw, line 43: Could you clarify the following?

For the same nF merging:

        if _f.nF == f.nF:  # match
            mrg.call_ += [merge(_f,f,mrg) if isinstance(_f,CoF) else f]  # if nested oF

If we have oF from cross_comp, and another oF from nested cross_comp within cross_comp, after the merging, both of their oF will be referencing thesame mrg?


For miss:

      else:  # miss: gate selects alternative bodies
            # not reviewed:
            g = CoF(nF=_f.nF, root=mrg, fo=0); g.call_=[_f,f]; g.fc=Ec_[_f.nF]; g.c=_f.c+f.c
            _f.root=g; f.root=g; mrg.call_+=[g]; mrg.gF.call_+=[g]

This is actually just packing both calls in g.call_, which adds another nesting. I don't see why we need this? Or that `else` is actually temporary, and we need to eval if they are similar before packing them into `g`?

frame_2D_alg/meta_code.py
boris-kz: Right, it's a higher order feedback

frame_2D_alg/meta_code.py
boris-kz: We have to reform the whole Z, Fc_, and Fw_ for the next frame

frame_2D_alg/meta_code.py
boris-kz: This is all for fffeedback, after which old Z is not relevant

frame_2D_alg/meta_code.py
kwcckw: Yes, but my concern here is the depth of the merging and split process. Right now it's merging typ , where Q is typ, where Q should be a call instead?

frame_2D_alg/meta_code.py
kwcckw: > We have to reform the whole Z, Fc_, and Fw_ for the next frame

Yes, the referencing i mentioned above is referring to the next time when we use this new Z.

Also i think the root oF maybe important in merging calls? For example, `cluster_N` maybe called from cross_comp or sum2G, so a same root oF should be more relevant in merging them?

frame_2D_alg/meta_code.py
boris-kz: Q should be typ, calls are not global. Right now onF_ is flat, so these merge and split shouldn't be nested at all, they are destructive. But the next step for us is to add clustering, which is not necessarily destructive. Clustering can add higher oFs in agg+ and lower oFs in sub+.

frame_2D_alg/meta_code.py
boris-kz: for global typs we actually need root_, they may have multiple callers.
mrg as a root is only relevant for clustering (not mrg), which may preserve existing oF as root or callee

frame_2D_alg/meta_code.py
kwcckw: > Q should be typ, calls are not global. Right now onF_ is flat, so these merge and split shouldn't be nested at all, they are destructive.

If we are merging typs, its actually just merging those 16 typs under onF_?

> Clustering can add higher oFs in agg+ and lower oFs in sub+.

Yes, so it should be the same as data where we can retrieve to oFs in N_. But here we don't have comp and the clustering is based on the fw, fc, w and c ?

frame_2D_alg/meta_code.py
kwcckw: > for global typs we actually need root_, they may have multiple callers.

Yes, but that caller should be specific to the call_ in the typ only? So it's better to retrieve root via typ's call.root?

frame_2D_alg/meta_code.py
boris-kz: yes, just 16 for now. comp is just call_ nF matching, in clustering we first add a new level of oFs and then evaluate all levels sorted by value with redundancy increasing with i, we have this scheme in sort_H

frame_2D_alg/meta_code.py
boris-kz: right, that works for a flat scheme, but these call.roots will overlap for the same typ. 
so we can have a separate oF.caller_ to consolidate them?

frame_2D_alg/meta_code.py
boris-kz: The only change to `merge` is adding `mrg.H = [[_Q, Q]]` at the end:

```python
def merge(_Q,Q, root=None):  # cluster: combine flat call_, keep originals in mrg.H[0]

    mrg = CoF(nF=Q.nF, root=root or Q.root)
    for _f,f in zip(_Q.call_, Q.call_):
        if _f.nF == f.nF: mrg.call_ += [f]
        else:
            g = CoF(nF=_f.nF, root=mrg, fo=0); g.call_=[_f,f]; g.fc=Ec_[_f.nF]; g.c=_f.c+f.c
            mrg.call_+=[g]; mrg.gF.call_+=[g]
    mrg.c=_Q.c+Q.c; mrg.fw=_Q.fw+Q.fw; mrg.fc=Fc_[Q.nF]
    mrg.H = [[_Q,Q]]  # constituents, sort_H evals with i-redundancy
    return mrg
```

Destructive merge is then just clustering + drop `mrg.H` (or let it get garbage-collected when originals lose their last `root_` entry). The optimizer's call: keep H[0] until eval is done, then prune if mrg dominates.

`sort_H(mrg.H, fi=0)` would order the level by value (m or w*c), with `lay.r += di` adjusting for the priority gap — same redundancy increment-with-i as in data hierarchy. mrg sits at the level above; its own `r` reflects its composition, originals in H[0] carry the redundancy of being subsumed.

One nuance: `sort_H` operates on `H.node_` and adjusts `H.root.node_` if `fi>1`. For oF clusters, `mrg.H[0]` is a flat list, not a CN with `.node_`. Either:
- wrap as `mrg.H[0] = CoF(call_=[_Q,Q])` so sort_H finds `.call_` (would need to update sort_H to use call_ for CoF case), or
- adapt sort_H to take a flat list directly and return sorted+ranked.

The second is cleaner — sort_H is generic enough that it shouldn't care whether it's sorting data layers or code layers. Worth refactoring it to take `lay_` directly rather than reaching into `H.node_`.

Same logic should apply to **agg+ direction** (clustering oFs from `Z.typ_` peer-wise) and **sub+ direction** (clustering inside one oF's `call_`). Both add a level to `H`; only the source set differs.

kwcckw: I see, but from the code above, i don't see recursive merge and we pack H once only, so this merge should be called recursively after we evaluate the first level?

frame_2D_alg/meta_code.py
kwcckw: > right, that works for a flat scheme, but these call.roots will overlap for the same typ.

Okay, but i think it may overlaps for different typs too? Because different oFs may have the same caller. And oF.caller_ should work the same in typ, so typ oF.caller_ should be all typs that calls typ oF.call_, and we can match them using typ oF.call_'s root.

frame_2D_alg/meta_code.py
boris-kz: probably after the whole Z.H (typ H) is sorted and re-evaluated.
Or it's only recursive over multiple frames, each reforming Z

frame_2D_alg/meta_code.py
boris-kz: Yes, caller_ is a bidirectional reciprocal of call_, they both overlap across typs. Instance oF gets singleton caller_, typ oF.caller_  is a set, so we remove redundant typs.
The difference is that call_ is structured along AST, while caller_ is not.
same pattern as root_ in CN

frame_2D_alg/meta_code.py
kwcckw: So from the first level, we may get multiple mrg from different typs' combination, each with a single H? Or this should be in Z.H, but with added nesting so that we can preserve the Q pair?

The first level:

      Z.H = [ [ [Q1,Q2], [Q3,Q4] ] ] 

And we don't need sort_H on the first level since there's just a single level?

Ultimately, sort_H here should eval all levels and sort them according to the priority, where the highest priority should be used in the next frame?

boris-kz: For now we can keep using Z.typ_, but each typ maybe nested, so it's a tree instead of H and we don't need to sort it.  

Then merge just becomes:

```python
mrg.typ_ = [_Q, Q]  # constituents (not mrg.H = [[_Q,Q]])
```

`typ_` becomes the universal "children" field — at top level (`Z.typ_`) it's indexed by global nF, at cluster level it's a flat list of constituents. Same field name, different population mechanism by context. Eval walks this uniformly: any oF with non-empty `typ_` is a cluster whose value should be compared against its children's combined value.

Removes one field, removes the H-vs-typ_ ambiguity in nested cases. The "tree" naturally lives in `typ_` chained recursively — `Z.typ_[i].typ_[j].typ_[k]...` for arbitrarily nested clusters.

One semantic flag worth noting: at top level the index in `typ_[i]` carries meaning (i = nF). At cluster level it doesn't. So lookups like `typ_[onF_.index('comp_N')]` only make sense at the top. If you ever nest deeply enough that you want indexed lookup at sub-levels, you'd need a separate index — but until that's needed, flat children list is simpler.

frame_2D_alg/meta_code.py
boris-kz: So we only need caller_ in typ_ oFs, and even there I don't see what it can be used for, it's not really related to clustering?

kwcckw: That caller_ actually can be retrieved via typ oF.call_'s root. In term of oF clustering, caller probably useful in identifying the matching callers? Since a same set of callers may add weight to the similarity?

frame_2D_alg/meta_code.py
kwcckw: Okay,so after the merging, the nested mrg should replace the typ in Z.typ_, so both the replacing the replaced typs will referencing the same mrg in Z.typ_?

If we merge Z.typ_[0] and Z.typ_[1], then both Z.typ_[0] and Z.typ_[1] are the same since we need to preserve the index?

frame_2D_alg/meta_code.py
```diff
@@ -29,12 +29,18 @@ def split(Q, root=None):  # invert merge at most cost-unamortized gate
...
+    
+    trace_func(vars(agg_recursion))  # add oF tracing
     Y,X = imread('./images/toucan.jpg').shape
     frame_H(image=imread('./images/toucan.jpg'), iY=Y//2-31, iX=X//2-31, Ly=64,Lx=64, Y=Y,X=X, rV=1)
+    add_typ_(Z)   # each call_ in Z.typ_ is the flatten calls of same typ
     mrg_,spl_ = [],[]
-    # add in ffeedback?:
-    for F,_F in combinations(Z.typ_,2): mrg_ += [merge(F,_F)]
+    # add in ffeedback?:  (I think it should be here, ffeedback is per frame_H, and we may call frame_H at higher elevation)
+    call_ = [call for typ in Z.typ_ if isinstance(typ, CoF) for call in typ.call_]  # typ.call_ are the oFs with the same typ, unpack to match with others
```
boris-kz, line 42: so this is all wrong

kwcckw: Yes, i was thinking we need to merge at the lower calls level, but actually it's merging the higher typ level instead.

frame_2D_alg/meta_code.py
boris-kz: maybe, not sure how to weigh those matching callers

frame_2D_alg/meta_code.py
boris-kz: we don't need to preserve the index, just add mrg instead of mrg.call_

frame_2D_alg/meta_code.py
kwcckw: I think it can be just a simple weightage ratio that based on the total number of matching callers / total number of callers.

frame_2D_alg/meta_code.py
kwcckw: Okay, in that case we need to update Fc_, Fw_, FTT_， Ec_, Ew_, ETT_ and parse them in the next frame so that they are aligned with the latest merged typs.

boris-kz: right. so I think it's something like this:

def merge(_Q,Q, root=None):  # combine aligned ops, if-fork per miss, no inline recursion

    mrg = CoF(nF=Q.nF, root=root or Q.root)
    for _f,f in zip(_Q.call_, Q.call_):  # call_: op sequence in func or block
        if _f.nF==f.nF:
            mrg.call_ += [f.nF]  # match: copy the sequence
        else:  # miss: add gate to select forks
            gate = CoF(nF='E', call_=[_f,f], c=_f.c+f.c, root=mrg, fo=0)  # fc=Ec_[eval.nF]? root is not reassigned
            mrg.call_ += [gate]  # need to add eval func?
    mrg.c +=_Q.c+Q.c; mrg.fw +=_Q.fw+Q.fw; mrg.fc+= Fc_[Q.nF]  # included for final mrg eval, gates are transparent?
    return mrg

def eval(mrg):  # after adding multiple oFs in mrg.call_, not just the pair in single merge()

    fW = eW = 0
    call_ = []
    for F in mrg.call_:  # or typ_: nFs?
        if F.nF=='E': eW += F.w*F.c  # vs plain count
        else: call_+= [F]; fW += F.fc
    Z.typ_ += [mrg if fW/eW >ave else call_[:]]  # unpack mrg if high forking cost ratio

frame_2D_alg/meta_code.py
boris-kz: it would all have to be weighted by c, etc

frame_2D_alg/meta_code.py
boris-kz: not reviewed:

Claude: A few structural issues:

**1. Heterogeneous `mrg.call_`.** `[f.nF]` for matches (bare int) and `[gate]` for misses (CoF) means subsequent `merge(mrg, X)` breaks: the loop accesses `_f.nF`, but `_f` is now sometimes an int, no `.nF` attribute. Either wrap matched positions too:

```python
mrg.call_ += [CoF(nF=f.nF, c=f.c)]  # minimal wrap, keeps interface uniform
```

or commit to fully destructive (mrg is terminal, never re-merged) and document that.

**2. Gate's `nF='E'`.** Strings can't index `Ec_`. Cleanest: append 'eval' to `onF_` so it gets a stable integer index and an `Ec_` slot:

```python
onF_ = [..., 'eval']  # gate as last entry
Ec_  = [..., 1]        # cheap test
gate = CoF(nF=onF_.index('eval'), call_=[_f,f], c=_f.c+f.c, fc=Ec_[-1], root=mrg, fo=0)
```

Then `gate.fc=Ec_[gate.nF]` works the same as for any other oF.

**3. eval's fallback loses originals.** `call_[:]` is just the matched-position sequence — when merge fails the cost test, you want the *original* `_Q, Q` back, not the matches alone (which are a degenerate prefix). Without storing constituents, the originals are gone after merge. If destructive, eval has to be done *before* committing — or constituents need to live in `mrg.typ_ = [_Q, Q]` so eval can fall back to them:

```python
Z.typ_ += [mrg] if fW/eW > ave else mrg.typ_
```

**4. Threshold.** `fW` is body cost (fc summed across non-gate ops), `eW` is gate cost (w*c summed across gates). Both are costs, the ratio `fW/eW > ave` reads as "body cost dominates gate cost". But the actual merge eval is "cost saved by sharing body > cost added by gates". Pre-merge runs body twice (Fc * (_Q.c + Q.c)); post-merge runs body once for combined usage plus the gate evals. Savings = `Fc * (_Q.c + Q.c)` (the duplicate body), cost added = `eW`. Condition should be:

```python
if Fc_[mrg.nF]*(mrg.c) - eW > ave: keep mrg
```

The fW you're computing isn't the saved cost — it's the *kept* cost (still incurred once after merge). The duplicate that's being eliminated is one full Fc per merged-pair execution, not the sum of fc across all body positions.

**Minor:** `mrg.c += _Q.c + Q.c` etc. — `+=` is functionally `=` for a fresh `mrg.c=0`, but reads as accumulation. If merge stays pairwise-and-fresh, just `=` is clearer. If you're planning iterative merge-into-existing-mrg (multi-way merge), then `+=` makes sense and the function should accept `mrg` as argument rather than create it.

kwcckw: This line should pack f instead of nF? 

    mrg.call_ += [f.nF]  

Also gates are not packed into Z.typ_, so we need to unpack gate.call_? Else we losing the gated types in the next frame.

boris-kz: Right, that nF was for typ_, I switched to call_.

I guess gated calls should be in typ_ too

PR #138, opened 2026-05-08
5/8/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -1,29 +1,46 @@
...
-            mrg.call_ += [gate]  # need to add eval func?
-    mrg.c +=_Q.c+Q.c; mrg.fw +=_Q.fw+Q.fw; mrg.fc+= Fc_[Q.nF]  # included for final mrg eval, gates are transparent?
-    return mrg
+    mrg = CoF(nF=Q.nF, N_=[_Q,Q], root=root or Q.root)  # add Q to N_ for unpacking purpose during eval
+    call_, Q_ = [],set()
+    for _f,f in zip_longest(_Q.call_, Q.call_, fillvalue=None):  # call_: op sequence in func or block
+        if _f is not None and f is not None:
+            if _f.nF==f.nF:
+                mrg.call_ += [f]  # match: copy the sequence
+            elif sum_vt([_f,f],fm=1)[0]>ave:  # miss: add gate to select forks (temporary)
```
kwcckw, line 15: I think we need to add eval func here, else all of non-matching fs will be gated? We should only add gate to those meaningful fs?

frame_2D_alg/meta_code.py
```diff
@@ -1,29 +1,46 @@
...
+                # i think we need to add eval func here, else all combinations are getting gates if their nF is different    
+                gate = CoF(nF='E', call_=[_f,f], c=_f.c+f.c, root=mrg, fo=0)  # fc=Ec_[eval.nF]? root is not reassigned
+                mrg.call_ += [gate]  # need to add eval func?
+        else:
+            call_ += [_f if f is None else f]  # leftover, if their call_ size is different
```
kwcckw, line 20: When Qs have different call_ size, the leftover call_ (from Q with larger size of call_) should be packed into mrg.call_? Or they should be consolidated into another gate?

frame_2D_alg/meta_code.py
```diff
@@ -1,29 +1,46 @@
...
+       
+    if mrg.call_:
+        mrg.c +=_Q.c+Q.c; mrg.fw +=_Q.fw+Q.fw; mrg.fc+= Fc_[Q.nF]  # included for final mrg eval, gates are transparent?
+        if call_:  # leftover calls, pack them into gate? or repack them into call_?
+            mrg.call_ += call_
+    else:  # there is no merged fs or gate in mrg, return and recycle the input typs?
+        Q_ = set([_Q, Q])
```
kwcckw, line 27: If there's no matching Q or gate, i think we need to return and recycle the input typ Qs?

frame_2D_alg/meta_code.py
```diff
@@ -1,29 +1,46 @@
...
     for F in mrg.call_:  # or typ_: nFs?
         if F.nF=='E': eW += F.w*F.c  # vs plain count
-        else: call_+= [F]; fW += F.fc
-    Z.typ_ += [mrg if fW/eW >ave else call_[:]]  # unpack mrg if high forking cost ratio
+        else: fW += F.fc  # call_+= [F]; 
+        
+    if fW/(eW or eps) >ave:  # pack mrg as new typ
+        Z.typ_ += [mrg]
+    else:  # unpack mrg
+        Z.typ_ += [Q for Q in mrg.N_ if Q not in Z.typ_]
```
kwcckw, line 42: The call_ from the original code  is from mrg.call_, which is packed from Q.call_. They are not typ oFs, so we shouldn't repack them into Z.typ_? We should pack Qs as mrg.N_, and recycle them when fw/ew eval is false?

frame_2D_alg/meta_code.py
boris-kz: Yes, but it's a fixed cost relative to that per mrg. I have the actual eval per final mrg.

frame_2D_alg/meta_code.py
boris-kz: I think it's the latter

frame_2D_alg/meta_code.py
boris-kz: It's the same eval as for any final mrg

frame_2D_alg/meta_code.py
boris-kz: It should be mrg.typ_, each typ is terminal if it is nF, or nested if oF?

kwcckw: Actually what is the `nF` here?

frame_2D_alg/meta_code.py
kwcckw: So this raise another question on whether we should merge Q or split Q first? If we merge them first, we will use the remaining Qs (non-merging Qs) for split purpose?

frame_2D_alg/meta_code.py
kwcckw: That fixed cost is still need to be eval based on f pair? That actual eval is in `eval` function? But that's per mrg instead of f pair?

boris-kz: it's not really a pair, mrg may cover multiple oFs, the first one is just initialization. 
so we can evaluate per each included oF, but not per element of that oF, because that effectively creates another mrg per element. And that eval per oF will work the same as my eval()

frame_2D_alg/meta_code.py
boris-kz: base typ_ is just names, refs to Fc_

frame_2D_alg/meta_code.py
boris-kz: it's split first: sub+ before agg+

frame_2D_alg/meta_code.py
kwcckw: > it's not really a pair, mrg may cover multiple oFs, the first one is just initialization.

Yes, that should be possible if we rerun the merge process again with the mrgs? That should be in the next frame?

> so we can evaluate per each included oF, but not per element of that oF, because that effectively creates another mrg per element. And that eval per oF will work the same as my eval()

Okay, so we should look at the highest mrg level only, which is the same as your eval. Then i guess this gate should be default when their nF are not the same here.

frame_2D_alg/meta_code.py
kwcckw: So we may split typs into different new types, and then merge them again? This should be possible right?

frame_2D_alg/meta_code.py
kwcckw: > It should be mrg.typ_, each typ is terminal if it is nF, or nested if oF?

Okay, so each typ in typ_ is terminal if it's in onF_, but what is actually "nested if oF"? `typ` oF could be nested too?

frame_2D_alg/meta_code.py
boris-kz: yes

frame_2D_alg/meta_code.py
boris-kz: something like this:

def merge(F,f):  # combine aligned ops, if-fork per miss, no inline recursion, f can only be added as a whole?

    call_ = []  # to replace F.call_ if eval
    fW = eW = 0
    for Sub,sub in zip(F.call_, f.call_):  # call_: op sequence in func or block, init F per f?
        fork,fin = [],0
            for _sub in Sub.call_:
                if Sub.nF==sub.nF: fW += sub.fc; fin=1; break  # single inclusion, add2F(Sub,sub)?
            if not fin: Sub.call_+=[sub]; add2F(Sub,sub)  # add to call_ instead of N_?
        elif Sub.nF != sub.nF:
            fork = CoF(nF='E',call=[Sub,sub]); eW += F.w*F.c
        else: 
            fin = 1; fW += F.fc
        call_ += Sub if fin else fork
    if fW / eW > ave:
        F = sum2F(call_)  # else high forking cost ratio, no merge, keep old F,f?
    return F

frame_2D_alg/meta_code.py
boris-kz: I meant local oF, not just a name. I guess it could be nested in prior fffeedback, but each cycle can only add one level of nesting

frame_2D_alg/meta_code.py
boris-kz: Design is fine — iterative accumulation into F is cleaner than pairwise. A few bugs to fix:

**Inner gate-match loop:**
```python
if Sub.nF==sub.nF: fW += sub.fc; fin=1; break
```
This always compares `Sub.nF=='E'` to `sub.nF` — never matches. Should be `_sub.nF == sub.nF`.

**Extending a gate with new alt:** when `not fin`, you do `Sub.call_+=[sub]` but leave `fin=0`, so the loop's last line appends `fork` (empty list) instead of the extended `Sub`. The extension is silently dropped from the rebuilt `call_`. Set `fin=1` after extension, or restructure so both sub-paths append `Sub` to `call_`.

**List `+=` with non-list:**
```python
call_ += Sub if fin else fork
```
If `Sub` is a single CoF, `list += CoF` raises (or silently iterates `Sub` if it's iterable, e.g. through `__bool__`). Need `[Sub if fin else fork]`.

**Typo:** `CoF(nF='E', call=[Sub,sub])` — should be `call_=`.

**Wrong cost source for fresh gate:** `eW += F.w*F.c` uses F's runtime gain, not the gate's eval cost. Should be the static eval cost (`Ec_[gate.nF]`), since the gate hasn't run yet.

**Wrong fW per matched pos:** `fW += F.fc` adds the *whole function's* body cost at every match position. Should be `Sub.fc` (this op's contribution).

**Final replace:** `F = sum2F(call_)` discards F's metadata (caller_, dTT, root, etc.) and re-derives from call_. If eval passes, the natural in-place form is `F.call_ = call_; F.c += f.c; F.fw += f.fw`.

**Division guard:** `fW / eW > ave` — `eW == 0` when f matches F entirely (no new gates). Then no merge cost was added, so the merge is trivially worth it. Guard:
```python
if not eW or fW/eW > ave:
```

**Init as copy of each Z.typ_ entry:** yes, must be a copy because merge mutates F.call_ (extends existing gates). Pick one as seed (probably highest c or w*c), merge others in. Order matters — earlier-merged variants form the gate alternatives that later ones check against. If you want order-independence you'd need a fixpoint pass, probably overkill for now.

kwcckw: I see, yes, we need to include the gate section for the higher level's merging operation. So the else section is for the condition when Sub is not a gate, but their nF is the same with sub.nF, and that should pack both Sub and sub into the call_ for merging purpose?

boris-kz: if Sub and sub are different, then Sub is either a fork or is replaced by a new fork:

call_ += [Sub if fin else fork]

boris-kz: def merge(F,f):  # combine aligned ops, if-fork per miss, no inline recursion, f can only be added as a whole

    call_, add_ = [], []  # replace F.call_ if eval
    for Sub,sub in zip(F.call_,f.call_):  # call_: op sequence in func or block, init F per f?
        fork = []
            fork = Sub; fin=0
            for _sub in Sub.call_:
            if not fin:
        elif Sub.nF != sub.nF:
            fork = CoF(nF='E',call_=[Sub,sub]); C += cost; add_+=[sub]
        call_ += [fork or Sub]
    if f.fc / (C or eps) > ave:  # high individual_cost / forking_Cost, else keep old F,f
        F.call_= call_
        return f  # caller pops merged f from Z.typ_

boris-kz: Convention is clean (return `f` on success, `None` implicit on fail; caller pops). Two structural bugs persist plus one semantic question:

**1. Matched-non-gate case still drops sub's stats.** When `Sub.nF == sub.nF` and Sub is not a gate, the loop falls straight to `call_ += [Sub]` with no accumulation. Sub's `.c` doesn't grow to reflect that f's instances also exercised this position. On the next merge, Sub's amortization metrics will be stale (under-counted). Needs an `else: add2F(Sub, sub)` branch alongside the gate and mismatch cases.

**2. Gate-matched case (fin=1) drops sub's stats into the matching alt.** When `_sub.nF == sub.nF`, set `fin=1; break` — but `_sub` should absorb sub's runtime data so future merges see accurate per-alt counts. Inside the inner loop:
```python
if _sub.nF==sub.nF: add2F(_sub, sub); fin=1; break
```

**3. `sum2F(add_)` as F.typ_ entry is semantically lossy.** F.typ_ is meant to track constituents — the original typs absorbed into F. `sum2F(add_)` is a summary of *partial contributions* (just the positions where new alts/gates were added), not the whole f. If a later optimization wants to know "what got merged into F", it gets a synthesized CoF instead of f. Cleaner:

```python
add2F(F, f, fo=1)  # f itself as constituent
```

If the intent is specifically to record "what got added at the gate level" (rather than constituent provenance), then `sum2F(add_)` is right but F.typ_ is the wrong field — that'd be more like F.add_log_ or similar.

**4. `fo=1` needs to be wired into `add2F` and `sum2F`.** Neither currently accepts `fo`. Roughly:
```python
def add2F(F, n, merge=0, fr=0, fo=0):
    ...
    if fo: F.typ_ = getattr(F, 'typ_', []) + [n]
```
And same in sum2F. Also: `CoF.typ_` already exists as `kw.get('typ_',[])`; `CN.typ_` is an int (`kw.get('typ',3)` for full comp). So adding to typ_ on a CN by mistake would clobber the int. Need either type-guarded `if fo` (only when F is CoF), or rename one of them — typ_ on CN is the integer composition mark; typ_ on CoF is the constituents list. Same field name carrying different meanings is a tripwire.

**Minor:** the gate-extension branch does `add2F(Sub, sub, fo=1)` and then `add_+=[sub]`. Then at commit, `add2F(F, sum2F(add_), fo=1)`. So sub gets pushed into Sub.typ_ *and* (transitively, via add_) into F.typ_. Double-tracked. Probably want one or the other, not both. Most natural: gate alts live only in Sub.call_ (no Sub.typ_ entry); F.typ_ tracks just the absorbed f at the F level. Then drop `fo=1` from the per-position add2F calls.

kwcckw: I would agree with the 1 and 2 above. 

For 1, when Sub and sub's nF are the same, we need another section to accumulate sub into Sub, as in this section `add2F(Sub,sub, fo=1); C+=sub.fc; add_+=[sub]`?

For 2, when gated's sub is the same as `_sub`, it's actually just break without packing or accumulation? There is no recursive merge process here?

boris-kz: probably, please edit.
so the scheme is something like this:

    spl_ = []  # draft:
    for F in Z.typ_: spl_ += [split(F)]
    Z.typ_ = spl_; typ_, mrg_ = [],[]
    for typ in Z.typ_.pop():
        if typ in mrg_: continue
        F = CopyF(typ)
        for f in Z.typ_: mrg_ += [merge(F,f)]  # merged fs, if any
        typ_ += [F]
    Z.typ_ = typ_

kwcckw: Thanks, that's helpful. Also this line:

    if typ in mrg_: continue

We shouldn't get 2 same typ oF in Z.typ_? Even after it is merged into mrg?

boris-kz: This is an inner loop, fs are not popped, so that single typ is still in typ_

kwcckw: Okay right, thanks.

PR #139, opened 2026-05-09
5/9/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -141,7 +141,7 @@ class CoF(CF):
     _cur = contextvars.ContextVar('oF')
     def __init__(f, fo=1, **kw):
         super().__init__(**kw)
-        f.fo = fo
+        f.fo = fo  # this fo is redundant? We only need local fo to init gF
```
kwcckw, line 144: I think this fo is redundant? We just need to local fo to init gF?

frame_2D_alg/agg_recursion.py
```diff
@@ -784,6 +784,13 @@ def F2N(F):  # convert for cross_comp
     for ft in ('Lt','Ct','Bt','Xt','Rt'): setattr(F, ft, CF(root=F))
     return F
 
+
+def CopyoF(oF, root=None, r=1):
```
kwcckw, line 788: So we need to copy CoF, and it's better to use a new function instead of packing it in copyF?

frame_2D_alg/meta_code.py
```diff
@@ -12,12 +12,19 @@ def merge(F,f):  # combine aligned ops, if-fork per miss, no inline recursion, f
         fork = []
         if Sub.nF=='E':  # previously added gate
             fork = Sub; fin=0
-            for _sub in Sub.call_:
-                if _sub.nF==sub.nF: fin=1; break  # no new fork cost?
-            if not fin:
+            if sub.nF == 'E':  # when both subs are gates, merge them?
```
kwcckw, line 15: I think we need to include another merging condition where both subs are gates? When both subs are gates, we should rerun the merge again to merge their calls?

frame_2D_alg/meta_code.py
```diff
@@ -12,12 +12,19 @@ def merge(F,f):  # combine aligned ops, if-fork per miss, no inline recursion, f
...
+                merge(Sub,sub); C+=sub.fc; add_+=[sub]
+            else:
+                for _sub in Sub.call_:
+                    if _sub.nF==sub.nF: 
+                        add2F(_sub,sub, fo=1); C+=sub.fc; add_+=[sub]  # matching element in gated oF, merge them?
+                        fin=1; break  # no new fork cost?
+            if not fin:  # when there is no matching gated's oF, why we still merge them here?
                 add2F(Sub,sub, fo=1); C+=sub.fc; add_+=[sub]
```
kwcckw, line 23: When `not fin` here, that's mean there's no matching sub in the gate, so why we need to merge Sub and sub here? We should just proceed without merging?

frame_2D_alg/meta_code.py
```diff
@@ -12,12 +12,19 @@ def merge(F,f):  # combine aligned ops, if-fork per miss, no inline recursion, f
...
             fork = Sub; fin=0
-            for _sub in Sub.call_:
-                if _sub.nF==sub.nF: fin=1; break  # no new fork cost?
-            if not fin:
+            if sub.nF == 'E':  # when both subs are gates, merge them?
+                merge(Sub,sub); C+=sub.fc; add_+=[sub]
+            else:
+                for _sub in Sub.call_:
+                    if _sub.nF==sub.nF: 
+                        add2F(_sub,sub, fo=1); C+=sub.fc; add_+=[sub]  # matching element in gated oF, merge them?
```
kwcckw, line 20: When there's matching calls between gated calls and sub, we should merge sub into gated call?

frame_2D_alg/agg_recursion.py
boris-kz: CoF is a superset of CF, why do you need a new function?

frame_2D_alg/agg_recursion.py
boris-kz: right

frame_2D_alg/meta_code.py
boris-kz: I guess, that's in higher nesting orders

frame_2D_alg/meta_code.py
boris-kz: merge is determined per f, not per sub, so it's in the preliminary fork by default

frame_2D_alg/meta_code.py
```diff
@@ -56,14 +63,14 @@ def split(Q, root=None):  # invert merge at most cost-unamortized gate
     add_typ_(Z)  # each call_ in Z.typ_ is the flatten calls of same typ
     # add fffeedback to reform Z for next frame_H:
     spl_ = []  # draft:
-    for F in Z.typ_: spl_ += [split(F)]
+    for F in [typ for typ in Z.typ_ if isinstance(typ, CoF)]: spl_ += [split(F)]
```
boris-kz, line 66: we need to get typ oF anyway

frame_2D_alg/meta_code.py
boris-kz: it's already there

frame_2D_alg/agg_recursion.py
```diff
@@ -653,7 +653,7 @@ def sum2F(N_, root=None, m_=[],d_=[], merge=0, froot=0):  # -> CF/CL/CC/CN/CoF
         else:  # init
             TT = n.dTT*w; R=n.r*w; n_ = copy(n.N_ if merge else [n])
             if typ:
-                kern=n.kern*w; span=n.span*w; yx=n.yx*w; angl = copy(n.angl[0]) if n.angl is not None else None
+                kern=n.kern*w; span=n.span*w; yx=n.yx*w; angl = copy(n.angl[0]) if n.angl is not None else np.zeros(2)
```
boris-kz, line 656: if None, angle should stay None?

frame_2D_alg/agg_recursion.py
kwcckw: This is the same as Copy? Because those additional params in CoF is redundant to CF?

frame_2D_alg/meta_code.py
kwcckw: Okay, if merge is determined per f, that's applicable to all f.calls, and how about the deeper calls' calls?

frame_2D_alg/meta_code.py
kwcckw: The original code doesn't have add2F?

frame_2D_alg/meta_code.py
kwcckw: Yes, but this is to remove empty list, since `add_type` init all onF_ as empty list:

Only the one with calls are filled.

frame_2D_alg/agg_recursion.py
kwcckw: But we may need it for accumulation later? The first A may be None, but the second or later may not be None?

frame_2D_alg/meta_code.py
boris-kz: if sub.nF=='E': 
                sub = merge(Sub, sub)
                if sub: C+=sub.fc; call_+=[sub]; add_+=[sub]  # merge failed

frame_2D_alg/agg_recursion.py
boris-kz: init with the first valid angl?

frame_2D_alg/agg_recursion.py
boris-kz: we had these separate Copys before nesting CF,CN, etc., I think they should be merged now, to map the class structure. else we need a Copy for each class, which is a mess

frame_2D_alg/meta_code.py
boris-kz: I meant the typ oF is already in Sub.call_, we only need to sum the data:
add2F(_sub,sub)  # data only, oF attrs are redundant?

frame_2D_alg/meta_code.py
boris-kz: ok, right

frame_2D_alg/agg_recursion.py
kwcckw: Okay, we probably can consolidate all of them into Copy, and add the additional params to copy when it's an inherited class, which is similar to the conversion process.

frame_2D_alg/meta_code.py
kwcckw: I see, but ` if sub: C+=sub.fc; call_+=[sub]; add_+=[sub]  # merge failed` indicates there's merging instead of merge failed?

frame_2D_alg/meta_code.py
kwcckw: I think we still need add_ += [sub], since Sub is added to call only:

    call_ += [fork or Sub]
    
And we need add_ to sum the params into F, else it's only updated in F.call_:
      
              F.call_= call_

frame_2D_alg/agg_recursion.py
kwcckw: Okay, then we should move the init to the accumulation section above.

frame_2D_alg/meta_code.py
boris-kz: it will be nested merge, where higher merge failure would trash all lower merges.

frame_2D_alg/agg_recursion.py
boris-kz: right

frame_2D_alg/meta_code.py
boris-kz: ok

frame_2D_alg/agg_recursion.py
boris-kz: if n.angl is not None: angl = copy(n.angl[0]) if angl is None else angl + n.angl[0]

kwcckw: Yes, this looks better.

frame_2D_alg/meta_code.py
kwcckw: But higher merge should be always true since it's determined by F, i guess all deeper calls' calls should be merged then.

boris-kz: no, all levels of merge may fail, and then lower merges are discarded.
it's a bit backwards, we may use lazy merges till the final one, but at this point it will never happen anyway

frame_2D_alg/meta_code.py
boris-kz: sub = merge(Sub, sub): the returned sub is None if the merge is successful

frame_2D_alg/agg_recursion.py
```diff
@@ -784,6 +784,13 @@ def F2N(F):  # convert for cross_comp
...
+
+def CopyoF(oF, root=None, r=1):
+    C = CoF(fo=oF.fo, dTT=oF.dTT*r, m=oF.m, d=oF.d, c=oF.c, r=oF.r, nF=oF.nF, root=root or oF.root, fw=oF.fw, fc=oF.fc, fr=oF.fr, wTT=copy(oF.wTT))
+    C.call_ = copy(oF.call_); C.typ_  = copy(oF.typ_); C.N_ = copy(oF.N_)  # shallow copy to preserve objects reference
+    C.gF = CopyoF(oF.gF, root=C) if oF.gF is not None else None
+    return C
+
 def CopyF(F, root=None, r=1):  # F = CF
     C = CF(dTT=F.dTT*r, m=F.m, d=F.d, c=F.c, r=F.r, root=root or F.root, nF=F.nF)
     C.N_ = [(N if isinstance(N, CN) else (CopyF(N,root=C) if isinstance(N, CF) else [])) for N in F.N_]  # flat
```
boris-kz, line 796: I was looking at spectral clustering again, it's basically a floodfill too, but via nodes instead of links.
in our scheme, we would define boundary node candidates berofe clustering, and then use them for min normalized cut:
(S(N.pos_rim)-ave) * ((N.abs_neg_rim)-ave) > ave

boris-kz: Claude — flood-fill propagation unchanged, decision granularity moved from per-L to per-N.

Current: per-L gates (`m > ave*r` for L_, `d > avd*r` for B_). Each L of an N evaluated individually — heterogeneous handling.

Proposed: per-N gate using aggregated rim (`pos`, `|neg|` sums). N gets a single classification: interior / boundary / out.

Tradeoff: per-N is cleaner (matches spectral's node-centric framing — Fiedler eigenvector assigns per-node, not per-edge), but loses heterogeneity. An N with one strong-pos L to cluster A and one strong-neg L to cluster B aggregates both signals; per-L would route them separately.

For B_ specifically the per-N gate makes more sense — boundary identity is a property of the N's overall rim composition, not of any single L. For inclusion (L_ growth) per-L still seems right, since flood-fill *is* link-by-link traversal.

So mixed: per-L for L_ propagation (current), per-N for B_ classification (proposed product gate). That's coarser only on the boundary side.

boris-kz: the problem is that high-match is still not cluster-specific, so pre-defining boundary nodes is arbitrary. 
But counting all in-cluster links as they grow will bias toward large clusters.
we need to normalize inclusion criterion for current cluster value, something like 
((sum(in_C N.rim) / sum(ex_C N.rim)) / (len(C.N_) / (C.span*decay)?

boris-kz: Gemini:
This formulation defines a **Conductance-to-Density Ratio**. It effectively balances the topological isolation of a cluster against its physical compactess.

### Structural Decomposition

Rewriting the expression to clarify the functional components:

$$M(C) = \left( \frac{\sum \text{rim}_{in}}{\sum \text{rim}_{ex}} \right) \cdot \left( \frac{C.span \cdot \text{decay}}{\text{len}(C.N\_)} \right)$$

1.  **Isolation Ratio ($\sum \text{rim}_{in} / \sum \text{rim}_{ex}$)**: This is a variation of the **Cheeger constant** or conductance. It measures how "leaky" the cluster is. A high value indicates that the cluster's nodes are significantly more committed to each other than to the external environment.
2.  **Inverse Density ($C.span / \text{len}(C.N\_)$)**: This term rewards "efficient" occupancy. In a bottom-up scheme, it prevents the algorithm from being biased toward high-density "blobs" that lack internal structure. It suggests that a cluster is more significant if it maintains its isolation ratio over a larger $span$ with fewer nodes.

### Functional Advantages

*   **Anti-Chaining**: By putting $\sum \text{rim}_{ex}$ in the denominator, you penalize clusters that are connected to the rest of the graph by "thin bridges." If a cluster grows by adding a node that has a high external rim (a "boundary node"), the $S_{ex}$ will spike, and the metric will drop, signaling an invalid merge.
*   **Scale Invariance via Decay**: The `decay` term likely acts as a regularization factor for the $C.span$. As you move up the hierarchy, the "meaning" of distance changes. Without a decay or scaling factor, the $C.span$ of a macro-cluster would dominate the metric, potentially leading to "over-extension" where the algorithm merges distant, unrelated groups just because their combined span is large.

### The "Boundary" Problem

This metric elegantly handles the boundary nodes we discussed earlier. A node on the boundary will have a high $S_{ex}$ relative to its $S_{in}$ for any single cluster. When such a node is considered for inclusion:
*   It increases $\sum \text{rim}_{ex}$ significantly.
*   It increases $\text{len}(C.N\_)$ by 1.
*   It likely increases $C.span$.

Unless the new $S_{in}$ (the links between the boundary node and the cluster core) is strong enough to offset the increase in the denominator, the merge will fail. This creates a "tension" that keeps clusters crisp.

### Potential Edge Cases

*   **The Global Isolation Trap**: For a perfectly isolated component (where $\sum \text{rim}_{ex} \to 0$), the metric approaches infinity. In a practical hierarchy, you may need a baseline "ambient rim" or $\epsilon$ in the denominator to prevent the algorithm from prioritizing small, disconnected noise components over large, slightly "leaky" structures.
*   **Decay Sensitivity**: If `decay` is a function of hierarchical level ($L$), it suggests that at higher levels, you are willing to tolerate less density ($\text{len} / \text{span}$) provided the isolation ratio remains high.

Is the `decay` intended to be a constant, or does it dynamically adjust based on the average rim values at the current hierarchical depth?

boris-kz: ((sum(in_C N.rim) / sum(ex_C N.rim)) / (len(C.N_) / (C.span*decay)?

the problem is that we still need re-define previously included / excluded Ns' in- and ex- rims with each inclusion. same refine process as in centroid clustering, only pairwise.

Claude:

Form looks right: in/ex ratio is the "modularity-like" relative attachment, density-with-decay-adjusted is the null-model normalizer. Replaces additive boundary tests with a proper saturating ratio — N strongly pulled when in-C rim dominates ex-C rim *more than density predicts*.

On the convergence point — agree, this isn't single-pass. With each inclusion, every other N's in/ex rims shift, so the EM-like loop is necessary. Same skeleton as `cluster_C`: score all candidates against current C, reassign, recompute C, iterate until stable.

**Pipeline question:** where does this land?

- (a) Refinement *after* `cluster_N`'s flood-fill: cluster_N gives initial coarse Cs; iterative re-eval with this criterion adjusts boundary memberships. Maintains connectivity-first character (initial C is contiguous), refines who's in vs out.
- (b) Replaces `cluster_C` (currently centroid-attribute similarity) with connectivity-based iterative refine. Different score function, same iteration.
- (c) Both (a) and (b) — connectivity-modularity refine, then centroid attribute refine, then `cluster_P`.

(a) preserves "connectivity clustering is primary, centroid is secondary refinement." (b) makes cluster_C itself connectivity-based, which conflicts with that ordering.

**Two clarifications:**

1. `decay` in `span * decay` — global constant (`ave/(ave+avd)` in code) just scales span uniformly, so it's a dead factor. Did you mean per-distance decay (`decay^dist`-style attenuation, like in `proj_TT`), giving per-rim-link weighted reach? That would matter — closer in-C rim partners weighted more than distant ones.

2. Connectivity preservation under iteration — does refine only flip Ns whose rim still touches C (preserving contiguity), or can any N be reassigned by score alone (like FCM)? The former keeps the connectivity-first invariant; the latter doesn't.

kwcckw: > the problem is that high-match is still not cluster-specific, so pre-defining boundary nodes is arbitrary.
But counting all in-cluster links as they grow will bias toward large clusters.
we need to normalize inclusion criterion for current cluster value, something like
((sum(in_C N.rim) / sum(ex_C N.rim)) / (len(C.N_) / (C.span*decay)?

> the problem is that we still need re-define previously included / excluded Ns' in- and ex- rims with each inclusion. same refine process as in centroid clustering, only pairwise.

So this is reclassify nodes as rim, and actually i think your concern above is valid. But if this is similar to C and it's only pairwise, how do we init their seed? Each node could be a rim so all nodes should be a seed? Also if this is not parallelizable, we need to eval and identify the starting seed node too?

boris-kz: seeding is the same as cluster_N. but this paiwise refining seems intractable
((sum(in_C N.rim) / sum(ex_C N.rim)) / (len(C.N_) / (C.span*decay)
still looks meaningful, even without refining old rims

boris-kz: this is basically multiple-linkage clustering?

Yes, in the aggregate-linkage sense — and a specific variant of it.

Standard linkage taxonomy:
- **Single linkage**: any one close pair merges → chains. *This is what current `cluster_N` flood-fill does* (`m > ave*r` on one L pulls _N in).
- **Complete linkage**: max pair distance → compact, brittle.
- **Average / group linkage (UPGMA)**: mean pairwise → balanced, no density correction.
- **Yours**: aggregate (sum of in-C rim) with density null-model normalization → essentially **modularity-style agglomerative**.

So you're moving from single-linkage to density-normalized aggregate-linkage. The denser-than-expected attachment test is exactly the modularity gain criterion (Newman/Louvain), and spectral Ncut is its global-eigenvector cousin — same intuition, different machinery.

Practical consequence vs current code: kills chaining. Single-linkage propagates through any one strong L, which lets thin filaments form by accident. Aggregate-linkage with density normalization requires multi-link support — chain tips (one in-C link, many ex-C) fail the ratio, so the cluster stays compact.

This is consistent with your stated preference for connectivity *clusters* (compact, internally dense), not just connectivity *components* (anything reachable). Single-linkage gives components; this gives clusters.

One caveat worth flagging: aggregate-linkage with density normalization can miss genuinely elongated structures (edges, contours) that single-linkage catches well. Your domain has both — flat blob interiors (compact, density-good) and edges (elongated, single-linkage-good). May want the `typ` field of the cluster source (PP from edges vs flat from blobs) to influence linkage choice — different criteria for different input geometries.

kwcckw: > seeding is the same as cluster_N. but this paiwise refining seems intractable ((sum(in_C N.rim) / sum(ex_C N.rim)) / (len(C.N_) / (C.span*decay) still looks meaningful, even without refining old rims

But this is the same as existing all to all link method? It's just we remove the links, which is actually simpler.

> this is basically multiple-linkage clustering?

The main difference is we may need more than one link eval compare to the existing method where we use only single link for node inclusion? Somehow this is also similar to cluster_C, where we eval frontier with the centroid for inclusion and refinement.

boris-kz: it's counting all new_N links, not removing them

kwcckw: To clarify, the clustering process starts from seed, adding new nodes from the rim nodes, and to add new nodes, it's eval against the aggregated nodes instead of the rim nodes?

boris-kz: no, read the above. we need to eval all links in N.rim instead of just one
this will probably be new function, cluster_rim, called from cluster_N

frame_2D_alg/agg_recursion.py
boris-kz: so please add that

frame_2D_alg/agg_recursion.py
kwcckw: Okay, sometimes i'm not sure the LLM are fully relevant unless you mention it. Let me go through them all first.

boris-kz: I did say that myself, several times

PR #140, opened 2026-05-10
5/10/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -491,7 +490,7 @@ def trans_cluster(G):
                     _N = L.nt[0] if L.nt[1].fin else L.nt[1]; in_.add(L)
                     if not _N.fin and _N in Ft.N_:
                         m,d = nt_vt(*L.nt)
-                        if m > ave * (r-1):  # cluster nt, L,C_ by combined rim density:
+                        if m > ave * (r-1) and cluster_rim(_N, N_, r):  # cluster nt, L,C_ by combined rim density:
```
kwcckw, line 493: So from my understanding, we eval each new `_N` based on all of their rim, where we have internal M for rim's N that's in N_, and external M for rim's N that's outside the N_. 

For decay, from previous claude, actually that should be just the distance?

    closer in-C rim partners weighted more than distant ones.

Also this "multiple-linkage clustering" is only applicable in clustering, so there shouldn't be any changes in the comp? Since the existing Ls are added via 1-to-1 eval of N pair.

boris-kz: Decay is global attr. That weighting is for len(N_) side, not directly for link vals. Yes, comp is not affected

kwcckw: Okay, in term of the eval of the `_N` inclusion, that `m > ave * (r-1)` should be removed now since that's only eval based on 1-to-1?

Another thing is the sequence of the eval `L_` and `_N`. A later `_N` may get more `N` in `N_`, which may have higher included M? So do we need to sort those Ns first?

frame_2D_alg/meta_code.py
```diff
@@ -61,9 +61,10 @@ def split(Q, root=None):  # invert merge at most cost-unamortized gate
     spl_ = []  # draft:
     for F in [t for t in Z.typ_ if isinstance(t,CoF)]: spl_+= [split(F)]
     Z.typ_= spl_; typ_,mrg_ = [],[]
-    for t in Z.typ_.pop():
+    while Z.typ_:
```
boris-kz, line 64: for i, t in enumerate(Z.typ_):
        if t in mrg_: continue
        F = Copy(t)
        for f in Z.typ_[i:]: mrg_ += [merge(F,f)]  # merged fs, if any
        typ_ += [F]

frame_2D_alg/agg_recursion.py
boris-kz: cluster_N is fine as is, rim eval is separate extension, like cluster_P to cluster_C but inline.
this is still connectivity clustering, all >ave evals are included, why do you want to sort them?

frame_2D_alg/agg_recursion.py
```diff
@@ -142,7 +142,6 @@ class CoF(CF):
     _cur = contextvars.ContextVar('oF')
     def __init__(f, fo=1, **kw):
         super().__init__(**kw)
-        f.fo = kw.get('fo',1)
```
boris-kz, line 145: why did you remove it?

kwcckw: We only need the local fo to init gF? Because gF is None when fo == 0?

frame_2D_alg/agg_recursion.py
kwcckw: > cluster_N is fine as is, rim eval is separate extension, like cluster_P to cluster_C but inline.

Okay, so if this is following the sequence of cluster_P, it will be added externally to cluster_N? So it should be eval per N_ before sum2G? Or it should after all sum2G to refine those Gs again?

boris-kz: no, this extension can be inline:

    def rim_connect(_N, N_,span):
        in_M = ex_M = 0.0
        N_ = set(N_)
        for L in _N.rim:
            __N = L.nt[0] if L.nt[1] is _N else L.nt[1]
            if __N in N_: in_M += L.m
            else:          ex_M += L.m
        return (in_M / (ex_M or eps)) / (span*decay)  # normalized N-to-N_ match, sum m for sum2G?

    if _N_ and sum(vt_(Ft.dTT, Ft.root.wTT * ttcN)) * wcN * (len(_N_) - 1) > (ave + avd) * (r + ccN *  (len(_N_) - 1))  > 0:  #| fL?
        for N in _N_:
            N.fin=0; N.exe=1; sum2F(N.rim,N.Rt)  # only if N was added in trans-cluster?
        G_=[]; Gt_=[]; in_= set()  # root attrs
        for N in _N_:  # form G per remaining N
            if N.fin or (Ft.root.root and not N.exe): continue  # no exemplars in Fg
            N_ = [N]; L_,B_ = [],[]; N.fin=1  # init G
            __L_= N.rim  # spliced rim
            while __L_:
                _L_ = []
                    _N = L.nt[0] if L.nt[1].fin else L.nt[1]; in_.add(L)
                    if not _N.fin and _N in Ft.N_:
                        m,d = nt_vt(*L.nt)
                            span = np.sqrt(len(N_))  # refine by rim connectivity / norm_approx_span
                            if span>3 and rim_connect(_N,N_,span) < ave * (r-1): continue

boris-kz: Trace looks right structurally — flood-fill unchanged, modularity added as second gate after per-link `m > ave*(r-1)`. Three issues:

**1. Dimensional mismatch on threshold:**

`rim_connect` returns `(in_M/ex_M) / (span*decay)` — a dimensionless density ratio. Comparing it against `ave*(r-1)` (which is m-scale, used elsewhere as match threshold) doesn't share units. The criterion's natural scale is around 1.0 (in_M roughly matches ex_M when N is on a boundary), not `ave*0.3`.

Either:
- threshold should be dimensionless, e.g. `< 1` or `< 1/(r-1)`-style
- or `rim_connect` should return something m-scaled (e.g. `in_M - ex_M*norm`) directly comparable to ave

**2. `L.m` sign in sums:**

`L.m` can be negative (below-ave match deviation). Summing a negative m into `in_M` weakens the in-cluster signal incorrectly — a weak in-C link reduces in_M numerically but conceptually it's a contrast link, belongs in ex_M or B_-side. `nt_vt` already filters `if l.m > 0`. Match that:

```python
if __N in N_:
    if L.m > 0: in_M += L.m
else:
    if L.m > 0: ex_M += L.m
```

Or split: positive m → in_M/ex_M; negative m or positive d → boundary tally (separate from this gate, used for B_).

**3. Span proxy:**

`np.sqrt(len(N_))` assumes 2D disk packing — fine as O(1) approximation, but drifts for elongated/chain shapes. If you want the modularity-style guard to *kill* chains, the real span should be larger than sqrt(len) for chains, making density artificially lower → easier to keep adding → defeats the purpose. Actual yx-based span (max pairwise or std) catches chains correctly.

Cheap actual span: track running `yx_min`, `yx_max` as N_ grows; span ≈ `np.hypot(yx_max - yx_min) / 2`. O(1) per inclusion.

**Minor:**

- `span > 3` skip is fine (modularity unstable for tiny C), but threshold here is in sqrt-len units (>3 ≈ len>9). If switching to actual span, recalibrate.
- `nt_vt(*L.nt)` already gates per-link m before reaching `rim_connect`; both gates redundant on the per-link side. The modularity gate doesn't need `m > ave*(r-1)` to fire first — it's a stronger N-level test. Could replace, not stack.

frame_2D_alg/agg_recursion.py
boris-kz: ok, right

frame_2D_alg/agg_recursion.py
kwcckw: For 2, i guess we only need absolute value? Because any inclusion should increase in_M value, and any exclusion should increase ex_M value?

boris-kz: no, ex_M is not active exclusion, just not already in cluster. 2 is fine, neg links should inhibit inclusion, contrast is defined per link, not per N

boris-kz: Claude:

Acknowledged on both. Dropping the boundary framing — the test gates inclusion only, contrast/B_ stays per-link.

Then his original `in_M += L.m` / `ex_M += L.m` (signed sums) is correct as written: negative in-link reduces `in_M`, inhibiting inclusion. Clean.

**One residual issue with signed sums:**

`ex_M` can go negative (N has contrast links to outside C → negative L.m on ex side). When `ex_M < 0`:
- `in_M / ex_M` flips sign
- comparison `< ave*(r-1)` no longer means what we want

Three options:

```python
if __N in N_: in_M += L.m              # signed, neg inhibits inclusion
else:         ex_M += max(0, L.m)      # (a) clamp: only positive outward pull counts
```

(a) keeps ex_M ≥ 0; conceptually "pull to outside" is only positive. Negative ex-links (contrast to outside) don't pull outward, so they don't belong in ex_M.

Alt:
```python
if __N in N_: in_M += L.m              # signed
elif L.m > 0: ex_M += L.m              # (b) skip negative ex entirely
```

(b) is the same as (a) functionally.

Or:
```python
in_M = sum(L.m for L in _N.rim if other(L) in N_)
ex_M = sum(L.m for L in _N.rim if other(L) not in N_ and L.m > 0)
```

(c) explicit and readable.

**Threshold scale (separate from boundary issue):**

The criterion `(in_M/ex_M) / (span*decay)` with `ex_M` clamped is dimensionless and centered roughly around 1 for marginal inclusion (in_M ~ ex_M*span*decay). Comparing against `ave*(r-1)` — which is ~0.3 for r=2 — admits everything with even modest in/ex imbalance. If that's the intended permissiveness of cluster_N (let cluster_C tighten later), fine. If you want this to do real work in cluster_N, threshold needs to scale with span, e.g. `< (r-1)` or fixed dimensionless cutoff.

Span proxy concern from before still stands: `sqrt(len)` blesses chains by underestimating their span.

frame_2D_alg/agg_recursion.py
```diff
@@ -784,30 +805,45 @@ def F2N(F):  # convert for cross_comp
...
+
+    C = cls(**attrs)
+
+    # CF -> CoF
+    if isinstance(N, CoF) and cls==CoF:
+        C.fw, C.fc, C.fr = N.fw, N.fc, N.fr
+        C.call_ = copy(N.call_); C.typ_  = copy(N.typ_); C.N_ = copy(N.N_)
+        C.gF = Copy(N.gF, root=C,cls=CoF) if N.gF is not None else None
+    else:  # create new class instances for non CoF
+        C.N_ = [(n if isinstance(n, CN) else Copy(n, root=C) if isinstance(n, CF) else []) for n in N.N_]
```
boris-kz, line 824: why do you need to copy those?

frame_2D_alg/agg_recursion.py
boris-kz: not sure we still need ex_M at all, that's a left-over from the attempt to pre-define boundary Ns?
it's still meaningful for competitive inclusion, but that's not done in floodfill

kwcckw: > not sure we still need ex_M at all, that's a left-over from the attempt to pre-define boundary Ns?

Okay, so sign flipping is an issue because it may causing the eval to become false. I think ex_M may still relevant because it may not be the leftover but it's just not included to N_ yet?

boris-kz: that's not what I meant. ex_M is only meaningful for competitive inclusion, floodfill is not competitive

kwcckw: I see, then we can remove the ex_M, and the sign flipping issue is no longer relevant now.

frame_2D_alg/agg_recursion.py
kwcckw: In prior CopyF:

      def CopyF(F, root=None, r=1):  # F = CF
          C = CF(dTT=F.dTT*r, m=F.m, d=F.d, c=F.c, r=F.r, root=root or F.root, nF=F.nF)
          C.N_ = [(N if isinstance(N, CN) else (CopyF(N,root=C) if isinstance(N, CF) else [])) for N in F.N_]  # flat
          return C

We only need to Copy CF but not CN, but my code above is wrong since `isinstance(n, CF)` is true for all other inherited classes. We should use if `n.typ == 0` instead.

boris-kz: just copy(N.N_), the elements are not modified anyway?

frame_2D_alg/agg_recursion.py
boris-kz: def rim_connect(_N, N_,span):
        in_M = sum([L.m for L in _N.rim if (L.nt[0] if L.nt[1] is _N else L.nt[1]) in N_])
        return in_M / (span*decay)  # normalized N-to-N_ match, sum m for sum2G?

kwcckw: I think we had this something similar before if it's just in_M (maybe for the pL earlier), and this probably can be unpacked too since it's just 2 lines of code.

frame_2D_alg/agg_recursion.py
kwcckw: Okay, thanks. I was confused with the copy of CF in H earlier. We are not suppose to copy the element in N_.

frame_2D_alg/agg_recursion.py
boris-kz: right

frame_2D_alg/agg_recursion.py
boris-kz: so it's the same for other forks?

kwcckw: Yes, since we shouldn't copy N_ in CoF too, so copy(N_) and copy(L_) should be default.

boris-kz: in general, the higher classes should simply add to attrs?

boris-kz: Looking at `Copy_`:

The mess has two causes:

1. **Each subclass branch sets attrs *after* construction**, even though `__init__` already accepts them via `kw.get`. So every CL/CC/CN copy first builds defaults in `__init__`, then overwrites them. Plus duplicated `C.N_=copy(N.N_); C.L_=copy(N.L_)` (set in attrs *and* again outside, *and* again in CoF branch).
2. **Self-referential fields** (`gF`, `Nt/Bt/Ct/Lt/Xt/Rt`, `H`) need `root=C`, so they genuinely must be post-init.

Your fix works for (1): each subclass branch just extends `attrs`, single `cls(**attrs)` call. Only the self-referential stuff stays post-init.

Sketch:

```python
def Copy_(N, root=None, r=1, cls=None, init=0, typ=None):
    cls = cls or type(N)
    a = dict(dTT=N.dTT*r, m=N.m, d=N.d, c=N.c, r=N.r, fb_=list(N.fb_),
             root=root or N.root, nF=N.nF, wTT=copy(N.wTT), w=N.w,
             typ=N.typ if typ is None else typ,
             N_=copy(N.N_), L_=copy(N.L_))
    if isinstance(N, CL):
        a.update(nt=list(N.nt), yx=copy(N.yx), kern=copy(N.kern), span=N.span,
                 angl=[copy(N.angl[0]), N.angl[1]] if N.angl else None)
    if isinstance(N, CC):
        a.update(m_=list(N.m_), d_=list(N.d_))
    if isinstance(N, CoF):
        a.update(call_=copy(N.call_), typ_=copy(N.typ_), fw=N.fw, fc=N.fc, fr=N.fr)
    if isinstance(N, CN) and not init:
        a.update(box=copy(N.box), mang=N.mang, exe=N.exe, fin=N.fin, root_=list(N.root_))
    C = cls(**a)
    # self-referential, need root=C:
    if isinstance(N, CoF) and N.gF is not None: C.gF = Copy_(N.gF, root=C, cls=CoF)
    if isinstance(N, CN):
        C.sub = N.sub; C.compared = copy(N.compared)
        if init:
            C.yx=[N.yx]; C.angl=[copy(N.angl[0]),N.angl[1]] if N.angl is not None else None
            C.L_=[l for l in N.rim if l.m>ave]; N.root=C; C.fin=0; C.N_=[N]
        else:
            for f in ('Nt','Lt','Bt','Ct','Xt','Rt'): setattr(C, f, Copy_(getattr(N,f), root=C))
            C.H = [Copy_(lev, root=C) for lev in N.H]
            C.angl = deepcopy(N.angl); C.yx = copy(N.yx)
    return C
```

Two snags worth flagging before you apply:

- **`CN.__init__` hardcodes `sub=0` and `compared=set()`** — they don't accept kw. So either keep them post-init (as above) or change `CN.__init__` to `n.sub = kw.get('sub',0)` and `n.compared = kw.get('compared',set())`. Latter is cleaner and folds them into the attrs dict.
- **`CN` `init=1` case overrides `angl/yx/L_/N_`** that CL.__init__ already set from kw. To avoid double-work you'd skip those keys in `a` when `init=1` and `isinstance(N,CN)` — but the current condition `isinstance(N,CN) and not init` for box/mang/etc. is the same pattern; in `init=1` mode you're constructing a fresh G, so it's fine to leave CL kwargs out entirely and let the post-init block fill them. Worth deciding which path is canonical.

Want me to also push `sub`/`compared` into `CN.__init__` kwargs?

kwcckw: Yes, that's the current `CF->CL->CN/CC` or `CF->CoF` structure. Is that a problem with this structure?

boris-kz: I posted Claude above

boris-kz: def Copy_(N, root=None, r=1, cls=None, init=0, typ=None):

    cls = cls or type(N)
    a = dict(dTT=N.dTT*r, m=N.m, d=N.d, c=N.c, r=N.r, fb_=list(N.fb_), root=root or N.root,
             nF=N.nF, wTT=copy(N.wTT), w=N.w, typ=N.typ if typ is None else typ,
             N_=copy(N.N_), L_=copy(N.L_))
    if isinstance(N, CL):
        a.update(nt=list(N.nt), yx=copy(N.yx), kern=copy(N.kern), span=N.span,
                 angl=[copy(N.angl[0]), N.angl[1]] if N.angl else None)
    if isinstance(N, CC):
        a.update(m_=list(N.m_), d_=list(N.d_))
    if isinstance(N, CoF):
        a.update(call_=copy(N.call_), typ_=copy(N.typ_), fw=N.fw, fc=N.fc, fr=N.fr)
    C = cls(**a)
    if isinstance(N, CoF) and N.gF is not None: C.gF = Copy_(N.gF, root=C, cls=CoF)
    if isinstance(N, CN):
        C.sub = N.sub; C.compared = copy(N.compared)
        if init:
            C.yx=[N.yx]; C.angl=[copy(N.angl[0]), N.angl[1]] if N.angl is not None else None
            C.L_=[l for l in N.rim if l.m>ave]; N.root=C; C.fin=0; C.N_=[N]
        else:
            for f in ('Nt','Lt','Bt','Ct','Xt','Rt'): setattr(C, f, Copy_(getattr(N,f), root=C))
            C.H = [Copy_(lev, root=C) for lev in N.H]
            C.angl=deepcopy(N.angl); C.yx=copy(N.yx); C.box=copy(N.box)
            C.mang=N.mang; C.exe=N.exe; C.fin=N.fin; C.root_=list(N.root_)
    for n in C.N_+C.L_: n.root = C
    return C

kwcckw: Okay, so the sub and compared should be accept kw. 

For the problem in 1, we can convert to that format (init all kw params first), but i think the existing structure is clearer since we only need to check isinstance once.

frame_2D_alg/meta_code.py
boris-kz: Claude:
What I see open:

- `init_wc` produces a list aligned with `onF_` (16 entries), matching `Fc_` in agg_recursion. But there's no wiring — agg_recursion's `Fc_` is still hand-set, meta_code would need to import-cycle-back, or agg_recursion needs to call into meta_code at startup.
- `get_wc` block mode takes hardcoded line numbers `(225, 243)`. No automatic block extraction (if/else, loops, conditional bodies) — but that's what `Ec_` per gate would need.
- `weights` dict is uniform per AST node type. Doesn't distinguish dTT-tensor ops from scalar ops, which differ by ~9x in actual cost.

What's the actual current intent — wiring init_wc into agg_recursion, automating block detection, refining the weights, or something else?

kwcckw: So init_wc creates weights based on onF_, while get_wc can be used in runtime to get weight of particular functions or block of code. 

We shouldn't need get_wc now since we don't need to retrieve their code weight during the runtime anymore. We still need init_wc to get the new weights if there's added onF_ or edited code structure.

PR #141, opened 2026-05-11
5/11/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -803,7 +806,7 @@ def Copy_(N, root=None, r=1, cls=None, init=0, typ=None):
             for f in ('Nt','Lt','Bt','Ct','Xt','Rt'): setattr(C, f, Copy_(getattr(N,f), root=C))
             C.H = [Copy_(lev, root=C) for lev in N.H]
             C.angl=deepcopy(N.angl); C.yx=copy(N.yx); C.box=copy(N.box); C.mang=N.mang; C.exe=N.exe; C.fin=N.fin; C.root_=list(N.root_)
-    for n in C.N_+C.L_: n.root = C
+    for n in C.N_+C.L_: n.root = C  # this should be optional? Because we may not need to update n.root to the copied C
```
kwcckw, line 809: I think this root assignment should be optional? Because we shouldn't update n.root in some cases, for example:

Nt.N_'s root shouldn't be updated to T.

frame_2D_alg/meta_code.py
```diff
@@ -39,13 +39,16 @@ def eval(mrg):  # after adding multiple oFs in mrg.call_, not just the pair in s
...
 
-    g = max(Q.gF.call_, key=lambda g: max(b.fc*b.c - b.fw*b.c for b in g.call_))
     Qs = []
-    for b in g.call_:
-        S = CoF(nF=Q.nF, root=root or Q.root); S.call_=[b if f is g else f for f in Q.call_]
-        S.gF.call_=[x for x in Q.gF.call_ if x is not g]
-        S.c=b.c; S.fw=sum(f.fw for f in S.call_ if isinstance(f,CoF)); S.fc=Fc_[Q.nF]
-        Qs += [S]
+    for gF in Q.call_:
+        if gF.nF == "E":
```
kwcckw, line 44: Could you clarify if the gF is still relevant?

Right now we add gate during the merging when nFs are different, so this split should be applied to those gates too?

Also the split workflow should be valid for the call, `g` with highest `b.fc*b.c - b.fw*b.c ` in gate.call_?

Then we should split g.call_ into multiple oFs, and those oFs should be repacked into gate.call_, and g should be removed from gate.call_?

frame_2D_alg/agg_recursion.py
```diff
@@ -231,7 +231,7 @@ def cross_comp(root, rr):  # core function mediating recursive rng+ and der+ cro
             if G_:
                 if not root.typ: F2N(root)  # promote @ 1st sub+ or agg+
                 root.H+= [sum2F(L_,root,froot=1)]  # lev: L_+ derivatives
-                root.Nt = sum2F(G_,froot=1); L=len(G_)-1  # or spliced C_?
+                root.Nt = sum2F(G_,froot=1); root.Nt.root = root; L=len(G_)-1  # or spliced C_?
```
boris-kz, line 234: root.Nt = sum2F(G_,root,froot=1); L=len(G_)-1
?

frame_2D_alg/agg_recursion.py
```diff
@@ -491,7 +491,7 @@ def trans_cluster(G):
                         m,d = nt_vt(*L.nt)
                         if m > ave * (r-1):  # cluster nt, L,C_ by combined rim density:
                             span = np.sqrt(len(N_))  # approx span
-                            if span > 3:  # refine by rim connectivity / norm span
+                            if span > 3:  # refine by rim connectivity / norm span (or this 3 should be ave too?)
```
boris-kz, line 494: eventually, I think 3 is fine for now. ultimately, these aves should be assigned automatically

frame_2D_alg/agg_recursion.py
```diff
@@ -686,7 +688,7 @@ def add2F(F, n, merge=0, fr=0, fo=0):  # unpack for batching in sum2F
         if isinstance(F,CoF) and isinstance(n,CoF): F.fw=n.m
     if merge <2:
         if fo: F.typ_+= [n]
-    else: F.N_ += (n.N_ if merge else [n])
+        else: F.N_ += (n.N_ if merge else [n])  # we shouldn't append or merge N_ when merge == 2
```
boris-kz, line 691: thanks

frame_2D_alg/agg_recursion.py
```diff
@@ -545,6 +546,7 @@ def cluster_C(Ft, E_,_r,_c):  # form centroids by clustering exemplar surround v
         for _C in _C_:  # C.m,d /rTT? sort / sum(_C.m_)?
             N__,n_,m_,d_,M,D,T,R,dTT,up = [],[],[],[],0,0,0,0, np.zeros((2,9)),0  # /C
             for n in _C.N_+_C._N_:  # current + frontier
+                if not hasattr(n, '_root_'): n.root_,n.m_,n.d_,n._root_,n._m_,n._d_ = [],[],[],[],[],[]  # n in rim but not in Ft.N_
```
boris-kz, line 549: if n not in Ft.N_: ?

frame_2D_alg/agg_recursion.py
boris-kz: so add froot in args?

frame_2D_alg/agg_recursion.py
kwcckw: We may have this loop in multiple iterations (in the refinement while loops), so we need to check with hasattr? Because they might be already added with those params in the prior iterations.

frame_2D_alg/agg_recursion.py
kwcckw: That froot is update N.root to root:
      
      if froot:

So we need to manually update F.root here.

frame_2D_alg/agg_recursion.py
kwcckw: Actually i don't see we need to update this root assignment after the copy? Their root should be remained the same since this is just a copy.

boris-kz: the root is for feedback, we don't need it if the elments are not updated.
so I guess it should be optional.

frame_2D_alg/agg_recursion.py
boris-kz: make it froot=2?

    if froot==1:
    elif froot==2: F.root=root 
    return F

frame_2D_alg/agg_recursion.py
boris-kz: we probably need to make a local copy of Ft.N_ and add all checked Ns, so we don't go over them again?

boris-kz: we already have a local copy, so

                if n not in N_: N_+=[n]; n.root_,n.m_,n.d_,n._root_,n._m_,n._d_ = [],[],[],[],[],[]

frame_2D_alg/agg_recursion.py
kwcckw: Alright, that should be better in case we need it again, thanks.

frame_2D_alg/agg_recursion.py
kwcckw: But that requires recursive checking of their rim, which is backward? Since we will get their rim again after that.

boris-kz: why is it backwards, all C.N_ rims should be checked for extension?

kwcckw: Sorry, i misunderstood earlier. I was thinking we will pre-init those params from their rim first. So we need to N_ = copy(Ft.N_), else we will be packing it into Ft.N_.

boris-kz: ok right

frame_2D_alg/meta_code.py
boris-kz: yes, we need new gF / new oF, both will collect new data

frame_2D_alg/meta_code.py
```diff
@@ -39,13 +39,16 @@ def eval(mrg):  # after adding multiple oFs in mrg.call_, not just the pair in s
 # not revised:
 def split(Q, root=None):  # invert merge at most cost-unamortized gate
```
boris-kz, line 40: this doesn't have to be reverse merge, we can split base oF too

kwcckw: > yes, we need new gF / new oF, both will collect new data

So if the split doesn't limited to gF, it could be all oFs now, including all typs and their nested calls?

The general idea is split the call with the highest cost in oF.call_ into multiple calls, and repack them into oF.call_?

In term of new gF/oF, are them redundant to the oF with typ == "E"? We only need either one of them since oF with typ == "E" should work the same?

boris-kz: typ == "E" is a simplification, evals may have various complexity levels.
yes, gFs are also oFs, but specifically assigned to root oF, with specific data.

the split was was never for gF itself, that's just an eval to call paired oF, which may be split.

kwcckw: Okay, in the prior gF structure, it's paired with oF, so i think it actually could be packed in oF.call_? Since all levels of oFs maybe split or merge, it could be treated as a different typ?

boris-kz: no, gF is always secondary to oF, which may not be gated: gF==None 
splitting is not directly related to gFs, it's just clustering oF.call_ by callees that are often called toghether

kwcckw: I see, so they are different kind of gates. That splitting oF should be renamed to something like a "group" for clarity?

boris-kz: we only have one kind of gate oFs now, they are not in onF_
we can call it cluster_call_

kwcckw: Okay, so we still need that 'E' for merging purpose since it's unrelated to the gate oF. Could you explain a bit on the general idea of the splitting here? It's just repack oF.call_'s calls into oF.call_?

boris-kz: something like this, we need to add oF.caller_:

def cluster_call_(rF, root=None):  # cluster rF callees that are called toghether?

    Sub_,sub_ = [],[]  # _C = rF.call_[0]; sub_ = [_C]
    _C = []
    for C in rF.call_[1:]:  # oFs, or all top-level AST items?
        if isinstance(C,CoF):  # call history eval, gated oFs only?
            if _C and C.caller_ != _C.caller_:  # pseudo, flesh it out
                if len(sub_) > 5:  # need real eval here
                    Sub_ += [sum2F(sub_)]; sub_ = [C]
                    continue  # else keep accumulating sub_?
        sub_ += [C]; _C = C
        '''
        S = CoF(nF=b.nF, root=root or gF.root); S.call_=[b if f is g else f for f in gF.call_]
        S.call_=[x for x in roF.gF.call_ if x is not g]
        S.c=b.c; S.fw=sum(f.fw for f in S.call_ if isinstance(f,CoF)); S.fc=Fc_[b.nF]
        Sub_ += [S]
        '''
    return Sub_  # may be empty

kwcckw: I see and that caller should be just all calls' root? 

>  C.caller_ != _C.caller_:

This should eval their length and also the sequence?

So this Sub_ is the split oFs, which should be replacing rF call_?

boris-kz: yes, for oF calls. I think most of ops in call_ are not oFs
also we need a deeper caller_ comparison than identity

Sub_ will be added to / replace Z.typ_ in the next frame

boris-kz: Claude:

Looked through the draft and surrounding code. Main conceptual issues, in priority order:

**1. `caller_` semantics undefined**
Tracer assigns single `oF.root = _CoF` per instance — each traced oF already has its full call path via `oF.root.root...` up to `Z`. So `caller_` is either:
- a cached path tuple (for fast deep comparison), or
- a list accumulated across multiple invocations of the same nF.

Your comment "deeper caller_ comparison than identity" only matters in the second case — if `caller_` is multi-invocation, then two CoFs with the same nF but different call histories should split. If it's the single trace path, then immediate `root is _root` already encodes it. Which one?

**2. What's actually in `rF.call_`?**
Tracer only appends CoFs (`_CoF.call_ += [oF]`). AST items aren't there. The `isinstance(C, CoF)` guard plus the "oFs, or all top-level AST items?" comment suggests the plan is to populate call_ with AST items too — but that's a separate change. Right now every C in the loop is a CoF, so the guard never fires.

**3. State bug in the reset branch**
```
if len(sub_) > 5:
    Sub_ += [sum2F(sub_)]; sub_ = [C]
    continue          # skips _C = C
```
After `continue`, `_C` still points to the C from before the reset. Next iteration compares against stale `_C`. Should be `sub_ = [C]; _C = C; continue`.

**4. Trailing `sub_` never flushed**
Last group is discarded. Needs `if sub_: Sub_ += [sum2F(sub_)]` after the loop.

**5. Eval `len(sub_) > 5` placeholder**
You flagged this. Real eval should compare group's accumulated fw against cost of forming a new typ (~Fc_ entry + sum2F cost). The `> 5` length cutoff has no value semantics.

**6. Output identity / merge compatibility**
`sum2F(sub_)` collapses mixed-nF items into one CoF with no clean nF. But `merge(F,f)` aligns by `Sub.nF == sub.nF`. New synthetic typs need either:
- a new nF id appended to onF_ (mutates the global indexing scheme — heavy), or
- merge() changed to align by something else (caller_ pattern? sub-sequence match?).

This isn't addressed yet and breaks the spl_ → Z.typ_ → merge pipeline in `__main__`.

**7. `__main__` type confusion**
```
spl_ += [cluster_call_(F)]   # list of Sub_ (lists)
Z.typ_ = spl_                # Z.typ_ becomes list-of-lists
for i, t in enumerate(Z.typ_):
    F = Copy_(t, cls=CoF)    # t is a list, not a CoF
```
`Copy_` expects a CoF. Either flatten spl_ or iterate the inner lists.

---

Clarifying questions before I touch the code:
- Is `caller_` a cached path-tuple (single-invocation), or a list across multiple invocations of the same nF?
- Should non-CoF AST items be added to call_ first (separate change to tracer), or is cluster_call_ CoF-only for now?
- For "deeper comparison than identity" — is the intent prefix match (shared ancestor depth), full-path equality, or LCS-style alignment?

kwcckw: That rF should be a typ? If yes, we may get multiple subs that have the same type for the next frame Z.typ_? 

> we need a deeper caller_ comparison than identity

Something like the prior ast features comparison?

boris-kz: cluster_call_ should generate new typs, possibly removing rF
not sure about comp


OK. Two follow-up clarifications before proposing the deeper comparison:

**A. When does caller_ accumulate?**

If `caller_` is "list across multiple invocations of same nF", then either:
- **Within one frame, post-aggregation:** rF.call_ is deduplicated per nF, each entry has caller_ = roots from all source instances. But then iterating rF.call_ loses the *sequence* needed for "called together" boundary detection.
- **Across frames:** each typ-CoF persists and its caller_ grows over multiple traces. cluster_call_ would then run on Z.typ_ entries, not raw Z.call_. But your `__main__` runs it on `Z.call_`, not `Z.typ_`.
- **Both:** raw per-invocation instances in call_ retain order, and each instance's caller_ is still a list because of cross-frame accumulation.

Which is intended?

**B. Once A is settled, options for deeper comparison:**

Given two callees C1, C2 with `caller_` lists of roots (each root itself has a `.root` chain back to Z):

1. **Path-prefix match** — count matching levels in `C1.caller_[i].root` chain vs `C2.caller_[j].root` chain, averaged. Captures "called from similar depth in similar contexts".
2. **Multiset overlap (Jaccard on nFs)** — `|set(c.nF for c in C1.caller_) ∩ set(c.nF for c in C2.caller_)| / |union|`. Cheap, ignores ordering, captures co-occurrence in same parent types.
3. **Sequence alignment over caller_** — LCS or similar on the ordered lists. Captures call-history *trajectory*, expensive.
4. **Frequency vector** — count nF occurrences in caller_, compare as TT-style vectors with vt_. Fits your existing match/diff machinery, gives m,d output directly usable downstream.

(4) is the most consistent with the rest of the codebase — it produces the same m,d shape that everything else generates, and the threshold becomes the same kind of `ave*cost` eval you use elsewhere. (2) is the cheap version of (4).

Lean toward (4)? Or is the "called together" notion more about *positional* adjacency in the parent's call_ (i.e., sequence-based, option 3)?

kwcckw: Each type is corresponding to a function, so if they are removed , we need to update all of those mappings such as the onF_, Fw_, etc, and probably need to recompute their ast cost during runtime.

boris-kz: That's for a new frame, we can use a whole new set of functions

kwcckw: Okay, at this moment the code is static, so the code generation to create new set of functions should be included at a higher level later?

boris-kz: Yes, we just need to form these functions

PR #142, opened 2026-05-12
5/12/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -52,8 +75,8 @@ def cluster_call_(rF, root=None):  # cluster rF callees that are called toghethe
     add_typ_(Z)  # each call_ in Z.typ_ is the flatten calls of same typ
     # add fffeedback to reform Z for next frame_H:
     spl_ = []
-    for F in Z.call_:
-        if isinstance(F,CoF): spl_ += [cluster_call_(F)]  # add eval?
```
kwcckw: So the cluster_call_ should be applied to flatten Z.call_? Or it's a typo and it should be Z.typ_ instead?

frame_2D_alg/meta_code.py
```diff
@@ -30,16 +30,39 @@ def merge(F,f):  # combine aligned ops, if-fork per miss, no inline recursion, f
...
+            
+        # sub_:
+        sub_eval = sum([sub.fc for sub in sub_+[C]])/(len(sub_) + 1) < ave * 10
+    
+        return caller_eval and sub_eval
+
+    Sub_,sub_,_C = [],[],[]
     for C in rF.call_[1:]:  # oFs, or all top-level AST items?
         if isinstance(C,CoF):  # call history eval, gated oFs only?
-            if _C and C.caller_ != _C.caller_:  # pseudo, flesh it out
```
kwcckw: The `C.caller_` here always have single root or caller unless it is a typ oF, because there should be only a single caller per oF?

frame_2D_alg/meta_code.py
```diff
@@ -30,16 +30,39 @@ def merge(F,f):  # combine aligned ops, if-fork per miss, no inline recursion, f
...
-                    continue  # else keep accumulating sub_?
-        sub_ += [C]; _C = C
-    return Sub_  # may be empty
+            if _C and comp_call(_C, C, sub_):  # pseudo, flesh it out
+                Sub_ += [sum2F(sub_)]; sub_ = [C]
+            else: sub_ += [C]
+            _C = C
+    if sub_: 
+        Sub_ += [sum2F(sub_)]  # last sub_
+    return Sub_  # may be empty  (each element have the same typ here)
```
kwcckw: If rF is a typ oF, their calls are all have the same typ. So we are actually splitting their calls into multiple typ oFs, but with the same typ?

frame_2D_alg/agg_recursion.py
```diff
@@ -142,6 +142,7 @@ class CoF(CF):
     def __init__(f, fo=1, **kw):
         super().__init__(**kw)
         f.call_ = kw.get('call_',[])  # top-level AST items
+        f.caller_ = kw.get('caller_',[])  # roots that call this oF
```
boris-kz, line 145: we can use f.root as a list, or f.root_ if root is used as higher oF in a nested oF?

frame_2D_alg/agg_recursion.py
```diff
@@ -160,6 +161,8 @@ def inner(*a, **kw):
             if oF.call_:
                 for i, F in zip((1,0),(oF, oF.gF)):
                     tree = flat_(oF); L=len(tree)-1
+                    if i:
+                        for call in tree: call.caller_ += [oF]   # add caller for flatten calls? Else actually root is already sufficient
```
boris-kz, line 165: probably pack it in list root, as above

frame_2D_alg/meta_code.py
boris-kz: single caller per call_[i], concat in multiple callers per typ_[i]?

frame_2D_alg/meta_code.py
boris-kz: typ can stay the same for now, but we generate nested numerical nFs here, using nF as a list?

frame_2D_alg/meta_code.py
boris-kz: should be Z.typ_

frame_2D_alg/agg_recursion.py
boris-kz: def add_typ_(oF):  # record oF vals for weighting, mapped to global FTT_

    for F in flat_(oF): 
        C = Copy_(F); C.root=[C.root]; typ_[F.nF] += [C]
    for i, F_ in enumerate(typ_):
        if F_: T =sum2F(F_,CoF()); T.nF=i; T.wTT=cent_TT(getattr(T,'rTT',T.dTT),T.r); typ_[i]=T
    oF.typ_ = typ_

frame_2D_alg/meta_code.py
boris-kz: actually no, we are splitting the call_, not typ_. typ_ is not oF AST

frame_2D_alg/agg_recursion.py
kwcckw: Yes, i guess root can be dynamic now where it's a list when it's a nested typ oFs with a list nFs.

frame_2D_alg/meta_code.py
kwcckw: Yes, i think we may get groups of different nFs now since input is oFs, we would need a list to pack them. I guess a different combination of numerical nFs would create a new typ.

frame_2D_alg/meta_code.py
kwcckw: Yes, so we use calls instead of typ, which mean it's always single caller?

frame_2D_alg/meta_code.py
kwcckw: If this is call_, so we should use the flatten call_, and regroup them into new typs?

boris-kz: sorry, the outer loop is Z.typ_, but we segment typ.call_, not flattened

frame_2D_alg/meta_code.py
boris-kz: we can map each call to its typ, then call.caller_ is typ.root:

    for F in flat_(oF):  # flattened call tree
        C = Copy_(F); C.root=[C.root]; typ_[F.nF] += [C]  # calls per typ
    for i, F_ in enumerate(typ_):
        if F_: 
            T = sum2F(F_,CoF())
            T.root=[F.root for F in F_]; T.nF=i; T.wTT=cent_TT(getattr(T,'rTT',T.dTT),T.r); typ_[i]=T

kwcckw: > call.caller_ is typ.root

Yes, thanks. This looks clear to me now.

boris-kz: def add_typ_(oF):  # record oF vals for weighting, mapped to global FTT_

    for i, F_ in enumerate(typ_):
        if F_:
    oF.typ_ = typ_

boris-kz: OK, two things to nail down before drafting:

**1. Iteration structure — confirming:**
```python
for T in Z.typ_:        # outer
    if T:
        for C in T.call_:   # per-instance (one invocation of nF=T's index)
            cluster_call_(C)   # walks C.call_, the children of this invocation
```
Each `C.call_` has children with mixed nFs, so caller_ lookups (`Z.typ_[child.nF].root`) are non-degenerate. Right?

**2. "Tolerant caller_ comparison":**

A real problem with strict lockstep comparison: `_C.caller_` and `C.caller_` are independent typ.root lists (callers of nF1 and nF2 respectively). Their orderings are unrelated — position k in one doesn't correspond to position k in the other. So positional `zip` similarity isn't meaningful.

Two readings of what you want:

**(a) Fork-as-cost-only:** alignment metric that counts mismatches as fork-cost, threshold decides match. The "fork" exists only in the accounting, not in cluster_call_'s output. Then the natural form is multiset/sequence alignment, not positional zip:
```python
def caller_sim(A, B):  # lists of root CoFs
    # project to nFs for "deeper than identity"
    a_, b_ = [r.nF for r in A], [r.nF for r in B]
    common = (Counter(a_) & Counter(b_)).total()
    forks = max(len(a_), len(b_)) - common
    return forks / max(len(a_), len(b_), 1) < tol
```

**(b) Fork-as-structure:** cluster_call_ actually emits gate CoFs (`nF='E'`) at mismatch positions inside sub_, mirroring merge()'s output. Then sub_ becomes a mixed sequence of plain Cs and fork-Cs, and Sub_ entries inherit this nested structure.

(a) is the natural fit for "decide group boundary"; (b) duplicates merge()'s job inside cluster_call_. Which?

**3. Multiset vs. positional:**
Even with (a), strict same-order matters only if invocation order in typ.root is semantically meaningful. It's just trace-order accumulation, so multiset/sorted comparison is more honest. Disagree?

**4. Identity vs nF projection:**
Comparing root CoFs by `is` vs by `.nF` matters for cross-frame: same `.nF` across different CoF objects means "same kind of caller". This is the actual "deeper than identity" — agreed?

Once these are settled I'll draft cluster_call_ + the bug fixes + the __main__ rewrite in one diff.

frame_2D_alg/meta_code.py
kwcckw: So when we segment typ.call_, all of the calls have the same typ, we need to redefine their typ with calls' calls?

boris-kz: who cares, we have to work out comparison first

frame_2D_alg/meta_code.py
boris-kz: Right, so the pivot: don't compare typ.root *lists*; compare the *bodies* (call_) of those root CoFs. The similarity question becomes "do the callers of nF1 and the callers of nF2 have similar call_ sequences?" — capturing "called together" via shared-body-shape rather than shared-identity.

Couple of levels this can land at, want to pick one:

**(a) Aggregate-multiset:** for each typ T, build `ctx(T) = Counter(child.nF for R in T.root for child in R.call_)`. Compare two typs by Counter overlap. Cheap, ignores order, captures "what kind of siblings does this typ have on average".

**(b) Summary-sequence alignment:** for each typ T, summarize R.call_ across R in T.root into one representative sequence (merge-style align across all Rs). Then compare two typs by another merge-style align of their summaries. Captures sequence; expensive (two layers of alignment) and the inner summary is fragile when caller bodies diverge.

**(c) Co-occurrence in caller bodies:** for each candidate pair (_C.nF, C.nF), scan `[R.call_ for R in Z.typ_[rF.nF].root]` — count how often the pair appears as adjacent (or near-adjacent) callees in a body. Directly measures "called together" empirically, no recursive alignment needed.

(a) is the cheap proxy, (c) is the most semantically direct, (b) is closest to the merge() pattern you quoted but doubles the cost.

Which one matches what you have in mind? My guess is (c), since it's the literal definition of "called together" and reuses the merge()-style position-aware comparison only at one level.

Also: does the comparison output need to be just a similarity scalar (for the boundary threshold), or actually produce fork structure (gate CoFs with nF='E') that goes into Sub_'s output? Earlier you tied the comparison "similar to merge()" — but merge builds output structure, whereas cluster_call_ in its draft form just groups.

boris-kz: this function splitting problem is not unique to cogalg?

No, it's a well-trodden problem in several literatures. The closest analogs:

**1. Profile-guided function clustering (PGO).** Compilers cluster co-called functions for code locality. Facebook's HFSort / HFSort+ (Ottoni & Maher, 2017) is the canonical reference: weighted call graph (edges = observed co-call frequency), greedy clustering by edge weight. Used in LLVM/BOLT, gcc-suite. The "called together" notion is exactly there.

**2. Sequence chunking / grammar induction.** Sequitur (Nevill-Manning & Witten, 1997), RePair, byte-pair encoding (BPE). Walk a sequence, replace recurring bigrams with new symbols, iterate. Your option (c) is essentially BPE on call traces. These don't use caller-context similarity, just direct co-occurrence — which is the simpler signal and works.

**3. Microservice decomposition.** Whole literature on partitioning monolith call graphs into services. Tools build call graphs (static or dynamic), then apply community detection (Louvain, modularity-based clustering). E.g., "Mono2Micro" (IBM), CARGO, Bunch.

**4. Code clone detection / refactoring suggestions.** "Extract method" tools (Sourcery, ReSharper, IntelliJ) use cohesion heuristics over local variable use and call sequences. More syntactic than what you want.

**5. Phase detection in execution traces.** SimPoint and similar — segment execution traces into phases via clustering BBV (basic block vectors). Sequence-position-aware, change-point style.

**Where this maps to your design choices:**
- Option (c) co-occurrence = BPE / Sequitur, well-studied, cheap, no recursion needed.
- Option (a) aggregate-multiset = HFSort-style edge weights without the graph clustering layer.
- Option (b) summary-sequence alignment = unusual; closest to multiple-sequence alignment in bioinformatics. Expensive and brittle.

**What's novel for cogalg, if anything:**
The actual algorithm choice isn't novel. What could be novel is using cogalg's own m/d/c metric and recursive aggregation as the similarity function instead of frequency counts — making the splitting decisions consistent with the rest of the system rather than bolting on standard clustering. But the structural problem (segment a call sequence by co-call statistics) is solved many times over.

Practical suggestion: start with (c) BPE-style or (a) HFSort-style. They're cheap, give immediate results, and you can swap in cogalg-native similarity later without restructuring.

frame_2D_alg/meta_code.py
boris-kz: So merge and into cluster_call_ both compare typ.call_s, but the former evaluates identity and the latter co-occurence, which requires tracking their callers.
That is internal and external similarity, we need to combine them? 
Another difference is is in the grain: merging the whole call_s, vs. parts of them, vs. none.

boris-kz: <html><head></head><body><p>Right, I was sloppy — those are genuinely different signals, not dual axes of one comparison.</p>
<p><strong>The two distinct signals:</strong></p>
<ul>
<li><strong>Internal (merge):</strong> are the two typs' <em>contents</em> structurally identical at the position level? Lockstep nF-equality of their call_s.</li>
<li><strong>External (cluster_call_):</strong> are the two typs <em>used</em> in similar contexts? Co-occurrence in caller bodies, requires caller_ tracking.</li>
</ul>
<p>A typ-pair can score high on one and low on the other — same computation in different contexts, or different computations called together. They answer different questions and drive different actions. Collapsing them into a single match scalar would dilute both.</p>
<p><strong>On the grain axis:</strong></p>

  | whole-typ | partial | none
-- | -- | -- | --
internal-id | merge() | (absent) | leave
external-co | (absent) | cluster_call_() | leave


<p>Two cells are empty:</p>
<ul>
<li>internal × partial = align two typs' call_s, merge the matching sub-sequence only, keep divergent parts as forks. This is what merge() <em>almost</em> does but commits the whole or nothing.</li>
<li>external × whole = two typs called from overlapping caller sets → merge them entirely on contextual grounds, regardless of internal content. Probably rare and weird.</li>
</ul>
<p>The internal × partial cell looks more useful — gradient between full merge and no merge.</p>
<p><strong>On "combining":</strong></p>
<p>Not as one scalar. What can be shared is the <strong>primitive</strong>: <code>comp_typ(A, B)</code> that returns <em>both</em> an internal-match score (from aligning A.call_ vs B.call_) <em>and</em> an external-match score (from comparing A.root list vs B.root list, with caller-body projection). The two scores then route to different actions: high internal → merge (whole or partial); high external → cluster grouping; neither high → leave.</p>
<p>So one primitive, two outputs, branching actions. Not unification of the operations, but extraction of the shared substrate.</p>
<p><strong>Scale-free reading:</strong></p>
<p>The same primitive runs at every level — call_ inside call_ inside call_ — producing internal and external scores at each level, each driving the appropriate action. That matches cogalg's general pattern (one comp, multiple match dimensions, scale-recursion).</p>
<p><strong>Open question for you:</strong>
Is the partial-internal cell worth filling now (gradient merge), or stay binary (merge/leave) and let cluster_call_ handle gradient externally? My read: partial-internal subsumes merge() and gives a cleaner gradient — merge() becomes the "100% internal match" special case. But that's a bigger refactor.</p></body></html>

kwcckw: > That is internal and external similarity, we need to combine them?

Yes, i think it should be done in a single comp. For internal similarity, other than using ast features, i guess we only can check their nF. For external similarity, that will be comparing their callers, probably just (number of same callers / total callers)?

> The two scores then route to different actions: high internal → merge (whole or partial); high external → cluster grouping; neither high → leave.

I think high internal and high external also should be merging/grouping calls here? It's just probably the grains will be different for each cases.


> Another difference is is in the grain: merging the whole call_s, vs. parts of them, vs. none

If we need to merge the whole call_s, that will be merging calls between two typ oFs?

frame_2D_alg/meta_code.py
kwcckw: So i would still think existing method (b) is the better choice since it preserve the sequence? The suggested (c) compares the occurrence of the pair from the root's calls sequence, which may not fully capture the sequence information.

frame_2D_alg/meta_code.py
boris-kz: whole call_ match will merge oFs,
partial call_ match will split oFs, if the gaps between matching sequences are significant

>(number of same callers / total callers)?

probably each weighted by c
so when different sequences are called toghether they should be spliced: there is no separate condition to call each?

kwcckw: > partial call_ match will split oFs, if the gaps between matching sequences are significant

Okay, so this may or may not merge between oFs now.

> so when different sequences are called together they should be spliced: there is no separate condition to call each?

Could you clarify further on this? Sequences are calls right? Why we need to call multiple calls? Or do you mean those sequences should be spliced or merged into an oF when we need to call them?

boris-kz: splicing is when we don't need a separate gate per sequence, they are called toghether anyway.
for spliting, it's >ave cost of adding forks, same as currently done in merge()

Claude:
this flips the framing in a useful way.

**Cost of separation as the eval:**

Right. Similarity is just a proxy. The real question is whether keeping two things separate is wasteful enough to justify the cost of merging (which includes producing forks or splices to carry over residual structure).

Compression criterion: merge if `separation_cost > merge_cost + residual_structure_cost`. That's it. Similarity feeds the calculation only as input to estimating overlap (which lowers separation cost — the more overlap, the more duplication you're paying for by staying separate).

**Where separation cost manifests:**

- **Across-seq (merge level):** keeping two near-identical typs separate means paying twice for the matched portions of their call_s — duplicate storage, duplicate downstream comparison work, duplicate updates on feedback. The residue when you merge is forks at the divergent positions. Forks are the *minimum* structure you must add to absorb the merge.

- **Within-seq (cluster_call_ level):** keeping two unconditionally-co-occurring callees separate means each gets its own gate/eval at every caller — duplicate gating overhead, duplicate dispatch, duplicate cost-accounting. The residue when you splice them is none (unconditional case) or a single shared gate (conditional case). Splicing fuses what would otherwise be N gates per caller into 1.

So your reading is exact: **forks = residual of partial across-seq merge; per-callee gating in callers = residual of refused within-seq splice.** Both are leftover separation that the merge/splice couldn't fully absorb.

**Why this framing is cleaner than similarity:**

1. Naturally handles partial-match cases. Similarity gives a 0.6 number that says nothing about whether to merge. Cost framing asks: "is the 60% overlap big enough to pay for itself given the 40% fork overhead?" — directly actionable.
2. Naturally handles the coincidence case. Two typs with identical content but disjoint callers still merit a clean merge (typ-level dedup saves storage). Two typs with identical content called from completely independent code paths still save the duplication — separation cost is content-duplication, not just context-duplication. Both motivations stack.
3. Aligns with compression: each merge/split/splice is justified iff savings > overhead. Same logic as data-level clustering in cogalg.

**Scale-free angle:**

This is the same criterion you apply at the data level — clusters form when keeping elements separate costs more than the structural overhead of grouping them. Promoting code clustering to the same criterion makes the two literally one operation: comp + cluster minimizes separation cost in both substrates.

**Practical eval form:**

```
sep_cost = overlap_size * unit_dup_cost       # what you pay by not merging
mrg_cost = structural_overhead + residue_cost # forks for merge, gate for splice
if sep_cost > mrg_cost: act  # merge / fork-merge / splice
```

Threshold disappears as a magic number — it's the cross-over point of two cost terms, both estimable from the same fc/fw accounting you already track.

**Question for the rewrite:**
Want merge() and cluster_call_() restructured to compute `sep_cost` and `mrg_cost` explicitly and act on the comparison, rather than the current `f.fc / C > ave` form? It'd unify the two operations under one eval and make the partial-merge / split / splice / fork-merge cases distinguishable by the same arithmetic.

boris-kz: So we need to combine cluster_call_ and merge, I haven't done that yet

kwcckw: I see, so we have 3 conditions when separation cost is high:
 
    merge / fork-merge / splice
    
For merge, it's combining both oFs, for fork-merge, it merging certain calls, but for splice, that's similar with fork-merge where we concatenate both calls?

Also when `if sep_cost > mrg_cost: act  # merge / fork-merge / splice` is false, the mrg_cost is high, we just need to leave as it is? Since the split and merge is actually in the same workflow now, we would need to split them first before merging those calls.

boris-kz: Splice is the same as merge, split is currently cluster_cal_. It's cost of forking in merge and cost gating (gF) in co-occurrence eval.

PR #143, opened 2026-05-13
5/13/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -28,25 +28,39 @@ def merge(F,f):  # combine aligned ops, if-fork per miss, no inline recursion, f
...
-        if isinstance(C, CoF):
-            if _C and comp_call_(_C, C) > ave:  # comp call history: C callers?
-                seg_ += [sum2F(seg)]; seg=[C]
-            else: seg += [C]
-            _C = C
-    if seg: seg_ += [sum2F(seg)]  # last seg
+        return m, d, M, D
+
+    seg_ = []; _C = call_[0]
+    for C in call_[1:]:  # AST?
```
kwcckw, line 48: I added a draft based on the prior conversation. 

I'm not quite sure on the costs below, could you check and correct it if needed?

frame_2D_alg/agg_recursion.py
```diff
@@ -142,6 +142,7 @@ class CoF(CF):
     def __init__(f, fo=1, **kw):
         super().__init__(**kw)
         f.call_ = kw.get('call_',[])  # top-level AST items
+        f.caller_ = kw.get('caller_',[])  # roots that call this oF
```
boris-kz, line 145: this is f.root

frame_2D_alg/agg_recursion.py
```diff
@@ -160,6 +161,8 @@ def inner(*a, **kw):
             if oF.call_:
                 for i, F in zip((1,0),(oF, oF.gF)):
                     tree = flat_(oF); L=len(tree)-1
+                    if i:
+                        for call in tree: call.caller_ += [oF]   # add caller for flatten calls? Else actually root is already sufficient
```
boris-kz, line 165: def add_typ_(oF):  # record oF vals for weighting, mapped to global FTT_

    for i, F_ in enumerate(typ_):
        if F_:

frame_2D_alg/meta_code.py
```diff
@@ -28,25 +28,39 @@ def merge(F,f):  # combine aligned ops, if-fork per miss, no inline recursion, f
         return f  # caller pops merged f from Z.typ_
 
 # draft, need to fold-in merge()
-def cluster_call_(rF, root=None):  # cluster rF callees if called toghether?
+def cluster_call_(call_, root=None):  # cluster rF callees if called toghether?
```
boris-kz, line 31: you are passing typ_, not call_

frame_2D_alg/meta_code.py
```diff
@@ -28,25 +28,39 @@ def merge(F,f):  # combine aligned ops, if-fork per miss, no inline recursion, f
...
 
     def comp_call_(_C, C):  # draft, need to include identity - cost of forking?
 
-        _caller_, caller_ = set(_C.root), set(C.root)
+        # internal
+        m = sum(1 for _c, c in zip(_C.call_, C.call_) if _c.nF == c.nF)
+        d = max(len(_C.call_), len(C.call_)) - m
```
boris-kz, line 37: min(len(_C.call_), len(C.call_))

kwcckw: In the code above, m is the number of matching nF, while d is the non-matching nFs.

Why it is the min of both calls' size here? `min(len(C.call), len(_C.call_))`

frame_2D_alg/meta_code.py
kwcckw: I'm parsing typ_ in the first call, but it maybe used in the fork-merge process below, using call_:

            elif _C.call_ and C.call_:  # fork-merge (flatten calls, remerge them again)
                seg_ += cluster_call_(_C.call_ + C.call_)
                
So it should regroup the flatten calls into typs again.

frame_2D_alg/agg_recursion.py
kwcckw: Thanks, this is probably from my prior pull request, i didn't update agg_recursion in the latest commit, not sure why it's here.

frame_2D_alg/agg_recursion.py
kwcckw: Sorry, this is not relevant, it's added from prior edits and it should be removed.

frame_2D_alg/meta_code.py
boris-kz: We only need forks while there can be matches, which is min len

frame_2D_alg/meta_code.py
boris-kz: It's two different loops, between call_s and within a call_, that distinction is still good

frame_2D_alg/meta_code.py
kwcckw: Okay right, the usage of zip truncates the longer call_, thanks.

frame_2D_alg/meta_code.py
kwcckw: Okay, so we need two different sections, one is eval between Z.typ_, and another is within each typ.call_?

For the sequence, it should be between call_s first or within a call_ first? I think the sequence matters since we may merge calls and the resulting merged call.call_ may have different size.

boris-kz: yes, so between call_s we compare aligned calls, and within a call_ it's consecutive calls.
these are guided by interupted vs. contiguous proximity.
I think external similarity (co-occurence) is more important for the first kind.

boris-kz: so between call_s we compare aligned calls, and within a call_ it's consecutive calls.
these are guided by interupted vs. contiguous proximity, respectively.

different aligned calls can be combined by adding forks, or just insert if they always have common caller?
identical consecutive calls can be combined in recursive expression.
different consecutive calls can be combined in common-caller sequences?

boris-kz: <html><head></head><body><p>clean 2×2 decomposition. Confirming structure, then three clarifications.</p>
<p><strong>The matrix:</strong></p>

  | identical content | different content
-- | -- | --
aligned (across-seq, interrupted) | merge (no residue) | fork  or  caller-conditional insert
consecutive (within-seq, contiguous) | loop / repeat-N CoF | splice into common-caller sub-sequence


<p>Two axes, four primitive actions, each justified by its own separation-cost arithmetic. Probably the right operation set.</p>
<p><strong>Three things to nail down:</strong></p>
<p><strong>1. "Insert if always common caller" — disjoint or shared?</strong></p>
<p>Two readings, opposite implications:</p>
<ul>
<li><em>Disjoint callers:</em> variation at position k is fully determined by caller identity → no runtime gate needed, just store both alternatives and let caller-dispatch pick. Caller-conditional, no fork cost.</li>
<li><em>Shared callers:</em> same caller can produce either variant → variation is data-conditional, needs a runtime gate (fork).</li>
</ul>
<p>The "no fork" outcome requires disjoint, not shared. Which did you mean? I read it as disjoint but the phrasing fits either.</p>
<p><strong>2. "Recursive expression" for identical consecutive — terminology check.</strong></p>
<p>If N copies of the same nF appear in a row, the natural compression is <code>(CoF, repeat=N)</code> — iteration, one CoF with a count. Genuine recursion (self-call) is a different structure (call_ contains a reference to the parent typ). Which:</p>
<ul>
<li>(a) repeat-count attribute on the CoF</li>
<li>(b) self-referential call_ entry</li>
<li>(c) both, depending on whether the repeats are static or data-dependent</li>
</ul>
<p>(a) is the cheap/common case. (b) only when the repeat count is dynamic. (c) probably the full answer.</p>
<p><strong>3. "Common-caller sequences" for different consecutive — granularity.</strong></p>
<p>Splice produces an extracted sub-typ. Question is who owns it:</p>
<ul>
<li><em>Multi-caller reuse:</em> the sub-sequence appears verbatim across multiple callers → extract once, replace each occurrence with a call to the extracted typ. Maximum compression, classic extract-method.</li>
<li><em>Per-caller composite:</em> the sub-sequence is private to one caller → splice in-place as a sub-CoF inside that caller's call_, no reuse.</li>
</ul>
<p>Multi-caller reuse is what "common-caller" naturally suggests if "common" means "shared across callers". Per-caller composite would be cheaper but compresses less. Which level is intended?</p>
<p><strong>One additional cell I'd add:</strong></p>
<p>You wrote "different aligned ... add forks OR insert if common caller". The identical-aligned case (top-left) doesn't need a fork at all — clean merge, zero residue. That's the easy case currently subsumed in merge(). Worth keeping explicit because the eval shape differs: identical-aligned has <code>residue_cost = 0</code>, making the merge essentially free relative to separation cost. Different-aligned pays for the fork; the trade-off matters.</p>
<p>If these three clarifications resolve cleanly, I think the design is ready for code. Want to settle them first or sketch the unified eval primitive that all four cells will draw from?</p></body></html>

kwcckw: > different aligned calls can be combined by adding forks, or just insert if they always have common caller?

I see, so this is between calls right? By adding forks, it's creating a new typ?

> identical consecutive calls can be combined in recursive expression.

So this is similar with our `_C` and `C` segmentation? Where consecutive same calls could be merged? And this should be within call?

> different consecutive calls can be combined in common-caller sequences?

I'm not sure if this is correct since many consecutive calls may share the same callers (like a callee in a func shared a same caller)? So we may need to add new eval other than the common caller?

boris-kz: no, we use forks in the same function all the time

identical consecutive calls is just adding call count instead of the calls.

>many consecutive calls may share the same callers

it depends on how long that sub-sequence is, and the callers may not be exactly the same 
so this clustering can be pretty complex

boris-kz: Right — splice → sub-oF is the messy case. Several independent dimensions:

**1. Length** — compression gain scales roughly linearly with sub-sequence length, but sub-oF overhead (construction, lookup, gating) is mostly fixed. So there's a minimum length threshold below which splice loses. Short runs (2-3 elements) usually aren't worth it.

**2. Caller-coverage degree** — full overlap (every caller runs the whole sequence) is the clean case. Partial overlap (caller A runs steps 1-5, B runs 2-4, C runs 1-3) is the common case. The sub-oF then needs internal gating per step.

**3. Position consistency** — does the sub-sequence always appear at the same position in caller bodies, or floats? Floating sub-sequences are harder to detect and harder to extract cleanly because the surrounding context isn't consistent.

**4. Variant within the sub-sequence** — even when the sequence appears in multiple callers, individual steps may differ across callers (caller A has step 3 = X, caller B has step 3 = Y). Handled by the same gate mechanism as merge's variant case, applied to a sub-oF step.

**How gF handles partial coverage:**

The extracted sub-oF carries its own gF tree. Per-step gates in gF record "step k fires if caller ∈ set S_k". Callers that run only a subset of steps pay the gate-eval cost for the skipped ones (cheap if caller-conditional). The sub-oF's `cost = base_cost + sum(per-step gate costs)`.

This matches the existing pattern — sub-oF is just another oF with its own gF — and reuses the same machinery as merge variants. Different axis, same primitive.

**Cost-of-separation view of splice:**

```
sep_cost = sum over callers c of (length covered by c × per-step cost without splice)
         + duplicate-storage cost across callers
mrg_cost = sub-oF construction + sum over steps k of gate_cost(S_k)
         + per-caller call-site replacement cost
extract if sep_cost > mrg_cost
```

Length appears as a multiplier on sep_cost; partial coverage shrinks sep_cost (less to dedupe) and grows mrg_cost (more per-step gates). The trade-off resolves naturally.

**Staging suggestion:**

Phase 1: only extract when length ≥ N and coverage is near-full. Simple sub-oF, no per-step gates. Cheap to implement, gives immediate compression on the easy cases.

Phase 2: enable partial coverage with per-step gating. Reuses merge's gate machinery.

Phase 3: floating-position detection (motif discovery). Hardest, deferred.

Phase 1 alone subsumes the original cluster_call_ draft and the splice notion for unconditional co-occurrence. Phases 2-3 are where the complexity Boris is flagging actually lives — and probably not worth coding until phase 1 produces enough sub-oFs to make the partial cases visible empirically.

**Question:** want me to draft phase 1 (the cleaned-up cluster_call_ + splice + add_typ_ + __main__ fixes), with phases 2-3 as docstring-only stubs? That gets a working pipeline today and leaves the complexity for incremental work.

kwcckw: > no, we use forks in the same function all the time
> 
> identical consecutive calls is just adding call count instead of the calls.
> 
> > many consecutive calls may share the same callers
> 
> it depends on how long that sub-sequence is, and the callers may not be exactly the same so this clustering can be pretty complex

Okay, for the call count, we don't have that now? So we need to add new params per call to indicate the caller count?

>  How gF handles partial coverage:

This partial coverage is quite complicated since we may have many possible of combinations here. Do we really need this partial coverage?

boris-kz: It's a general problem, clustering is never clean. Whenever we have forks in call_, each fork is potential segmentation point, that's how we design functions manually too. 

caller count is actually len(typ.root), not sure that matters

kwcckw: > It's a general problem, clustering is never clean. Whenever we have forks in call_, each fork is potential segmentation point, that's how we design functions manually too.

Yes, but this may get lengthy when we have a long sequence. For example, if we have 4 calls, from 1 to 4, we need to check multiple sequences, something like 12,123,1234,23,234,34, and this may get really long if we have more calls. This probably can be done recursively to match all the possible combinations.

> identical consecutive calls is just adding call count instead of the calls.
> caller count is actually len(typ.root), not sure that matters

But if identical consecutive count is adding call count, we need to add the count? Unless typ.root is not unique so preserve duplicated calls' count.

boris-kz: just think about how we currently decide to add forks in the same comp function vs splitting it into separate functions. we just need to automate this process.

never mind those identical calls, it's not relevant

frame_2D_alg/meta_code.py
boris-kz: so it seems that merge is still relevant for combining aligned call_s, cluster_call_ is for single call_ only.
we just need to add offset:

def merge(F,f):  # combine aligned ops, if-fork per miss, no inline recursion, f can only be added as a whole

    call_, add_ = [], []  # replace F.call_ if eval
    for i, (Sub,sub) in enumerate(zip(F.call_,f.call_)):  # call_: op sequence in func or block, init F per f?
        fork = []
            fork = Sub; fin=0
            if sub.nF=='E':
                sub = merge(Sub, sub)  # add2F(Sub,ssub_)?
                if sub: C+=sub.fc; call_+=[sub]; add_+=[sub]  # merge failed
            for _sub in Sub.call_:
            if not fin:
        elif Sub.nF != sub.nF:
            fork = CoF(nF='E',call_=[Sub,sub]); C += cost; add_+=[sub]
        call_ += [fork or Sub]
    call_ += F.call_[i+1:] if len(F.call_)>len(f.call_) else f.call_[i+1:]  # offset
    
    if f.fc / (C or eps) > ave:  # high individual_cost / forking_Cost, else keep old F,f
        F.call_= call_
        return f  # caller pops merged f from Z.typ_

frame_2D_alg/meta_code.py
kwcckw: > just think about how we currently decide to add forks in the same comp function vs splitting it into separate functions. we just need to automate this process.

I see, this is helpful. So we usually convert the forks into func if it's too long or we need to run it in multiple sections. I guess we can include this similar rules , so we must check all the combinations here to identify if there's multiple duplicated sequences.

frame_2D_alg/meta_code.py
kwcckw: This is packing the additional longer calls in call_ right? If sub is longer, i think we need to add those additional calls into add_ to allow add2F adds their params?

frame_2D_alg/meta_code.py
boris-kz: it's a cost of forking vs. the cost of adding new function. 
but the problem is how to automatically define such new function in the first place

frame_2D_alg/meta_code.py
boris-kz: ok, right:

    off = F.call_[i+1:] if len(F.call_)>len(f.call_) else f.call_[i+1:]
    call_ += off; add_ += off

frame_2D_alg/meta_code.py
kwcckw: > but the problem is how to automatically define such new function in the first place

We can treat each function is a sequence, sot that this line in oF:

   out = func(*a, **kw)

Should be running a series of functions in a sequence?

frame_2D_alg/meta_code.py
kwcckw: I think add_ is only relevant if f.call_ is longer than F.call_ since we add `add_` into F, so:

      off = F.call_[i+1:] if len(F.call_)>len(f.call_) else f.call_[i+1:]
      call_ += off; 
      if len(f.call_)>len(F.call_): add_ += off

boris-kz: right

frame_2D_alg/meta_code.py
```diff
@@ -28,25 +28,39 @@ def merge(F,f):  # combine aligned ops, if-fork per miss, no inline recursion, f
         return f  # caller pops merged f from Z.typ_
 
 # draft, need to fold-in merge()
-def cluster_call_(rF, root=None):  # cluster rF callees if called toghether?
+def cluster_call_(call_, root=None):  # cluster rF callees if called toghether?
 
     def comp_call_(_C, C):  # draft, need to include identity - cost of forking?
```
boris-kz, line 33: something like this:

def cluster_call_(Z, root=None):  # cluster rF callees if called toghether?

    def comp_callers(_T, T):

        caller_, _caller_ = set(T.root), set(_T.root)  # always typs?
        olp = _caller_ & caller_
        off = list(_caller_-olp) + list(caller_-olp)
        return M/D

    for T in Z.typ_: 
        seg_,seg,_C = [],[],[]
        for C in T.call_:  # AST?
            if isinstance(C,CoF):
                if not _C: continue  # init _C only
                _t,t = Z.typ_[_C.nF], Z.typ_[C.nF]
                if comp_callers(_t, t) > ave:  # cost of forking < cost of adding new function?
                    seg_ += [sum2F(seg)]; seg=[C]
                else: seg += [C]
                _C = C
        if seg: seg_ += [sum2F(seg)]  # last seg
        return seg_  # may be empty

frame_2D_alg/meta_code.py
boris-kz: I meant new functions, not in current onF_

frame_2D_alg/meta_code.py
boris-kz: sorry, I haven't gone through that yet.

frame_2D_alg/meta_code.py
kwcckw: Okay, right now each onF maps to a function, i think we need to init it as a sequence (we should have only a single function sequence in the first iteration), so that the subsequent merge or split functions will have the same structure.

frame_2D_alg/meta_code.py
kwcckw: So cluster_call_ only requires the comparison between calls' callers but not the calls?

boris-kz: right, because consecutive calls won't be the same anyway, it' just that the callee may be recursive

frame_2D_alg/meta_code.py
boris-kz: the question is how to form and evaluate that new sequence

kwcckw: Probably a new class seq? Then oF will call sequence instance to run all of their sequence functions.

PR #144, opened 2026-05-14
5/14/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -32,43 +32,38 @@ def merge(F,f):  # combine aligned ops, if-fork per miss, no inline recursion, f
...
 
 # draft
-def cluster_call_(Z, root=None):  # cluster T callees if called toghether?
+def cluster_call_(Z, root=None):  # cluster T callees if called together?
 
-    def comp_callers(_T, T):
+    def comp_callers(_C, C):
 
+        if _C.nF == C.nF: _T,T = Z.typ_[_C.root.nF], Z.typ_[C.root.nF]  # if same typ, get their caller's typ?
+        else:             _T,T = Z.typ_[_C.nF], Z.typ_[C.nF]  # caller_ = typ.root
```
kwcckw, line 40: The existing code gets typs from the calls of the same T:

     _t,t = Z.typ_[_C.nF], Z.typ_[C.nF] 

So both `_t` and `t` are always the same, and we need to get their root's typ if both calls' nF are the same?

frame_2D_alg/meta_code.py
```diff
@@ -32,43 +32,38 @@ def merge(F,f):  # combine aligned ops, if-fork per miss, no inline recursion, f
...
-                        merged = merge(_C, C)
-                        if merged: seg_ += [_C]
-                    elif _C.call_ and C.call_:  # fork-merge (flatten calls, remerge them again)
-                        seg_ += cluster_call_(_C.call_ + C.call_)            
-                '''
-        if seg: seg_ += [sum2F(seg)]  # last seg
-        return seg_  # may be empty
+                if isinstance(_C, CoF): # init _C only
+                    # we should terminate only if < ave?
+                    if comp_callers(_C,C) < ave:  # cost of forking < cost of adding new function?
```
kwcckw, line 62: We should terminate seg here only if the comp < ave? Else it should continue to splice the calls?

frame_2D_alg/meta_code.py
boris-kz: this is for consecutive calls, they can't have the same nF?

frame_2D_alg/meta_code.py
```diff
@@ -32,43 +32,38 @@ def merge(F,f):  # combine aligned ops, if-fork per miss, no inline recursion, f
...
+        seg_ += [new_typ]
+
+    seg_ = []
     for T in Z.typ_:
-        seg_, seg, _C = [],[],[]
+        if not T: continue
+        seg, _C = [],[]
         for C in T.call_:  # AST?
             if isinstance(C,CoF):
-                if not _C: continue  # init _C only
```
boris-kz, line 50: we are finding an overlap between callers, through the typ of a call.
that's why I have comp_callers(_t,t)

frame_2D_alg/meta_code.py
```diff
@@ -80,9 +75,7 @@ def comp_callers(_T, T):
     frame_H(image=imread('./images/toucan.jpg'), iY=Y//2-31, iX=X//2-31, Ly=64,Lx=64, Y=Y,X=X, rV=1)
     add_typ_(Z)  # each call_ in Z.typ_ is the flatten calls of same typ
     # add fffeedback to reform Z for next frame_H:
-    spl_ = []
-    for F in Z.typ_: spl_ += cluster_call_(F)  # add eval?
-    Z.typ_ = spl_
+    Z.typ_ = cluster_call_(Z)
```
boris-kz, line 78: thanks

frame_2D_alg/meta_code.py
```diff
@@ -32,43 +32,38 @@ def merge(F,f):  # combine aligned ops, if-fork per miss, no inline recursion, f
...
-                if sep_cost > mrg_cost: # merge / fork-merge
-                    if mi > me:  # merge or splice
-                        merged = merge(_C, C)
-                        if merged: seg_ += [_C]
-                    elif _C.call_ and C.call_:  # fork-merge (flatten calls, remerge them again)
-                        seg_ += cluster_call_(_C.call_ + C.call_)            
-                '''
-        if seg: seg_ += [sum2F(seg)]  # last seg
-        return seg_  # may be empty
+                if isinstance(_C, CoF): # init _C only
```
boris-kz, line 60: _C is always CoF in my scheme?

frame_2D_alg/meta_code.py
boris-kz: my comp_callers was supposed to return costs, but it's actually a gain

frame_2D_alg/meta_code.py
kwcckw: But cluster_call_ is per typ? So C.root (those calls' typ) will be the same.

frame_2D_alg/meta_code.py
kwcckw: Okay, for cost, it will be the other way, which is D/M?

frame_2D_alg/meta_code.py
kwcckw: Those calls are in typ.call_, which have the same nF? Because right now `cluster_call_` is per typ's calls?

frame_2D_alg/meta_code.py
kwcckw: Yes, actually this is redundant and it should be removed.

frame_2D_alg/meta_code.py
```diff
@@ -32,43 +32,38 @@ def merge(F,f):  # combine aligned ops, if-fork per miss, no inline recursion, f
     else: return F
 
 # draft
-def cluster_call_(Z, root=None):  # cluster T callees if called toghether?
+def cluster_call_(Z, root=None):  # cluster T callees if called together?
 
-    def comp_callers(_T, T):
+    def comp_callers(_C, C):
```
boris-kz, line 37: we are finding an overlap between callers, through the typ of a call

frame_2D_alg/meta_code.py
boris-kz: if not _C: _C = C; seg = [C]; continue

frame_2D_alg/meta_code.py
boris-kz: doesn't matter, we can't have the same oF called twice in a row

frame_2D_alg/meta_code.py
boris-kz: > But cluster_call_ is per typ? So C.root (those calls' typ) will be the same.

this is supposed to be mapping T.root, which is a list

frame_2D_alg/meta_code.py
kwcckw: The problem is we are segmenting T.call_, which all have the same nF:

          for T in Z.typ_:
              seg_, seg, _C = [],[],[]
              for C in T.call_:  # AST?
                  if isinstance(C,CoF):
                      if not _C: continue  # init _C only
                      _t,t = Z.typ_[_C.nF], Z.typ_[C.nF]  # caller_ = typ.root
                      if comp_callers(_t,t) > ave:  # cost of forking < cost of adding new function?
                          seg_ += [sum2F(seg)]; seg=[C]
                      else: seg += [C]
                      
Their nF will be different only if `_C` is the last call in prior typ, and `C` is the first call in the new typ, is this intended?

boris-kz: calls can't have the same nF, they are different functions
typ is one of their callers, nF has nothing to do with that

frame_2D_alg/meta_code.py
boris-kz: it doesn't really matter at this point

frame_2D_alg/meta_code.py
```diff
@@ -32,43 +32,38 @@ def merge(F,f):  # combine aligned ops, if-fork per miss, no inline recursion, f
...
+        new_typ = sum2F(seg); new_typ.nF = [call.nF for call in new_typ.call_]  # new nF is a list packing multiple nFs
+        seg_ += [new_typ]
+
+    seg_ = []
     for T in Z.typ_:
-        seg_, seg, _C = [],[],[]
+        if not T: continue
+        seg, _C = [],[]
         for C in T.call_:  # AST?
             if isinstance(C,CoF):
```
boris-kz, line 59: Non-oFs don't have caller records, so they should be included in seg by default?

Claude:
Then the inner loop becomes:

```python
for C in T.call_:
    if isinstance(C, CoF):
        if not _C: _C = C; seg += [C]; continue        # first CoF anchors seg
        _t, t = Z.typ_[_C.nF], Z.typ_[C.nF]
        if comp_callers(_t, t) > ave: seg += [C]       # shared callers → same seg
        else: seg_ += [sum2F(seg)]; seg = [C]          # break
        _C = C
    else:
        seg += [C]                                     # non-oF: ride along
```

Two notes on this:

- Using `seg += [C]` (not `seg = [C]`) in the init branch lets any leading non-CoFs that already went into `seg` via the `else` stay with the first segment instead of being discarded.
- Non-CoFs between two CoFs get attributed to whichever segment is current at that moment. If `_C` and `C` end up split into different segs, the intervening non-CoFs stay with the *earlier* seg (because they were appended before the split decision fires on the next CoF). That seems right — they were called in that context.

Combined with the earlier two fixes (polarity + return placement), the full revision:

```python
def cluster_call_(Z, root=None):

    def comp_callers(_T, T):
        caller_, _caller_ = set(T.root), set(_T.root)
        olp = _caller_ & caller_
        off = list(_caller_-olp) + list(caller_-olp)
        return M/(D or eps)

    out = []
    for T in Z.typ_:
        if not T: continue
        seg_, seg, _C = [],[],[]
        for C in T.call_:
            if isinstance(C, CoF):
                if not _C: _C = C; seg += [C]; continue
                _t, t = Z.typ_[_C.nF], Z.typ_[C.nF]
                if comp_callers(_t, t) > ave: seg += [C]
                else: seg_ += [sum2F(seg)]; seg = [C]
                _C = C
            else:
                seg += [C]
        if seg: seg_ += [sum2F(seg)]
        out += seg_
    return out
```

Still open: does `T.root` actually carry the caller typs at this stage? `add_typ_` would need to populate it. Want to check what it produces, or is that already wired?

frame_2D_alg/meta_code.py
kwcckw: We group calls into the same T in add_typ_:

          for i, F_ in enumerate(typ_):
              if F_:
          oF.typ_ = typ_
          
Specifically this line:

So all F in F_ has the same nF, and that's in T.call_, added from sum2F in add_typ_.

The code below is per T:

    for C in T.call_:
    
That T.call_ are actually `F_` from `add_typ_`, which all of them have the same typ as T.

The calls should be different only if they are not retrieved from typ oF (T) or we are running the func with the combined calls from multiple T.call_, such as Z.call_.

frame_2D_alg/meta_code.py
boris-kz: def cluster_call_(Z):  # cluster T callees if called together?

    def comp_callers(_T, T):
        caller_, _caller_ = set(T.root), set(_T.root)
        olp = _caller_ & caller_
        off = list(_caller_-olp) + list(caller_-olp)
        return M/(D or eps)
    T_= []
    for T in Z.typ_:
        if not T: continue
        seg_, seg, _C = [],[],[]
        for C in T.call_:
            if isinstance(C, CoF):
                if not _C: _C = C; seg += [C]; continue
                _t, t = Z.typ_[_C.nF], Z.typ_[C.nF]
                if comp_callers(_t,t)> ave: seg += [C]
                else: seg_ += [sum2F(seg)]; seg = [C]
                _C = C
            else: seg += [C]
        if seg: seg_ += [sum2F(seg)]  # last seg, add eval?
        if seg_: T_ += seg_  # segmented T, add eval?
        else:    T_ += [T]   # recycled T
    Z.typ_ = T_

frame_2D_alg/meta_code.py
boris-kz: ok, so the problem is sum2F(F_,CoF()), it converts F_ to T.call_?
that F_ should be N_ instead?

boris-kz: Agree with the diagnosis. `sum2F(F_, CoF())` hits the `elif fO: F.call_ = n_` branch (line 669), so T.call_ becomes F_ — the same-nF instances, not their callees. That's why cluster_call_ sees identical nFs across `T.call_`.

Conceptual fix: F_ are *instances of the same function*, so they're a cluster → T.N_ = F_. The heterogeneous callee sequence lives one level down, inside each F.call_.

But then T.call_ has to come from somewhere for merge()/cluster_call_ to operate on. Two options:

1. Align-and-sum: T.call_[k] = sum2F([F.call_[k] for F in F_]) — position-wise merge across F_, same as what merge() does pairwise. Gives a proper "averaged" callee sequence with real varying nFs.

2. Representative: T.call_ = F_[0].call_ (or any one — all same function, same structure). Cheaper, no alignment cost, but loses the per-position aggregation.

(1) seems right since downstream merge() expects aggregated stats per call position.

Either way, sum2F's fO branch needs distinguishing: "sum N CoF instances of same function" (→ N_=n_, call_=aligned-merge) vs the current "build a CoF from a list of callees" (→ call_=n_).

Question: do you want me to change sum2F to branch on this, or just fix it locally inside add_typ_ (swap T.N_/T.call_ after the sum2F call)?

boris-kz: Checked all sum2F call sites that hit the fO branch (input is CoF list):

- **line 188** `sum2F(F_,CoF())` — same-nF instances → the bug we're discussing. Wants N_, not call_.
- **line 164** `sum2F(tree, oF)` — tree of mixed-nF callees, but result is fed straight to `add2F(root, F, 2)` (stats-only merge into oF). The call_ of the temp F is never read.
- **line 191** `sum2F([t for t in typ_ if t], CoF())` — same: temp result consumed by add2F for stats. call_ incidental.
- **meta_code line 29** `add2F(F, sum2F(add_), fo=1)` — same pattern, fo=1 appends to typ_, only stats used.
- **meta_code lines 54/57** `sum2F(seg)` — seg is a consecutive sub-sequence of an existing call_; the resulting CoF represents a "compound callee" wrapping seg as its body. *Here* call_=seg is semantically correct.

So out of 5 fO-branch call sites: 1 legitimate use (seg), 1 bug (188), 3 don't care.

Cleanest: flip sum2F default to `F.N_ = n_` in the fO branch, add a `fcall=0` kwarg for the seg case. Or just patch 188 locally and leave sum2F alone.

Lean toward the local patch — only one real call site benefits from the rename, and the seg-case "call_=segment-as-sub-function-body" reading is fine to keep as the default semantics.

kwcckw: > that F_ should be N_ instead?

But typ oF. call_ supposes to pack same nF's oF only, we will never get any callees with different nF here?

This line in add_typ_ is actually flatten Z's nested calls:

    for F in flat_(oF): typ_[F.nF] += [F]
    
So N_ won't solve the problem here, unless we are clustering across multiple Ts with different callees.

boris-kz: it's a call_, AST, it can't be a sequence of the same oFs, that doesn't make any sense

kwcckw: > it's a call_, AST, it can't be a sequence of the same oFs, that doesn't make any sense

Yes, for those typ oF, previously we packed those same type's calls in oF.call_, which should be in oF.N_ now?

To get different calls per typ oF, we only can get it via those same typ's oF's root now?

boris-kz: same nF → same function body → same call_ structure. So:

```python
T = sum2F(F_,CoF()); T.nF=i; T.wTT=cent_TT(getattr(T,'rTT',T.dTT),T.r)
T.N_ = T.call_; T.call_ = F_[0].call_  # instances → N_, representative callees → call_
typ_[i]=T
```

kwcckw: I see, and this line:

    T.call_ = F_[0].call_

Should be getting all F_'s call_ instead?

boris-kz: read above, they all have the same call_

frame_2D_alg/meta_code.py
boris-kz: Two things:

**1. The "recycled T" branch almost never fires.** As written, `seg_` ends up empty only when `T.call_` is empty. Any non-empty `T` produces at least one `sum2F(seg)` from the last-seg block, so `T_ += seg_` always wins. That means even a T with zero actual breaks gets rebuilt as a fresh `sum2F` wrapper around its own contents — wasteful, and the new CoF's identity is fresh, breaking any external `.root` references pointing at the original T.

Guard on real segmentation:

```python
if len(seg_) > 1: T_ += seg_    # actual split
else:             T_ += [T]     # no break ever fired, keep original
```

**2. The two `# add eval?` points.** Both are gating questions about whether a candidate seg is *worth* making into its own typ. The natural criterion is the same one in `merge()`: segment cost vs forking cost. A 1-call seg is almost never worth its own typ. Sketch:

```python
# at break:
if sum(c.fc for c in seg) > ave * (Ec_[0] or eps):
    seg_ += [sum2F(seg)]
seg = [C]
# at end:
if seg and sum(c.fc for c in seg) > ave * (Ec_[0] or eps):
    seg_ += [sum2F(seg)]
```

But the exact form depends on what `Ec_[0]` represents at this scope — same fork cost as merge, or a different constant for split? merge's eval is `f.fc / C > ave` (individual / forking cost). Splitting flips the question: is `seg.fc` large enough to justify the new fork's cost? If yes, split; if no, the seg gets folded into the next one (i.e. don't break).

If you want them symmetric to merge, the rejected-seg shouldn't be dropped on the floor — it should merge into the *next* seg rather than become its own. Right now the code on break does `seg_ += [...]; seg = [C]`, which commits the old seg regardless of whether splitting pays off. The eval should be: *should we break here at all?*, not *should we keep this seg?*. Reorganized:

```python
if comp_callers(_t,t) > ave: seg += [C]
elif sum(c.fc for c in seg) > ave * Ec_[0]:        # break only if seg worth keeping
    seg_ += [sum2F(seg)]; seg = [C]
else: seg += [C]                                    # too small to split off, absorb
_C = C
```

That makes the eval part of the split decision rather than a post-hoc filter.

frame_2D_alg/meta_code.py
kwcckw: > same nF → same function body → same call_ structure. So:

Actually this may not true all the time because call_ is added in runtime, and if some forks are not run, their calls will be differ?

boris-kz: ok, so we need to get typ.call_ from mapping function AST?

Claude — AST is the right canonical source. The runtime call_ is path-dependent: gates skip, loops repeat, so neither F_[0].call_ nor a union of F_.call_'s is reliable.

meta_code already has the ast machinery (`get_ops`, `init_wc`). Extending to extract the callee sequence per function is straightforward: walk the FunctionDef, collect `ast.Call` nodes in source order, map name → nF via `onF_.index(name)`.

One choice left: what's *in* T.call_ at each AST position?

- **(a) skeleton only**: list of nF indices (or stub CoFs with just .nF set). cluster_call_ works directly (only reads .nF). merge() needs .fc too, but that's just `Fc_[nF]` from globals.
- **(b) summed**: at AST position k, sum2F all F.call_ entries from F_ that actually ran position k. Gives per-position stats, but needs runtime→AST alignment which is fragile under conditionals/loops.

(a) is cleaner: T is the canonical *function*, its stats live in T itself; T.call_ is just the structural skeleton over nFs. Per-instance variance is already captured by T's summed dTT/c/r.

Go with (a)?

me: we need to convert all AST into CoFs, by the same segmentation process as in cluster_call_?
Yes — unified pipeline:

1. AST-extract: per nF, walk FunctionDef → linear sequence of ast.Call → mapped to callee nFs
2. Segment via cluster_call_ logic: compare consecutive callees by caller-overlap (Z.typ_[nF].root), wrap contiguous high-overlap spans as sub-CoFs
3. T.call_ = segmented tree

So cluster_call_ stops being a runtime-only operation; it's the canonical structurer of any callee sequence, AST or runtime. And since AST is the trusted source, runtime call_ becomes obsolete for this purpose — we still need F_ for summed stats in T itself, but T.call_ is built from AST alone.

One wrinkle: segmentation needs `Z.typ_[nF]` populated to look up caller_ for comp_callers. That's a chicken-and-egg if we're building Z.typ_ via add_typ_. Two-pass: first pass builds T's with stats but flat call_; second pass segments each T.call_ now that all typs exist with their .root sets.

Or: segment lazily at first cluster_call_ invocation, after add_typ_ finishes.

boris-kz: `onF_` is in agg_recursion (line 58), so the builder belongs there too — no circularity. Draft:

```python
import ast  # add to imports

def build_typ_skel(paths=("agg_recursion.py","comp_slice.py","slice_edge.py")):
    nodes = {}
    for p in paths:
        with open(p) as f: tree = ast.parse(f.read())
        for n in ast.walk(tree):
            if isinstance(n, ast.FunctionDef) and n.name in onF_: nodes[n.name]=n
    call_seq = [[] for _ in onF_]
    for name, fn in nodes.items():
        i = onF_.index(name)
        for n in ast.walk(fn):
            if isinstance(n, ast.Call):
                f=n.func; c = getattr(f,'id',None) or getattr(f,'attr',None)
                if c in onF_: call_seq[i] += [onF_.index(c)]
    callers = [[] for _ in onF_]
    for i, seq in enumerate(call_seq):
        for c in seq:
            if i not in callers[c]: callers[c] += [i]
    return call_seq, callers

TYP_SKEL = None  # built lazily
```

Then add_typ_ (diff vs current):

```python
def add_typ_(oF):
    global TYP_SKEL
    if TYP_SKEL is None: TYP_SKEL = build_typ_skel()
    call_seq, callers = TYP_SKEL
    for F in flat_(oF): typ_[F.nF] += [F]
    for i, F_ in enumerate(typ_):
        if F_:
            T = sum2F(F_,CoF()); T.nF=i; T.wTT=cent_TT(getattr(T,'rTT',T.dTT),T.r)
            T.N_ = T.call_                          # instances → N_
            typ_[i] = T
    for i, T in enumerate(typ_):                    # wire structure after all T's exist
        if T:
            T.call_ = [typ_[k] for k in call_seq[i] if typ_[k]]
            T.root  = [typ_[k] for k in callers[i]  if typ_[k]]
    oF.typ_ = typ_
    if any(typ_): add2F(oF, sum2F([t for t in typ_ if t], CoF()))
```

Notes / open questions:

- `ast.walk` is BFS, gives surface order (parent before child). For `g(h(x))` you get [g, h]. Switch to a recursive `lineno/col_offset` traversal if you want strict source order.
- Dedup in callers but not in call_seq — a function calling `comp_N` twice gets two entries (matches the AST). Intentional?
- T.call_ entries are *shared* references into typ_. cluster_call_'s `sum2F(seg)` reads only, fine. If anything later mutates a T in place expecting local ownership, this aliases.
- T.fc isn't propagated through sum2F's fO branch — should we set `T.fc = Fc_[i]` explicitly in add_typ_? merge() reads it.

frame_2D_alg/agg_recursion.py
```diff
@@ -1017,8 +1017,7 @@ def expand_lev(_iy, _ix, elev, T):  # seed tile is pixels in 1st lev, or Fg in h
         T_, PV__,C,R = [],np.zeros([Ly,Lx]),0,0  # tiles, maps to level frame
```
boris-kz, line 1017: move all class defs, etc, to meta_code, that's where they will be modified, it's the definition of "meta_code"? agg_recursion will be input-processing code only.

Yes, feasible. Clean partition:

**meta_code (type/primitive layer)**
- Globals: `ave, avd, decay, wY, wX, wYX, wM..wA, wT, wTT, onF_, Fc_/Fw_/FTT_ (+unpacks), Ec_/Ew_/ETT_, eps, Z`
- Helpers: `eps_`, `prop_F_`, `flat_`, `extend_box`, `sort_H`, `eval`, `mdecay`
- Classes: `CF, CL, CC, CN, CoF` (with trace/gate)
- Primitives: `vt_, sum_vt, cent_TT, sum2F, add2F, add_H, F2N, Copy_`
- Existing AST: `weights, get_ops, init_wc, get_wc`
- New: `build_typ_skel`, `add_typ_`, plus existing `merge`, `cluster_call_`

**agg_recursion (algorithm layer)** — top: `from meta_code import *` (or named)
- `cross_comp`, `comp_N_`, `comp_C_`, `comp_N`, `comp_F`, `base_comp`, `comp_derT`, `comp`, `comp_A`
- `get_exemplars`, `cluster_N`, `cluster_C`, `cluster_P`
- `sum2G`, `comb_Ft`, `add_Nt`, `add_Lt`
- `vect_edge`, `trace_edge`, `ffeedback`, `proj_N`, `frame_H`

Layering becomes: frame_blobs → slice_edge → comp_slice → meta_code → agg_recursion.

Gotchas to flag:

- `CF(CBase)` — meta_code imports `CBase` from frame_blobs. Fine.
- `@trace_func` (CoF.trace) decorates agg_recursion functions; that's applied at runtime in meta_code's `__main__` via `trace_func(vars(agg_recursion))` (already there). So decorator definition in meta_code, application crosses module — still one-way.
- `onF_` references function names defined in agg_recursion and other files. meta_code doesn't import them; resolution is by name string only. Fine.
- `vt_` is in onF_ but also called directly inside sum2F / sum_vt. Both move to meta_code together, so internal calls stay intra-module.
- Stray AST `__main__` block in meta_code (lines 60-77 *and* 144-152) — consolidate into one after the move.
- `from comp_slice import w_t` (line 9 of agg_recursion) — check if `w_t` is used outside agg_recursion functions; if used by anything moving to meta_code, that import moves too.

**### **generalize: all classes to meta_code, making it the bottom of the stack:****

```
meta_code (classes, primitives, AST)
  ↓
frame_blobs (uses CBase, defines image ops)
  ↓
slice_edge (uses CP)
  ↓
comp_slice (uses CdP)
  ↓
agg_recursion (algorithms)
```

Classes to migrate:
- frame_blobs: `CBase`, `CN`, `CBlob`, `CrNode_`
- slice_edge: `CP`
- comp_slice: `CdP`
- agg_recursion: `CF, CL, CC, CN, CoF`

Two gotchas:

1. **Name collision: `CN`** is defined in both frame_blobs (line 71) and agg_recursion (line 112). Once both live in meta_code, one must rename. Independent of this refactor but forced by it.

2. **Subclass-in-lower-module pattern**: frame_blobs defines `CrNode_(CN)` *within* itself — so it inherits from a class also defined there. If both move to meta_code, that's fine. But it suggests CBase + CN are the "framework" and CBlob/CrNode_/CP/CdP are "framework instantiations" — debatable whether the latter belong in meta_code or with their algorithm code. Drawing the line strictly: all class definitions → meta_code, even the ones tightly coupled to one algorithm.

kwcckw: > move all class defs, etc, to meta_code, that's where they will be modified, it's the definition of "meta_code"? agg_recursion will be input-processing code onl

But we are not really changing the code right? It's just the sequence or func forking maybe different?

frame_2D_alg/meta_code.py
kwcckw: >Non-oFs don't have caller records, so they should be included in seg by default?

All calls should have caller, except Z, the root oF, so i don't see this is an issue? We don't have any non-oFs anyway?

frame_2D_alg/meta_code.py
kwcckw: Okay, but i think that TYP_SKEL could be a local, since we only need it in add_typ_. 

Also for T.root, it should be just all T.call_'s root? I think that's actually simpler and clearer.

frame_2D_alg/agg_recursion.py
boris-kz: some consolidation: frame_blobs doesn't need CN, AST is only used in meta_code, etc.

frame_2D_alg/meta_code.py
boris-kz: yes, but add_typ_ should be in meta_code too

frame_2D_alg/meta_code.py
boris-kz: currently the call_ should be defined by AST, which has a lot non-oF objects.
so they should all be clustered into oFs

frame_2D_alg/agg_recursion.py
kwcckw: > yes, but add_typ_ should be in meta_code too

Yes, but do we really need to move all core funcs into meta_code? If it's due to the import problem, i think we can pack it in another file like agg_base_functions, then we can import it from meta_code or agg_recursion.

boris-kz: Input-processing functions stay in agg_recursion, the data flow is:

agg_recursion runs → @trace populates each CoF's dTT, fw, fc, m, d, c, r
add_typ_ aggregates per-nF → Z.typ_[i] with summed stats + AST skeleton
merge / cluster_call_ read those stats (comp_callers uses c.w * c.c; merge's accept condition uses f.fc / C) — decide structural changes
ffeedback writes results back to global Fw_/FTT_
next agg_recursion run uses updated filters/structure

So performance dependence is data-flow through CoF stats, not import-level. Decision logic in meta_code; validation in the next run. One-way import preserved.

ffeedback only uses `vt_, FTT_, np` and the H structure — no comp/cluster ops. It's filter adjustment, pure meta. Moves to meta_code.

`trace_func` (line 1075) similarly — it's the wiring that applies CoF.traced over a module dict. Belongs in meta_code next to the CoF class and decorator. The `__main__` block in agg_recursion (line 1084) then just does `trace_func(vars(agg_recursion))` after `from meta_code import trace_func`.

Net effect — meta_code ends up owning the entire meta loop: type definitions + primitives + AST + add_typ_ + merge + cluster_call_ + ffeedback + trace_func. 
agg_recursion is purely the algorithm functions that get traced.

kwcckw: > Net effect — meta_code ends up owning the entire meta loop: type definitions + primitives + AST + add_typ_ + merge + cluster_call_ + ffeedback + trace_func.
agg_recursion is purely the algorithm functions that get traced.

This should be almost the same as current structure, but what is the type definitions + primitives? Is it the onF_?

boris-kz: there should be no exposed primitives in call_, they should all be packed in onF_now, through the same cluster_call_

kwcckw: To clarify, the primitives are those basic single line ops such as +, -, / etc? So they will be packed into a func now?

boris-kz: yes, because we need to collect callers and statististics,
something like this:

boris-kz: def cluster_AST(Q):  # initial wrap all primitives with oFs for Z.call_

    seg_, C = [],[]
    for c in Q:
        if C:
            if isinstance(c, CoF): seg_ += [C]; C = c
            else: C.call_ += [c]
        elif isinstance(c, CoF): C=c
        else: C = CoF(call_=[c])
    if C: seg_ += [C]
    return seg_

boris-kz: Logic works, but two concerns:

**Mutation of input CoFs.** When `c=CoF_A` then primitives follow, `C.call_ += [c]` appends to `CoF_A.call_` directly. If CoF_A is `Z.typ_[k]` (a shared typ ref), you're polluting the typ with primitives that semantically belong to this call site, not to the typ itself. Two callers of cluster_AST would cross-contaminate.

Fix: copy on transition:
```python
if isinstance(c, CoF): seg_ += [C]; C = CoF(call_=list(c.call_), nF=c.nF)  # or use Copy_(c)
```

**Asymmetric structure.** Trace `[p1,p2,CoF_A,p3,p4]`:
- result: `[CoF(call_=[p1,p2]), CoF_A(call_=[...orig, p3,p4])]`

First primitive run is a fresh wrapper; later runs are glued onto the *preceding* typ-CoF. Downstream code reading seg_ can't tell from shape which elements are typs vs wrappers — both look like CoFs with call_.

If "primitives belong to the call they follow" is intentional (locality semantics), this is correct — but then the leading run before any CoF is an orphan, treated differently. Could prepend an implicit anchor CoF, or accept asymmetry as "code before first call is its own block".

Which reading? Locality (current) or standalone (CoFs unmolested, primitive runs as separate wrappers)?

kwcckw: > yes, because we need to collect callers and statististics, something like this:
> 
> def cluster_AST(Q): # initial wrap all primitives with oFs?
> 
> ```
> seg_, C = [],[]
> for c in Q:
>     if C:
>         if isinstance(c, CoF): seg_ += [C]
>         else: C.call_ += [c]
>     elif isinstance(c, CoF): C=c 
>     else: C = CoF(call_=[c])
> return seg_
> ```

This should be also static right? This Q should be the tree from ast.walk, packing all the nodes of a function?

kwcckw: For example, using cross_comp:

          N_,G_ = root.N_,[]; fC = N_[0].typ==2  # root is Ft, converted below, rc=rdn+olp, comp N_|B_|C_:
          # comp_ s update root oF:


This line `N_,G_ = root.N_,[]; fC = N_[0].typ==2  # root is Ft, converted below, rc=rdn+olp, comp N_|B_|C_:` should be packed separately as a CoF too? Or we need to limit that to meaningful ops only, such as the ops that only involves arithmetic?

boris-kz: Actually, Z.call_ is already all oFs, only typ.call_s can have primitives? So we may not need this wrapper

kwcckw: Yes, primitives should be per call, so without this wrapper, the smallest unit function are actually those functions that defined in onF_, which we may only combine them into different combinations later.

boris-kz: Z.call_ is already all oFs

PR #145, opened 2026-05-15
5/15/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -1049,13 +1038,17 @@ def expand_lev(_iy, _ix, elev, T):  # seed tile is pixels in 1st lev, or Fg in h
             F.H += [lev := sum2F(tile_)]  # include top lev, same vals as F?
             if cross_comp(lev, rr=elev)[0]:  # spec->tN_,tC_,tL_, proj comb N_'L_?
                 elev += 1
+                # this feedback should be in meta_code only now?
+                '''
                 if rV > ave:
                     add_typ_(Z)  # typ_ maps to Fw_,Fc_,FTT_
```
kwcckw, line 1044: So this entire section and ffeedback should be moved to meta_code now?

frame_2D_alg/meta_code.py
```diff
@@ -94,6 +134,28 @@ def cluster_AST(Q):  # initial wrap all primitives with oFs for Z.call_
...
         for f in Z.typ_[i+1:]: mrg_ += [merge(F,f)]
         typ_ += [F]
+        
+    
+    if not isinstance(frame, list):
+        rV,rTT = ffeedback(frame)
+        
+        # we need to remove or add FTT_, Fw_ based on the new Z.typ_ here?\
+        # then recompute Fc_ using AST.walk in runtime again
+        for i, tF in enumerate(Z.typ_):
```
kwcckw, line 144: So after the Z.typ_'s calls merging, and after the ffeedback, we need to add or remove Fw_ and FTT_ here. For Fc_, that requires us to recompute those new typ's AST weight again in runtime?

frame_2D_alg/agg_recursion.py
boris-kz: yes, because all these globals are meta to the active code

frame_2D_alg/meta_code.py
```diff
@@ -77,14 +79,52 @@ def cluster_AST(Q):  # initial wrap all primitives with oFs for Z.call_
     if C: seg_ += [C]
     return seg_
 
+
+
+def add_typ_(oF):  # record oF vals for weighting, mapped to global FTT_
```
boris-kz, line 84: I have that in 246

frame_2D_alg/meta_code.py
boris-kz: yes, but it's runtime on a new data, the results from the old data are not affected

frame_2D_alg/agg_recursion.py
```diff
@@ -179,17 +179,6 @@ def gate(gain,cost, dTT=None):  # gain = evaled_block_w - default_block_w
         return gain>T
```
boris-kz, line 179: classes should be in meta_code too, so it doesn't import anything from agg_recursion?

frame_2D_alg/agg_recursion.py
boris-kz: this code should stay here, only add_type and ffeedback defs are moved to meta_code

ffeedback's body is pure meta (wTT ratios across levels, FTT_ lookup) — no algorithm semantics. It belongs in meta_code on content grounds.

The fact it's *called from* frame_H doesn't conflict — frame_H stays in agg_recursion and just imports ffeedback. Same pattern as add_typ_, merge, cluster_call_.

So: ffeedback → meta_code. The line 1054 call becomes `rV,FTT_ = ffeedback(F)` where ffeedback comes from `from meta_code import ...`.

Caveat: ffeedback references `FTT_` globally (line 1072). Since FTT_ moves to meta_code too, it's resolved there at module-level — no parameter passing needed.

Anything that's reflective/aggregative/filter-tuning → meta_code. agg_recursion stays the algorithm body that gets reflected upon.

frame_2D_alg/meta_code.py
```diff
@@ -77,14 +79,52 @@ def cluster_AST(Q):  # initial wrap all primitives with oFs for Z.call_
...
+    _wTT = frame.H[1].wTT
+    for lev in frame.H[2:]:  # sum ratios between consecutive-level TTs, top-down frame expansion levels, not lev-selective or sub-lev recursive
+        rTT += np.divide(_wTT,lev.wTT)
+        _wTT = lev.wTT
+    rM = rD = 0
+    rm, rd = vt_(rTT,wTT);
+    return rM+rD, rTT
+
+
 if __name__ == "__main__":
```
boris-kz, line 120: so all that should be in agg_recursion

frame_2D_alg/agg_recursion.py
kwcckw: But we also need this rTT, rV from this section to update Fw_, Fc_ and FTT_?

frame_2D_alg/meta_code.py
kwcckw: The resulting Fw_, Fc_, FTT_ should be recomputed for new data? Since the Z.typ_ will be different after the merging process.

frame_2D_alg/meta_code.py
kwcckw: Yes, but your version involves AST, where it actually could be simpler ：

     T.call_ = list(set([call for F in F_ for call in F.call_]))
 
 We just need to get all non duplicated calls from F.call_, instead of getting it from AST, since your code also getting the runtime calls by using if typ_[k]:
 
     T.call_ = [typ_[k] for k in call_seq[i] if typ_[k]]

frame_2D_alg/agg_recursion.py
kwcckw: Yes, and those global aves, wTT, Fc_, Fw_ should be moved here too.

frame_2D_alg/agg_recursion.py
boris-kz: they can be imported from meta_code

frame_2D_alg/meta_code.py
boris-kz: it has to be AST to get Z.call_ first, it doesn't fall from the sky

frame_2D_alg/meta_code.py
boris-kz: yes

frame_2D_alg/meta_code.py
boris-kz: Z represents frame_H instance:

- Remove `'frame_H'` from `onF_` (line 58). Length drops 16→15.
- `Fc_, Fw_, FTT_, Ec_, Ew_, ETT_` and their unpacks (lines 63-69) all shrink by one — drop `cFrm/wFrm/ttFrm`. Any reference to those constants in algorithm code (e.g., line 1036-1037 `wTT=ttFrm`, `(ave+avd)*(R+cFrm*L)`) needs replacement, since the frame-level cost/weight now lives in Z directly (as Z.c/Z.w).
- Z.call_ = `cluster_AST(frame_H_ast_seq)` — built once in add_typ_ (or wherever Z's typ_ gets initialized) since frame_H's body is static.
- Z.typ_ remains, now sized 15.
- The @trace decorator should *not* wrap frame_H (it'd self-wrap Z). Either exclude it explicitly in trace_func, or rely on `'frame_H' not in onF_` after the removal — the current check `if name in onF_` then handles it automatically.

One conceptual loose end: runtime stats. AST gives Z.call_ as one entry per source call site; runtime traces produce many instances per site (especially inside the while loop). Aggregation per nF goes into Z.typ_, but per-source-position stats inside Z.call_ have nowhere to come from. Likely fine — Z.call_ is structural, Z.typ_ holds the numbers — but worth being explicit that Z.call_ entries don't carry runtime accumulation.

boris-kz: Forward locality — each wrapper holds [trigger prims..., the typ call]:

```python
def cluster_AST(Q):  # primitives are triggers for following typ CoF; bundle them as one CoF

    seg_, seg = [], []
    for c in Q:
        seg += [c]
        if isinstance(c, CoF): seg_ += [CoF(call_=seg)]; seg = []
    if seg: seg_ += [CoF(call_=seg)]  # trailing prims with no following call
    return seg_
```

Trace `[p1, p2, CoF_A, p3, p4, CoF_B, p5]` →

`[CoF(call_=[p1,p2,CoF_A]), CoF(call_=[p3,p4,CoF_B]), CoF(call_=[p5])]`

The typ-CoF is now an *element inside* the wrapper's call_, not the wrapper itself. So no mutation worry (the typ ref is referenced, not modified).

Open: is a trailing trigger-run with no following call meaningful? It happens for return statements, post-loop cleanup, etc. — primitives that don't gate any onF_ call. Either keep it (current code), or drop it as non-actionable.

kwcckw: > it has to be AST to get Z.call_ first, it doesn't fall from the sky

Okay, so this is a bit different with my understanding earlier. I was thinking we need to pack those the actual runtime calls in T.call_, but with your add_type:

    T.call_ = [typ_[k] for k in call_seq[i] if typ_[k]]

This is actually packing their callees' typ oF rather than the calls oF.

boris-kz: Z.call_ is all possible oF calls, packed with their glue logic in cluster_AST
Z.typ_ is a set of all gated oFs in flattened call_

two orthogonal views on the same runtime:

- **Z.call_**: structural / source-order. cluster_AST output over frame_H's AST callees. Each element = CoF(call_=[triggers..., typ_ref]). Reflects "what gets called from the entry point, in what order, gated by what".

- **Z.typ_**: statistical / per-nF. `flat_(oF)` walks the entire runtime tree, bins every CoF instance by its nF, sum2F per bin. Reflects "how many times each onF_ function ran and with what aggregate stats".

Symmetric at the typ level too:
- **T.call_**: cluster_AST output over T's own function body AST — structural skeleton of its body, with trigger-bundled sub-calls.
- **T.N_**: the F_ instance list for nF, runtime sample population that summed into T.

So every CoF has the same pattern: `.call_` = static structure from AST, `.N_` = runtime sample population, scalar fields (dTT, c, r, fc, fw) = aggregated stats. Z is just the entry-level instance of this pattern.

One implication worth flagging: nothing in Z.call_ is itself in Z.typ_, but Z.call_'s typ references *point into* Z.typ_. So Z.call_ and Z.typ_ share refs, just organize them differently — same data, two indexes.

boris-kz: Two things:

**Is gF call-specific?** Yes. Line 147: `__init__` auto-creates a fresh `.gF` per CoF (`fo=1` default), and the @trace decorator (line 156) creates a new oF + new oF.gF on every invocation. So gF is per-instance, never shared. A typ in Z.typ_ has its own .gF too — that one's the aggregate.

**Should triggers go in .gF instead of .call_?** Yes — that's exactly the gF semantics. Pre-call primitives are evaluation/conditional code that gates whether (or how) the following typ-call runs. That maps cleanly onto the existing gate channel.

Restructured cluster_AST:

```python
def cluster_AST(Q):  # primitives are gates for the following typ; attach to its .gF

    seg_, gate = [], []
    for c in Q:
        if isinstance(c, CoF):
            site = CoF(nF=c.nF, root=Z)        # fresh call-site, .gF auto-created
            site.gF.call_ = gate               # accumulated triggers → gate channel
            seg_ += [site]; gate = []
        else:
            gate += [c]
    if gate: seg_ += [CoF(call_=gate)]         # trailing prims, no follower
    return seg_
```

Notes:

- Site is a fresh CoF; the typ ref is preserved via `.nF` only, no mutation of `Z.typ_[c.nF]`.
- This means each AST occurrence of a typ-call gets its own site object — multiple call sites for the same typ stay distinct, with their own gates.
- The trailing case (primitives with no following typ-call) — strictly speaking, these aren't gates for anything. Could drop them; could keep them as a residual `.gF`-only CoF (`CoF(); ret.gF.call_=gate`). Your call.

Open: does `site` need its `.gF.call_` to be CoF-wrapped primitives, or are raw AST items fine there too? The same question we've been deferring — depends on whether downstream code (merge, comp_callers) reads gate contents structurally or just statistically.

kwcckw: > Z.call_ is all possible oF calls, packed with their glue logic in cluster_AST
Z.typ_ is a set of all gated oFs in flattened call_

Thanks, this is clear now. I misunderstood earlier and thought Z.call_ are those calls added from runtime, while it should be static and it's the all possible oF calls instead.

But the existing code is only getting the possible typ call from runtime, because typ_ is added from flat_(Z), which is added from runtime calls:

      for F in flat_(oF): typ_[F.nF] += [F]

Then from this line:

     T.call_ = [typ_[k] for k in call_seq[i] if typ_[k]]

So `k` is a fork but if it's not run from the runtime, if typ_[k] will be false and it won't be packed to T.call_.

boris-kz: I think that's fine for typ_, it will still have empty slots for oFs that were never called, that's all we need?

kwcckw: > Two things:
> 
> **Is gF call-specific?** Yes. Line 147: `__init__` auto-creates a fresh `.gF` per CoF (`fo=1` default), and the @trace decorator (line 156) creates a new oF + new oF.gF on every invocation. So gF is per-instance, never shared. A typ in Z.typ_ has its own .gF too — that one's the aggregate.
> 
> **Should triggers go in .gF instead of .call_?** Yes — that's exactly the gF semantics. Pre-call primitives are evaluation/conditional code that gates whether (or how) the following typ-call runs. That maps cleanly onto the existing gate channel.
> 
> Restructured cluster_AST:
> 
> ```python
> def cluster_AST(Q):  # primitives are gates for the following typ; attach to its .gF
> 
>     seg_, gate = [], []
>     for c in Q:
>         if isinstance(c, CoF):
>             site = CoF(nF=c.nF, root=Z)        # fresh call-site, .gF auto-created
>             site.gF.call_ = gate               # accumulated triggers → gate channel
>             seg_ += [site]; gate = []
>         else:
>             gate += [c]
>     if gate: seg_ += [CoF(call_=gate)]         # trailing prims, no follower
>     return seg_
> ```
> 
> Notes:
> 
>     * Site is a fresh CoF; the typ ref is preserved via `.nF` only, no mutation of `Z.typ_[c.nF]`.
> 
>     * This means each AST occurrence of a typ-call gets its own site object — multiple call sites for the same typ stay distinct, with their own gates.
> 
>     * The trailing case (primitives with no following typ-call) — strictly speaking, these aren't gates for anything. Could drop them; could keep them as a residual `.gF`-only CoF (`CoF(); ret.gF.call_=gate`). Your call.
> 
> 
> Open: does `site` need its `.gF.call_` to be CoF-wrapped primitives, or are raw AST items fine there too? The same question we've been deferring — depends on whether downstream code (merge, comp_callers) reads gate contents structurally or just statistically.

The `Q` here is a list of preprocessed AST function, something like this?

      Q = [
          Primitive("diff = x - y"), 
          Primitive("magnitude = abs(diff)"), 
          CoF(nF="comp_pixel")
      ]
      
 So gate is packing those primitives only?

boris-kz: yes, because there is a gate for each oF. 
but right now gF is just an eval oF, I guess we need to add other promitives in it too

boris-kz: Right — currently gF.call_ collects runtime eval CoFs (one per `gate()` invocation, line 178). We'd be mixing two categories:

- Existing: eval CoFs (runtime gate samples — each a measurement of one gating decision)
- New: AST primitives (static source of the gating code)

Cleanest unification — same pattern we just established for every CoF:

- `gF.call_` = **static** AST primitives (and the explicit `gate()` site as one of them)
- `gF.fw / gF.fc / gF.dTT` = **runtime accumulated** scalars from each gate decision
- `gF.N_` = the runtime sample list, if you want individual samples preserved

Implication: line 178 (`gF.call_+=[g]`) becomes scalar accumulation:

```python
gF.fw += gw; gF.fc += Ec; gF.dTT += g.dTT  # if g.dTT exists
gF.N_ += [g]                                 # optional: keep samples
```

instead of appending to call_.

So gF stops being a runtime-instance container and becomes a proper CoF: structural call_ from AST, stats from aggregation. Same shape as Z, same shape as every T — fractal.

One downstream consideration: anything that currently iterates `gF.call_` expecting eval CoFs (with `.w, .fc, .span`) now sees a mix of AST primitives and one eval site. Need to audit those readers. The `for i, F in zip((1,0),(oF, oF.gF))` block at line 161 doesn't iterate gF.call_, just sums it via flat_, so probably fine if flat_ tolerates primitives.

boris-kz: gF is still attached to oF, so gF.call_ was redundant, but now we can use it for the primitives?


OK, so just the swap: gF.call_ was redundant as a runtime eval-sample list (the parent oF already aggregates that), so reuse the slot for AST primitives — its actual semantic content.

Then line 178's `gF.call_+=[g]` drops (or moves to scalar accumulation on gF/oF), and gF.call_ holds only the static primitives from cluster_AST. cluster_AST stays as you drafted.

So my over-elaboration with .N_ etc was unnecessary — gF doesn't need a sample list, the redundancy was the whole point.

frame_2D_alg/meta_code.py
```diff
@@ -14,8 +14,10 @@ def merge(F,f):  # combine aligned ops, if-fork per miss, no inline recursion, f
         if Sub.nF=='E':  # previously added gate
             fork = Sub; fin=0
             if sub.nF=='E':
-                sub = merge(Sub, sub)  # add2F(Sub,ssub_)?
-                if sub: C+=sub.fc; call_+=[sub]; add_+=[sub]  # merge failed
+                subt = merge(Sub, sub)  # add2F(Sub,ssub_)?
```
boris-kz, line 17: thanks

frame_2D_alg/meta_code.py
```diff
@@ -53,7 +55,7 @@ def seg_eval(seg, seg_):
         seg_, seg, _C = [],[],[]
         for C in T.call_:
             if isinstance(C, CoF):
-                if not _C: _C = C; seg += [C]; continue
+                if isinstance(_C, list): _C = C; seg += [C]; continue
```
boris-kz, line 58: why not  if not _C: ?

frame_2D_alg/meta_code.py
kwcckw: > I think that's fine for typ_, it will still have empty slots for oFs that were never called, that's all we need?

Yes, but right now onF_ is using a set of defined names, and later when we have new typs after the merging, we need a new onF_ structure? Probably a nested list of base nFs (for example, (0,1,2) in onF_?

frame_2D_alg/meta_code.py
kwcckw: The existing CoF return false when their call_ is empty:

But they are still relevant as long they are not first init empty list?

frame_2D_alg/meta_code.py
```diff
@@ -94,6 +134,28 @@ def cluster_AST(Q):  # initial wrap all primitives with oFs for Z.call_
...
+        rV,rTT = ffeedback(frame)
+        
+        # we need to remove or add FTT_, Fw_ based on the new Z.typ_ here?\
+        # then recompute Fc_ using AST.walk in runtime again
+        for i, tF in enumerate(Z.typ_):
+            if tF: Fw_[i] = tF.fw/tF.c; FTT_[i] = tF.wTT
+        ave/=rV; avd/=rV; Fw_ = np.array(Fw_) / rV  # Fc_ is fixed
+        for i in range(len(FTT_)): FTT_[i] = FTT_[i] * rTT / rV
+    
+        agg_recursion.Z = Z
```
boris-kz, line 149: I don't get this?

frame_2D_alg/meta_code.py
boris-kz: ok

frame_2D_alg/meta_code.py
boris-kz: that's only if typ_ is nested too, then we probably need nF = [root.nF,curr.nF]

kwcckw: > gF is still attached to oF, so gF.call_ was redundant, but now we can use it for the primitives?
> 
> OK, so just the swap: gF.call_ was redundant as a runtime eval-sample list (the parent oF already aggregates that), so reuse the slot for AST primitives — its actual semantic content.
> 
> Then line 178's `gF.call_+=[g]` drops (or moves to scalar accumulation on gF/oF), and gF.call_ holds only the static primitives from cluster_AST. cluster_AST stays as you drafted.
> 
> So my over-elaboration with .N_ etc was unnecessary — gF doesn't need a sample list, the redundancy was the whole point.

I see, so CoF should be packed as seg, while primitive will be packed in their gF.call_ now.

> that's only if typ_ is nested too, then we probably need nF = [root.nF,curr.nF]

Yes, so we need to update `build_typ_skel`  to deal with that since now it assumes onF_ is flat.

frame_2D_alg/meta_code.py
kwcckw: This is actually parsing back the updated params such as Z, FTT_, etc into agg_recursion so that agg_recursion uses the updated params in the next run.

frame_2D_alg/meta_code.py
boris-kz: some refactoring:

def add_gF(Q, root=None): # pack primitives in next oF.gF

    if root is None: root = Z
    C_, gate = [],[]
    for c in Q:
        if isinstance(c, CoF):
            C = CoF(nF=c.nF,root=root)  # fresh call-site, .gF auto-created
            C.gF.call_ = gate; C_ += [C]  # gF is accumulated triggers
            gate = []
        else: gate += [c]
    return C_

frame_2D_alg/meta_code.py
boris-kz: I think you are talking about Z, it represents instance of frame_H, which is the root onF now

kwcckw: If i'm not mistaken, we need to parse the new Z to the next frame? So here i'm parsing it back to agg_recursion, so that the next frame uses the new Z?

frame_2D_alg/meta_code.py
kwcckw: So Q is packed sequentially, all the prior primitives are packed to a later CoF here? While it should be the other way? Prior CoF should pack all the subsequent primitives until the end of code of there's new CoF?

frame_2D_alg/meta_code.py
boris-kz: agg_recursion should get new Z per loop of frame_H, in ffeedback?

                if rV > ave:
                    if elev== max_elev:
                        Z,rV,FTT_ = ffeedback(F)  # from top lev
                        add_call_typ_(Z)  # typ_ maps to Fw_,Fc_,FTT_

frame_2D_alg/meta_code.py
boris-kz: gate is the gF of the next oF

frame_2D_alg/meta_code.py
kwcckw: > agg_recursion should get new Z per loop of frame_H, in ffeedback?
> 
> ```
>             if rV > ave:
>                 if elev== max_elev:
>                     Z,rV,FTT_ = ffeedback(F)  # from top lev
>                     add_call_typ_(Z)  # typ_ maps to Fw_,Fc_,FTT_
> ```

But this is called form agg_recursion, we need the additional `cluster_call_`, `merge`, etc to create new typs, pack it in Z and then maps the Fw_, Fc_ FTT_? This should be done in meta_code?

frame_2D_alg/meta_code.py
kwcckw: Thanks, i was confused earlier.

image

frame_2D_alg/meta_code.py
boris-kz: yes, we need to add something like frame_HH for that, it's not agg_recursion

kwcckw: And that frame_HH should call meta_code, wrapping both existing agg_recursion and meta_cde? Because without meta_code, we won't be able to merge any new typs.

boris-kz: it will be an oF, not a module

kwcckw: Something like each frame is a call , packed within a higher call - frame_HH's oF?

boris-kz: Right

PR #146, opened 2026-05-16
5/16/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -210,29 +227,39 @@ def add_gF(Q, root=None): # pack primitives in next oF.gF
...
+        for x in seq:
+            if isinstance(x, int):
+                if isinstance(typ_[x], CoF): out.append(typ_[x])   # skip empty typ_
+            else: out.append(x)
+        return out
+
     global TYP_SKEL
-    if TYP_SKEL is None: TYP_SKEL = build_typ_skel()
+    if TYP_SKEL is None: TYP_SKEL = build_typ_skel()  # we don't need this TYP_SKEL? We will recompute this in every new frame anyway 
     call_seq, callers, entry_seq = TYP_SKEL
```
kwcckw, line 248: The existing call_seq here is a nested list, each packing primitives or CoF from each typ. 

So each typ's seq is something like this, the primitives and int, where int represent the CoF:

   [<_ast.Assign object at 0x000002D1774F69D0>, <_ast.For object at 0x000002D17DB2B970>, <_ast.If object at 0x000002D17DB03FA0>, 13, 13, <_ast.Assign object at 0x000002D179803550>, <_ast.Assign object at 0x000002D17DADD670>, 2, <_ast.For object at 0x000002D1033D1130>, <_ast.If object at 0x000002D103528D90>, <_ast.Assign object at 0x000002D1035282B0>, <_ast.Return object at 0x000002D1034D6940>]

I added `resolve` function above to pack their typ into sequence if it's an int.

For primitives, they should be added to gate in add_gF. For consistency, i think we can convert it into CoF too? Something like the `prim` function above?

      def prim(primitive, root):  # static primitive computation
      
          # get weights
          def get_ops(node): return sum(weights.get(type(n), 0) for n in ast.walk(node))
          
          # get base weight from vt_
          def get_base():  
              global BASE
              if BASE is None:
                  import importlib, inspect
                  src = inspect.getsource(importlib.import_module('agg_recursion').vt_)
                  BASE = get_ops(ast.parse(src).body[0])
              return BASE
      
          p = CoF(nF=root.nF, root=root, fo=0)      
          p.fc = p.fw = get_ops(primitive) / get_base()  # ast cost
          p.ast = primitive                              # pack ast? To identify between primitive CoF and  calls
          return p

frame_2D_alg/agg_recursion.py
```diff
@@ -1040,6 +1040,7 @@ def expand_lev(_iy, _ix, elev, T):  # seed tile is pixels in 1st lev, or Fg in h
                 elev += 1
                 if rV > ave:
                     if elev== max_elev:
+                        from meta_code import add_call_typ_, ffeedback, add_gF
```
boris-kz, line 1043: why do you import the same thing in every loop?

frame_2D_alg/meta_code.py
```diff
@@ -210,29 +227,39 @@ def add_gF(Q, root=None): # pack primitives in next oF.gF
...
+        out = []
+        for x in seq:
+            if isinstance(x, int):
+                if isinstance(typ_[x], CoF): out.append(typ_[x])   # skip empty typ_
+            else: out.append(x)
+        return out
+
     global TYP_SKEL
-    if TYP_SKEL is None: TYP_SKEL = build_typ_skel()
+    if TYP_SKEL is None: TYP_SKEL = build_typ_skel()  # we don't need this TYP_SKEL? We will recompute this in every new frame anyway
```
boris-kz, line 247: this code update is supposed to be sparser than filter update in current ffeedback, so it will be per frame_H in frame_HH, not per frame in frame_H

frame_2D_alg/meta_code.py
boris-kz: why did you add this prim?

def add_gF(Q, root=None): # pack primitives in next oF.gF

    if root is None: root = Z
    C_, gate = [],[]
    for c in Q:
        if isinstance(c, CoF):
            C.gF.call_ = gate; C_ += [C]; gate = []
        else: gate += [c]
    return C_

all data is added in CoF:

            gF.wTT = ETT_[gF.nF]; gF.fw = Ew_[gF.nF]; gF.fc = Ec_[gF.nF]; _CoF.gF.N_+= [gF]

boris-kz: >So each typ's seq is something like this, the primitives and int, where int represent the CoF:

[<_ast.Assign object at 0x000002D1774F69D0>, <_ast.For object at 0x000002D17DB2B970>, <_ast.If object at 0x000002D17DB03FA0>, 13, 13, <_ast.Assign object at 0x000002D179803550>, <_ast.Assign object at 0x000002D17DADD670>, 2, <_ast.For object at 0x000002D1033D1130>, <_ast.If object at 0x000002D103528D90>, <_ast.Assign object at 0x000002D1035282B0>, <_ast.Return object at 0x000002D1034D6940>]

what is the point of all that?

I think we just need to store python code in gF.call_?

kwcckw: > why did you add this prim?

This prim is creating CoF to pack the code level primitive as CoF, which should be useful if we need to use them for comparison.

> > So each typ's seq is something like this, the primitives and int, where int represent the CoF:
> 
> [<_ast.Assign object at 0x000002D1774F69D0>, <_ast.For object at 0x000002D17DB2B970>, <_ast.If object at 0x000002D17DB03FA0>, 13, 13, <_ast.Assign object at 0x000002D179803550>, <_ast.Assign object at 0x000002D17DADD670>, 2, <_ast.For object at 0x000002D1033D1130>, <_ast.If object at 0x000002D103528D90>, <_ast.Assign object at 0x000002D1035282B0>, <_ast.Return object at 0x000002D1034D6940>]
> 
> what is the point of all that?
> 
> I think we just need to store python code in gF.call_?

Those are actually the primitives, the `c` that's packed in gate in `gate += [c]`. It might be useful for merge or comp if we want to go down to lower levels?

boris-kz: add_gF takes the raw build_typ_skel output (ints + ast.stmts), mints the call-site CoF when it sees an int, accumulates stmts as the gate:

```python
def add_gF(Q):  # int → call-site CoF; stmts → its .gF.call_ (python code)
    C_, gate = [], []
    for c in Q:
        if isinstance(c, int):
            C = CoF(nF=c); C.gF.call_ = gate; C_ += [C]; gate = []
        else: gate += [c]
    return C_
```

No resolve, no typ_ lookup. gF.call_ holds ast.stmt nodes (the actual python code that gates the call) — exactly the primitive content, no wrapping. The call-site CoF carries `.nF`; stats are looked up from Z.typ_[nF] downstream when needed.

build_typ_skel can then drop the typ-ref-as-CoF idea entirely — it just returns sequences of `int | ast.stmt`, which is what it already produces. The `callers` reverse-index still works (scans for ints).

frame_2D_alg/agg_recursion.py
kwcckw: This is within `if elev== max_elev:`, so it should be called only once? But anyway, this should be just temporary since there is cyclic import problem with the existing functions.

boris-kz: ok, right

frame_2D_alg/meta_code.py
kwcckw: Okay thanks. So this is simpler without converting the primitives.

boris-kz: in CoF, this is not needed:

        gF.wTT = ETT_[gF.nF]; gF.fw = Ew_[gF.nF]; gF.fc = Ec_[gF.nF]; _CoF.gF.N_+= [gF]

gF should summed from gF.call_ in add_gF?
that should be per call, typ.gF is meaningless?

kwcckw: > gF should summed from gF.call_ in add_gF?

But gF.call_ are primitives, we need to get their ast weight if we need to get those params from primitives, as in the workflow in my current commit.

> that should be per call, typ.gF is meaningless?

I think it should be per typ, because all calls with the same typ have the same gF since gF is added based on the static ast analysis via `build_typ_skel`?

boris-kz: the same typ can be called in different onFs, with different gating?
yes, we do need those ast weights, so your prim is just to compute gF.fc?

boris-kz: Claude: On the E* init — depends which gF you mean, there are two populations:

Runtime gF (created in traced, per invocation): gate() accumulates onto it during the run, gF.fw += gw; gF.fc += Ec. add_gF runs after the run (in ffeedback). So the runtime gF has no measured call_ when gate() fires — it needs the E* baseline, or fw/fc start at 0 and you lose the structural prior. Keep the init here.
Static call-site gF (created in add_gF from AST): its call_ holds real primitives. fc = sum of primitive costs — measured, not the Ec_ prior. So for these, the E* init is redundant; add_gF should set gF.fc = sum(cost(s) for s in gate) directly.

So not "only in add_gF" — both, different objects. traced keeps E* as the runtime/bootstrap prior; add_gF overwrites with the measured sum for the structural gFs.
The deeper point: Ew_/Ec_/ETT_ are placeholders that exist because you had no measured gate structure. Once add_gF gives real per-gate costs, the E* priors are pure bootstrap — first run only, then ffeedback should be feeding measured values back into them (same way Fw_ gets tF.fw/tF.c). If that feedback loop for E* isn't wired, that's the actual gap — not the init line itself.
Which gF did you mean — or is the question whether E* should survive at all once add_gF is in?

kwcckw: > the same typ can be called in different onFs, with different gating? 

But the problem is `build_typ_skel` gets only static gating information, but not the runtime. I think we would need runtime call oF to retrieve that information, not sure how exactly we can trace that yet.

> yes, we do need those ast weights, so your prim is just to compute gF.fc?

Yes, but we just need;

    p.fc = p.fw = get_ops(primitive) / get_base() 
    
And then sum the fc of all primitives into gF.fc.

boris-kz: gating costs are static per calling oF, that's all we need.

frame_2D_alg/meta_code.py
```diff
@@ -1,7 +1,50 @@
...
+    ast.Subscript: 5,  # index resolution into container
+    ast.GeneratorExp: 3,  # lazy wrapper, inner loop body counted as child nodes
+    ast.While: 1,  # condition re-evaluation overhead per iteration, body counted separately
+    ast.For: 1,  # iterator protocol: __iter__ + __next__ overhead, body counted separately
+    ast.ListComp: 1,  # same iteration overhead as For + list.append + allocation
+    ast.SetComp: 2,  # same as ListComp + hashing per element
+    ast.Call: 3,  # frame creation + arg binding + return: overhead beyond the callee body itself
+}
+
+BASE = None  # unit weight from vt_
```
boris-kz, line 26: that BASE is always 1?

frame_2D_alg/meta_code.py
boris-kz: >p.fc = p.fw = get_ops(primitive) / get_base()

so with
```
costs = {
    ast.Assign: 2,  # bind name: trivial
    ast.Attribute: 5,  # single dict lookup on object
    ast.UnaryOp: 2,  # apply one operator to one operand
    ast.BoolOp: 1,  # short-circuit decision between already-evaluated values
    ast.Compare: 2,  # compare already-evaluated operands
    ast.If: 1,  # test + pick branch, body ops counted separately
    ast.IfExp: 2,  # same as If
    ast.BinOp: 2,  # apply operator to two already-evaluated operands
    ast.AugAssign: 2,  # read + op + store, but op and target counted separately
    ast.Subscript: 5,  # index resolution into container
    ast.GeneratorExp: 3,  # lazy wrapper, inner loop body counted as child nodes
    ast.While: 1,  # condition re-evaluation overhead per iteration, body counted separately
    ast.For: 1,  # iterator protocol: __iter__ + __next__ overhead, body counted separately
    ast.ListComp: 1,  # same iteration overhead as For + list.append + allocation
    ast.SetComp: 2,  # same as ListComp + hashing per element
    ast.Call: 3,  # frame creation + arg binding + return: overhead beyond the callee body itself
}
```
we need to map these costs to each op in gF.call_ and sum them in gF.fc?

boris-kz: Claude:
Mechanically yes — the dict is already designed for walk-and-sum (the "body counted separately" / "inner loop body counted as child nodes" comments assume `ast.walk` recurses into children):

```python
def stmt_cost(s): return sum(NODE_COST.get(type(n),0) for n in ast.walk(s))
# in add_gF: C.gF.fc = sum(stmt_cost(s) for s in gate)
```

`type(n)` with `.get(...,0)` — AST nodes don't subclass within this set, so exact-type lookup is right. The static call-site gF starts at fc=0 (CoF default), so it's a clean assignment, no double-count with the traced E* init (different gF objects).

But one substantive problem, not a detail:

**It's iteration-blind.** `For: 1, While: 1, ListComp: 1` — the loop multiplier vanishes. A loop body that runs once and one that runs 10⁴ times get the same fc. For most non-trivial glue (which is loops over node lists — exactly what this codebase does everywhere), the static sum systematically undercounts true compute cost, often by orders of magnitude. So gF.fc here is a **structural-complexity** measure ("how much code"), not a **runtime-cost** estimate ("how much compute").

That matters because ffeedback's whole premise is weighting by *actual* cost. If gF.fc feeds a gain-vs-cost decision, an iteration-blind cost makes the decision wrong for anything loop-heavy. The iteration count is runtime-only — it has to come from the gate()/N_ side and multiply the body cost. Static AST can give you the per-iteration body cost; it can't give you the count.

Secondary: non-onF_ helper calls get a flat 3 (frame overhead), their bodies uncounted. Fine if helpers are cheap, undercounts if not.

So: fine to compute and store this, but be explicit that gF.fc-from-AST is structural complexity. If ffeedback needs real cost, this number needs an iteration multiplier from runtime before it's trustworthy.

kwcckw: > gating costs are static per calling oF, that's all we need.

Okay right, and that should be added to all call oF (in add_gF) since a same typ might be called in multiple sections, and each of them have different gating.

boris-kz: def add_gF(T): # per typ,call_, pack primitives in next oF.gF.call_

    C_, gate = [],[]
    for c in T.call_:
        if isinstance(c,CoF):
            c.gF.call_ = gate
            c.gF.fc = sum(costs.get(type(n),0) for p in gate for n in ast.walk(p))
            C_ += [c]; gate = []
        else: gate += [c]
    return C_

Each p in gate is an ast.stmt — a tree, not one node — so walk it and sum per-node lookups. Double comprehension: outer over the gate statements, inner over each statement's nodes. costs is your module-level dict.
This drops the build_typ_skel .cost attachment — no precompute, add_gF does the lookup directly. Still static (AST-only, no runtime), just computed at add_gF time instead of skel-build time. Recomputed each add_call_typ_ call, but that's cheap and add_call_typ_ is coarse/infrequent anyway.

kwcckw: > > p.fc = p.fw = get_ops(primitive) / get_base()
> 
> so with
> 
> ```
> costs = {
>     ast.Assign: 2,  # bind name: trivial
>     ast.Attribute: 5,  # single dict lookup on object
>     ast.UnaryOp: 2,  # apply one operator to one operand
>     ast.BoolOp: 1,  # short-circuit decision between already-evaluated values
>     ast.Compare: 2,  # compare already-evaluated operands
>     ast.If: 1,  # test + pick branch, body ops counted separately
>     ast.IfExp: 2,  # same as If
>     ast.BinOp: 2,  # apply operator to two already-evaluated operands
>     ast.AugAssign: 2,  # read + op + store, but op and target counted separately
>     ast.Subscript: 5,  # index resolution into container
>     ast.GeneratorExp: 3,  # lazy wrapper, inner loop body counted as child nodes
>     ast.While: 1,  # condition re-evaluation overhead per iteration, body counted separately
>     ast.For: 1,  # iterator protocol: __iter__ + __next__ overhead, body counted separately
>     ast.ListComp: 1,  # same iteration overhead as For + list.append + allocation
>     ast.SetComp: 2,  # same as ListComp + hashing per element
>     ast.Call: 3,  # frame creation + arg binding + return: overhead beyond the callee body itself
> }
> ```
> 
> we need to map these costs to each op in gF.call_ and sum them in gF.fc?

Yes, but I guess the unit cost of vt_ is no longer relevant here since that's for functions?

frame_2D_alg/meta_code.py
kwcckw: This is for functions, I guess we don't need that for primitives.

frame_2D_alg/meta_code.py
boris-kz: we can build-up func costs from prim costs, starting from funcs that contain only primitives
add_gF works per-typ, and a typ's AST is static, so extraction can be a per-typ helper called inside add_call_typ_:

pythondef typ_AST(nF):  # statement-level int|stmt seq for one onF_ function
    fn = ast.parse(inspect.getsource(FUNCS[nF])).body[0]
    seq = []
    for stmt in fn.body:
        names = [getattr(c.func,'id',None) or getattr(c.func,'attr',None)
                 for c in ast.walk(stmt) if isinstance(c,ast.Call)]
        onF = [n for n in names if n in onF_]
        seq += [onF_.index(n) for n in onF] if onF else [stmt]
    return seq

drop build_typ_skel as a named batch function. The per-typ extractor stays. You still need FUNCS (nF → function object) built once via the importlib/inspect lookup — that one bit of the old build_typ_skel survives, just as a lookup table instead of a full pre-pass.
Net: one small typ_AST(nF) + a FUNCS table, no build_typ_skel. The int→CoF resolution and entry (frame_H) handling still need a home — inside add_call_typ_? Confirm and I'll fit them.

boris-kz: def add_call_typ_(T):

    for F in flat_(T): typ_[F.nF] += [F]          # bin runtime instances, T.call_ still trace tree
    for i, F_ in enumerate(typ_):
        if F_:
            t = sum2F(F_,CoF()); t.nF=i; t.wTT=cent_TT(getattr(t,'rTT',t.dTT),t.r)
            t.N_ = t.call_                         # instances → N_
            t.call_ = add_gF(typ_AST(i))           # static AST structure
            typ_[i] = t
    T.typ_ = typ_
    T.call_ = add_gF(typ_AST(T.nF))                # T's own static structure
    if any(typ_): add2F(T, sum2F([t for t in typ_ if t], CoF()))

This needs add_gF to take the AST sequence directly, not T. Your add_gF(T) reads T.call_ — so either keep that and do t.call_ = typ_AST(i); t.call_ = add_gF(t) (two steps), or change add_gF's signature to add_gF(seq). The seq also still needs ints→call-site-CoF resolution before add_gF's isinstance(c,CoF) works — fold that into typ_AST or add_gF. Which?

kwcckw: > we can build-up func costs from prim costs, starting from funcs that contain only primitives

Yes, so by using that method, it's no longer use vt_ as unit cost but using the native cost instead?

boris-kz: you already used vt_ as unit cost to get your weights{}?
here we are just matching stmnts with weghts

kwcckw: Yes, but that vt_ unit cost should be wrong here because that's for functions? Here it's per ops, so yes, it's just using the default weights from the dict to match with the stmnts.

frame_2D_alg/meta_code.py
```diff
@@ -210,29 +227,39 @@ def add_gF(Q, root=None): # pack primitives in next oF.gF
...
+            if isinstance(x, int):
+                if isinstance(typ_[x], CoF): out.append(typ_[x])   # skip empty typ_
+            else: out.append(x)
+        return out
+
     global TYP_SKEL
-    if TYP_SKEL is None: TYP_SKEL = build_typ_skel()
+    if TYP_SKEL is None: TYP_SKEL = build_typ_skel()  # we don't need this TYP_SKEL? We will recompute this in every new frame anyway 
     call_seq, callers, entry_seq = TYP_SKEL
-    oF.call_ = add_gF(call_seq, root=Z)
```
boris-kz, line 224: >Yes, but that vt_ unit cost should be wrong here because that's for functions? Here it's per ops, so yes, it's just using the default weights from the dict to match with the stmnts.

ok, right.
so how about this:

def typ_AST(nF, typ_):  # static body scan: typ refs | primitive stmts

    seq = []
    # .body[0]: FunctionDef from parsed Module; .body: that function's stmts
    for s in ast.parse(inspect.getsource(globals()[onF_[nF]])).body[0].body:
        oF_ = [n for n in (getattr(c.func,'id',None) or getattr(c.func,'attr',None)
               for c in ast.walk(s) if isinstance(c,ast.Call)) if n in onF_]
        seq += [typ_[onF_.index(n)] for n in oF_] if oF_ else [s]
    return seq

boris-kz: `ast.parse(src)` → `Module`. `Module.body` is the list of top-level nodes; for a single-function source that's `[FunctionDef]`, so `.body[0]` grabs the `FunctionDef`.

`FunctionDef` is the whole `def name(args): ...` node — it bundles `name`, `args`, `decorator_list`, `returns`, and `body`. Its `.body` is the list of statements *inside* the function.

```
def f():      <- FunctionDef (the def header + everything under it)
    x = 1     <- FunctionDef.body[0]
    g()       <- FunctionDef.body[1]
```

Two `.body` = two nested containers: module holds the def, def holds the stmts. First `.body[0]` reaches into the module to get the def; second `.body` reaches into the def to get its statement list.

kwcckw: Sorry I'm not in my desk now, will check again later.

PR #147, opened 2026-05-17
5/17/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -171,10 +170,10 @@ def get_func_node(tree, name):
 def typ_AST(nF, typ_):  # static body scan: typ refs + primitive stmts
 
     seq = []
-    for stmt in ast.parse(inspect.getsource(globals()[onF_[nF]])).body[0].body:  # .body[0]: FunctionDef, .body: function stmts
```
kwcckw, line 174: So instead of globals, we need to get the func's primitives from agg_recursion.

frame_2D_alg/meta_code.py
```diff
@@ -171,10 +170,10 @@ def get_func_node(tree, name):
...
-        oF_ = [n for n in (getattr(c.func, 'id', None) or getattr(c.func, 'attr', None)
-                 for c in ast.walk(stmt) if isinstance(c, ast.Call)) if n in onF_]
-        seq += [typ_[onF_.index(n)] for n in oF_] if oF_ else [stmt]
+    for stmt in ast.parse(inspect.getsource(getattr(agg_recursion, onF_[nF]))).body[0].body:
+        oF_ =  [n for n in (getattr(c.func, 'id', None) or getattr(c.func, 'attr', None)
+                for c in ast.walk(stmt) if isinstance(c, ast.Call)) if n and n in onF_]
+        seq += [typ_[onF_.index(n)] for n in oF_ if isinstance(typ_[onF_.index(n)], CoF)] if oF_ else [stmt]  # skip empty typ? or we need to preserve it?
```
kwcckw, line 176: So do we need to preserve the empty list here when the typ is empty? It's still present in the code, just we don't run it in the prior run.

frame_2D_alg/meta_code.py
```diff
@@ -196,10 +195,13 @@ def add_call_typ_(T):
         if F_:
             t = sum2F(F_,CoF()); t.nF=i; t.wTT=cent_TT(getattr(t,'rTT',t.dTT), t.r)
             t.N_ = t.call_  # instances → N_
-            t.call_ = add_gF(t, typ_AST(i))  # static AST structure
+            t.root = [F.root for F in F_]   # this is missed out?
```
kwcckw, line 198: We still need their root here?

frame_2D_alg/meta_code.py
```diff
@@ -196,10 +195,13 @@ def add_call_typ_(T):
...
             typ_[i] = t
+    # after all typs are filled
+    for i, t in enumerate(typ_):
+        if isinstance(t, CoF): t.call_ = add_gF(t, typ_AST(i, typ_)) # static AST structure
     T.typ_ = typ_
-    T.call_ = add_gF(T, typ_AST(T.nF))  # T's own static structure
+    T.call_ = add_gF(T, [call for t in typ_ if isinstance(t, CoF) for call in t.call_])  # T's own static structure (Z should covers all, so just typ_?)
```
kwcckw, line 204: For Z, i'm not sure if we need to add their gF? Or we can just skip this?

frame_2D_alg/meta_code.py
boris-kz: thanks

frame_2D_alg/meta_code.py
boris-kz: yes but we don't need to use .index, nF is an index?

frame_2D_alg/meta_code.py
boris-kz: right

frame_2D_alg/meta_code.py
boris-kz: Z.gF is frame_H gating: pre-recursion sequence in frame_H itselt?

boris-kz: The thread (`r3254587306`) is kwcckw asking whether Z needs its own gF, after he rewrote the line to concatenate every typ's `call_`:

```python
T.call_ = add_gF(T, [call for t in typ_ if isinstance(t,CoF) for call in t.call_])
```

Your reply ("Z.gF is frame_H gating: pre-recursion sequence in frame_H itself") is the correct call, and it means kwcckw's change should be rejected — revert to:

```python
T.call_ = add_gF(T, typ_AST(T.nF))   # T's own static structure
```

Why the concat is wrong, conceptually: Z is the `frame_H` node — one specific function, the program entry. Its `call_` must be `frame_H`'s own body. Gluing all `t.call_` together makes Z.call_ a multiset of every function's statements with no ordering and no nesting — not a structure. The typs are already separate nodes in `Z.typ_`; Z doesn't re-aggregate them, it *is* frame_H. `typ_AST(Z.nF)` parses frame_H, the recursion-entry CoFs become `Z.call_`, and the primitives preceding them become each CoF's `gF.call_` — that is exactly the "pre-recursion sequence." So: don't skip Z's gF, and don't synthesize it from typ_ — it comes from frame_H.

One wrinkle worth knowing before you sign off the thread. frame_H's recursion entries (`cross_comp`, `vect_edge`, `proj_N`) sit *inside* the nested defs `base_tile`/`expand_lev` and inside the `while elev` loop — none are top-level statements of frame_H. `typ_AST` iterates `.body[0].body` (top-level stmts only) and uses `ast.walk` per stmt, so it *will* find those calls, but it hoists them all out of whichever top-level stmt encloses them (the `while`, or a `def`). add_gF then can't split per-CoF gates inside that stmt — the entire while-loop is one opaque `gate` blob.

So "pre-recursion sequence" is currently recoverable only at the granularity of frame_H's top-level statements (`elev=0`, the two defs, `global`, `while`). If you want the gate to be precisely the statements before the recursive `cross_comp(lev, rr=elev)`, `typ_AST` has to descend into the loop body rather than treating the loop as one stmt. Do you want that descent, or is top-level-statement granularity enough for Z?

frame_2D_alg/meta_code.py
kwcckw: Yes, but we don't use nF here? We are checking oF_, which packing the string names of onF_.

frame_2D_alg/meta_code.py
kwcckw: I see, so this is the gating on it's own recursion, thanks for clarifying.

frame_2D_alg/meta_code.py
boris-kz: use it

frame_2D_alg/meta_code.py
boris-kz: we only need to add gF.call_, that former static oF.call_ was for AST, which is just oF.body?

def add_typ_(R):  # root oF, alayws Z?

    for F in flat_(R): typ_[F.nF] += [F] # bin runtime instances, T.call_ still trace tree
    for i, F_ in enumerate(typ_):
        if F_:
            T.N_ = T.call_  # instances → N_
            T.root = [F.root for F in F_]  # all callers
            typ_[i] = T
    for i, t in enumerate(typ_):  # skip empty typs
        if isinstance(t,CoF): add_gF(t, typ_AST(i,typ_))  # t.gF.call_: static AST primitives
    if any(typ_):
        add_gF(R, typ_AST(R.nF, onF_))  # adds R.gF, onF_[0]=frame_H?
    R.typ_ = typ_

boris-kz: Claude: 
Agreed, and it means your non-assignment is right — I was wrong to flag it as a bug last turn.

The static `oF.call_` as a stored CoF sequence is redundant: it's just the function body, fixed source, re-parseable any time via `typ_AST`. `call_` should stay the runtime trace — the valued structure carrying the dTT/c/r samples. A parallel static copy adds nothing. So drop `t.call_ =` / `R.call_ =`; `add_gF` runs for its side effect only and discarding `C_` is correct.

`gF` is the genuine exception, and that's why you keep it: the gate primitives are *not* in the runtime trace at all — `@trace` only records CoF calls, never the primitive code between them. The only way the gate exists on the object is to derive it from AST and attach it. So `typ_AST` + `add_gF` exist purely to populate `gF.call_`/`gF.fc`.

One thing to resolve before this is clean — and it's the part that decides whether "static call_ is redundant" fully holds:

`typ_AST(i, typ_)` parses function `i`, and for every onF call inside it emits `typ_[that_nF]` — the shared typ aggregate. `add_gF` then does `c.gF.call_ = gate` on that shared object. So a typ called from N parent bodies gets its `gF` written N times across the loop — **last typ_AST call wins**. The gate is inherently per-call-site (the setup before `cross_comp` in `frame_H` ≠ before `cross_comp` in `cluster_N`), but you're storing one `gF` per typ.

So which is it:

- If one representative gate per typ is acceptable → fine as-is, but know that's what's happening.
- If `gF` should aggregate across call sites (like dTT, c, r do) → `add_gF` must *accumulate* into `gF`, not assign.
- If gates must stay distinct per call site → then `typ_AST` has to emit fresh per-site CoFs, and that per-site list *is* a real non-redundant static `call_` — contradicting "redundant." That's the version from a few iterations back.

Which gate model do you want? That determines whether `add_gF`'s plain assignment stays or becomes an accumulate.

boris-kz: typ callers are in typ.root, so those gates per caller should be there too?
then root[i] = [caller,gF]?

Claude: 
that kills the last-write-wins: the gate stops being one shared `typ.gF` and becomes per `(typ, caller)`, which is what it actually is.

Two precision points:

1. `T.root = [F.root for F in F_]` is per runtime *instance* — one entry per F, duplicated when a caller hits the typ in a loop. For `root[i] = (caller, gF)` you need it deduped to caller typs (`typ_[F.root.nF]`), otherwise the same `(caller,gF)` repeats and the gate gets multiply counted downstream.

2. The gate is per call *site*, not per caller. One caller body can call the same typ from several sites with different preceding primitives. `(caller, gF)` collapses those — per caller you either sum the gates into one gF (consistent with the rest of the aggregate model) or root entries are really per-site. Sum it, unless you want them split — which is the clarifying question.

Sourcing matters: the gate is only knowable by parsing the *caller's* body. So build root from the `typ_AST` pass, not the runtime `[F.root for F in F_]` — when `typ_AST(caller_i)` finds a call to typ k, append `(typ_[i], gate)` to `typ_k.root`. typ_AST then yields structure + gates together; `add_gF` no longer writes a shared `typ.gF`.

One thing to decide: `[F.root for F in F_]` (runtime) gives which callers actually fired and how often; typ_AST (static) gives all call sites + gates. They differ — a static call site may be gated off at runtime. Do you want root to be the static caller set (gates, full structure) and fold runtime frequency into each caller's `c`, or keep both lists?

kwcckw: Okay, so when we run `add_gF(R, typ_AST(R.nF, onF_))`, it's adding gF to R only right?

>  root[i] = [caller,gF]?

Okay right, because we may get multiple gates for calling the same function from different section of code.

Also for root, right now it's packing runtime info, we probably just need another var to pack static root info?

boris-kz: something like this?

def add_typ_(R):  # root oF, always Z?

    def typ_AST(nF, typ_):  # static body scan: typ refs + primitive stmts
        seq = []
        for stmt in ast.parse(inspect.getsource(getattr(agg_recursion, onF_[nF]))).body[0].body:  # .body[0]: FunctionDef, .body: function stmts
            oF_ =  [n for n in (getattr(c.func, 'id', None) or getattr(c.func, 'attr', None)
                    for c in ast.walk(stmt) if isinstance(c, ast.Call)) if n and n in onF_]
            seq += [typ_[onF_.index(n)] for n in oF_ if isinstance(typ_[onF_.index(n)], CoF)] if oF_ else [stmt]
            # keep typ slots for indexing
        return seq

    def add_gF(T, stmts):  # per typ.call_, pack primitives in next oF.gF.call_
        C_, gate = [],[]
        for c in stmts:
            if isinstance(c,CoF):
                c.gF.fc = sum(costs.get(type(n),0) for p in gate for n in ast.walk(p))
                c.gF.call_ = gate; C_ += [c]; gate = []
            else: gate += [c]
        return C_

    for F in flat_(R): typ_[F.nF] += [F] # bin runtime instances, T.call_ still trace tree
    for i, F_ in enumerate(typ_):
        if F_:
            T.N_ = T.call_; root_ = []  # instances → N_
            for F in F_:  # gate per caller:
                r = F.root; root_ += [[r, add_gF(None, typ_AST(r.nF, onF_))]]
            T.root = root_
            typ_[i] = T
    for i, t in enumerate(typ_):  # skip empty typs
        if isinstance(t,CoF): add_gF(t, typ_AST(i,typ_))  # t.gF.call_: static AST primitives
    if any(typ_):
        add2F( R, sum2F([t for t in typ_ if t], CoF()))
        add_gF(R, typ_AST(R.nF, onF_))  # adds R.gF, onF_[0]=frame_H?
    R.typ_ = typ_

frame_2D_alg/meta_code.py
boris-kz: Ok gF should be per caller, so we just need to convert oF call_ to [(prim or [gate, (oF or subcall_))]. 
gate can be a set of primitives or gF. subcall_ is forks with a common gate, so call_ can be a tree. 
same form for oF.body, just static. 
Gated oFs are ops we collect statistics on, while residual primitives can be simple core or gating ops. No more oF.gF

boris-kz: Actually, gate oF needs to collect statistics on gated oFs.
So the latter should be packed in the call_ of the former.
gate oF.call_ is a mix of primitives and gated oFs, same as any call_, no formal distinction?
the distinction between call_ and body is that the body is static and comprehensive, 
while call_ is only statistics-collecting oFs, with root as a caller?

boris-kz: Claude — so the correction is two-fold, not just static-vs-valued: `call_` is also prim-stripped.

`body` = the full static `[prim|oF]` tree. `call_` = `body` projected onto its `oF` nodes only, valued. Primitives carry no runtime stats and their cost is re-derivable from syntax, so they're dead weight in `call_` — drop them, keep the valued oFs.

The consequence worth being explicit about: nesting in `call_` then reflects only *oF*-gates. A primitive `if` (non-onF condition) is gone, so the oFs it gated rise to the enclosing scope — primitive-gated branches flatten in `call_`. That's consistent with the earlier point: only stats-bearing gates are oFs, only they need to appear in the stats structure; primitive gating, like primitive cost, lives in `body` and is consulted from there when needed.

So: `call_` = oF tree, valued, oF-gate nesting only. `body` = full tree, static, source of structure + primitive cost. `N_` = raw instance samples, separate.

kwcckw: Okay, so in short, gF is no longer needed and body is static info on primitives and all possible callees, while call_ is runtime executed callees only? For root, it will be in both body and call to pack all possible callers and runtime callers?

boris-kz: root is probably relevant only for nested oFs, not their callers.
The callers are local to the call anyway, and serve as gate, so they should collect all data.
I don't know if we need to represent them in callee oFs, there is no separate feedback?

boris-kz: `root` = the structural parent in the `call_` tree — the enclosing/gating oF whose `call_` this oF sits in. It's the up-edge, inverse to `call_`'s down-edge, and it's an occurrence-level attribute: each occurrence has exactly one structural parent.

So a typ aggregate has no single `root` — `T.root = [F.root for F in F_]` should go. A typ is a pure sum (`sum2F`); "callers" isn't a typ-level fact. The caller of an occurrence just *is* its structural parent, and that parent — acting as gate — already contains the occurrence via its `call_` and aggregates its stats bottom-up. No callee→caller back-reference is needed for that.

And right — no separate feedback path needs one either. `ffeedback` is global filter adjustment propagated through the level structure, not a per-edge signal routed from a callee back to specific callers. A back-ref would only be required if feedback were per-caller; it isn't.

One consumer to check before dropping the caller-list outright: `cluster_call_` — its whole job was comparing callers' bodies. With the list gone, "callers of typ k" = reverse-scan `typ_` for any `call_` referencing k, then read those callers' `call_`. That's the same re-derivable-structure principle you've applied throughout — fine, as long as `cluster_call_` does that scan rather than expecting a stored list.

frame_2D_alg/meta_code.py
boris-kz: anyway, all that looks like a pile of nonsense to me

frame_2D_alg/meta_code.py
kwcckw: > The callers are local to the call anyway, and serve as gate, so they should collect all data.

Yes, they are local to call, so it will be callee.root = caller only? Where their gate info should be in caller's body?

boris-kz: callee has multple callers, and each can collect info per call, no need to represent them in callees

kwcckw: Okay, we had this previously for comparison because we need their caller from the call to identify if both calls are identical in some way?

frame_2D_alg/meta_code.py
kwcckw: This is to get the typ oF based on oF_, where oF_ is the names (string) retrieved from ast. So typ_ may contain empty list, and we can't really use their nF here.

frame_2D_alg/meta_code.py
boris-kz: Claude: Yes — `_CoF` in CoF is the caller. 
The contextvar holds the current caller; `@trace` appends each call into `_CoF.call_`, so a caller's `call_` ends up with one CoF per call it made, each carrying that call's dTT/c/r. `root` is just that `_CoF`. The per-call collection mechanism already exists — no new machinery.

But be precise about what it gives: `_CoF` nests by *dynamic enclosing function*, not by *gate*. When `comp_N_` runs inside an `if` condition, it returns before the branch executes — so the branch's calls trace with `_CoF` set to the caller function, landing as *siblings* of `comp_N_`, not as its children. So "gate oF collects stats on gated oFs" is not what the tracer produces.

That nesting is a static overlay: take the flat-per-function trace `_CoF` gives, add the gate structure from `body`/typ_AST, and re-parent gated oFs under their gate oF. Trace supplies values plus the dynamic-caller tree; `body` supplies gate structure; `call_` is the merge.

So caller-collects-per-call is done — it's `_CoF`. Gate-oF nesting is derived, not traced.

Claude
Right — then there's no separate gate notion and no static overlay. The dynamic enclosing function *is* the gate, `_CoF` is that function, and the caller's `call_` collecting one valued CoF per call it makes is the whole mechanism. `comp_N_` in an `if` condition isn't a sub-gate — it's just one of frame_H's calls. frame_H is the gate of everything in its body, and `frame_H.call_` already holds them all, siblings, valued. `root` = `_CoF` = that gate.

So drop what I was building: no re-parenting gated oFs under a within-function gate oF. `call_` is flat per function — its oF calls as siblings — and the tree is purely function→callee recursion: a call's `nF` → `typ_[nF]`, whose `call_` is that function's calls.

The within-function conditional structure — which calls a given `if` co-gates — isn't in `call_` and doesn't need to be. It lives in `body`, static, consulted only if `cluster_call_`/valuation ever needs that grouping. Same split as before: `body` = structure, `call_` = values. `_CoF` is complete; nothing to add.

frame_2D_alg/meta_code.py
boris-kz: replace strings with indices in oF_
or we can use typ instead

frame_2D_alg/meta_code.py
kwcckw: Okay, so we need the other way around later, from oF to their calls (caller-calls) instead of the existing call to retrieve their caller.

boris-kz: yes, that's already  `_CoF in CoF`.
but it seems we do need `caller_` for comp_callers

kwcckw: That `_CoF` is added in runtime, but caller_ is all possible callers for a typ oF? So caller_ should be static and can be retrieved from ast.

boris-kz: no, caller_ is dynamic, we need actual callers

kwcckw: Then it's just single caller? Because the runtime call should have only single caller?

boris-kz: no

frame_2D_alg/meta_code.py
boris-kz: anyway, it seems this is all dead now:

`ast.walk` survives in `get_ops` / `init_wc` / `get_wc` — summing `costs.get(type(n),0)` over each function's `FunctionDef` to get its static op-count weight (`wc`/`fc`). That's intrinsically syntactic; the trace can't produce it, since cost is a property of the code, not the run.

`typ_AST` and `add_gF` are now dead, and their walks go with them:

- `typ_AST` built the static `[prim|oF]` sequence for `call_`. But `call_` now comes from the `_CoF` runtime trace — valued oFs, flat per function, tree via `nF` recursion. No parse needed.
- `add_gF` packed gates. The gate is the dynamic enclosing function, so there's no gate-sequence to build and nothing to pack.

The only thing that could revive a structural parse is conditional: if `cluster_call_`/valuation ever needs the within-function `if`-grouping or the *never-fired* call sites — neither is in the trace, both are in source. But nothing currently active needs that, so don't build it speculatively.

So: keep the cost walks, drop `typ_AST` and `add_gF` entirely.

kwcckw: Okay, so we only retrieve their cost when needed, rather than pre-compute them now?

frame_2D_alg/meta_code.py
kwcckw: Okay, so that should be retrieve-able from typ oF only since it collects all callers of same typ calls.

frame_2D_alg/meta_code.py
boris-kz: I guess it's just typ.fc

PR #148, opened 2026-05-18
5/18/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -82,7 +82,7 @@ def seg_eval(seg, seg_):
             else: seg += [C]
         if seg: seg_eval(seg, seg_)  # last seg, add eval?
         if seg_: T_ += seg_  # segmented T, add eval?
-        else:    T_ += [T]   # recycled T
+        else:    T_ += [T]   # recycled T (seg is always filled, so T_ may not be relevant?)
```
kwcckw, line 85: This line is redundant? Because we won't get T. with empty call_, and seg always will be filled anyway.

frame_2D_alg/meta_code.py
```diff
@@ -54,18 +54,18 @@ def merge(F,f):  # combine aligned ops, if-fork per miss, no inline recursion, f
...
     D = sum([c.w * c.c for c in off])
     return M/(D or eps)  # match if same_callers / diff_callers > ave?
 
-def cluster_call_(Z):  # cluster T callees if called together?
+def cluster_call_():  # cluster T callees if called together? (Z is global and can be retrieved within the func anyway)
 
     def seg_eval(seg, seg_):
         if sum(c.fc for c in seg) > ave * Ec_[0]:  # base eval cost
-            T = sum2F(seg); T.nF = [call.nF for call in T.call_]  # new_nF = old_nF_?
+            T = sum2F(seg); T.nF = [call.nF for call in T.call_]  # new_nF = old_nF_? (only need list when there's different typ in call_?)
```
kwcckw, line 68: The existing code group calls and concatenate their nF as the new typ.nF.

As a result, we may get something like new typ.nF = [2,2,2], or new typ.nF = [3,3,3]. When all of their nFs are the same, i guess we can remove the group to become new typ.nF = 2 or new typ.nF = 3?

kwcckw: Also i think the next step we need to update onF_, Fc_, Fw_, FTT_ to align with the new typs?

boris-kz: if oF is nested, then nF should be nested too: a list of levels, not a list of elements.
so a group nF is nF of the first element, same slot in top oF_, and element nFs are indices in group.typ_? 
yes, update onF_, Fc_, Fw_, FTT_

frame_2D_alg/meta_code.py
boris-kz: seg_eval should remove weak seg, so it may return empty

frame_2D_alg/meta_code.py
```diff
@@ -158,7 +135,7 @@ def add_typ_(R):  # root oF, always Z?
             T = sum2F(F_,CoF()); T.nF=i; T.wTT=cent_TT(getattr(T,'rTT',T.dTT), T.r)
             T.N_ = T.call_; root_ = []  # instances → N_
             T.caller_ = [F.root for F in F_]  # for comp_callers only?
-            T.root = root_
+            T.fc = get_ops(ast.parse(inspect.getsource(getattr(agg_recursion, onF_[i]))).body[0])
```
boris-kz, line 138: thanks

frame_2D_alg/meta_code.py
kwcckw: I see, so a nested nF should be unique and form their own typ since it's very unlikely to get the same nested structure multiple times. 

> so a group nF is nF of the first element, same slot in top oF_, and element nFs are indices in group.typ_?

Actually i don't quite get this, could you include an example? Also could you clarify if the group here is referring to the new typ after the seg_eval?

frame_2D_alg/meta_code.py
kwcckw: Okay right, thanks.

frame_2D_alg/meta_code.py
boris-kz: group oF is a new typ, and it's own typ_ is group elements, so their nFs are indices in that new typ_

boris-kz: def seg_eval(seg, seg_):
        if sum(c.fc for c in seg) > ave * Ec_[0]:  # base eval cost
            T = sum2F(seg); nF = T.nF = seg[0].nF  # 1st onF slot
            onF_[nF+1: nF+len(seg)+1].pop()  # remove nested oFs
            for i, F in enumerate(seg): F.nF = i  # nested indices 
            seg_ += [T]
        else: seg[:] = []

kwcckw: > nF = T.nF = seg[0].nF  # 1st onF slot

The new typ.nF is the first call.nF?

> onF_[nF+1: nF+len(seg)+1].pop() 

`seg` might contain partial typ.call_ only,  so we shouldn't pop `onF_` based on the size of seg?

boris-kz: actually we don't need all this indexing, it's just a new T_:

def cluster_call_():  # cluster T callees if called together, for global Z only?

    def seg_eval(seg):
        if sum(c.fc for c in seg) > ave * Ec_[0]:  # base eval cost
            # add co-occurence (external match) eval here
            for i, F in enumerate(seg): F.nF = i  # nested indices
            return sum2F(seg)
    T_ = []
    for T in Z.typ_:
        if not T: continue
        seg_,seg, _C = [],[],[]
        for C in T.call_:
            if isinstance(C, CoF):
                if isinstance(_C, list):  # init
                    _C = C; seg += [C]; continue
                _t, t = Z.typ_[_C.nF], Z.typ_[C.nF]
                if comp_callers(_t,t) > ave:
                    seg += [C]  # add consecutive externally matching calls
                else:  # init seg with consecutive externally missing call
                    seg_+= [seg_oF] if (seg_oF := seg_eval(seg)) else []
                    seg = [C]; _C = C
            else: seg += [C]
        if seg:  # last seg
            seg_ += [seg_oF] if (seg_oF := seg_eval(seg)) else []
            T_ += [seg_ or T]  # eval segmented T?
    return CoF(typ_ = T_)  # new Z for new input, no other attrs yet?

boris-kz: `onF_` mutation is gone — good, that kills the global-corruption bug. Remaining issues:

**1. `T_ += [seg_ or T]` nests a list — concrete bug.** `seg_` is itself a list of composites. Wrapping it in `[]` appends it as one element, so `T_` becomes `[[oF,oF], T, [oF], ...]`. Next round `for T in Z.typ_` hits a list → `T.call_` AttributeError. You want:
```python
T_ += seg_ or [T]
```
(extend with the segments; fall back to `[T]`). This is the `if seg_: T_+=seg_ else: T_+=[T]` from the original, correctly collapsed.

**2. `_C` no longer advances in the match branch.** It's updated only in init and in the `else` (miss) branch. So `comp_callers(_t, t)` always compares against the *segment anchor*, never the immediate predecessor. The prior version advanced `_C` every CoF. The comment says "consecutive matching calls" (implies chain), the code does anchor-comparison. Both are defensible for co-occurrence clustering — anchor gives more homogeneous segments — but pick it deliberately; right now comment and code disagree.

**3. The placeholder eval needs the match values you're discarding.** `comp_callers` is called pairwise and its result thrown away after `> ave`. To make `seg_eval` a real compression criterion (your `# add co-occurence eval here`), you need the *accumulated* match across the segment — so retain each `comp_callers` value as the segment grows (e.g. carry a running sum on `seg`, or recompute inside `seg_eval`). `sum(fc) > ave*Ec_[0]` alone is still cost-only.

**4. Partial data loss persists (acknowledged by your "eval segmented T?").** If a T splits into 3 segments and only 1 passes `seg_eval`, the other 2 segments' callees are dropped — `seg_` is truthy so the whole-T fallback doesn't fire. Recycle is all-or-nothing; there's no "kept segments + leftover callees" path.

**5. Unresolved from before:** `nF` overloaded (global slot on composites from `sum2F` vs. local 0..n on members) breaks `comp_callers` on a second pass; `c.fc` in the sum still crashes if `call_` holds non-CoF items. If `call_` is genuinely CoF-only per the line-144 comment, the `else: seg += [C]` branch is dead — resolve which.

kwcckw: When we have run this section:

      def seg_eval(seg):
          if sum(c.fc for c in seg) > ave * Ec_[0]:  # base eval cost
              # add co-occurence (external match) eval here
              for i, F in enumerate(seg): F.nF = i  # nested indices
              return sum2F(seg)
              
 What would be the nF of this new typ? We might form multiple new typs within a single T, so all of them should have unique nF?

boris-kz: never mind nF, that's deeply secondary, the problem is computing co-occurence.
it can't be done in T.call_, the only caller is always T

boris-kz: Agreed — the `for C in T.call_` iteration is the wrong frame for co-occurrence.

Inside one T, every callee is called by T by construction. So `comp_callers(_t, t)` for two callees of the same T always has T in the overlap — a constant offset. The discriminating signal is the *rest* of each typ's caller set, and that's already typ-level: `_t = Z.typ_[_C.nF]` is the canonical typ, `_t.root` aggregates callers across the whole namespace, not T-locally. So you don't need the T loop to compute it — and restricting to consecutive pairs within `T.call_` is comparing call-order *adjacency*, not co-occurrence.

The two notions are conflated in the current code:
- **Co-occurrence / "called together"** = which oFs share caller sets. A global statistic over `Z.typ_`. Comparator `comp_callers` (caller-set overlap) is correct for this.
- **Call-order adjacency** = which oFs are called consecutively by some T. That's what `T.call_` sequence position gives you — a pipeline/composition signal, and `comp_callers` is the *wrong* comparator for it.

If you want co-occurrence, drop the outer `for T` and the `seg_`/`_C` linear-segmentation machinery entirely. It becomes: cross-comp `Z.typ_` entries pairwise on caller sets, cluster oFs with high mutual caller-overlap into composites — i.e. connectivity clustering over the oF set keyed by shared callers, the project's primary clustering mode. Linear segmentation by consecutive position is the wrong shape; there's no meaningful "segment" of a co-occurrence graph.

So `cluster_call_` should iterate `Z.typ_` directly. The only thing lost is order-adjacency — and if you do want pipeline patterns too, that's a separate pass with a different comparator, not `comp_callers`.

kwcckw: > it can't be done in T.call_, the only caller is always T

So T is the typ oF, where T.call_'s caller should be different since a same function might be called from multiple sections?

But anyway, actually between T.call_ looks better to me since it more reasonable to cluster calls from different T, because within T.call_ should be the same func, which mostly will be clustered all the time.

boris-kz: def cluster_call_():  # cluster T callees if called together, for global Z only?

    _T_ = []
    for t in Z.typ_:
        if t: t.grp = [t]; t.caller_ = set(t.caller_); _T_ += [t]; t.V = 0  # pairwise accum of membership value?
    V = 0; Olp,Off = set(),set()
    C = -Fc_[0]  # default cost of new representation, placeholder
        if _T.grp is T.grp: continue  # the pair was merged
        v,olp,off = comp_callers(_T,T)
        if v > ave:
            V += v; Olp.update(olp); Off.update(off)  # caller overlap and offset, not sure we need it
            _T.grp += T.grp; T.grp = _T.grp; T.V+=v; _T.V+=v  # concat new roots of externally matching typs
    grp_ = set(t.grp for t in _T_)
    for T_ in grp_:
        v,c = 0,-Fc_[0]
        for t in T_: v+= t.v; c+= t.fc  # external match and compression of element representation
    if V > ave - C:
        G_ = []
        for T in _T_:
            if T.grp in G_: continue
            grp = T.grp; gV = sum(t.v for t in grp); gC = sum(t.fc for t in grp) - 1  # if G.fc = 1
            if gV > ave- gC:  # caller overlap gain - (ave - grp_compression)
                G_ += [grp]
            else: G_ += grp; V-= gV; C-= gC  # revert grouping
        if V > ave - C:
            G_ = [sum2F(grp) if isinstance(grp,list) else grp for grp in G_]  # keep discrete Ts
            # form new Z for new input, vals added in CoF traced:
            return CoF(typ_=G_)  # or Z.w = V+ C*ave?

kwcckw: So this is actually clustering Ts, and if T is empty, we should just skip them?

Also we don't need this section? And the v in this section has redundancy since ts in the same group added the same v bidirectionally.

      for T_ in grp_:
          v,c = 0,-Fc_[0]
          for t in T_: v+= t.v; c+= t.fc  # external match and compression of element representation

For the new typ here, i guess we just need to reassign their nF based on their sequence?
          
    G_ = [sum2F(grp) if isinstance(grp,list) else grp for grp in G_]  # keep discrete Ts

boris-kz: Claude:
Corrected `cluster_call_` (your last pasted version is the base; the file copy is stale so I'm not touching it):

```python
def cluster_call_():  # cluster T callees if called together, for global Z only?

    _T_ = []
    for t in Z.typ_:
        if t: t.grp = [t]; t.caller_ = set(t.caller_); t.V = 0; _T_ += [t]
        if _T.grp is T.grp: continue  # already merged
        v,olp,off = comp_callers(_T,T)
        if v > ave:
            g, t_grp = _T.grp, T.grp        # capture before repointing
            for x in t_grp: x.grp = g       # repoint whole group of T, not just T
            g += t_grp
            _T.V += v; T.V += v
    grp_ = []
    for t in _T_:                            # distinct groups by identity (lists unhashable)
        if not any(t.grp is g for g in grp_): grp_ += [t.grp]
    V,C, G_ = 0, -Fc_[0], []                 # C: default cost of new representation
    for grp in grp_:
        gV = sum(t.V for t in grp) / 2       # un-double: each pair counted on both members
        gC = sum(t.fc for t in grp) - 1      # if G.fc = 1
        if gV > ave - gC:                    # caller-overlap gain > ave - grp compression
            G_ += [grp]; V += gV; C += gC
        else: G_ += grp                      # revert: keep discrete Ts
    if V > ave - C:
        return CoF(typ_=[sum2F(grp) if isinstance(grp,list) else grp for grp in G_])
```

Changes:
- **Union fixed.** `g, t_grp = _T.grp, T.grp` captures both lists first, then repoints *every* member of T's group to `g` before `g += t_grp`. Capturing `t_grp` first avoids the self-concat that `g += T.grp` would cause once `T.grp` itself gets repointed to `g`.
- **`t.V`** used consistently (was `t.v` on read).
- **Dead `grp_` loop removed**, replaced by a correct distinct-groups collection — your original `set(t.grp ...)` crashed on unhashable lists; identity-`any` does what you wanted.
- **Accounting unified.** Global `V` pre-sum dropped; `V` and `C` accumulate per-group from accepted groups only. `gV` divided by 2 so it's in the same units as a single-counted pair total — fixes the off-by-2 revert.
- **Rejected-group dedup fixed** — iterating distinct `grp_` instead of `_T_` means a rejected multi-member group dissolves to discrete members exactly once.

Left as-is: `olp,off` unpacked but unused (your call — drop from `comp_callers` return if not needed). And `comp_callers` itself must read `caller_`, not `root` — verify, since I can't see its current version.

kwcckw: So to prevent redundancy, i guess we can skip the bidirectional assignment here?

    _T.V += v; T.V += v
    
We always eval with summed v from the group anyway, so i think it's fine if we only assign it in either one of the T.

boris-kz: not sure, that won't be accurate per T, we may need that membership value elsewhere

def cluster_call_():  # cluster Ts if called together, for global Z only?

    _T_ = []
    for t in Z.typ_:
        if t: t.grp = [t]; t.caller_ = set(t.caller_); t.V = 0; _T_ += [t]
        if _T.grp is T.grp: continue  # already merged
        v, olp,off = comp_callers(_T,T)
        if v > ave:
            _g,g = _T.grp,T.grp
            for t in g: t.grp = g  # repoint T.grp
            g += g; _T.V += v; T.V += v  # pairwise accum of membership value
    G_= []
    V = 0; C = -Fc_[0]  # grp cost if grp.nF=comp_N_ 
    for t in _T_:
        if t.grp[0] is not t: continue  # eval each group once, at its seed
        grp = t.grp; gV = 0; gC = -Fc_[0]  # membership and cost per grp 
        for t in grp: gV += t.V; gC += t.fc
        gV /= 2  # pairwise redundancy
        if gV > ave - gC: G_ += [grp]; V += gV; C += gC
        else: G_ += grp;
    if V > ave - C:
        # form new Z for new input, vals added in CoF traced:
        typ_ = [sum2F(grp) if isinstance(grp, list) else grp for grp in G_]
        return CoF(typ_=typ_)  # or Z.w = V+ C*ave?

kwcckw: Okay, that gV/2 is actually accurate too, just we won't be able to retrieve the per pair value without the link here.

boris-kz: we probably need to store summed membership V per grp_oF, that's meaningfull for clustering?

Yes — and it's not free, which is the point. `sum2F(grp)` only *node-sums* the members; `gV` is a *relational* quantity — the sum of inter-member caller-match — so it lives in the pairwise comparisons and isn't recoverable from the node-sum. If you don't store it, the composite loses the information that justified forming it.

It's the same structure as everywhere else in the project: a clustered graph carries summed-link value alongside summed-node value. The grp-oF is a graph of member typs; the matching caller-pairs are its links; `gV` is their summed match. So it belongs on the composite, exactly analogous to a `cross_comp` graph storing its link sum.

Placement: put it on the composite as `.V` — uniform with the member `.V` you already added (member `.V` = that typ's accumulated pair-match; composite `.V` = the group total). Symmetrically, `gC` is the group's compression — also a group-level property worth keeping if anything downstream evals the composite.

Two practical notes:
- It must be set explicitly — `.V` is an ad-hoc attr, `sum2F` won't carry it. And by the time you build `typ_` in the final comprehension, `gV` is out of scope. So form the composite in the accept branch where `gV` is still live, rather than in a trailing comprehension:
  ```python
  if gV > ave - gC:
      oF = sum2F(grp); oF.V = gV; oF.fc... 
      G_ += [oF]; V += gV; C += gC
  else: G_ += grp
  ```
  then drop the `sum2F(... if isinstance ...)` comprehension; `G_` already holds composites + discrete typs.
- Keep `.V` distinct from whatever `CoF.traced` writes on the next run — `.V` is a static formation value, traced adds dynamic run-values. Don't let them alias the same slot.

The one thing to decide: is stored `.V` meant to be *consumed* by the next-level `cluster_call_` (a cohesion prior when merging composites), or just diagnostic? If the former, it should feed `comp_callers` or the eval — currently nothing reads it.

kwcckw: > we probably need to store summed membership V per grp_oF, that's meaningfull for clustering?

It will be used in the next frame? Within comp_callers as an additional component?

frame_2D_alg/meta_code.py
```diff
@@ -54,18 +54,18 @@ def merge(F,f):  # combine aligned ops, if-fork per miss, no inline recursion, f
...
     return M/(D or eps)  # match if same_callers / diff_callers > ave?
 
-def cluster_call_(Z):  # cluster T callees if called together?
+def cluster_call_():  # cluster T callees if called together? (Z is global and can be retrieved within the func anyway)
 
     def seg_eval(seg, seg_):
         if sum(c.fc for c in seg) > ave * Ec_[0]:  # base eval cost
-            T = sum2F(seg); T.nF = [call.nF for call in T.call_]  # new_nF = old_nF_?
+            T = sum2F(seg); T.nF = [call.nF for call in T.call_]  # new_nF = old_nF_? (only need list when there's different typ in call_?)
             seg_ += [T]
```
boris-kz, line 69: not sure ATM, but we keep all meaningful attrs. so

def cluster_call_():  # cluster Ts if called together, for global Z only?

    def comp_callers(_T, T):
        # compute rel value of callers overlap
        olp = _T.caller_ & T.caller_
        off = list(_T.caller_- olp) + list(T.caller_- olp)
        return M / (D or eps)
        # match if same_callers / diff_callers > ave?
    _T_ = []
    for t in Z.typ_:
        if t: t.grp = [t]; t.caller_ = set(t.caller_); t.V = 0; _T_ += [t]
        if _T.grp is T.grp: continue  # already merged
        v = comp_callers(_T,T)
        if v > ave:
            _g,g = _T.grp,T.grp
            for t in g: t.grp = _g  # repoint T.grp
            _g += g; _T.V += v; T.V += v  # pairwise accum of membership value
    G_= []
    V = 0; C = -Fc_[0]  # grp cost if grp.nF=comp_N_
    for t in _T_:
        if t.grp[0] is not t: continue  # eval each group once, at its seed
        grp = t.grp; gV = 0; gC = -Fc_[0]  # membership, compression / grp
        for t in grp: gV += t.V; gC += t.fc  # add links?
        gV /= 2  # norm for pairwise redundancy
        if gV > ave-gC: G_ += [[grp,gV,gC]]; V += gV; C += gC
        else:           G_ += grp
    if V > ave - C:
        typ_ = []  # form new Z.typ_ for new input, vals added in CoF traced:
        for grp in G_:
            if isinstance(grp,list):
                g,v,c = grp; T=sum2F(g); T.memb=v; T.compr=c  # downward reciprocals of V,C?
                typ_ += [T]
            else: typ_ += [grp]  # keep old T
        Z = CoF(typ_=typ_); Z.memb, Z.compr = V,C  # membership and compression summed per Z?
        return Z

boris-kz: Claude
Structurally this is sound now — union, accumulation, final loop, and the `.memb`/`.compr` rename all check out, no crashes. Two things remain:

**Singleton groups pass on phantom compression.** A typ that matched nothing has `grp=[t]`, `t.V=0`, so `gV=0`. The only way it then satisfies `gV > ave-gC` is via `gC = t.fc - Fc_[0]` — i.e. claimed compression. But `sum2F([t])` wraps one typ in a composite: zero abstraction, the member is retained inside, net cost goes *up*, not down. The `gC` formula assumes k members collapse to one slot; for k=1 there's no collapse. Since `gV=0` means no cohesion, a singleton should never form a composite — gate the accept on `len(grp) > 1` (or equivalently `gV > 0`). Currently a lone expensive typ gets needlessly wrapped.

**`set(t.caller_)` assumes every typ carries `caller_`.** You committed to `caller_` over `.root` — confirm it's actually populated on typs (in `add_typ_`/`CoF.traced`), or `set(t.caller_)` is an AttributeError on the first typ. This is the one dependency I can't see.

On your `# add links?`: nothing to add. The matching caller-pairs aren't materialized as link objects — each `v` is already folded into the members' `.V`, and the composite's `T.memb` *is* the summed link value. The links' representation cost is what the single `Fc_[0]` composite-cost abstracts. You'd only "add links" if you wanted to keep them discretely, which would contradict storing the summed `memb`.

kwcckw: Okay, right now we use only comp_callers, which is external, I guess we can include some additional internal features such as v or their combined body primitives later.

boris-kz: Right, but we have to consider the sequence too. It will be similar to merge

PR #149, opened 2026-05-19
5/19/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -62,32 +77,31 @@ def comp_callers(_T, T):
...
-        if t.grp[0] is not t: continue  # eval each group once, at its seed
-        grp = t.grp; gV = 0; gC = -Fc_[0]  # membership, compression / grp
-        for t in grp: gV += t.V; gC += t.fc  # add links?
+    for grp in G_:
+        gV = 0; gC = -Fc_[0]  # membership, compression / grp
+        for t in grp: gV += t.V; gC -= t.fc  # add links?  (gC supposes to be begative when we ave-gC below?)
         gV /= 2  # norm for pairwise redundancy
-        if gV > ave-gC: G_ += [[grp,gV,gC]]; V += gV; C += gC
-        else:           G_ += grp
+        if gV > ave-gC: grp[:] = [grp[:],gV,gC]; V += gV; C += gC
```
kwcckw: I restructured the code a bit so that G_ packs only seed now.

frame_2D_alg/meta_code.py
```diff
@@ -52,8 +52,23 @@ def merge(F,f):  # combine aligned ops, if-fork per miss, no inline recursion, f
         return F,f
     else: return F
 
-def cluster_call_(Z):  # cluster Ts if called together, for global Z only?
-
+def cluster_call_():  # cluster Ts if called together, for global Z only?
+
+    global Z
+
+    def comp_sequence(_call_, call_):  # or pack in comp_callers?
```
kwcckw: > Right, but we have to consider the sequence too. It will be similar to merge

So i think it could be simpler than merge, where we just need to sum their weight as M if nFs are the same, else sum weight to D? For the body weight, it should be fc, which is the cost from ast?

frame_2D_alg/meta_code.py
```diff
@@ -62,32 +77,31 @@ def comp_callers(_T, T):
...
+
     if V > ave - C:
         typ_ = []  # form new Z.typ_ for new input, vals added in CoF traced:
         for grp in G_:
-            if isinstance(grp,list):
+            if isinstance(grp[0],list):
                 g,v,c = grp; T=sum2F(g); T.memb=v; T.compr=c  # downward reciprocals of V,C
                 typ_ += [T]
-            else: typ_ += [grp]  # keep old T
+            else: typ_ += grp  # keep old T  (grp is a list of typs, so remove nested [])
```
boris-kz: this is actually a single T, I just used grp as a generic name

frame_2D_alg/meta_code.py
boris-kz: this similarity only matters if we actually merge Ts in grp_oF?

boris-kz: Correct — and that exposes a category error in the eval. sum2F(g) bags the members into G.call_; it doesn’t factor anything. Executing G still walks all k elements, so its real cost is still ~Σfc, not Fc_[0]. So gC = Σfc - Fc_[0] measures a collapse that the clustering never performs.

Co-occurrence clustering can’t compress computation by itself. “Called together” (shared callers) is a locality/selection signal — it’s the HFSort kind of grouping: it tells you which typs to place together, not how to do less work. Grouping by co-occurrence gives you a candidate set, nothing more.

Real compute compression needs the members’ op sequences to be aligned and factored — shared ops kept once, divergences turned into forks. That’s exactly what your other merge(F,f) does over call_ sequences. sum2F sums values; merge factors structure. Only the merged result has a genuinely smaller body, so only then is a sub-Σfc cost real.

So the pipeline is two stages, currently conflated:

	•	cluster_call_ = selection — co-occurrence grouping, eval on cohesion (gV) only. No gC here; it’s not compressing.
	•	merge over each group’s members = compression — aligns op sequences, and it already has its own gain eval (f.fc / C > ave).

Two ways to fix it:

	1.	Drop gC from cluster_call_’s eval (cohesion-only selection), build the composite by folding members with merge instead of sum2F, and let merge’s internal eval decide+measure the actual saving.
	2.	Keep cluster_call_ purely as a grouper that returns candidate groups, and run merge as a separate downstream pass.

Either way, gC only becomes meaningful once the composite is the merged (op-factored) result — not the sum2F bag. As written, the clustering eval is rewarding compression that doesn’t exist.

kwcckw: I see, so we need really `merge` similar operation at the same time. We probably can fold comp_callers in merge?

frame_2D_alg/meta_code.py
kwcckw: Each grp in G is either `[grp,gV,gC]` or the list t added from `grp = [t]`? So typ_ should be merged with [t]?

boris-kz: else:           G_ += grp

that unpacks grp

frame_2D_alg/meta_code.py
boris-kz: more like adding cost compression sequence in comp_callers:

    for i, (Sub,sub) in enumerate(zip(F.call_,f.call_)):  # call_: op sequence in func or block, init F per f?
        fork = []
            fork = Sub; fin=0
            if sub.nF=='E':
                subt = merge(Sub, sub)  # add2F(Sub,ssub_)?
                if isinstance(subt, tuple):
                    sub = subt[1]; C+=sub.fc; call_+=[sub]; add_+=[sub]

merging itself is for final Z in cluster_call_

frame_2D_alg/meta_code.py
kwcckw: Okay, so this is different with my code because my new code doesn't unpack grp into G_. There is just

    if gV > ave-gC: grp[:] = [grp[:],gV,gC]; V += gV; C += gC

without the else section.

boris-kz: how is that better?

kwcckw: With this update all groups will be in G_, so we can skip the additional process to get grp from seed again.

frame_2D_alg/meta_code.py
kwcckw: Okay, so we need something similar as your posted code above, where we walkthrough both calls side by side, but the `merge` should be replaced with the grouping sequence in cluster_call_?

boris-kz: Structure is a graded continuum; every pair partially aligns, so internal similarity is always a meaningful graded number, not a filter. That's a point *for* computing it per pair.

And computing both in `comp_callers` is coherent: external (caller-set overlap) and internal (body alignability) are both graded pairwise quantities, the loop visits every pair anyway. With actual body-merge they play distinct roles — external = how many call sites benefit, internal = how much per site (call-collapse + the body-factor shrink). Both needed; the gain is `|callers|` × `(per-invocation saving)`.

But your #4 — summing pairwise similarities to eval the group merge — is where it breaks. **Neither external nor internal decomposes into a sum of pairwise terms.**

External: the group gain is `(k−1) × |∩ of all k caller sets|` — callers of the *complete* group. The pairwise sum overcounts: a caller present in m of the k members is counted `C(m,2)` times across pairs, but contributes once (or zero, if m<k) to the k-way intersection. Superadditive on shared callers.

Internal: the k-way merge is one aligned skeleton with forks, path-dependent. An op common to m members appears in `C(m,2)` pairwise alignments but is stored/run once in the k-way result. So the pairwise sum overcounts shared ops by the same `C(m,2)` factor.

The error isn't random — it scales with how much is shared, i.e. it's *largest exactly for the cohesive groups you most want to evaluate correctly*. So the pairwise sum systematically overstates the merge gain of good groups.

Resolution — and it's consistent with what you said earlier ("merging itself is for final Z"):
- Pairwise `comp_callers` → **selection only**. It's the connectivity signal that grows the group. Overcounting is fine here; it just biases toward growing, which connectivity clustering wants.
- The **eval** is the actual k-way merge at final-Z. That computes the real external intersection and the real factored body exactly. Accept/reject the composite on *that*, not on `gV`.

One consequence: if the k-way merge is the eval, pairwise *internal* similarity is redundant — the merge computes it exactly, and `n²` partial pairwise merges get discarded. External pairwise (cheap set ops on `caller_`) is enough to grow the group. Keep pairwise internal only if you want a pre-filter to avoid spending k-way merges on groups that co-occur but won't factor — that's a merge-cost tradeoff, your call. But it shouldn't be in the *eval*.

frame_2D_alg/meta_code.py
```diff
@@ -62,32 +77,31 @@ def comp_callers(_T, T):
...
     for _T,T in combinations(_T_,2):
         if _T.grp is T.grp: continue  # already merged
-        v = comp_callers(_T,T)
+        m, d = comp_sequence(_T.call_, T.call_); v = comp_callers(_T,T); v += m/(d or eps)
         if v > ave:
             _g,g = _T.grp,T.grp
             for t in g: t.grp = _g  # repoint T.grp
             _g += g; _T.V += v; T.V += v  # pairwise accum of membership value
-    G_= []
+        if _T.grp not in G_: G_ += [_T.grp]
```
boris-kz: this is packing grps before they are complete and merged, so you have a full duplication?

frame_2D_alg/meta_code.py
boris-kz: > Okay, so we need something similar as your posted code above, where we walkthrough both calls side by side, but the `merge` should be replaced with the grouping sequence in cluster_call_?

yes, but the process should be the same, we just add forks from all group elements, regardless of order?

frame_2D_alg/meta_code.py
boris-kz: Claude:

I read the PR thread (16 comments) and compared the diff against the current `cluster_calls` in `meta_code.py`. Summary of where Chee's update stands relative to Boris's review:

**What Boris asked for, across the thread**

The thread converges on a clear two-stage split (r3265787866, r3266334877, anchor r3266393201):

- pairwise comp (`comp_callers` + body comparison) = **selection only** — connectivity signal that grows groups; overcounting is fine here.
- the actual **k-way merge** at final-Z = **eval** — "add forks from all group elements, regardless of order." Only the merged, op-factored composite has a real cost, so accept/reject must happen on *that*, not on summed pairwise values.

**What the PR actually does**

1. `comp_sequence` — right direction, wrong role. It adds internal (body-alignability) similarity, recursive over nested `call_`, which matches Boris's "compute both in `comp_callers`" point. But it's pure *comparison* — it returns M,D, never produces a factored skeleton with forks. The k-way merge Boris kept asking for (and confirmed in the anchor comment) is still not implemented. Chee did the easy half (internal similarity feeds selection) and skipped the hard half.

2. The core error is **unaddressed**. The PR still does `_T.V += v; T.V += v` then `gV += t.V`, and evals groups on `gV > ave-gC`. That is exactly the pairwise-sum eval Boris said is broken in r3266334877: a caller/op shared by m members is counted C(m,2) times in the pairwise sum but contributes once to the k-way result — superadditive overcount, worst for the cohesive groups you most want to judge. Adding `comp_sequence` into `v` just adds a *second* quantity that doesn't decompose into a sum either. So the PR makes the selection signal richer but leaves the eval wrong.

3. `gC += t.fc` → `gC -= t.fc` — symptom-patching. Boris's actual position is to **drop gC from `cluster_call_`'s eval entirely** (r3265787866: Σfc measures "a collapse the clustering never performs"). Flipping the sign just points the wrong number the other way; the variable shouldn't be in the selection eval at all.

4. In-loop `G_` packing (`if _T.grp not in G_: G_ += [_T.grp]`) — Boris flagged this as duplication (r3266373771) and he's right. `grp` lists mutate and merge during the loop; appending a group reference before all merges complete leaves orphaned absorbed sub-groups in `G_` alongside the live merged group. Chee's defense (r3266331973: "all groups will be in G_, skip getting grp from seed") is wrong — the current seed-based collection (`if t.grp[0] is not t: continue`) is correct *because* it reads final state.

**Minor**

`M += _c.fc * c.c + c.fc * c.c` multiplies both terms by `c.c` — first term should presumably be `_c.fc * _c.c`. And weighting AST cost (`fc`) by `.c` mixes a count into a static cost comparison; `min` would be the cleaner match measure (Chee's own "or use min?" flags this).

**Bottom line**

The PR is a reasonable incremental step on *selection* but does not implement the change Boris actually requested. The eval is still a sum of pairwise terms; it should be the order-independent k-way merge that adds forks from all group members. Until that merge exists, `gV`/`gC` are measuring a compression that never happens.

Want me to sketch the k-way merge (fold members into one fork-skeleton, eval on its real factored cost), or first fix the `G_` duplication so the group set is clean before tackling the merge?

frame_2D_alg/meta_code.py
kwcckw: Yes, in term of group elements, their order shouldn't be matter because we will check their calls side by side anyway? All elements will be checked and merged anyway regardless or their order?

boris-kz: right, after we have an efficient grp. the problem is getting these grps in one pass, we need to normalize for grp size, similar to cluster_N: 

                                iM = sum([L.m for L in _N.rim if (L.nt[0] if L.nt[1] is _N else L.nt[1]) in N_])
                                if iM / (span*decay) < ave * (r-1): continue  # normalized N-to-N_ match, sum for sum2G?

frame_2D_alg/meta_code.py
kwcckw: Okay right, those they maybe merged. Sorry, my edits here are no longer relevant.

frame_2D_alg/meta_code.py
kwcckw: That's mean the comparison and grouping should be limited by their call_ size, weight and cost? 

This ` iM = sum([L.m for L in _N.rim if (L.nt[0] if L.nt[1] is _N else L.nt[1]) in N_])` will be summing their cost, and `if iM / (span*decay) < ave * (r-1): continue` will be the ratio of cost to their group size?

boris-kz: normalized, not limited. not sure on the exact mechanincs

Claude:

Half-right. Bootstrap *must* be pairwise — nothing else exists before a group, granted. But "chicken-and-egg" doesn't license the current structure, which resolves it the *crudest* possible way: pairwise transitive closure fully defines the group (`v>ave` on any pair → union), and the group-wise eval is a pure post-hoc accept/reject. That's single-linkage. Its failure mode is chaining — A~B~C~…~Z all merged because each *pairwise* link passed, even though A and Z share nothing — and that's exactly where the size-bias and the overcounted `gV` come from.

The resolution to chicken-and-egg isn't *ordering* (form, then eval) — it's *interleaving*. And the `cluster_N` snippet you just pasted is itself the example: `iM/(span*decay)` is a group-wise density check, but it runs *inside* accretion, against the *current partial* `N_` — not on a finished group. The cluster is defined *by* the eval-gated growth. There's never a fully-formed group sitting around waiting to be evaluated; there's a running group-vs-candidate check at every accretion step. That's how connectivity clustering escapes the regress.

So for `cluster_call_` the fix is structural, not just a normalization:
- Don't do one transitive union-all pass. Grow incrementally — seed, then add typs in order of pairwise `v` strength.
- Gate each accretion by a check against the *current* group: does the candidate keep the group's current caller-intersection large? does its body still align with the current merged skeleton? Pairwise *proposes* the accretion; the group-wise-against-partial check *accepts* it.
- Then there's no raw pairwise sum to normalize — each step evaluates density against the current group, which never runs away with size.

This is also why your earlier `gV`/`gC` debate was unresolvable: you were trying to make a post-hoc whole-group number do a job that belongs distributed across the growth steps.

The other standard escape is the centroid pass — `cluster_C`: form provisional groups, compute group stats, reassign typs to their best group, iterate to fixpoint. The project already does connectivity-then-centroid for data (`cluster_N` → `cluster_C`); `cluster_call_` currently has neither the incremental gate nor the refinement pass — it's just single-linkage. Mirror the existing two-stage pattern and the chicken-and-egg is handled the same way it already is on the data side.

kwcckw: Okay, and in cluster_N, we normalize it to span * decay, but here for grp size, the max grp size is Z's flatten calls, so we probably can use that?

boris-kz: it's not directly relevant, cluster_N is a floodfill, here the oFs are not local

kwcckw: So we need to use the suggested workflow, where we increase grp size incrementally? From the descriptions above, it's actually very similar with cluster_C, where we compute C first, and then compare and match it with the rest of the candidates.

boris-kz: there some locality via caller co-occurrence, but that seems to give a dimension per caller, not terribly helpful

kwcckw: Okay right that will be extremely complex if we have a new dimension per caller, and impossible to compute their similarity when we have large number of dimensions.

kwcckw: So we can try this approach to grow from a seed at one time? It's just it's not parallelizable.

    - Don't do one transitive union-all pass. Grow incrementally — seed, then add typs in order of pairwise v strength.
    - Gate each accretion by a check against the current group: does the candidate keep the group's current caller-intersection large? does its body still align with the current merged skeleton? Pairwise proposes the accretion; the group-wise-against-partial check accepts it.
    - Then there's no raw pairwise sum to normalize — each step evaluates density against the current group, which never runs away with size.

boris-kz: first we need add comp_body to comp_call, evaluating additional forks as in merge, but without actual merging

boris-kz: Here, mirroring `merge`'s accounting position-by-position but stripping every side effect — no `add2F`, no `call_`/`add_` build-up, no destructive recursion. Recurses through `comp_body` instead of `merge`:

```python
def comp_body(F, f):  # project merge cost without committing; mirror merge accounting, no mutation
    C = 0; cost = Ec_[0]
    for Sub, sub in zip(F.call_, f.call_):
        if Sub.nF == 'E':
            if sub.nF == 'E':
                c = comp_body(Sub, sub)
                if sub.fc / (c or eps) > ave: C += sub.fc; continue  # inner would accept
            for _sub in Sub.call_:
                if _sub.nF == sub.nF: break
            else: C += sub.fc                                   # no matching fork → would add
        elif Sub.nF != sub.nF: C += cost                        # would fork
    return C
```

Hook into `comp_callers` so selection has both signals:

```python
def comp_callers(_T, T):
    olp = _T.caller_ & T.caller_
    off = list(_T.caller_- olp) + list(T.caller_- olp)
    M = sum([c.w * c.c for c in olp]); D = sum([c.w * c.c for c in off])
    v = M / (D or eps)                 # external: caller overlap
    c = comp_body(_T, T)               # internal: projected merge cost
    return v, c
```

Three things worth flagging:

**Tail asymmetry.** `merge` handles length mismatch (`f.call_[i+1:]` etc.) — added to `add_` if f is longer, skipped if F is longer. Mirror left out of `comp_body` above to match merge's `C` accounting (merge doesn't `C += tail*cost`). If you want stricter projection, add `C += abs(len(F.call_)-len(f.call_)) * cost` before the return. Trade-off: faithful-to-merge vs accurate cost.

**`body` vs `call_`.** Per your earlier point, this should walk *bodies*, not `call_`s. As written it follows `merge` and uses `call_`, since `body` isn't separately exposed. When body lands as its own attr, swap `F.call_` → `F.body` in both functions — the walk logic is unchanged.

**Recursion cost.** `comp_body` recurses through every gate at every pairwise comp — so it's not a cheap pre-filter, it does most of merge's traversal work just to project. That's the cost of getting the internal signal during selection. If it becomes prohibitive, two options: (a) limit recursion depth (gate-level only, no sub-gate descent — coarse projection); (b) cache `comp_body(_T,T)` keyed on `(id(_T),id(T))` since the same pair gets compared once per `combinations`.

kwcckw: Okay, and the cost , c here could be merged into v?  Else this comp_body actually can be called outside of comp_callers?

frame_2D_alg/meta_code.py
```diff
@@ -52,8 +52,23 @@ def merge(F,f):  # combine aligned ops, if-fork per miss, no inline recursion, f
...
     else: return F
 
-def cluster_call_(Z):  # cluster Ts if called together, for global Z only?
-
+def cluster_call_():  # cluster Ts if called together, for global Z only?
+
+    global Z
+
+    def comp_sequence(_call_, call_):  # or pack in comp_callers?
+        M = D = 0
```
boris-kz: yes, they should be separate, and comp_body should be on AST, draft:

def comp_body(_n, n):  # project AST-merge cost, recursive, no commit

    if type(_n) is not type(n): return get_ops(_n) + get_ops(n)
    C = 0
    _ch, ch = list(ast.iter_child_nodes(_n)), list(ast.iter_child_nodes(n))
    for a, b in zip(_ch, ch): C += comp_body(a, b)
    for x in _ch[len(ch):] + ch[len(_ch):]: C += get_ops(x)
    return C

boris-kz: I pushed that and Claude draft of complete linkage clustering, haven't reviewed either one

kwcckw: Okay, and cost should be used separately in eval? V > ave * cost? Actually how do we really decide if cost should be multiplied with ave or just ave+cost?

boris-kz: That C is actually cost compression, so it's probably ave/C

PR #150, opened 2026-05-20
5/20/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -56,13 +56,40 @@ def merge(F,f):  # combine aligned ops, if-fork per miss, no inline recursion, f
 def cluster_calls():  # cluster Ts if called together, for global Z only?
 
     def comp_body(_n, n):  # project AST-merge cost, recursive, no commit
+        '''
```
kwcckw, line 59: For comp_body, suppose it should compare the ast primitives of T in sequence. 

So i think we should comp their node sequences even when both Ts' nF are different since they may have the same or aligned nodes?

Also for similar nodes, i think should just skip them? Since their deeper ops will be the same anyway.

For different nodes, right now i'm getting their cost , or it should be more precise to run comp_body recursively on them?

boris-kz: yes, we can add conditional oF node unpacking, if different but the larger sequence is similar

kwcckw: Do you mean add conditional oF node unpacking if most of their sequence is similar but with different nF?

But without oF node unpacking first, we can't know if their sequence is similar?

boris-kz: when most of the outer sequence is similar, the one oF is in. So it will be breadth-first

frame_2D_alg/meta_code.py
```diff
@@ -56,13 +56,40 @@ def merge(F,f):  # combine aligned ops, if-fork per miss, no inline recursion, f
...
         for a, b in zip(_ch, ch): C += comp_body(a, b)
         for x in _ch[len(ch):] + ch[len(_ch):]: C += get_ops(x)
         return C
+        '''
+        
+        def get_node_c(node):  # cost of a single AST node, including onF_
+            if isinstance(node, ast.Call):
+                if isinstance(node.func, ast.Name):      name = node.func.id
+                if isinstance(node.func, ast.Attribute): name = node.func.attr
+                if name in oF_: return get_ops(ast.parse(inspect.getsource(getattr(agg_recursion, oF_[oF_.index(name)]))).body[0])
```
boris-kz, line 73: these names should be indices

kwcckw: Yes thanks, i was thinking too much. It should be just `name` instead of ` oF_[oF_.index(name)]`

frame_2D_alg/meta_code.py
kwcckw: I see, so we need a separate cost set for that, for example, cost_longer, cost_shorter and cost_difference_nF. 

If final cost is high, we need to run comp_body on those pair node with different nF again, and then sum their nested C with the cost_logner and cost_shorter again.?

frame_2D_alg/meta_code.py
boris-kz: I can't believe the mess you pushed here. It's easier for me to do everything from the scratch

frame_2D_alg/meta_code.py
boris-kz: all you need here is the cost of forks required to combine the bodies,
vs the cost of overlap between the two

kwcckw: Okay, then we not really need the recursive operation, we should just get the cost from both nodes here.

frame_2D_alg/meta_code.py
kwcckw: We need this function because a node could be a call to the functions in onF_, so instead of getting their cost from dict `costs`, we need to use `get_ops` to unpack and retrieve the function's cost.

boris-kz: that's a separate function to precompute compexity of all nFs in Z.typ_,
and then put it in the same dict

kwcckw: Okay, but we already have that in init_Wc, it's just those cost are normalized with unit cost vt_, we just need to change the return:

     return [(round(get_ops(funcs[name]) / base), get_ops(funcs[name])) for name in oF_], base


Then:

    fct_, base = init_wc(("agg_recursion.py", "comp_slice.py", "slice_edge.py"))
    fc_, fcn_ = zip(*fct_)
    
To recompute their c when we merged new typ, it will be just `get_ops(get_ast(new_T))`.

boris-kz: yes, and that get_ops should be unpacked

frame_2D_alg/meta_code.py
boris-kz: it can be recursive

kwcckw: > when most of the outer sequence is similar, the one oF is in. So it will be breadth-first

So here actually when the outer sequence is similar, we only will perform recursive operation to find the more refined differences within the pair of nodes?

Which mean we need to pack those nodes pair, eval C for similarity and perform the recursive operation again.

frame_2D_alg/meta_code.py
boris-kz: >we already have that in init_Wc,

we already have Fc_, just replace the vals

boris-kz: def comp_body(_n, n, C=3):  # estimate AST-merge cost compression, init mean C=3, accum from recursive unpack

        if isinstance(n, CoF):
            if isinstance(_n, CoF):
                if n.nF==_n.nF: C+= Fc_[n.nF]
                else:
                    C-=2  # fork compression cost, ast.IfExp
                    if C > 3:  # mean compression value
                        for _sub,sub in zip(_n.body,n.body):  # eval bodies before merging?
                            C = comp_body(_sub, sub, C)
                        if len(_n.body) != len(n.body): C-=2  # single tail fork cost
            else: C-=2
        elif type(_n) is type(n): C+= costs.get(type(n),0)
        else: C-=2
        return C

boris-kz: Same bug as before, just shifted. Default C=3, charge -=2 → C=1, gate `> 3` fails (1 not > 3), no walk. Every top-level pair returns 1.

The problem is gating *after* charging the structural fork: post-charge C is always at least 2 below default, so a gate set at default never passes on the first call. Two consistent options:

**Gate before the charge** — clean semantics, "explore iff at-or-above mean evidence":
```python
            else:
                if C >= 3:  # at/above mean → worth exploring branches
                    C -= 2
                    for _sub, sub in zip(_n.body, n.body): C = comp_body(_sub, sub, C)
                    if len(_n.body) != len(n.body): C -= 2
                else: C -= 2  # fork charged regardless
```

**Or keep gate-after-charge but offset the threshold by the fork cost**:
```python
                if C > 1:  # i.e. C+2 > 3, pre-charge above mean
```
Same effect, less self-documenting — the relation between gate and default is hidden in an offset.

Either way, the default and threshold are linked: if mean changes from 3 to M, both default and gate threshold (`C >= M` or `C > M-2`) must update. Naming it as a single constant would make that explicit and prevent the next drift.

boris-kz: def comp_body(_n, n, C=4):  # estimate AST-merge cost compression, init mean C=3, accum from recursive unpack

        if isinstance(n, CoF):
            if isinstance(_n, CoF):
                if n.nF==_n.nF: C+= Fc_[n.nF]
                elif C > 3:  # mean compression value
                    for _sub,sub in zip(_n.body,n.body):  # eval bodies before merging?
                        C = comp_body(_sub, sub, C)
                    if len(_n.body) != len(n.body): C-=2  # single tail fork cost
                else: C-=2  # fork compression cost, ast.IfExp
            else: C-=2
        elif type(_n) is type(n): C+= costs.get(type(n),0)
        else: C-=2
        return C

frame_2D_alg/meta_code.py
```diff
@@ -145,6 +172,9 @@ def ffeedback(frame):  # adjust filters: all aves *= rV, ultimately differential
 def get_ops(node):
     return sum(costs.get(type(n), 0) for n in ast.walk(node))
 
+def get_ast(T):  # get ast primitives of T, and T.nF must be in onF_
+    return ast.parse(inspect.getsource(getattr(agg_recursion, oF_[T.nF]))).body[0]
+
 def init_wc(paths): # compute weights based on operations in function, independent of deeper callees
```
boris-kz, line 178: def get_Fc_(paths):

    funcs = {}
    for path in paths:
        with open(path, encoding="utf-8") as f:
            tree = ast.parse(f.read(), filename=path)
        for node in ast.walk(tree):
            if isinstance(node, ast.FunctionDef) and node.name in oF_:
                funcs[node.name] = node
    return [round(sum(costs.get(type(p),0) for p in ast.walk(funcs[name]))) for name in oF_]

boris-kz: Yes, it produces a usable proxy for per-function structural complexity — walk_sum tracks manual Fc_ order pretty well (cluster_N highest, cross_comp/ffeedback/slice_edge lowest), just ~40x scaled. Correlation is real, but two things worth knowing:

**Scale.** Outputs are ~40x your hand-set Fc_ values. So either rescale `Fc_` to this range, or divide. If you intend these to interoperate with `Ec_` / `ave`, the scale matters — the absolute number isn't meaningful alone, only ratios.

**Uncounted node types.** Most "uncounted" entries (Load/Name/Store/Constant) are AST wiring with no runtime cost — correctly omitted. `Add`/`Mult`/etc. are operator objects already absorbed into BinOp/AugAssign — also correct. But some real ones are missing from `costs`:Reasonable additions to `costs`: `ast.Return` (1), `ast.Slice` (2), `ast.Starred` (1), `ast.Lambda` (3), `ast.NamedExpr` (2). No Try/With/Yield/DictComp/Raise in oF_ bodies, so the dict isn't significantly underspecified.

**Bottom line:** The output ordering matches your hand-set ordering well enough that you can drop the manual `Fc_` and use this. The 40x scale shift is the only thing that requires attention — either rescale these or rescale anything that compares against them (`Ec_`, gain thresholds).

Want me to also check the ratio comp_N_:comp_C_:comp_N etc. against your manual values to see if any pairs are inverted?

frame_2D_alg/meta_code.py
kwcckw: So the mean C of 3 is set based on the dict `costs` values? 

And now this C is actually the "saving" rather than cost because i can see C is reduced for single tail or type incompatibility cost, while it's incremented for same type or same nF, which makes more sense for Ave/C now.

frame_2D_alg/meta_code.py
kwcckw: This is simpler now, but we don't need vt_ as base unit anymore? Or that should be applied when needed later?

boris-kz: no, base units are from AST. But we need to build Fc_ in the order of composition, first oFs that only contain primitives, etc

frame_2D_alg/meta_code.py
boris-kz: yes, C is just another ave: mean compression across oFs

frame_2D_alg/meta_code.py
boris-kz: funcs = {}  # name → FunctionDef, populated by parse_funcs(paths)

def parse_funcs(paths):

    for path in paths:
        with open(path, encoding="utf-8") as f:
            tree = ast.parse(f.read(), filename=path)
        for node in ast.walk(tree):
            if isinstance(node, ast.FunctionDef) and node.name in oF_:
                funcs[node.name] = node

def add_typ_(R):

    for F in flat_(R): typ_[F.nF] += [F]
    for i, F_ in enumerate(typ_):
        if F_:
            T.N_ = T.call_
            T.fc = Fc_[T.nF]
            typ_[i] = T
    for i, T in enumerate(typ_):  # second pass: static body refs other Ts
        if T: T.body = build_body(funcs[oF_[i]], typ_)
    if any(typ_):
    R.typ_ = typ_

# draft
def build_body(node, typ_):

    out = []
    children = node.body if isinstance(node, ast.FunctionDef) else ast.iter_child_nodes(node)
    for child in children:
        if isinstance(child, ast.Call) and isinstance(child.func, ast.Name) and child.func.id in oF_:
            T = typ_[oF_.index(child.func.id)]
            if T: out += [T]
            continue
        if type(child) in costs: out += [child]
        out += build_body(child, typ_)
    return out

kwcckw: Okay, so instead of manually insert the onF_, we should init it by checking through all the functions in agg_recursion? Because some functions are not in onF_ (like comp_dert, comp, etc) , and they should be added since they are part of the nodes and contains only primitives.

boris-kz: yes, and we can get the Fc_ from Z.typ_

frame_2D_alg/meta_code.py
boris-kz: Got it — one-sided. Compression = cost of `n` saved by merging into `_n`. Restructured:

```python
def comp_body(_n, n, C=0):  # compression of n given _n: cost of n absorbed by merge
    if isinstance(n, CoF) and isinstance(_n, CoF):
        if n is _n: C += Fc_[n.nF]                      # full reuse, save whole body
        elif n.fc > Cmin:                                # only descend if n is non-trivial
            for i, sub in enumerate(n.body):
                if i < len(_n.body): C = comp_body(_n.body[i], sub, C)
                else: C -= mismatch                      # n has tail past _n
        else: C -= mismatch
    elif type(_n) is type(n): C += costs.get(type(n),0)
    else: C -= mismatch
    return C
```

Changes vs original:

- **`C=0` init.** Compression starts at zero; positive means savings. Magic-number init dropped.
- **Threshold is local (`n.fc > Cmin`), not path-accumulated.** Whether to descend depends on whether *n* itself is worth analyzing, not on what came before. Names: `Cmin` (descent threshold), `mismatch` (fork penalty).
- **Iterate over `n.body`, not zip.** zip drops the tail; we need it because tail-of-n is new code added to _n.
- **One-sided length penalty.** Only `n` longer costs (`n` has unmatched trailing ops). `_n` longer is free — n is a subset, _n keeps its extras at no cost to n.
- **`n is _n` instead of `n.nF == _n.nF`.** Same T = same object since `body` holds refs to `Z.typ_[i]`. Cleaner.

**Three things to flag:**

**A. Asymmetry now matters for `cluster_calls`.** `comp_body(_T, T) ≠ comp_body(T, _T)`. Pick one of:
- `min(comp_body(a,b), comp_body(b,a))` — conservative; worst-case compression
- average — symmetric proxy
- pick direction by convention (e.g. larger-fc as `_n`, smaller as `n` → "absorb small into large")

**B. Fork cost when descending into diff-nF CoFs.** Currently if `n.fc > Cmin` we descend but never charge for the fork itself. Two diff-nF Ts being merged requires a runtime dispatch (an `if`/branch in merged code). Probably want:
```python
elif n.fc > Cmin:
    C -= mismatch  # fork dispatch cost
    for i, sub in ...
```

**C. `Fc_[n.nF]` on full-reuse is the callee's *whole* body cost.** That's the right magnitude for "saved by reuse". But `n.fc` is already that value (`T.fc = Fc_[T.nF]` in add_typ_). So:
```python
if n is _n: C += n.fc
```
is equivalent and doesn't rely on `Fc_` being in scope here.

Want me to wire `c` into `cluster_calls`'s gate now (current code discards it)?

frame_2D_alg/meta_code.py
kwcckw: Okay, and i think we need to pack their cost in every call? Because we might split or merge them later, and we can just retrieve them in each call to sum the cost into the group. We can't always retrieve the cost from global Fc_ because global Fc_ might contain only grouped typ in subsequent frames, where it's no longer contain the merged function's cost.

frame_2D_alg/meta_code.py
boris-kz: Not sure this is better, the whole cluster_calls is not fully revised

frame_2D_alg/meta_code.py
boris-kz: split or merge is per frame_HH, pretty far from per call
Fc_ is fine, we just need to fill it

frame_2D_alg/meta_code.py
kwcckw: The code above uses mismatch, that's a constant? It should be the cost of forking, which is ` ast.IfExp: 2`?

Also it may get negative C in this scheme, which means there's more cost in merging forks, so we are no able to use Ave/C anymore. We need to add eval so that when C is negative, it skips the merging?

boris-kz: Yes, 2. For eval, it's V+C > ave. Maybe C*coef

kwcckw: Yes, i think we need another coef, something like an ave? Because native C values are generally very large, we need to scale it down to the ave level.

PR #151, opened 2026-05-21
5/21/2026

kwcckw:


frame_2D_alg/meta_data.py
```diff
@@ -0,0 +1,26 @@
+from copy import copy, deepcopy
+import numpy as np
```
kwcckw, line 2: So there's a cyclic import issue, and i think the best solution now is to create another file to import it from both agg_recursion and meta_code.

frame_2D_alg/meta_code.py
```diff
@@ -221,11 +222,13 @@ def build_body(node, typ_):
             if T: out += [T]
             continue
         if type(child) in costs: out += [child]
-        out += build_body(child, typ_)
+        out += build_body(child, typ_)  # this should be nested? Now it's packed flat
```
kwcckw, line 225: Right now the body is packed flat with the node's body, and i think it should be nested to preserve the hierarchy? Else we actually can't really know if the elements within `out` is from child or not.

frame_2D_alg/meta_code.py
```diff
@@ -188,12 +179,22 @@ def get_Fc_(paths): # compute weights based on operations in function, independe
...
     for path in paths:
         with open(path, encoding="utf-8") as f:
             tree = ast.parse(f.read(), filename=path)
-        for node in ast.walk(tree):
-            if isinstance(node, ast.FunctionDef) and node.name in oF_:
+        for node in tree.body:  # only function definitions, skips those class def
+            if isinstance(node, ast.FunctionDef):
                 funcs[node.name] = node
+    # replaces oF_ with all func names
+    oF_ = list(funcs.keys())
```
kwcckw, line 190: So i edited this func to pack the function under `def` only, so that it skips those funcs under classes. Also it should replaces the original oF_?

frame_2D_alg/meta_code.py
```diff
@@ -221,11 +222,13 @@ def build_body(node, typ_):
...
             continue
         if type(child) in costs: out += [child]
-        out += build_body(child, typ_)
+        out += build_body(child, typ_)  # this should be nested? Now it's packed flat
     return out
 
 if __name__ == "__main__":
     # move to agg_recursion
+    parse_funcs(["agg_recursion.py"])  # populate funcs
+    # we actually should build those Fc_, Ec_ using the func here?
```
kwcckw, line 231: In actual workflow, after we get the new oF_ from parse_funcs, we should rebuild those Fc_, Ec_, Fw_, FTT_, again because they need to be aligned with the new oF_?

Then the build_body should be static and it should be run here to get body of all possible oF_, rather than typ_ since the existing typ_ is obtained via runtime, which may not contain all functions.

frame_2D_alg/agg_recursion.py
```diff
@@ -7,7 +7,15 @@
 from frame_blobs import frame_blobs_root, imread, comp_pixel, CBase
 from slice_edge import slice_edge
 from comp_slice import comp_slice, w_t
-from meta_code  import oF_,gF_,Fw_,Fc_,FTT_, Ew_,Ec_,ETT_
+from meta_data import (
```
boris-kz, line 10: I don't want to keep jumping across files

frame_2D_alg/meta_data.py
boris-kz: meta_code doesn't need much from agg_recursion, we just need to move class defs etc here.
as we discussed

frame_2D_alg/meta_code.py
boris-kz: yes

frame_2D_alg/meta_code.py
boris-kz: only the primitives can be nested. but yes, I guess we do need hierarchy to align them

frame_2D_alg/agg_recursion.py
```diff
@@ -633,7 +635,10 @@ def sum2F(N_, root=None, m_=[],d_=[], merge=0, froot=0):  # -> CF/CL/CC/CN/CoF
         if typ==2: F.m_,F.d_=m_,d_; F.m,F.d=sum(m_),sum(d_)
         else:      F.m, F.d = vt_(TT)
         if typ==3: F.box = box
-    elif fO: F.call_ = n_; F.fw = Fw; F.m, F.d = vt_(TT)
+    elif fO: 
+        F.call_ = n_; F.fw = Fw; F.m, F.d = vt_(TT)
+        # add merging of body here?
```
boris-kz, line 640: merging is for summing across typs? why do you need it?

frame_2D_alg/meta_code.py
boris-kz: >it skips those funcs under classes

I don't get this?

frame_2D_alg/meta_code.py
boris-kz: Right. Flat pre-order can align positions by accident (sibling count matches per parent) but loses the "this branch has no counterpart" signal — `If` vs `For` should fork once, not produce N noisy leaf mismatches.

**Body as a tree.** Each entry is one of three kinds:

- **T ref** (CoF) — terminal, opaque; comp_body recurses through its own `.body`
- **`(ast_type, [children])`** — composite node, substitution happens inside
- **bare `ast` node** — costed leaf (BinOp, Compare, etc., no interesting children left after filtering)

**build_body, recursive:**

```python
def build_body(node, typ_):  # returns T | (type, [children]) | node | None
    if isinstance(node, ast.Call) and isinstance(node.func, ast.Name) and node.func.id in oF_:
        T = typ_[oF_.index(node.func.id)]
        if T: return T
    children = [r for c in ast.iter_child_nodes(node) if (r := build_body(c, typ_)) is not None]
    if children: return (type(node), children)
    if type(node) in costs: return node
    return None
```

Top of `add_typ_` second pass:
```python
T.body = [r for c in funcs[oF_[i]].body if (r := build_body(c, typ_)) is not None]
```

**comp_body updated to three kinds:**

```python
def comp_body(_n, n, C=0):
    if isinstance(n, CoF) and isinstance(_n, CoF):
        if n is _n: C += n.fc
        elif n.fc > Cmin:
            C -= mismatch
            for i, sub in enumerate(n.body):
                if i < len(_n.body): C = comp_body(_n.body[i], sub, C)
                else: C -= mismatch
        else: C -= mismatch
    elif isinstance(n, tuple) and isinstance(_n, tuple) and n[0] is _n[0]:
        C += costs.get(n[0], 0); C -= mismatch  # fork cost for descending
        for i, sub in enumerate(n[1]):
            if i < len(_n[1]): C = comp_body(_n[1][i], sub, C)
            else: C -= mismatch
    elif type(_n) is type(n) and isinstance(n, ast.AST):
        C += costs.get(type(n), 0)
    else: C -= mismatch
    return C
```

**Notes:**

- `(ast.If, [...])` vs `(ast.For, [...])` → falls to mismatch (`n[0] is _n[0]` fails). One fork, correct.
- `(ast.If, [a,b,c])` vs `(ast.If, [a,b,c,d])` → match type, recurse on first 3 pairs, tail `d` adds mismatch. Correct.
- Bare leaf branch only fires for nodes that had no surviving children — leaves like `Compare`, `BinOp` with all-Constant/Name args.
- `Cmin` threshold uses `n.fc`, not the tuple-level analogue. If a composite's recursion is so cheap that we want to gate it, add a size check on `n[1]`. Probably not needed.
- The composite `mismatch` charge on every descent might over-penalize. Currently I charge once per composite-match (fork-to-enter). If you want only top-level descent to charge, lift it out and only charge at CoF→body descent. Depends on whether each branching ast node represents a real merge fork or just structural sharing.

One open question: **should the fork charge `mismatch` on every composite descent, or only on CoF descent?** The composite descent is "we're inside a matched control structure, peeking at sub-ops"; that's not really a runtime branch. Probably:

```python
elif isinstance(n, tuple) and isinstance(_n, tuple) and n[0] is _n[0]:
    C += costs.get(n[0], 0)
    for i, sub in enumerate(n[1]):
        if i < len(_n[1]): C = comp_body(_n[1][i], sub, C)
        else: C -= mismatch
```

(no `-= mismatch` on descent into matched composite). Fork charge only on CoF→body descent (and on type mismatches).

Which fork model?

frame_2D_alg/meta_data.py
kwcckw: Okay, so everything in agg_recursion should be moved to meta_code? Which we not really need it later if we move everything to meta_code?

frame_2D_alg/meta_code.py
kwcckw: The existing code add those def under class def too, for example, `__init__` and `__bool__`, so we should skip those.

frame_2D_alg/agg_recursion.py
kwcckw: We need to sum their body when we sum group into typ, within cluster_calls?

        if isinstance(grp,list):
            g,v,c = grp; T=sum2F(g); T.memb=v; T.cmpr=c  # downward reciprocals of V,C

frame_2D_alg/meta_code.py
kwcckw: So we can use this?
        
        T ref (CoF) — terminal, opaque; comp_body recurses through its own .body
        (ast_type, [children]) — composite node, substitution happens inside
        bare ast node — costed leaf (BinOp, Compare, etc., no interesting children left after filtering)
        
For children, it could be simplified using a nested list ? I'm not sure if we need the ast_type.

frame_2D_alg/agg_recursion.py
boris-kz: that's within a type, so the body is the same, nothing to merge

frame_2D_alg/meta_data.py
boris-kz: class defs, constants, and operations on them: that's the definition of meta code.
active code (functions) stays in agg_recursion
we discussed that

frame_2D_alg/meta_code.py
boris-kz: each fork and node in that nested list is an ast primitive, that's ast_type?

frame_2D_alg/meta_code.py
boris-kz: ok right

frame_2D_alg/meta_data.py
kwcckw: Okay, thanks for clarifying again.

So this is actually still cyclic because we need to import functions from agg_recursion to meta code, while constants, etc from meta_code to agg_recursion?

frame_2D_alg/meta_code.py
kwcckw: Yes, it's an unpacked list of child's primitives, so that `ast_type` in `(ast_type, [children])` is not really needed.

frame_2D_alg/agg_recursion.py
kwcckw: For cluster_calls, each element in grp is a typ (we init t.grp=[t]), where their body should be different?

    for t in Z.typ_:
        if t: t.grp = [t]; t.caller_ = set(t.caller_); t.V = 0; _T_ += [t]
        v,c = comp_callers(_T,T), comp_body(_T,T)
        pairs += [[v,c,_T,T]]

boris-kz: that's for the next frame_H, so the data should stay empty

frame_2D_alg/meta_data.py
boris-kz: it's fine, modified functions are for next instance of agg_recursion, they won't affect current instance

frame_2D_alg/meta_code.py
boris-kz: ok so

    def comp_body(_n, n, C=0):  # estimated n-merge cost compression, init mean C=3, accum from recursive unpack

        if isinstance(n, CoF):
            if isinstance(_n, CoF):
                if n is _n: C += n.fc
                elif n.fc > ave_C:  # mean compression value
                    C -= 2  # fork compression cost, ast.IfExp
                    for _sub, sub in zip(_n.body, n.body): C = comp_body(_sub, sub, C)
                    if len(n.body) > len(_n.body): C -= 2  # single tail fork cost
                else: C -= 2
            else: C -= 2
        elif isinstance(n, tuple):  # AST fork
            if isinstance(_n, tuple) and n[0] is _n[0]:
                C += costs.get(n[0], 0)
                for _sub, sub in zip(_n[1], n[1]): C = comp_body(_sub, sub, C)
                if len(n[1]) > len(_n[1]): C -= 2
            else: C -= 2
        elif type(_n) is type(n): C += costs.get(type(n),0)  # AST leaf
        else: C -= 2
        return C
?

frame_2D_alg/meta_code.py
boris-kz: actually Fc_ is redundunt to oF.fc?

frame_2D_alg/agg_recursion.py
kwcckw: This is for the next frame_H, but body is static and it's from the code instead of the data? We may use sum2F to sum group into new typ:

     g,v,c = grp; T=sum2F(g); T.memb=v; T.cmpr=c  # downward reciprocals of V,C

frame_2D_alg/meta_code.py
kwcckw: Yes, they should be the same. We can move it to Fc_ for consistency, since we access Fw via the global too.

frame_2D_alg/meta_code.py
kwcckw: Not sure if ` n.fc > ave_C` is the better way because we may need different ave for different typ n?

frame_2D_alg/agg_recursion.py
boris-kz: that's before that new typ is used in active code, in the next instance of agg-recursion

frame_2D_alg/meta_code.py
boris-kz: more like ave_C * n.fc

frame_2D_alg/agg_recursion.py
kwcckw: Yes, so do you mean in the next instance of agg_recursion, we should recompute their body instead of merge them?

frame_2D_alg/meta_code.py
kwcckw: > more like ave_C * n.fc

Okay, this makes more sense so ave_C will be a multiplier instead.

frame_2D_alg/agg_recursion.py
boris-kz: the body is merged, the data in merged oF is recomputed from new inputs

frame_2D_alg/meta_code.py
boris-kz: just use oF_[i].fc

frame_2D_alg/meta_code.py
boris-kz: Fname_ = {}  # name → FunctionDef

def parse_funcs(paths):

    for path in paths:
        with open(path, encoding="utf-8") as f:
            tree = ast.parse(f.read(), filename=path)
        for node in tree.body:  # only top-level function definitions
            if isinstance(node, ast.FunctionDef):
                Fname_[node.name] = node

def add_typ_(R):

    def build_body(node):
        if isinstance(node, ast.Call) and isinstance(node.func, ast.Name) and node.func.id in oF_:
            T = typ_[oF_.index(node.func.id)]
            if T: return T
        sub_ = [r for t in ast.iter_child_nodes(node) if (r := build_body(t)) is not None]
        if sub_: return (type(node), sub_)
        if type(node) in costs: return node

    def set_fc(body):
        fc = 0
        for n in body:
            if isinstance(n, CoF): fc += 3
            elif isinstance(n, tuple): fc += costs.get(n[0],0) + set_fc(n[1])
            else: fc += costs.get(type(n),0)
        return fc

    for F in flat_(R): typ_[F.nF] += [F]
    for i, F_ in enumerate(typ_):
        if F_:
            T.N_ = T.call_
            typ_[i] = T
    for i, T in enumerate(typ_):  # second pass: static body refs other Ts
        if T:
            T.body = build_body(Fname_[oF_[i]])
            T.fc = set_fc(T.body)
    if any(typ_):
    R.typ_ = typ_

kwcckw: This function should get all function names first, and then only compute their fc.

Right now this function is getting function name and the function's ast node, so we need to add build_body into this line:

    funcs[node.name] = node
    
into:

    funcs[node.name] = (node, build_body(node))

Then the funcs will be a dict, mapping each function name to their ast node and fc, something like this: `funcs['cross_comp'] =  [node, fc]`

This should be static so we do not have CoF instance for the fc assignment yet. Also same changes need to be done in build_body since that's actually designed for runtime typ_.

boris-kz: I think it's fine to have it in add_typ_, I posted edits above

kwcckw: Okay, ideally this `parse_funcs` should be run first to init all func's costs and weights, and then agg_recursion only will be executed? Or we stick with the current manual initialization of oF_ first?

parse_funcs -> init cost|weight -> agg_recursion -> add_typ_ -> cluster_call -> merge -> repeat?

boris-kz: That would run add_typ_ on module AST instead of Z.call?

kwcckw: Yes, and the input will be the function nodes, added in `parse_funcs` above:

      def parse_funcs(paths):
            
            for path in paths:
                with open(path, encoding="utf-8") as f:
                    tree = ast.parse(f.read(), filename=path)
                for node in tree.body:  # only top-level function definitions
                    if isinstance(node, ast.FunctionDef):
                        Fname_[node.name] = node

Each node is the typ body, so we can use them to generate cost and build the costs and weights.

boris-kz: so

Module-level (`meta_code.py`):

```python
Fname_ = {}  # name → FunctionDef
typ_   = []  # parallel to oF_, populated by add_typ_

def parse_funcs(paths):
    for path in paths:
        with open(path, encoding="utf-8") as f:
            tree = ast.parse(f.read(), filename=path)
        for node in tree.body:
            if isinstance(node, ast.FunctionDef):
                Fname_[node.name] = node

def build_body(node):
    if isinstance(node, ast.Call) and isinstance(node.func, ast.Name) and node.func.id in oF_:
        T = typ_[oF_.index(node.func.id)]
        if T is not None: return T
    sub_ = [r for t in ast.iter_child_nodes(node) if (r := build_body(t)) is not None]
    if sub_: return (type(node), sub_)
    if type(node) in costs: return node

def set_fc(body):
    fc = 0
    for n in body:
        if isinstance(n, CoF): fc += 3
        elif isinstance(n, tuple): fc += costs.get(n[0],0) + set_fc(n[1])
        else: fc += costs.get(type(n),0)
    return fc

def add_typ_():  # static, one-time after parse_funcs
    typ_.clear(); typ_.extend([None] * len(oF_))
    for i, name in enumerate(oF_):
        if name in Fname_:
            T = CoF(nF=i); T.caller_ = []; typ_[i] = T
    for i, T in enumerate(typ_):
        if T is not None:
            T.body = build_body(Fname_[oF_[i]])
            T.fc = set_fc(T.body)
```

In `CoF.traced` (agg_recursion.py), add accumulation after the existing post-call block:

```python
@staticmethod
def traced(func):
    @wraps(func)
    def inner(*a, **kw):
        _CoF = CoF._cur.get()
        oF = CoF(nF=oF_.index(func.__name__), root=_CoF)
        oF.wTT = FTT_[oF.nF]; oF.fw = Fw_[oF.nF]; oF.fc = Fc_[oF.nF]; _CoF.call_ += [oF]
        if oF.call_:
            if oF.fw*L > ave*(oF.fc*L):
                sum2F(tree, oF); wtt = getattr(oF,'rTT',oF.dTT); oF.wTT = cent_TT(wtt,oF.r)
                oF.w += np.mean(oF.wTT)
        import meta_code  # deferred to avoid cycle
        if meta_code.typ_:
            T = meta_code.typ_[oF.nF]
            if T is not None:
                add2F(T, oF, merge=2)
                if isinstance(_CoF.nF, int):
                    _T = meta_code.typ_[_CoF.nF]
                    if _T is not None and _T not in T.caller_: T.caller_.append(_T)
        CoF._cur.set(_CoF)
        return out
    inner.wrapped = True
    return inner
```

Call-site change at line 1020:
- `add_typ_(Z)` → `add_typ_()`, and move it out of the loop to a one-time call near startup (after `parse_funcs([...])`). Inside the loop it just thrashes typ_, wiping accumulated runtime aggregates.

Three notes:

1. **`typ_.clear()` + `extend()` instead of reassignment.** Keeps the module binding stable so deferred imports / cross-module references stay valid.
2. **Subtree-fold ordering.** The accumulation happens *after* the conditional `sum2F(tree, oF)` block, so if folding fired, T accumulates the folded subtree, not the raw oF. If you want T to track raw per-call data only, move the accumulation block before the fold.
3. **`Z` skip.** `_CoF.nF` is the string `'Z'` for the global Z root, so `isinstance(_CoF.nF, int)` filters it out — Z isn't recorded as a caller. Correct.

boris-kz: make `oF_` itself the list of Ts with names attached:

```python
class CoF(CF):
    def __init__(f, **kw):
        super().__init__(**kw)
        f.name = kw.get('name','')
        f.call_ = kw.get('call_',[])
        f.body  = kw.get('body',[])
        f.fw,f.fc,f.fr = [kw.get(x,0) for x in ('fw','fc','fr')]
```

`meta_code.py`:

```python
_names = ['comp_N_','comp_C_','comp_N','comp_F','get_exemplars','cluster_N','cluster_C','cluster_P',
          'cross_comp','frame_H','vect_edge','trace_edge','ffeedback','proj_N','comp_slice','slice_edge']
oF_ = [CoF(nF=i, name=n) for i,n in enumerate(_names)]
oF_idx = {T.name: T.nF for T in oF_}  # name → position, used at trace time

def add_typ_():  # static, one-time after parse_funcs
    for T in oF_:
        if T.name in Fname_:
            T.caller_ = []
            T.body = build_body(Fname_[T.name])
            T.fc = set_fc(T.body)

def build_body(node):
    if isinstance(node, ast.Call) and isinstance(node.func, ast.Name) and node.func.id in oF_idx:
        T = oF_[oF_idx[node.func.id]]
        if T.body: return T  # populated check
    sub_ = [r for t in ast.iter_child_nodes(node) if (r := build_body(t)) is not None]
    if sub_: return (type(node), sub_)
    if type(node) in costs: return node
```

Call-site changes:

- `oF_.index(func.__name__)` → `oF_idx[func.__name__]`
- `name in oF_` → `name in oF_idx`
- `typ_[i]` → `oF_[i]`
- The CoF.traced accumulation block becomes:
  ```python
  T = oF_[oF.nF]
  if T.body:
      add2F(T, oF, merge=2)
      if isinstance(_CoF.nF, int):
          _T = oF_[_CoF.nF]
          if _T.body and _T not in T.caller_: T.caller_.append(_T)
  ```

**Two concerns:**

1. **"Populated" check.** Was `typ_[i] is not None`; now `T.body` (truthy iff non-empty list). Safe unless a real function legitimately produces empty body (no costed ops + no oF_ calls — unlikely for any real function in oF_, but check). Could use `T.fc` instead, or add an explicit flag.

2. **oF_idx duplicates `_names`.** Could derive from oF_ as shown, but if any name has nF mismatch with position (shouldn't happen here), it'd silently break. Minor.

Net effect: `typ_` variable disappears; all references go through `oF_`. The cost is the `oF_idx` dict for fast name lookup at trace time. Worth it — name lookup is hot path.

frame_2D_alg/meta_code.py
```diff
@@ -188,12 +179,22 @@ def get_Fc_(paths): # compute weights based on operations in function, independe
...
         with open(path, encoding="utf-8") as f:
             tree = ast.parse(f.read(), filename=path)
-        for node in ast.walk(tree):
-            if isinstance(node, ast.FunctionDef) and node.name in oF_:
+        for node in tree.body:  # only function definitions, skips those class def
+            if isinstance(node, ast.FunctionDef):
                 funcs[node.name] = node
+    # replaces oF_ with all func names
+    oF_ = list(funcs.keys())
+    # ['eps_', 'prop_F_', 'flat_', 'vt_', 'sum_vt', 'cross_comp',
```
boris-kz, line 191: Right — `nF` is the canonical identifier; `name` was redundant. Use nF everywhere; only need name-string lookup at the two static touchpoints (trace decoration + AST parse).

```python
_names = ['comp_N_','comp_C_','comp_N','comp_F','get_exemplars','cluster_N','cluster_C','cluster_P',
          'cross_comp','frame_H','vect_edge','trace_edge','ffeedback','proj_N','comp_slice','slice_edge']
oF_     = [CoF(nF=i) for i in range(len(_names))]
oF_idx  = {n: i for i,n in enumerate(_names)}      # name → nF, static
Fname_  = [None] * len(_names)                      # FunctionDef per nF

def parse_funcs(paths):
    for path in paths:
        with open(path, encoding="utf-8") as f:
            tree = ast.parse(f.read(), filename=path)
        for node in tree.body:
            if isinstance(node, ast.FunctionDef) and node.name in oF_idx:
                Fname_[oF_idx[node.name]] = node

def add_typ_():
    for i, T in enumerate(oF_):
        if Fname_[i] is not None:
            T.caller_ = []
            T.body = build_body(Fname_[i])
            T.fc = set_fc(T.body)

def build_body(node):
        i = oF_idx.get(node.func.id)
        if i is not None and oF_[i].body: return oF_[i]
    sub_ = [r for t in ast.iter_child_nodes(node) if (r := build_body(t)) is not None]
    if sub_: return (type(node), sub_)
    if type(node) in costs: return node
```

CoF stays as-is (no `name` attr). `CoF.traced` captures nF once at decoration time:

```python
@staticmethod
def traced(func):
    nF = oF_idx[func.__name__]  # decoration time, one lookup
    @wraps(func)
    def inner(*a, **kw):
        _CoF = CoF._cur.get()
        oF = CoF(nF=nF, root=_CoF)
        oF.wTT = FTT_[nF]; oF.fw = Fw_[nF]; oF.fc = Fc_[nF]; _CoF.call_ += [oF]
        if oF.call_:
            if oF.fw*L > ave*(oF.fc*L):
                sum2F(tree, oF); wtt = getattr(oF,'rTT',oF.dTT); oF.wTT = cent_TT(wtt,oF.r)
                oF.w += np.mean(oF.wTT)
        T = oF_[nF]
        if T.body:
            add2F(T, oF, merge=2)
            if isinstance(_CoF.nF, int):
                _T = oF_[_CoF.nF]
                if _T.body and _T not in T.caller_: T.caller_.append(_T)
        CoF._cur.set(_CoF)
        return out
    inner.wrapped = True
    return inner
```

Net: `typ_` and `name`-on-CoF both gone. Only `oF_idx` (16-entry dict) survives as the string-lookup bridge, used only at module-load and during AST parse.

frame_2D_alg/meta_code.py
kwcckw: > make oF_ itself the list of Ts with names attached:

We probably can try this, but the names should be generated from parse_funcs. Then oF_ is a global typ_, and we can skip it in Z.

boris-kz: right, and we don't actually need Z?
also don't need 
Fc_ = [13,11,18,5,4,22,20,12,3,13,14,11,3,5,4,3]; cN_,cC_,cN,cF, cE,ccN,ccC,ccP, cX,cFrm,cVct,cTrc,cBac,cPrj,cCS,cSE = Fc_
Fw_ = copy(Fc_); wN_,wC_,wN,wF, wE,wcN,wcC,wcP, wX,wFrm,wVct,wTrc,wBac,wPrj,wCS,wSE = Fw_  # ave gain/call, init = cost
FTT_= [deepcopy(wTT) for _ in range(16)]; ttN_,ttC_,ttN,ttF, ttE,ttcN,ttcC,ttcP, ttX,ttFrm,ttVct,ttTrc,ttBac,ttPrj,ttCs,ttSE = FTT_


all that can be accessed via oF_

boris-kz: Mixed answer:

**Fc_, Fw_, FTT_: yes, redundant.** Each `oF_[i]` is a T with `.fc`, `.fw`, `.wTT` attrs already. Replace:
- Line 146: `oF.wTT = FTT_[nF]; oF.fw = Fw_[nF]; oF.fc = Fc_[nF]` → `T = oF_[nF]; oF.wTT = T.wTT; oF.fw = T.fw; oF.fc = T.fc`
- Line 1022: `Fw_[i] = tF.fw/tF.c` → `oF_[i].fw /= oF_[i].c` (in-place; `tF` *is* `oF_[i]` now)
- Line 1023 vectorized scaling: loop over oF_ instead of `np.array(Fw_) / rV`.

**Z: no, not redundant.** Z is the *runtime* context root — where top-level traced calls register their parent via `CoF._cur`. oF_ entries are *type* representatives. Z accumulates a runtime call tree per frame; oF_[i] accumulates aggregates across all calls of type i. Different roles, both needed.

**Two snags worth flagging before you consolidate Fc_/Fw_/FTT_:**

1. **Named-scalar unpacks at lines 14-16, 58-60.** `cN_, cC_, ... = Fc_` etc. give you scalar handles used throughout agg_recursion. If you drop the arrays, these unpacks need rewriting:
   ```python
   cN_,cC_,cN,cF, cE,ccN,ccC,ccP, cX,cFrm,cVct,cTrc,cBac,cPrj,cCS,cSE = [T.fc for T in oF_]
   ```
   But this only works *after* add_typ_ runs (T.fc is 0 before). Currently the unpacks happen at module load with the hard-coded list. Either:
   - Keep `_Fc_ = [13,11,18,...]` as a local-only init constant, do unpack from it, then push into Ts in add_typ_. Drop module-level Fc_ name; named scalars remain.
   - Or move the named-scalar unpacks out of module-load into a function called after add_typ_.

2. **Vectorized scaling** (line 1023): `np.array(Fw_) / rV`. With oF_ as the source, this becomes a loop:
   ```python
   for T in oF_:
       if T.body: T.fw /= rV; T.wTT /= rV
   ave/=rV; avd/=rV
   ```
   Minor cost — 16 entries, not vectorization-critical.

Net: drop Fc_/Fw_/FTT_ from meta_code, replace with attrs on oF_[i]. Keep named scalars by unpacking from a local constant list at meta_code init, separate from the arrays. Z stays.

Confirm you want to do this, or hold off?

kwcckw: > right, and we don't actually need Z? also don't need Fc_ = [13,11,18,5,4,22,20,12,3,13,14,11,3,5,4,3]; cN_,cC_,cN,cF, cE,ccN,ccC,ccP, cX,cFrm,cVct,cTrc,cBac,cPrj,cCS,cSE = Fc_ Fw_ = copy(Fc_); wN_,wC_,wN,wF, wE,wcN,wcC,wcP, wX,wFrm,wVct,wTrc,wBac,wPrj,wCS,wSE = Fw_ # ave gain/call, init = cost FTT_= [deepcopy(wTT) for _ in range(16)]; ttN_,ttC_,ttN,ttF, ttE,ttcN,ttcC,ttcP, ttX,ttFrm,ttVct,ttTrc,ttBac,ttPrj,ttCs,ttSE = FTT_
> 
> all that can be accessed via oF_

Okay, this is actually better, but for Z, we still need it to pack the runtime call tree?

boris-kz: I think so

frame_2D_alg/meta_code.py
```diff
@@ -115,12 +105,13 @@ def comp_callers(_T, T):
...
     # draft: complete-linkage agglomeration on pairs
-    qual = {frozenset({p[2], p[3]}) for p in pairs if p[0] > ave}  # qualifying pairs
+    qual = {frozenset({p[2], p[3]}) for p in pairs if p[0]+p[1] > ave}  # qualifying pairs
     pairs.sort(key=lambda p: -p[0])  # best v first
     for v, c, _T, T in pairs:
         if v <= ave: break
         if _T.grp is T.grp: continue
-        if all(frozenset({a, b}) in qual for a in _T.grp for b in T.grp):
+        match = sum([1 for a in _T.grp for b in T.grp if frozenset({a, b}) in qual])
+        if match/len(T.grp)*len(T.grp) > 0.75:  # using all is too strict?
```
boris-kz, line 114: I haven't checked that yet

kwcckw: Okay, we probably can come back to this after the oF_structure is finalized.

PR #152, opened 2026-05-22
5/22/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -143,8 +145,9 @@ def traced(func):
         @wraps(func)
         def inner(*a, **kw):
             _CoF = CoF._cur.get()
-            oF = CoF(nF=oF_.index(func.__name__), root=_CoF)
-            oF.wTT = FTT_[oF.nF]; oF.fw = Fw_[oF.nF]; oF.fc = Fc_[oF.nF]; _CoF.call_ += [oF]
+            oF = CoF(nF=iF_[func.__name__], root=_CoF)
+            # we need fw per oF now? add that in add_typ_?
+            oF.wTT = FTT_[oF.nF]; oF.fw = oF_[oF.nF].fc; oF.fc = oF_[oF.nF].fc; _CoF.call_ += [oF]
```
kwcckw, line 150: So local call's fc and fw should be updated from global oF_ now, which mean we need to add fw into global oF too? They should be init from add_typ_:

      def add_typ_():
          for i, T in enumerate(oF_):
              if nF_[i] is not None:
                  T.caller_ = []
                  T.body = build_body(nF_[i])
                  T.fc = set_fc([T.body])

frame_2D_alg/meta_code.py
```diff
@@ -164,7 +164,7 @@ def add_typ_():
         if nF_[i] is not None:
             T.caller_ = []
             T.body = build_body(nF_[i])
-            T.fc = set_fc(T.body)
+            T.fc = set_fc([T.body])  # body is not iterable but we iter it in set_fc, so make it a list for the first run?
```
kwcckw, line 167: So the returned body is not iterable because it is either a CoF, node or (type, sub), and we need to convert it into list here so that set_fc is able to loop them and also the deeper sub in the same way:

      def set_fc(body):
          fc = 0
          for n in body:
              if isinstance(n, CoF): fc += 3
              elif isinstance(n, tuple): fc += costs.get(n[0],0) + set_fc(n[1])
              else: fc += costs.get(type(n),0)
          return fc

frame_2D_alg/meta_code.py
```diff
@@ -185,10 +185,20 @@ def set_fc(body):
 if __name__ == "__main__":
     # move to agg_recursion
     parse_funcs(["agg_recursion.py"])  # populate Fname_
-    trace_func(vars(agg_recursion))  # add oF tracing
     Y,X = imread('./images/toucan.jpg').shape
+    add_typ_()
+    # temporary:
+    # update agg_recursion's globals before running them
+    agg_recursion.oF_,agg_recursion.iF_,agg_recursion.nF_ = oF_,iF_,nF_
```
kwcckw, line 192: So this is temporary but in actual case, from higher level of code, we need to assign oF_, iF_ and nF_ into agg_recursion before running them. 

Also looks like Z is actually not needed. The CoF of frame_H below is actually representing the Z from each frame?

boris-kz: Yes, I forgot about that frame_H oF, thanks!

frame_2D_alg/agg_recursion.py
boris-kz: how is fw different from w?

frame_2D_alg/agg_recursion.py
```diff
@@ -55,8 +55,10 @@
 wY, wX = 64, 64; wYX = np.hypot(wY,wX)
 wM,wD,wi, wG,wI,wa, wL,wS,wA = 10, 10, 20, 20, 5, 20, 2, 1, 1  # dTT weights = reversed relative ave, update from wTT_ after feedback
 wT = np.array([wM,wD,wi, wG,wI,wa, wL,wS,wA]); wTT = np.array([wT,wT*avd])  # default for comp_N_?
-cN_,cC_,cN,cF, cE,ccN,ccC,ccP, cX,cFrm,cVct,cTrc,cBac,cPrj,cCS,cSE = Fc_  # replace with oF_[i].fc, same below:
-wN_,wC_,wN,wF, wE,wcN,wcC,wcP, wX,wFrm,wVct,wTrc,wBac,wPrj,wCS,wSE = Fw_  # ave gain/call, init = cost
+# below are updated from meta_code now
```
boris-kz, line 58: you didn't update their uses here?

cN_,cC_,cN,cF, cE,ccN,ccC,ccP, cX,cFrm,cVct,cTrc,cBac,cPrj,cCS,cSE = (
wN_,wC_,wN,wF, wE,wcN,wcC,wcP, wX,wFrm,wVct,wTrc,wBac,wPrj,wCS,wSE) \
= [F.fc for F in oF_]  # ave gain/call, init = cost
ttN_,ttC_,ttN,ttF, ttE,ttcN,ttcC,ttcP, ttX,ttFrm,ttVct,ttTrc,ttBac,ttPrj,ttCs,ttSE = [F.dTT for F in oF_]

frame_2D_alg/meta_code.py
boris-kz: from meta_code  import oF_,iF_,nF_?

frame_2D_alg/meta_code.py
boris-kz: you are not actually doing anything here? 
we need a list components, not a single component wrapped in a list?

frame_2D_alg/agg_recursion.py
kwcckw: They are the same? Just like we unpack c from Fc_?

frame_2D_alg/agg_recursion.py
kwcckw: That's already updated from meta_code, right before we call frame_H:

    agg_recursion.oF_,agg_recursion.iF_,agg_recursion.nF_ = oF_,iF_,nF_
    # cost and weight init from iF_ now
    for oF, cname, wname in zip(oF_,
                               ('cN_','cC_','cN','cF','cE','ccN','ccC','ccP','cX','cFrm','cVct','cTrc','cBac','cPrj','cCS','cSE'),
                               ('wN_','wC_','wN','wF','wE','wcN','wcC','wcP','wX','wFrm','wVct','wTrc','wBac','wPrj','wCS','wSE')):
        setattr(agg_recursion, cname, oF.fc)
        setattr(agg_recursion, wname, oF.fc)

    trace_func(vars(agg_recursion))  # add oF tracing (after parsing iF_, oF_ and nF_ into agg_recursion)

frame_2D_alg/meta_code.py
kwcckw: Right now that's having cyclic import issue, so we can't import meta_code from agg_recursion.

frame_2D_alg/agg_recursion.py
boris-kz: I think my way is clearer

frame_2D_alg/agg_recursion.py
boris-kz: c is data count, fc is function complexity
ok, fw is a process weight, that's different from local cluster weight?
actually that data w is just membership, it's not related to wTT

frame_2D_alg/meta_code.py
boris-kz: it should be agg_recursion importing from meta_code.
I told you to move classes to meta_code
meta_code should not have main

frame_2D_alg/meta_code.py
```diff
@@ -145,7 +145,7 @@ def trace_func(module_dict, module_name=None):
 
     if module_name is None: module_name = module_dict.get('__name__')
     for name, obj in list(module_dict.items()):
-        if name in oF_:
+        if name in iF_.keys():  # should be iF_ now
```
boris-kz, line 148: thanks

frame_2D_alg/agg_recursion.py
kwcckw: The problem is the globals will be run whenever we import agg_recursion, and oF_ is not exist yet.

boris-kz: agg_recursion should not be imported, it's an end user

frame_2D_alg/agg_recursion.py
```diff
@@ -143,8 +145,9 @@ def traced(func):
         @wraps(func)
         def inner(*a, **kw):
             _CoF = CoF._cur.get()
-            oF = CoF(nF=oF_.index(func.__name__), root=_CoF)
-            oF.wTT = FTT_[oF.nF]; oF.fw = Fw_[oF.nF]; oF.fc = Fc_[oF.nF]; _CoF.call_ += [oF]
+            oF = CoF(nF=iF_[func.__name__], root=_CoF)
```
boris-kz, line 148: oF = CoF(nF=iF_[func.nF], root=_CoF)?

frame_2D_alg/meta_code.py
kwcckw: > it should be agg_recursion importing from meta_code. I told you to move classes to meta_code meta_code should not have main

Okay, sorry, i haven't move those yet. So from meta_code, we only need sum2F from agg_recursion, and to prevent cyclic issue , i think we can create a sum2F that's specifically for meta_code? Something like sum2T?

frame_2D_alg/agg_recursion.py
kwcckw: This `func` is a function object, and they are not CoF, so we can't use func.nF here.

frame_2D_alg/agg_recursion.py
kwcckw: Okay  sorry, this is clear now.

frame_2D_alg/meta_code.py
kwcckw: This is to make it compatible with ` set_fc`.

      def set_fc(body):
          fc = 0
          for n in body:
              if isinstance(n, CoF): fc += 3
              elif isinstance(n, tuple): fc += costs.get(n[0],0) + set_fc(n[1])
              else: fc += costs.get(type(n),0)
          return fc
          
 `set_fc` assumed body is iterable, but from this line ` T.body = build_body(nF_[i])`, `build_body` returns node, oF[i] or a tuple of (type, sub_), so they are actually not iterable if we use `set_fc(T.body)`. That `for n in body` loop is actually for nested call via `elif isinstance(n, tuple): fc += costs.get(n[0],0) + set_fc(n[1])`.

frame_2D_alg/meta_code.py
boris-kz: sum2F, add2F here should be different, they will be summing only fw and fc?
also we need to import agg_recursion AST to get initial bodies, but that's just initialization, not cyclic

frame_2D_alg/agg_recursion.py
boris-kz: above:
func.nF = iF_[func.__name__]  # decoration time
but I guess it's not needed?

frame_2D_alg/meta_code.py
boris-kz: the nested call is a single fork only, we need to unpack set_fc in build_body?

boris-kz: Yes, redundant traversal. Fold cost into build_body — return `(result, cost)` so every recursion both shapes and weighs the same nodes. Definitionally consistent.

```python
def build_body(node):  # returns (result, cost): T | (type, sub_) | node | None, fc
        i = oF_idx.get(node.func.id)
        if i is not None and oF_[i].body: return oF_[i], 3  # Call overhead at this site
    sub_, sub_fc = [], 0
    for t in ast.iter_child_nodes(node):
        r, c = build_body(t)
        if r is not None: sub_.append(r); sub_fc += c
    if sub_:    return (type(node), sub_), costs.get(type(node),0) + sub_fc
    if type(node) in costs: return node, costs[type(node)]
    return None, 0
```

In add_typ_:

```python
def add_typ_():
    for i, T in enumerate(oF_):
        if Fname_[i] is not None:
            T.caller_ = []
            body_, fc = [], 0
            for c in Fname_[i].body:
                r, k = build_body(c)
                if r is not None: body_ += [r]; fc += k
            T.body, T.fc = body_, fc
```

`set_fc` goes away entirely.

Two notes:

1. **`type(node)` cost is added at composite level**, not leaf. So `(ast.If, [...])` contributes `costs[ast.If]=1` plus children. A bare leaf (no surviving children, type in costs) contributes only `costs[type]`. Same as set_fc semantics — confirm by tracing once.

2. **T refs charge 3 at their parent's accumulation**, not via their own `T.fc`. So if comp_N_ calls comp_N twice, comp_N_'s fc gets +6 (two callsites), while comp_N's fc is independent. Correct — "independent of deeper callees".

frame_2D_alg/agg_recursion.py
kwcckw: I see, i think that's redundant since we can get the same thing from `iF_[func.__name__]` now.

frame_2D_alg/meta_code.py
kwcckw: > sum2F, add2F here should be different, they will be summing only fw and fc?

Yes, and those params such as caller_ should be added in runtime.

> also we need to import agg_recursion AST to get initial bodies, but that's just initialization, not cyclic

I think there's no need to import there, we just need to read the file from path, just like parse_funcs `parse_funcs(["agg_recursion.py"])`.

frame_2D_alg/agg_recursion.py
kwcckw: I see, thanks for clarifying again. I confused c with fc earlier. 

> ok, fw is a process weight, that's different from local cluster weight?

Process weight should be code based, while local cluster weight is from data?

boris-kz: right

frame_2D_alg/meta_code.py
boris-kz: ok right. but we need to call cluster_calls in agg_recursion, per frame_H

frame_2D_alg/meta_code.py
kwcckw: Okay, so this can be simpler by just retrieving their body and cost at the same time, then we can skip the additional unpacking process to retrieve the cost again.

boris-kz: not just that, right now set_fc is not doing anything

kwcckw: The existing set_fc assigns their cost based on body node type? 

    def set_fc(body):
        fc = 0
        for n in body:
            if isinstance(n, CoF): fc += 3
            elif isinstance(n, tuple): fc += costs.get(n[0],0) + set_fc(n[1])
            else: fc += costs.get(type(n),0)
        return fc

The edits above assigns cost and unpacks body at the same time, which allow us to skip set_fc now.

boris-kz: ok, packed in:

def F_body_():

    def build(node):  # AST → CoF | (type,sub_) | ast_leaf | None
            i = nF_.get(node.func.id)
            if i is not None and oF_[i].body: return oF_[i]
        sub_ = [ret for t in ast.iter_child_nodes(node) if (ret := build(t)) is not None]
        if sub_: return (type(node), sub_)
        if type(node) in costs: return node

    def set_fc(n):  # nested structure → cost
        if isinstance(n, CoF): return 3
        if isinstance(n, tuple): return costs.get(n[0],0) + sum(set_fc(c) for c in n[1])
        return costs.get(type(n), 0)

    for i, F in enumerate(oF_):
        if not F.body: continue
        F.caller_ = []
        F.body = [r for c in F.body if (r := build(c)) is not None]
        F.fc = sum(set_fc(n) for n in F.body)

kwcckw: Okay, but wouldn't it be better to get body and cost at the same time? Since the unpacking process is the same.

But your code above is clearer and we can reuse the set_fc when needed.

boris-kz: That's what I thought, but Claude says we can't do it inline, the structure is built yet

frame_2D_alg/agg_recursion.py
boris-kz: I am trying to refactor sum2F and add2F for CoFs only:



```
def sum2O(N_, root=None):  # for w,c,r, fw,fc,fr only?

    fc_ = np.array([n.fc for n in N_], dtype=float); fC = fc_.sum(); w_ = fc_/fC  # N = N_[0]
    # unfinished, use w_ for N summing?
    return CoF(call_=N_, root=root, fC=fC  )  # fw = Fw

def add2O(F, n, merge=0):  # unpack for batching in sum2F

    if F.c:  # sum subtree gain, fixed cost, extensive: no weighting?
        C=F.c+n.c; _w,w = F.c/C, n.c/C; F.r = F.r*_w + n.r*w; F.c = C; F.fw += n.m
    else:
        F.c=n.c; F.r=n.r; F.fw=n.m
    F.N_ += (n.N_ if merge else [n])
    return F
```

frame_2D_alg/meta_code.py
kwcckw: That's weird, the code is static, so the code structure must already exist? Let me check again when I have the access to my computer later

frame_2D_alg/agg_recursion.py
kwcckw: Okay, I will go through them again later.

boris-kz: so I moved classes and meta-data to meta_code, not finished

PR #153, opened 2026-05-23
5/23/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -113,15 +113,15 @@ def traced(func):
...
-            _CoF.call_ += [oF]
-            CoF._cur.set(oF); out = func(*a, **kw)
+            if _CoF is not None: _CoF.call_ += [oF]
+            token = CoF._cur.set(oF); out = func(*a, **kw)
             if oF.call_:
                 tree = flat_(oF); L=len(tree)-1
                 if oF.fw*L > ave*(oF.fc*L):  # conditional?
                     sum2O(tree, oF); wtt = getattr(oF,'rTT',oF.dTT);  oF.wTT = cent_TT(wtt,oF.r)
-            CoF._cur.set(_CoF)
+            CoF._cur.reset(token)
```
kwcckw, line 124: So this changes is needed when we use frame_H as the root oF, and their context is empty. Explanation from LLM:

When you use `token = CoF._cur.set(oF)`, the token acts like a perfectly preserved bookmark of whatever the context was right before you changed it.

Here is why `reset(token)` is superior to doing `CoF._cur.set(_CoF)`:

- It handles the "Empty" state perfectly:
  
  If `frame_H` is the very first function called, the context is empty. If you manually restore it by doing `.set(None)`, the `ContextVar` now holds the value `None`.

  But if you use `.reset(token)`, it strips the value entirely, returning the `ContextVar` to its true, natively empty state.

- It is recursion-proof:
  
  In your `agg_recursion` script, functions call other traced functions, pushing new `oF` objects onto the context stack.

  The token guarantees that when a deeply nested function finishes, it pops exactly its own `oF` off the stack, restoring the exact parent `oF`, preventing cross-talk between different branches of your execution tree.

frame_2D_alg/meta_code.py
```diff
@@ -296,13 +296,12 @@ def set_fc(n):  # nested structure → cost
         return costs.get(type(n), 0)
 
     for i, F in enumerate(oF_):
-        if not F.body: continue
```
kwcckw, line 299: Why we need to check F.body here? Their body is empty before the `build` below and we need `build` below to fill their body?

Also looks like oF.body is nested all the time because all functions must have child? So body's structure is something like `(func def, sub_)`.

frame_2D_alg/meta_code.py
```diff
@@ -296,13 +296,12 @@ def set_fc(n):  # nested structure → cost
...
         F.caller_ = []
-        F.body = [r for c in F.body if (r := build(c)) is not None]
-        F.fc = sum(set_fc(n) for n in F.body)
+        F.body = build(nF_[i])
+        F.fc = set_fc(F.body)  # each main body is sub anyway, where the typ is function def: (func def, sub_)
 
 def sum2O(N_, root=None):  # for w,c,r, fw,fc,fr only?
-
+    # we need to update c across all functions now? Only some functions such as cluster_N, cluster_C are updated now
     c_ = np.array([n.c for n in N_], dtype=float); C = c_.sum(); w_ = c_/C
```
kwcckw, line 305: For data count c here, we may not assign them across most of the functions, so they will stay empty? Or we need to revert it to one of the prior scheme where we init all function's c with caller's c?

kwcckw:
> but Claude says we can't do it inline, the structure is built yet

So i checked with claude, but looks like we can fold set_fc into build?

Also could you show me the response from the claude?

 
image

frame_2D_alg/agg_recursion.py
```diff
@@ -916,12 +916,13 @@ def ffeedback(frame):  # adjust filters: all aves *= rV, ultimately differential
     trace_func(vars())
     Y,X = imread('./images/toucan.jpg').shape
     # frame = agg_frame(0, image=imread('./images/toucan.jpg'), iY=Y, iX=X)
-    frame = frame_H(image=imread('./images/toucan.jpg'), iY=Y//2 -31, iX=X//2 -31, Ly=64,Lx=64, Y=Y, X=X, rV=1)
+    parse_funcs(["agg_recursion.py","comp_slice.py","slice_edge.py"])  # populate nF_
+    F_body_()  # fill body
+    # not sure if we need this Z, we only need oF_ and access calls from oF_?
```
boris-kz, line 921: yes, just make it oF_[0], so frame_H doesn't need to return it?

frame_2D_alg/meta_code.py
boris-kz: ok, can we have a more meaningfull term than that "token"? like __oF?

frame_2D_alg/meta_code.py
boris-kz: right

frame_2D_alg/meta_code.py
boris-kz: yes data should be empty, I just copied it

boris-kz: actually this called inside CoF too, there all data should be summed

frame_2D_alg/agg_recursion.py
kwcckw: But oF_[0] is their typ, and we may call frame_H in expand_lev again? This Z is actually the frame_H's call instead of typ.

frame_2D_alg/meta_code.py
kwcckw: Yes, but they are summed from callees, those calls may not have c too?

frame_2D_alg/meta_code.py
kwcckw: Yes, it's the prior context, so i think _oF should be fine too.

frame_2D_alg/meta_code.py
boris-kz: they should, these calls process data

frame_2D_alg/agg_recursion.py
boris-kz: ok, nF_[0]?

boris-kz: actually oF_ is fine, it does accumulate data across frames?

kwcckw: Yes, it's just we need this Z or per frame_H's oF if we need the per frame_H tree hierarchy.

frame_2D_alg/meta_code.py
kwcckw: So we should add them? Functions like comp_N, comp_F doesn't assign c to oF, so oF.c should be the sum of c of the processed data within the function?

boris-kz: yes just add fdata=1 in CoF

frame_2D_alg/agg_recursion.py
boris-kz: ok, add H to Frm, use H[i] in frame_H:

`_names = ['frame_H',  # root function
          'comp_N_','comp_C_','comp_N','comp_F',  # comp_ functions
          'get_exemplars','cluster_N','cluster_C','cluster_P',  # clust_ functions
          'cross_comp','vect_edge','trace_edge','ffeedback','proj_N','comp_slice','slice_edge']  # combined, ancillary
          # + eval Fs?`

`oF_ = [CoF(nF=i) for i in range(len(_names))]; oF_[0].H = []`

Then it doesn't need a return?

frame_2D_alg/meta_code.py
kwcckw: > yes just add fdata=1 in CoF

Actually how does fdata enable the accumulation of c? The c accumulation is only valid when fdata = 1? But all functions' oF should accumulate c?

boris-kz: not if we merge oFs

frame_2D_alg/agg_recursion.py
kwcckw: So this H should be filled when oF is the root frame_H, and it should pack their calls in nested list? So that each element is calls from each frame?

kwcckw: So this H should be filled when oF is the root frame_H, and it should pack their calls in nested list? So that each element is calls from each frame?

boris-kz: right

frame_2D_alg/meta_code.py
kwcckw: So this c accumulation should be done in `traced`? I thought it should be done in function body, where we need to add them manually? Then merge oFs shouldn't be an issue too.

boris-kz: it just adds a lot of clutter, that's why we have traced

kwcckw: It's the same in traced because each function returns different data structure, i don't see how we can add c easily from the returned vars of all functions?

boris-kz: what's the problem here:

                if oF.fw*L > ave*(oF.fc*L):  # conditional?
                    sum2O(tree,oF,1); wtt = getattr(oF,'rTT',oF.dTT);  oF.wTT = cent_TT(wtt,oF.r)

kwcckw: This is actually using the default fw and fc? Since their fw and fc should be updated in sum2O below?

boris-kz: yes, if oF.call_:  should be enough for now, we can add deeper filter later

frame_2D_alg/meta_code.py
boris-kz, line 293: try unpacking it, doesn't make any sense

frame_2D_alg/agg_recursion.py
boris-kz: we already have that, new F subsumes the prior one into its H:

            F.H =[]; F.N_= tile_

frame_2D_alg/meta_code.py
boris-kz, line 226: why do we need frozenset here?

frame_2D_alg/agg_recursion.py
kwcckw: But we still can't access their call tree from root F.H here, unless we return it?

frame_2D_alg/meta_code.py
kwcckw: Yes, so actually set_fc should be folded into build, this is actually redundant to unpack and set their fc again.

kwcckw: So `frozenset` is used here because it checks if the pair are in T.grp regardless their (p1, p2) order.

For example, if we use tuple, (1,2) == (2,1) = False, but with frozenset, frozenset((1,2)) == frozenset((2,1)) = True.

boris-kz: As it is, it doesn't make any sense if you unpack it, just try

frame_2D_alg/agg_recursion.py
boris-kz: return F, that's all we need

frame_2D_alg/meta_code.py
boris-kz: it doesn't make any sense anyway, which is clear if you unpack it

boris-kz: it's too opaque.  I think we just need to pack pairs in T.rim, like links

kwcckw: Okay, then we shouldn't unpack both build and set_fc and remove the functions since both of them are needed for recursive operations, so it's good as it is now.

kwcckw: Alright, also for their eval, using `all` is too strict? Or it must be strict because the typ merging should be rare?

boris-kz: you don't get it, it's trash anyway

boris-kz: none of that makes any sense, precisely because the damn Claude made it opaque

kwcckw: So we don't need the nested sub_? Since the cost of child node should be replaced by the cost of calling the child function or code. Which mean everything here should be unpacked to a flat body and the cost summing process should be straightforward without going into the nested structure.

kwcckw: Okay, so this is actually similar with the clustering in trace_edge, and the T.grp here is actually the Gt. We can reuse the similar structure, but with added eval to cluster them if the link is added by default.

boris-kz: the difference here is that this is complete linkage, so grp should be a set of links, not nodes?

boris-kz: the primitives should be nested, else we can't align them for comparison.
we just need to skip that top func def?

kwcckw: But if grp is set of links, we might get overlapping nodes in different group? Since the final T is summed by nodes instead of their links, and links is only good for the linkage eval?

boris-kz: get the nodes from links
I pushed partial draft, will check tomorrow

kwcckw: Okay, that top func def is just for code consistency, we actually can remove it.

kwcckw: Okay, so if we are grouping links, all those overlapping nodes should be grouped under the same new T until there's no more overlapping?

PR #154, opened 2026-05-24
5/24/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -281,22 +287,17 @@ def build(node):
...
-        if type(node) in costs: return node
-    # this is wrong:
-    def set_fc(n):  # nested structure → cost
-        if isinstance(n, CoF): return 3
-        if isinstance(n, tuple): return costs.get(n[0], 0) + sum(set_fc(c) for c in n[1])
-        return costs.get(type(n), 0)
+            if i is not None and oF_[i].body: return oF_[i], 3
+        sub_ = [rett for t in ast.iter_child_nodes(node) if (rett:= build(t)) is not None]
+        if sub_: sub_, fc_ = zip(*sub_); return (type(node), sub_), sum(fc_)+costs.get(type(node), 0)
+        if type(node) in costs: return type(node), costs.get(type(node), 0)
```
kwcckw, line 293: We not really need `set_fc` and the cost can be folded in `build`, then both cost and body can be retrieved from `build` only. 

Also there's a mistake in my line 293 above, it should be:

    return node, costs.get(type(node), 0)

I was confused earlier and it should be node instead of their type.

frame_2D_alg/meta_code.py
```diff
@@ -227,28 +228,33 @@ def comp_callers(_T, T):
...
+        V, _T, T = link
+        if _T.grp is not T.grp: 
+            olp = T.rim & _T.rim  
+            if sum([V for V,_,_ in olp]) > ave * len(olp):  # full linkage's olp V?
+                _T.grp.add(link); _T.grp.update(T.grp) 
+                T.grp = _T.grp  # group T into _T           
+
+    grp_ = np.unique([T.grp for T in _T_ if T.grp])  # unique grp
+    # with link method, T without links will be recycled?
+    _T_ = [T for T in _T_ if not T.grp]
```
kwcckw, line 245: In this method, we add link to grp whenever the linkage V's eval is true. Therefore, many T.grp will be empty if all eval are false, and they will be recycled, as in prior scheme?

frame_2D_alg/meta_code.py
```diff
@@ -227,28 +228,33 @@ def comp_callers(_T, T):
...
-        for t in grp: gV += t.V; gC += t.fc  # add links?
-        gV /= 2  # norm for pairwise redundancy
-        if gV > ave-gC: G_ += [[grp,gV,gC]]; V += gV; C += gC
-        else:           G_ += grp
+    for grp in grp_:
+        N_, L_ = [], []; gV = gC = 0  # membership, compression / grp
+        for link in grp: 
+            V, _T, T = link
+            # overlapping Ns between G_ here, so we need to eval T's G here, and merge them?
+            N_ += [_T, T]; L_ += [link]
```
kwcckw, line 254: If the clustering is via links here (grouping links instead of nodes), we may get overlapping Ts from different link. So i think we need additional step here to check for existing Gs in _T and T, and then merge those Gs if they are available?

frame_2D_alg/meta_code.py
```diff
@@ -113,6 +113,7 @@ def traced(func):
         def inner(*a, **kw):
             _CoF = CoF._cur.get(None)
             oF = CoF(nF=iF_[func.__name__], root=_CoF)
+            oF_[iF_[func.__name__]].call_ += [oF]  # we need to pack call in typ oF?
```
boris-kz, line 116: right

frame_2D_alg/meta_code.py
boris-kz: thanks

frame_2D_alg/meta_code.py
boris-kz: right

frame_2D_alg/meta_code.py
```diff
@@ -227,28 +228,33 @@ def comp_callers(_T, T):
...
-        # eval olp_V?
-        _g, g = _T.grp, T.grp
-        for t in g: t.grp = _g
-        _g += g
-    # old:
+    L_.sort(key=lambda link: link[0], reverse=True)  # complete-linkage agglomeration on pairs, replace with centroids if max n pairs?
+    for link in L_:
+        V, _T, T = link
+        if _T.grp is not T.grp: 
+            olp = T.rim & _T.rim
```
boris-kz, line 238: complete linkage is T.rim in grp, not in _T.rim. 
so it's not a flood fill anymore, it's really a centroid-based clustering, with centroid distributed over grp
that means recursive refining. I haven't worked that out yet

kwcckw: Okay, but if it's a centroid based clsutering, each grp will be init with single T? Then we compare each new `_T` with the grp and refine them again? So totally similar with existing cluster_C?

boris-kz: Probably init with exemplars, else all grps will be the same, we need to break symmetry

boris-kz: Yes — that framing actually works.

Average linkage = centroid clustering where the “centroid” is the empirical distribution of the cluster (uniform weight 1/|A| on each member) rather than the mean point. The cluster-cluster distance becomes the expected cross-pair distance: E_{a∼A, b∼B}[d(a,b)]. That’s exactly average linkage.

Precise relationship, for squared Euclidean:
mean cross-pair sq dist = ||c_A − c_B||² + var(A) + var(B)

So average linkage = centroid distance plus intra-cluster spread terms. Centroid linkage drops the spread; average linkage retains it. The “distribution” of the cluster carries the variance that a single centroid throws away.

Two senses of “distributed” worth keeping distinct:

	•	Distributed representation (spread, multi-point): average linkage, your intended sense. Each member contributes equally to the cluster’s representation; the cluster is its empirical distribution.
	•	Distributed membership (fuzzy/soft, FCM): each point has graded membership in multiple clusters. Different axis.

Average linkage is (1) without (2): hard membership, distributed representation. FCM is (2) with centroid: soft membership, point representation. You could combine them — soft membership and distributed representation — which would be expected pairwise distance under fuzzy memberships.

So the clean statement: average linkage is centroid clustering with the centroid replaced by the full member set, and centroid-distance replaced by mean cross-pair distance. The spread terms are the price of not compressing.

kwcckw: Okay, so if we follow centroid clustering method and using soft membership here, each new typ might have overlapping typ nodes?

boris-kz: soft membership is tentative, we can assign exclusively to max V after the clusters are formed

boris-kz: Claude: both points hold, with caveats.

**More accurate**: average linkage retains the spread information centroids discard, and it's monotonic. Centroid linkage has the additional problem of dendrogram **inversions** — merged-cluster centroids can be closer to a third cluster than the pre-merge clusters were, giving non-monotonic merge heights. Average linkage doesn't do this. So it's both more faithful (keeps variance) and better-behaved (monotonic).

**Cheaper up to a crossover**: per cluster-pair comparison, the cost picture is:

- Average linkage: ~|A|·|B| pair comparisons (each pair distance from already-computed lateral comp).
- Centroid: ~|A|+|B| work to form/update centroids, then ~1 comparison between them.

So average linkage wins while |A|·|B| is less than the cost of forming + comparing centroids. Crossover depends on the cost ratio between a pair comparison and a centroid formation. In your setting, centroid formation isn't `mean(vec)` — it's `sum2F`/`add2F` over nested CL/CC structures, which is real structured work. So the crossover sits at meaningfully larger cluster sizes than the trivial vector case would suggest.

Important refinement: this analysis assumes you're computing pair distances on demand. If you already have all pairwise link values from connectivity clustering (which you do — that's what `cluster_N` produced), then average linkage can be evaluated *for free* using existing links and a Lance-Williams update on the pairwise matrix. The centroid path requires constructing something new (the centroid CF). So the crossover is pushed even further out.

Where centroid wins: at large |A|·|B|, especially when you need to compare a cluster against *many* others — the centroid amortizes (form once, compare many times) while average linkage pays per comparison. Also when you need a persistent compressed representation (you want the cluster as an object, not just a relation).

Practical fit with your hierarchy: connectivity clustering (`cluster_N`) gives you links → average linkage is the natural follow-up, cheap. Centroid refinement (`cluster_C`) is justified when clusters are large enough that the amortization wins, or when you want the centroid as a downstream object (further comparison, hierarchy promotion).

kwcckw: Okay, so there should be a pruning section, where each node will be only assigned to their max V's G? This V should be the comparison m between each node and the cluster?

boris-kz: > Okay, so there should be a pruning section, where each node will be only assigned to their max V's G? This V should be the comparison m between each node and the cluster?

yes, so we need v_ per node

kwcckw: That's actually similar with m_ in cluster_C, i guess we need root_ and d_ too?

`root_` for root typ checking, while `d_` probably for pruning purpose?

boris-kz: Yes, the reasoning is correct, with one clarification about where symmetry actually bites.

The setup: oFs have no spatial/structural adjacency — any oF can call/co-occur with any other. So there's no graph to do connectivity clustering on; you fall back to all-pairs comparison (which is what `combinations(_T_,2)` already does).

The symmetry argument is specifically a **centroid-init** problem. If you tried centroid clustering with random/uniform initial centroids, every centroid is statistically the global mean → every iteration moves them toward the same global mean → all groups end identical. Exemplars break that: pick K distinguished nodes as initial centers, then assign/refine. Standard k-means++ logic, just applied here.

The current `cluster_oF_` doesn't have the symmetry problem because it's complete-linkage agglomerative: link *values* provide the ordering, the first merge is determined by the highest-V link, etc. Asymmetry comes from the weights, not from initialization. So you only need exemplars when you switch off complete-linkage to centroids — which is exactly the comment on line 234 ("replace with centroids if max n pairs").

Natural exemplar choices from already-computed quantities:
- highest `T.V` (sum of rim values) — most-connected node
- highest `T.fc` (individual compression) — most-salient node alone
- farthest-first: pick highest-V, then iteratively pick the node with lowest comp_callers+comp_body overlap to already-picked exemplars — this is the symmetry-breaking analog of k-means++

One thing worth flagging: at small N (the regime where avg linkage was cheaper than centroids), you don't need exemplars at all — just keep complete/avg linkage on the full pair set. Exemplars are only needed once you commit to a centroid representation because N got too big. So exemplar-init pairs naturally with the crossover from average-linkage to centroid that we discussed earlier — it's the same transition.

Is the question whether to add exemplar init to `cluster_oF_` now, or is this conceptual prep for when N grows large enough that the centroid path becomes necessary?

boris-kz: > 

yes except we don't need d_ because there is no spatial direction

kwcckw: Alright, and i guess this shouldn't be parallelizable since it starts from the highest V's exemplar.

boris-kz: The principle: in complete cross-comp the V matrix has statistically uniform off-diagonal entries — every node's rim is a sample from the same pair-V distribution. So:

- **Row aggregates don't discriminate**: `T.V`, mean rim V, max rim V are approximately equal across all T. They look like noise around a global constant.
- **Inter-cluster aggregates don't discriminate either**: mean cross-pair V between any two subsets of comparable size is the same global constant. So avg-linkage as a *merge* criterion is useless — every potential merge has the same cost.
- **Diagonal is the only node-distinguishing signal**: `T.fc` (individual function compression) is set from T's content alone, not from any comparison. It's the only quantity that varies *per node* independent of the comparison population.

Consequences:

1. **Exemplars must be picked from the diagonal**, i.e. `T.fc` (individual saliency), not `T.V` (rim sum). My earlier suggestion to NMS-sort by `T.V` was wrong in this regime — `T.V` is uniform.

2. **Off-diagonal V is for assignment, not ranking**. Once exemplars are picked by `T.fc`, each non-exemplar T is assigned to the exemplar E with highest V(T, E). The pairwise V matrix discriminates *which exemplar T is closest to*, not which pair to merge first.

3. **No agglomerative path is meaningful here.** Sort-by-V agglomeration relies on V differences carrying cluster signal; in uniform off-diagonal V they're just noise. The first merge is arbitrary, the rest chains off that noise.

4. **Rim-overlap logic from sparse case doesn't transfer.** `T.rim ∩ _T.rim` in complete cross-comp is just the single direct link between T and _T — not shared neighbors. The whole `olp` test in current code is sparse-rim mechanics misapplied to a dense graph.

So the right shape: K-medoids / affinity-propagation style. Pick exemplars by diagonal NMS, assign others by V(T, exemplar). No agglomeration, no avg-linkage between clusters, no `T.rim` intersections.

The avg-linkage discussion still applies *within* the partitional path: a cluster's "distributed centroid" is its member set; an incoming node's score against cluster A is the mean V(T, m) over m ∈ A. That works because for a *specific* A (not a random subset), the mean isn't uniform — it picks up A's coherence around its exemplar.

Is `T.fc` set such that the NMS on it produces meaningfully sparse exemplars, or does it also collapse in practice? That determines whether even this approach works, or whether you need another diagonal-like quantity.

boris-kz: Me: 
T.fc is only meaningful if we include oF fcs nested in it's body, but it's not good for picking exemplars.
The oF_ itself is sort of a seed exemplar, it is manually designed to start self-improvement process. 
So I can probably pick exemplars manually too.

Claude: 
If exemplars are part of the design layer rather than something to discover, structure becomes:

1. Tag specific oFs as exemplars in the initial setup (e.g. an `exe` flag on `CoF` at construction, set for hand-picked seeds).
2. For each non-exemplar T: compute mean V(T, m) over m in each exemplar's current member set; assign T to the highest-scoring exemplar group.
3. Optionally iterate: as members are added, the "distributed centroid" (member-set mean) shifts; re-score and reassign until stable. Usually one or two passes is enough.

The avg-linkage discussion now lives entirely in step 2 — comparing a candidate T against a group's distributed member set rather than against the exemplar alone is what gives you the spread-aware assignment. So avg-linkage is the *assignment metric*, not the merge criterion.

This also drops the V matrix sort and the `T.rim & _T.rim` test — both were doing sparse-rim work in a context that doesn't have sparse rims. What remains needed from the comp pass is the full pairwise V (just for assignment lookup), not the rim structures or row sums.

One question for the design: hard partition (each T joins one exemplar group) or fuzzy (each T has graded membership weighted by V to each exemplar)? Fuzzy fits "distributed centroid" thinking — if a group's identity is its distribution, a member's contribution can also be distributed. Hard partition is simpler and fits the existing `T.grp` set semantics. Which is your intent here?

kwcckw: > T.fc is only meaningful if we include oF fcs nested in it's body, but it's not good for picking exemplars.
The oF_ itself is sort of a seed exemplar, it is manually designed to start self-improvement process.
So I can probably pick exemplars manually too.

`fc` is function complexity, and i don't see how can be used to determine exemplar too, so i think it must be based on V. 

The manual exemplar selection should be still best to based on the highest V's T and V>ave? Since we will compare them all to all first anyway, and not all T should be an exemplar. 

> This also drops the V matrix sort and the T.rim & _T.rim test — both were doing sparse-rim work in a context that doesn't have sparse rims. What remains needed from the comp pass is the full pairwise V (just for assignment lookup), not the rim structures or row sums.

Not sure on this, so this suggests revert to simpler single link eval rather than the rim overlapping method?

boris-kz: V is actually the same for all oFs, unless we manually add initial grp per exemplar
it's ave linkage, not slingle linkage, but with lower weights outside that initial grp

kwcckw: Okay, the average linkage still needs to compare each new T to the mean T's cluster, which is similar to cluster_C too. 

Also for the exemplar initialization, it will be selected manually from certain Ts now? But in the subsequent frames with new typs, we still need to automate this process?

boris-kz: not mean, pairwise, just read the above

kwcckw: Okay, so actually the average linkage is using the pairwise V as the filter to decide whether the new node will be added?

For example, the potential new node will be used to generate pairwise V with all member of the existing cluster, and then those Vs will be used to eval the membership of new node?

boris-kz: yes, it's mean_V > ave for membership, instead of all_Vs in complete linkage, I misundestood that before

kwcckw: Alright, and since we use mean_V > ave here, we shouldn't get soft membership now?

boris-kz: It's soft while iterating, that >ave is for final assignment

kwcckw: > It's soft while iterating, that >ave is for final assignment

So during the refinement or iteration, we still need to compute their membership score, and that is using the same mean V too?

boris-kz: Yes, that's mean of member to all other members

PR #155, opened 2026-05-25
5/25/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -224,10 +224,13 @@ def comp_callers(_T, T):
         return M / (D or 1e-7)  # match if same_callers / diff_callers > ave?
 
     # draft:
-    E_ = []; T_ = oF_
+    # temporary, preset:
+    E_ = sorted(oF_, key=lambda t: t.fc, reverse=True)[:int(len(oF_)/2)]  # high fc as exemplars? 
+    for T in oF_: T.grp = set([T])
```
kwcckw, line 229: So exemplars should be preset, and it's just random? Or we can init them with the high fc's T? Since a more complex T should have a higher chance with more body primitives in the grouping?

frame_2D_alg/meta_code.py
```diff
@@ -236,33 +239,37 @@ def comp_callers(_T, T):
             else:   t.ex_[o] = V
```
kwcckw, line 239: If this is outside the iterative loops, fin will never be true? So we need to compare them again in each new iteration?

frame_2D_alg/meta_code.py
```diff
@@ -236,33 +239,37 @@ def comp_callers(_T, T):
...
     # not reviewed:
     for _ in range(4):  # iterative reassignment
+        grpt_ = [[T.grp, 0, 1, 0, 1] for T in T_]   # init [grp, inV, inC, exV, exC], or only from exemplars?
```
kwcckw, line 242: So this section is actually quite difficult to read. From my understanding, initially each T is assigning their internal and external V to the corresponding compared `_T`'s group (this is still fuzzy).

Then, T.grp will be updated to the highest V's overlapping group, and this is eval by their pair V only. 

In the final step, from exemplars, each T that's same group as exemplar will be eval based on mean pairwise V for the final membership. The rest of the no-grouped Ts probably should be recycled.

Also we have internal and external V per T pair, but i think most of the time we will need both internal and external V for evaluation?

frame_2D_alg/meta_code.py
boris-kz: nothing is random here, and fc only matters for sub-clustering. I outlined classes of the oFs:

_names = ['frame_H',  # root function

          'comp_N_','comp_C_','comp_N','comp_F',  # comp_ functions
          'get_exemplars','cluster_N','cluster_C','cluster_P',  # clust_ functions
          'cross_comp','vect_edge','trace_edge','ffeedback','proj_N','comp_slice','slice_edge']  # combined, ancillary

frame_2D_alg/meta_code.py
boris-kz: it's preset, in_ is the same oF class: comp_, clust_, composite, eval

frame_2D_alg/meta_code.py
boris-kz: it's always both, but external V *=.5
yes, final is internal only
Exemplars are just initializations

frame_2D_alg/meta_code.py
kwcckw: Sorry, i'm a bit confused here. So do you mean those functions under names are exemplars? Or we need another separated process to determine them first so we can preset them before the first iteration?

frame_2D_alg/meta_code.py
kwcckw: I see, so we probably need to add this additional comp_, clust_, composite or eval categories into each oF, and use them for the comparison here? For subsequent frames, when we have combined typs, we will need new category?

frame_2D_alg/meta_code.py
kwcckw: Thanks, also there is no comp within the iterations, so the refinement will be purely based on their in_ and ex_ in each iteration, where in_ and ex_ should be updated based on the new grouping of Ts in each iteration?

boris-kz: Right. I will set both exemplars and grps myself

kwcckw: Also i think the existing code in cluster_oF_ is actually not quite readable, so it's better to replace the usage of dict?

boris-kz: How? It may be just a habit, we are not used to dicts

kwcckw: Yes, other than the code structure is different with existing agg_recursion, i think mostly is due to the naming. For example, `m` should be `t` for type, and gV should be grp_ since a mapping list.

frame_2D_alg/meta_code.py
boris-kz: comp_N is an exemplar for the first line, cluster_N for the second, cross_comp for third
the lines are intitial grps. something like that

kwcckw: I see, but we can't do this for the subsequent frames right? So in the next frame, we still need an automated process to determine the exemplars?

boris-kz: the next frame_H will have nested oFs, their roots are already effective centroids, so we don't need exemplars

frame_2D_alg/meta_code.py
```diff
@@ -236,33 +239,37 @@ def comp_callers(_T, T):
...
-            full = {**T.in_, **T.ex_}
-            T.in_, T.ex_, T.grp = {}, {}, bG
-            for m,v in full.items():  # repartition both endpoints
-                if T in m.in_: m.in_.pop(T)
-                else: m.ex_.pop(T)
-                same = m.grp is bG
-                (T.in_ if same else T.ex_)[m] = v
-                (m.in_ if same else m.ex_)[T] = v
-            change = 1
+            for ind, tv in zip((0,2),(T.in_, T.ex_)):  # accumulate internal and external V based on target group
```
boris-kz, line 245: you keep doing this global grpt_, which is a total mess.
we only need T.grpt

frame_2D_alg/meta_code.py
```diff
@@ -224,10 +224,13 @@ def comp_callers(_T, T):
...
+    # temporary, preset:
+    E_ = sorted(oF_, key=lambda t: t.fc, reverse=True)[:int(len(oF_)/2)]  # high fc as exemplars? 
+    for T in oF_: T.grp = set([T])
+
+    T_ = oF_
     for t in T_:  # preset initial T.grp and .exe
         t.caller_ = set(t.caller_); t.in_,t.ex_ = {},{}  # _T.V s in and outside local grp
-        if t.exe: E_ += [t]
     for _T,T in combinations(T_,2):  # full pairwise V, stored per-node
         V = comp_callers(_T,T) + comp_body(_T,T)
```
boris-kz, line 235: this V should be V*(min( _T.fc,T.fc)) , no need to keep them separate?

V = (comp_callers(_T,T) + comp_body(_T,T)) * min(_T.fc,T.fc)

frame_2D_alg/meta_code.py
```diff
@@ -224,10 +224,13 @@ def comp_callers(_T, T):
...
 
     # draft:
-    E_ = []; T_ = oF_
+    # temporary, preset:
+    E_ = sorted(oF_, key=lambda t: t.fc, reverse=True)[:int(len(oF_)/2)]  # high fc as exemplars? 
+    for T in oF_: T.grp = set([T])
+
+    T_ = oF_
     for t in T_:  # preset initial T.grp and .exe
         t.caller_ = set(t.caller_); t.in_,t.ex_ = {},{}  # _T.V s in and outside local grp
```
boris-kz, line 233: `t.in_,t.ex_ = {},{}; t._in_,t._ex_ = {},{} `

frame_2D_alg/meta_code.py
boris-kz: I think we should follow the pattern of cluster_P here, that's also complete static matrix refining.
but with gT_ instead of C_, and pairwise in_,ex_ reassignment instead of C_ computation and comparison

Drafted following cluster_P shape — snapshot `_in_`/`_ex_`, reassign, repartition, convergence gate:

```python
def cluster_oF_():  # partitional, exemplar-seeded, FCM-style iterative reassignment

    def comp_body(_n, n, C=0):  # unchanged
        ...
    def comp_callers(_T, T):    # unchanged
        ...

    E_ = []
    for t in oF_:  # preset t.gt = [grp,V] (shared per-group ref) and t.exe
        t.caller_ = set(t.caller_)
        t.in_,t.ex_ = {},{}; t._in_,t._ex_ = {},{}
        if t.exe: E_ += [t]

    for _T,T in combinations(oF_,2):  # initial partition by preset gt
        V = (comp_callers(_T,T) + comp_body(_T,T)) * min(_T.fc, T.fc)
        fin = _T.gt is T.gt
        for t,o in ((_T,T),(T,_T)):
            if fin: t.in_[o] = V
            else:   t.ex_[o] = V

    while True:
        for t in oF_: t._in_,t._ex_ = t.in_,t.ex_; t.in_,t.ex_ = {},{}
        # reassign each non-exe to best gt by mean V over candidate group
        moved = 0
        for T in oF_:
            if T.exe: continue
            gV = {}  # gt -> [sumV, count]
            for d in (T._in_, T._ex_):
                for m,v in d.items():
                    if m.gt not in gV: gV[m.gt] = [0,0]
                    gV[m.gt][0] += v; gV[m.gt][1] += 1
            bGT = max(gV, key=lambda g: gV[g][0]/gV[g][1])
            if bGT is not T.gt: T.gt = bGT; moved = 1
        # repartition in_/ex_ from new gt (V values from snapshot)
        for _T,T in combinations(oF_,2):
            V = _T._in_.get(T, _T._ex_.get(T))
            fin = _T.gt is T.gt
            for t,o in ((_T,T),(T,_T)):
                if fin: t.in_[o] = V
                else:   t.ex_[o] = V
        # convergence: total intra-V change
        Mt  = sum(sum(t.in_.values())  for t in oF_) / 2
        _Mt = sum(sum(t._in_.values()) for t in oF_) / 2
        if not moved or Mt - _Mt < ave: break

    new_oF_ = []
    for E in E_:
        N_ = [t for t in oF_ if t.gt is E.gt]
        gV = sum(sum(t.in_.values()) for t in N_) / 2
        gC = sum(t.fc for t in N_)
        nT = sum2O(N_); nT.memb = gV; nT.cmpr = gC
        new_oF_ += [nT]
    return new_oF_
```

Mapping to cluster_P:
- `_md__` / `md__` ↔ `_in_/_ex_` / `in_/ex_` (snapshot vs current partition)
- `base_comp(C,N)` recomputed each iter ↔ V values don't change, only their partition by gt
- `sum2F(N_, root, m_, d_)` reforming C ↔ no group reformation; gt is shared ref, persists
- `Mt*(dM+dD)` improvement gate ↔ `Mt - _Mt` (simpler; no D analog since V is symmetric)

Open items:
1. **Initial `t.gt` setup**: needs `E.gt = [E, 0]` for each exemplar at oF_ construction, and `t.gt = nearest_E.gt` for non-exemplars via raw V (one-pass bootstrap). Should bootstrap live inside `cluster_oF_` as a pre-loop step, or external?
2. **gt[1] update**: V slot in gt isn't used in the reassignment math. Update it from group's intra-V per iteration for downstream consumers? Or leave for whoever sets gt initially?
3. **Convergence threshold**: `Mt - _Mt < ave` is too loose; cluster_P uses `Mt*(dM+dD)*wcP*L > ave*(r+ccP*L)`. Want me to lift that form, or keep simple-difference?

boris-kz: I need to redefine in_,ex_, it's more like pairwise forks (diffs) from comp_body?

claude

Yes, the reframe is right. In fuzzy mode, in_/ex_ should be per-pair structural attributes, not group-partition labels.

The split that makes sense: comp_body returns (M, D) instead of a single C — M for matching tokens, D for fork/diff costs (the `-= 2` accumulations are already conceptually D, just collapsed into C with a sign). Same for comp_callers: (M, D) instead of M/D ratio. Per pair, you then have a match component and a diff component:

- `t.in_[o] = m_pair` — alignment between t and o (positive evidence for clustering)
- `t.ex_[o] = d_pair` — divergence/forks between t and o (negative evidence)

Both fixed at cross-comp time, never repartitioned. The group structure lives in *weights*, not in which dict a pair lives in.

What fuzzy refinement updates: soft membership weights `u(T, gt)` for each T and each candidate gt. The pairwise m/d are fixed; only the u's iterate. Group affinity becomes a weighted aggregate:

  `aff(T, gt) = Σ_m u(m, gt) · in_[m]  −  α · Σ_m u(m, gt) · ex_[m]`

Then update u(T, gt) ∝ aff(T, gt), normalize across gts per T. Exemplars stay pinned at u(E, E.gt) = 1. Iterate until u's stabilize. Final hard assignment = argmax u per T.

This is closer to cluster_P now, but for a different reason than I forced earlier: cluster_P iterates m/d weights against fixed N_; this iterates u-weights against fixed pairwise (m, d). Both are "FCM-style on a fixed substrate."

Before drafting: a few decisions worth pinning down so I don't guess wrong on style and structure:

1. **Membership representation**: `T.u_ = {gt: weight}` per node, or a global `u__` matrix? Per-node dict matches the existing `caller_`, `in_`, `ex_` style.

2. **α in `aff = M − α·D`**: a global constant like `ave`, or learned per-iteration?

3. **Normalization**: standard FCM uses `u_ij = 1 / Σ_k (d_ij/d_ik)^(2/(p-1))` with fuzzifier `p`. Or simpler — normalize raw aff to sum to 1 per T. Which form fits your accounting?

4. **comp_body split into (M, D)**: I can sketch the modification — D accumulates the `-= 2` cost terms as `+= 2`, M accumulates the positive contributions. Want me to do that, or do you want to write it yourself given it's a small diff to existing comp_body?

kwcckw: > you keep doing this global grpt_, which is a total mess.
we only need T.grpt

Okay right, the method above with _ex and _in is actually looks better. Also in this section:

            gV = {}  # gt -> [sumV, count]
            for d in (T._in_, T._ex_):
                for m,v in d.items():
                    if m.gt not in gV: gV[m.gt] = [0,0]
                    gV[m.gt][0] += v; gV[m.gt][1] += 1
            bGT = max(gV, key=lambda g: gV[g][0]/gV[g][1])
            if bGT is not T.gt: T.gt = bGT; moved = 1


I think we not even need gV? `T._in_` and `T._ex_` should be unique, so we just need to get the best `gt` by finding the max V across the combined T._in_ and T_ex_ .



>  t.in_[o] = m_pair — alignment between t and o (positive evidence for clustering)
>  t.ex_[o] = d_pair — divergence/forks between t and o (negative evidence)

> Both fixed at cross-comp time, never repartitioned. The group structure lives in weights, not in which dict a pair lives in.

Their value is fixed at runtime, but the positional information such as the in_ or ex_ could be differ when we move Ts to different groups? For example, a pair could be in `t._ex_` in the first iteration, but it should be in `t._in_` when the other t moved to the same group as t?

boris-kz: Sorry, that in/ex split is just reinforcing similarity that we already know, which is degenerative?

boris-kz: Yes,

Within-level iteration over fixed pairwise V can do two things:
- **Smooth / diffuse** the similarity signal (spectral-style, transitive similarity through the graph). Generates info not directly in raw V but bounded by what V *implicitly* contains via second-order paths.
- **Reinforce** the ranking (FCM-style, your "degenerative" case). Converges to argmax_E V(T, E), same fixed point as one-pass.

Neither adds *structural* depth — they're both projections of the existing pairwise matrix. For an algorithm whose purpose is to build a deeper hierarchy of representations, this is degenerative in the meaningful sense: no new compression emerges, no new comparison substrate exists.

The genuine refinement path requires **substrate change**, which means forming composite objects whose internal structure is emergent:
- Merge member ASTs into a composite body where shared substructure is preserved and divergence is marked as fork
- `comp_body(T, composite)` then produces (m, d) that aren't recoverable from any `comp_body(T, member_i)` — the composite has forks where no individual member does
- This new (m, d) is genuine information, not a re-projection

That's what cluster_C / cluster_P do via `sum2F` — they form structural centroids and re-comp against them. Avg-linkage has no equivalent: it operates on pre-computed pairwise V and never constructs a new comparison target.

So the conclusion is structural, not just operational:

- **Avg-linkage is a terminal operation**, not a recursive one. It's good for one-shot bucketing of an existing similarity matrix. It doesn't deepen.
- **Recursion / deepening requires compositing**, which moves you to centroid-style clustering by necessity. The composite IS the new substrate.

For cluster_oF_ specifically, this means:
- One-pass hard assignment (T → argmax aggregated-V over gt members) is the right shape for this level
- No within-level iteration buys anything structurally meaningful
- Refinement happens at the **next level up**, when these composites (formed via sum2O / AST-merge) get re-clustered. At that level the substrate has genuinely changed: nodes are composites with new ASTs, pairwise V is computed against new structures, fresh forks emerge

So the hierarchical recursion is: one-shot avg-linkage at each level + composite formation between levels. The composite formation is where structure gets built; clustering at each level just buckets the current population. This matches the broader CogAlg pattern: connectivity / similarity bucketing is cheap at-level; centroid / composite formation is the deepening operation.

Concretely for the code: cluster_oF_ collapses to a single pass — compute pairwise (m, d), each non-exe picks argmax-affinity gt, sum2O forms composites, done. The while-loop goes away. Iteration only re-enters when the composites get re-clustered at the next call up.

kwcckw: > Sorry, that in/ex split is just reinforcing similarity that we already know, which is degenerative?

We do need that in and ex information since they will be used separately? And that in and ex will be reassigned during the iteration, which is important. By using a single V and without new comp, we can't really refine them iteratively?

boris-kz: it will just keep existing grps, biased toward their exemplars. That's what degenerative means

frame_2D_alg/meta_code.py
boris-kz:
```python
def cluster_oF_():  # avg-linkage init via exemplars, cluster_P-like fuzzy refinement

    def comp_body(_n, n, C=0): ...   # unchanged
    def comp_callers(_T, T): ...      # unchanged

    # --- init: cross-comp all pairs, cache pairwise V (single value, no m/d) ---
    E_ = []
    for t in oF_:
        t.caller_ = set(t.caller_); t.V_ = {}  # pairwise V to every other oF
        if t.exe: E_ += [t]
    for _T,T in combinations(oF_,2):
        V = (comp_callers(_T,T) + comp_body(_T,T)) * min(_T.fc, T.fc)
        _T.V_[T] = V; T.V_[_T] = V

    # one-shot avg-linkage: each non-exe joins exemplar with max V to it
    # (avg over members trivially reduces to nearest-exemplar at init since member=[E])
    G_ = [[E] for E in E_]
    for T in oF_:
        if T.exe: continue
        G_[np.argmax([T.V_[E] for E in E_])] += [T]
    C_ = [sum2O(G) for G in G_]  # initial composites from one-shot grouping

    # --- refinement: re-comp T against each composite, reform composites, iterate ---
    # this is non-degenerative because each C is an AST-merge of its members,
    # so comp_body(C,T) produces V that's NOT recoverable from any T.V_[member]
    Ln, Lc = len(oF_), len(C_); L = Ln*Lc
    _v__ = np.zeros((Ln, Lc))
    while True:
        v__ = np.zeros((Ln, Lc))
        for j,T in enumerate(oF_):
            for i,C in enumerate(C_):  # fresh V per iter: substrate (C) changed
        C_ = [sum2O(oF_, v__[:,i]) for i in range(Lc)]  # reform: every T contributes weighted by its affinity
        Vt, dV = v__.sum(), np.abs(v__ - _v__).sum()
        if Vt * dV * wcP*L <= ave * (Ln + ccP*L): break  # cluster_P convergence pattern
        _v__ = v__

    # --- finalize: hard assign via argmax, form output composites ---
    new_oF_ = []
    for i in range(Lc):
        N_ = [T for j,T in enumerate(oF_) if v__[j].argmax() == i]
        if not N_: continue
        gV = v__[:,i].sum(); gC = sum(t.fc for t in N_)
        nT = sum2O(N_); nT.memb = gV; nT.cmpr = gC
        new_oF_ += [nT]
    return new_oF_
```

Notes on what shifted from earlier drafts:

- **in_/ex_ dropped.** They were a group-partition view; here every T compares to every C in every iter, so the matrix `v__` is natural. No partition because there's no "outside" — full Lc-way fuzzy weights per T.
- **V is single-valued.** No (m, d) split since comp_body returns one C.
- **Substrate change is the composites.** Each iter, `C_` is reformed via `sum2O` weighted by `v__`, so `comp_body(C,T)` is computing against a different AST each pass — that's the new information, not a re-projection of the pairwise V cache.
- **Exemplars aren't pinned** explicitly; they retain influence through their strong self-V in the initial `v__` and their structural anchor in their group's `sum2O`. If they drift too far, that's a sign the exemplar choice was wrong.

Open implementation questions:

1. **`sum2O(N_, weights)` signature**: cluster_P calls `sum2F(N_, root, m_, d_)` but in current `sum2F` the actual aggregation weight is `n.c`, not `m_/d_` — those just get stored on the centroid. Two interpretations: (a) `sum2O` needs a real weighted-aggregate path for fuzzy refinement to differentiate composites, or (b) intrinsic-weighted aggregation is intentional and the per-composite differentiation comes only through stored membership profiles. Which one matches your intent? If (a), `sum2O` needs the weights to drive the merge.

2. **`comp_body(C, T)` for composite C**: does C from `sum2O` have a body field that `comp_body` recurses into? If `sum2O` produces an AST-merged body (shared subtrees preserved, divergent subtrees marked fork), this works directly. If C only has summary stats (no body), then comp_body has to fall back to comparing T against C's *members* — which collapses back to the degenerative re-projection.

(2) is the load-bearing one. The non-degenerative claim depends entirely on `sum2O` producing a composite with a real merged body. If it doesn't, the iteration adds nothing over one-shot.

frame_2D_alg/meta_code.py
kwcckw: I see, then i think we should generate new V with comps within the iterations, and probably remove the exemplars, similar as cluster_P. But shouldn't be normal for them to be biased towards the exemplars? Else we wouldn't need exemplars in the first place.

boris-kz: It doesn't add any info. the only way to form novel clusters is to form centroids, as above

kwcckw: > Refinement happens at the next level up, when these composites (formed via sum2O / AST-merge) get re-clustered. At that level the substrate has genuinely changed: nodes are composites with new ASTs, pairwise V is computed against new structures, fresh forks emerge

Do you referring to this? So the actual refinement should be across levels, which is between frames here?

boris-kz: It's not levels, just loops, like in cluster_P.
I pushed a draft

kwcckw: Thanks, that's much clearer now. Also i think exemplars is not needed since we assign weight with every oF anyway? Also right now the final oF is limited to the number of `C_ ` , which is same as the exemplars length.

boris-kz: Yes, now we need to initialize with manual grps, not exemplars

PR #156, opened 2026-05-26
5/26/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -245,16 +248,17 @@ def comp_callers(_T, T):
...
             for i,C in enumerate(C_):  # fresh V per iter: substrate (C) changed
                 v__[j,i] = (comp_callers(C,T) + comp_body(C,T)) * min(C.fc, T.fc)
-        C_ = [sum2O(oF_, v__[:,i]) for i in range(Lc)]  # reform: every T contributes weighted by its affinity
+        C_ = [sum2O(oF_, w_=v__[:,i]) for i in range(Lc)]  # reform: every T contributes weighted by its affinity
         Vt, dV = v__.sum(), np.abs(v__ - _v__).sum()
-        if Vt * dV * wcP*L <= ave * (Ln + ccP*L): break  # cluster_P convergence pattern
+        if Vt * dV * wcO*L <= ave * (Ln + ccO*L): break  # cluster_P convergence pattern
         _v__ = v__
     new_oF_ = []  # finalize:
     for i in range(Lc):
```
kwcckw, line 256: >  Also right now the final oF is limited to the number of C_ , which is same as the exemplars length.

So Lc is the length of exemplars, and in this scheme, the new oF should be limited to this length only, is this intended? Also following the method in cluster_P, all Ts must be assigned to one of the exemplar cluster, so there's no recycled Ts at all.

frame_2D_alg/meta_code.py
```diff
@@ -245,16 +248,17 @@ def comp_callers(_T, T):
...
-        N_ = [T for j,T in enumerate(oF_) if v__[j].argmax() == i]  # hard assign
+        NV_ = [(T,v__[j,i])  for j,T in enumerate(oF_) if v__[j].argmax() == i]  # hard assign
+        N_, V_= zip(*NV_)
         if not N_: continue
         gV = v__[:,i].sum(); gC = sum(t.fc for t in N_)
-        nT = sum2O(N_); nT.memb = gV; nT.cmpr = gC
+        nT = sum2O(N_, w_=np.array(V_)); nT.memb = gV; nT.cmpr = gC
```
kwcckw, line 261: If we follow cluster_P, we should get the membership weight for each N_, and this w_ mainly will be used to adjust fw only? I don't think we will get dTT in type oF. For fr, i'm not sure yet, could you clarify?

frame_2D_alg/meta_code.py
```diff
@@ -245,16 +248,17 @@ def comp_callers(_T, T):
         for j,T in enumerate(oF_):
             for i,C in enumerate(C_):  # fresh V per iter: substrate (C) changed
                 v__[j,i] = (comp_callers(C,T) + comp_body(C,T)) * min(C.fc, T.fc)
-        C_ = [sum2O(oF_, v__[:,i]) for i in range(Lc)]  # reform: every T contributes weighted by its affinity
+        C_ = [sum2O(oF_, w_=v__[:,i]) for i in range(Lc)]  # reform: every T contributes weighted by its affinity
         Vt, dV = v__.sum(), np.abs(v__ - _v__).sum()
-        if Vt * dV * wcP*L <= ave * (Ln + ccP*L): break  # cluster_P convergence pattern
+        if Vt * dV * wcO*L <= ave * (Ln + ccO*L): break  # cluster_P convergence pattern
```
kwcckw, line 253: For code level wcO and ccO, it should be just a constant? Or they need to follow the same feedback process as data w and c?

boris-kz: constant for now, they are higher-order than oF filters, should be updated per something like frame_HHH

frame_2D_alg/meta_code.py
boris-kz: yes, fixed per level. recycling should be the final step: unpack weak grps after refining.

frame_2D_alg/meta_code.py
```diff
@@ -290,18 +294,24 @@ def build(node):  # AST → CoF | (type,sub_) | ast_leaf | None
...
                 t, fc = rett; F.body += [t]; F.fc += fc
 
-def sum2O(N_, root=None, fdata=0):  # for w,c,r, fw,fc,fr only?
+def sum2O(N_, root=None, w_=None, fdata=0):  # for w,c,r, fw,fc,fr only?
 
-    c_ = np.array([n.c for n in N_], dtype=float); C = c_.sum(); w_ = c_/C
-    fc_ = np.array([n.fc for n in N_], dtype=float); fC = fc_.sum(); w_ = fc_/fC  # N = N_[0]
-    # unfinished, use w_ for N summing?
-    return CoF(call_=N_, root=root, fC=fC  )  # fw = Fw
+    if fdata:  # this should be for data only?
```
boris-kz, line 305: right

frame_2D_alg/meta_code.py
boris-kz: fr is for nested oFs, because nested levels are redundant.
fw is probably for nesting too, it's a higher oF membership weight  
for data we have plain w

kwcckw: Okay, so we need to increase fr whenever we pack new T with sum2O? 

For fw, it will be increased based on their fc? For every new nested level, fw is increased by fc? The new oF.fw = existing fw + fc?

frame_2D_alg/meta_code.py
kwcckw: I see, so there should be another section to check before this line:

    nT = sum2O(N_, w_=np.array(V_)); nT.memb = gV; nT.cmpr = gC
    
 That should be eval based on their V

frame_2D_alg/meta_code.py
boris-kz: oF.fw is actually current oF.V, specific to the final grp

kwcckw: Okay, since the process weight is V here, we probably can assign it in sum2O?

boris-kz: yes, assign to all element oFs in oG.N_? this is static, so oG.call_ should be spliced from these oFs?

kwcckw: Yes, so oG.call_ should be those oFs since data should be packed in oG.N_. Each element in V_ is corresponding to each N_'s fw, and oG.fw should be summed from V_ and reassigned again in the next frame?

boris-kz: no, the "data" here is oFs, oG is the higher order oF

kwcckw: Sorry, those calls should be in call_ only now, i was confused with the prior N_'s scheme earlier.

frame_2D_alg/meta_code.py
boris-kz, line 230: actually, initial exemplars are probably more arbitrary than initial groups, so

```
for t in oF_: t.caller_ = set(t.caller_); t.V_ = {}  # pairwise V to every other oF
for _T,T in combinations(oF_,2):
    V = (comp_callers(_T,T) + comp_body(_T,T)) * min(_T.fc, T.fc)
    _T.V_[T] = V
    T.V_[_T] = V
grp_= []; _T = oF_[0]; grp=[_T]
for T in oF_[1:]:
    if T.typ==_T.typ: grp+=[T]
    else: grp_+=[grp]; grp=[T]
    _T=T
C_ = [sum2O(grp) for grp in grp_]  # initial composites

```

boris-kz: Claude — initial groups are free from oF_ source ordering, no exemplar tagging needed. Full draft slotting your init into the refinement:

```python
def cluster_oF_():  # init by typ-adjacency, refine cluster_P-like

    def comp_body(_n, n, C=0): ...   # unchanged
    def comp_callers(_T, T): ...      # unchanged

    # cross-comp all pairs, cache pairwise V (carried for downstream; unused in init)
    for t in oF_: t.caller_ = set(t.caller_); t.V_ = {}
    for _T,T in combinations(oF_,2):
        V = (comp_callers(_T,T) + comp_body(_T,T)) * min(_T.fc, T.fc)
        _T.V_[T] = V; T.V_[_T] = V

    # init: contiguous same-typ runs in oF_ -> initial groups
    grp_ = []; _T = oF_[0]; grp = [_T]
    for T in oF_[1:]:
        if T.typ == _T.typ: grp += [T]
        else: grp_ += [grp]; grp = [T]
        _T = T
    grp_ += [grp]   # !! your snippet drops the final group
    C_ = [sum2O(g) for g in grp_]

    # refinement: comp each T against each reformed composite
    # substrate (C_) changes per iter via sum2O re-aggregation -> non-degenerative
    Ln, Lc = len(oF_), len(C_); L = Ln*Lc
    _v__ = np.zeros((Ln, Lc))
    while True:
        v__ = np.zeros((Ln, Lc))
        for j,T in enumerate(oF_):
            for i,C in enumerate(C_):
        C_ = [sum2O(oF_, v__[:,i]) for i in range(Lc)]  # weighted re-aggregation
        Vt, dV = v__.sum(), np.abs(v__ - _v__).sum()
        if Vt * dV * wcP*L <= ave * (Ln + ccP*L): break  # cluster_P convergence
        _v__ = v__

    # final hard assignment via argmax
    new_oF_ = []
    for i in range(Lc):
        N_ = [T for j,T in enumerate(oF_) if v__[j].argmax() == i]
        if not N_: continue
        gV = v__[:,i].sum(); gC = sum(t.fc for t in N_)
        nT = sum2O(N_); nT.memb = gV; nT.cmpr = gC
        new_oF_ += [nT]
    return new_oF_
```

Three things worth confirming:

1. **Pairwise V cache (`t.V_`)** is built but never used in this function — only init uses typ-adjacency, refinement comps against composites. Keeping it for downstream consumers, or drop?

2. **Same `sum2O` weighted-aggregation question** as before. `sum2O(oF_, v__[:,i])` assumes the second arg is per-member weights driving the AST-merge. If `sum2O` currently doesn't take weights, every C in C_ ends up identical at each iter (since all are aggregated from the full oF_), and refinement collapses. Whether `sum2O` produces meaningfully different composites per group is load-bearing.

3. **Refinement could be skipped entirely** if typ-adjacency is the intended final partition. The whole loop earns its keep only when typ alone undersegments — e.g., when two functions share typ but are structurally distant, or two functions of different typ are structurally close. Is that condition expected in practice, or is typ the right partition already and refinement is overkill at this level?

frame_2D_alg/meta_code.py
boris-kz: yes, 
    
    for t in oF_: t.caller_ = set(t.caller_); t.V_ = {}  # pairwise V to every other oF
    for _T,T in combinations(oF_,2):
        V = (comp_callers(_T,T) + comp_body(_T,T)) * min(_T.fc, T.fc)
        _T.V_[T] = V
        T.V_[_T] = V
    grp_= []; _T = oF_[0]; grp=[_T]
    for T in oF_[1:]:
        if T.typ==_T.typ: grp+=[T]
        else: grp_+=[grp]; grp=[T]
        _T=T
    grp_ += [grp]
    C_ = [sum2O(g) for g in grp_]  # initial composites
    Ln,Lc = len(oF_),len(C_); L = Ln*Lc
    _v__ = np.zeros((Ln, Lc))
    # cluster_P analog:
    while True:
        v__ = np.zeros((Ln, Lc))
        for j,T in enumerate(oF_):  # refine by comp T x G_
            for i,C in enumerate(C_):
        C_ = [sum2O(oF_, w_=v__[:,i]) for i in range(Lc)]  # weighted re-aggregation
        V, dV = v__.sum(), np.abs(v__-_v__).sum()
        if V*dV*(wcO*L) <= ave * (Ln+ ccO*L): break  # convergence
        _v__ = v__
    new_oF_ = []  # final hard assign
    for i in range(Lc):
        t_ = [T for j,T in enumerate(oF_) if v__[j].argmax() == i]
        if t_:
                nT = sum2O(t_); nT.memb = gV; nT.cmpr = sum(t.fc for t in t_) / t.fc
                new_oF_ += [nT]
    return new_oF_

frame_2D_alg/meta_code.py
kwcckw: > actually, initial exemplars are probably more arbitrary than initial groups, so

But we have all different typs in the first iteration, and this `if T.typ==_T.typ: grp+=[T]` is only true in the subsequent frames. So actually it should be the same as cluster_P in the first iteration, where each T should init their own single T's grp.

boris-kz: no, the typ here should be set manually: comp_, clust_, root_, conv_

kwcckw: I see, thanks. I misunderstood it as nF earlier.

So when we have new composite oF, how do we really set their typ? Or we will be comparing their nested oF's typ instead?

boris-kz: we don't have a process for composite oF yet, right now it's only 5 of them, no need for clustering:

`_names = ['frame_H','cross_comp','trace_edge',                 # root_
          'comp_N_','comp_C_','comp_N','comp_F',               # comp_
          'get_exemplars','cluster_N','cluster_C','cluster_P', # clus_
          'ffeedback','proj_N',                                # proj_
          'vect_edge','comp_slice','slice_edge']               # prep_`

boris-kz: Next, we need to add splitting / divisive clustering per oF, not sure how much of cluster_oF_ we can reuse there

kwcckw: > Next, we need to add splitting / divisive clustering per oF, not sure how much of cluster_oF_ we can reuse there

That's something similar with sub+? But why we need that when we only need the new oFs (oG) in the next frame? Divisive clustering is just recluster those oG.call_ again?

boris-kz: it's oF, not oG, the split() we had before, except for the body. 
it's symmetrical to clustering, for the same reason: reduce the costs

`typ_= ['root_','root_','root_','comp_','comp_','comp_','comp_','clus_','clus_','clus_','clus_','proj_','proj_','prep_','prep_','prep_']`
`oF_ = [CoF(nF=i,typ=typ) for i,typ in enumerate(typ_)]  `

boris-kz: Re-mapping reuse for intra-body mining (single oF, find recurring subtrees, fold into sub-CoFs):

**Reusable, mostly as-is:**
- `comp_body` — natural fit for pairwise subtree similarity. Already returns a compression score, which is exactly the merge criterion.
- Hard-assign tail logic — straightforward.

**Not reusable / different:**
- `comp_callers` — gone. Subtrees aren't called.
- Initial typ grouping — replace with grouping by root `type(node)` (AST node class). `comp_body` already returns ≤ ave_C for mismatched roots, so this is just an efficiency prefilter.
- EM loop with `_v__/v__` — questionable for this task. BPE/Sequitur-style mining is exact-match dominated; the EM refinement is for fuzzy centroids over a continuous similarity. For body mining you mostly want either "same subtree shape" or "same with k-position diffs", which is closer to hashing-then-grouping than to fuzzy centroids. Could still run a fuzzy pass after, but it's optional, not the spine.
- `sum2O` — needs a variant: build a sub-CoF from AST-subtree members, no `caller_`, body = the shared subtree, `fc` = subtree cost × occurrences saved.

**Missing piece, the actual hard part:** enumeration. cluster_oF_ has its population given (oF_). Body mining must produce one. Two enumerators to choose between:

- **Linear (BPE-like):** scan body as a flat sequence, look for repeated contiguous n-grams of length ≥2. Misses non-contiguous recurrence but is cheap and matches what AST sequences actually look like at the top level of a function.
- **Tree (subtree-set):** walk every `(type, sub_)` node at every depth, hash by shape, group identical/near-identical. Catches all recurrence but explodes; needs a floor like `fc > ave_C` and depth/size limits to be tractable.

Probably want tree, with linear as the cheap early pass.

**Conceptual issues to settle before writing:**

1. **Destructive or not?** Per your existing principle (merge/split destructive, clustering non-destructive) — folding a recurring subtree into a sub-CoF reference rewrites the body. Is that a merge-class op, or a new clustering-class op that leaves the original body intact and stores the discovered patterns alongside?

2. **Floor.** What's the minimum subtree size/cost worth extracting? `fc > ave_C` is the obvious gate from existing code, but for short bodies even ave_C might be too high.

3. **Order vs cluster_oF_.** Intra-body mining before `cluster_oF_` shrinks bodies and surfaces shared sub-CoFs as alignable units across functions — probably what you want. After is also coherent (refine each composite's representative). Which?

4. **CoF-ref elements in body.** Already-called CoFs (the `oF_[i], 3` branch in `build`) — are they mineable like AST tuples, or treated as atoms during intra-body mining?

Lean toward: pick tree enumeration + non-destructive (store extracted sub-CoFs in a new attribute like `oF.sub_call_` rather than rewriting `body`) + run before cluster_oF_ + treat CoF refs as atoms. Confirm or push back.

kwcckw: > for the same reason: reduce the costs

So the fc is expected to be reduced when we recluster them since it should have higher compression? But now new oF's fc is summed from their calls's fc, which is the same? We need to add another process to optimize their fc during the reclustering process?

boris-kz: I am talking about splitting original oFs
smaller functions take less memory and processing, so we need to split the code forks that are rarely called together

kwcckw: I see, that makes sense now since after the split, those oFs should get lower fc since it will be repack to global oF_ instead of the original oF.call_.

boris-kz: Just think about how we split some functions manually

Claude:
Re-reading: the population for divisive sub+ is `oF.call_` (the CoFs called by this oF), not the AST primitives in body. The primitives just don't carry the right signal — only CoFs have `caller_`. So body is the scope but call_ is the clustered set.

If that's right, then **cluster_oF_ is almost entirely reusable**, just re-parameterize the population:

```
cluster_oF_(oF_)            → cluster_call_(F.call_)
comp_callers, comp_body     → unchanged, both apply
typ grouping seed           → likely skip (call_ is small, typ already homogeneous within a single F)
EM loop                     → unchanged
sum2O                       → unchanged (CoF members carry caller_)
hard-assign                 → unchanged
```

Essentially `cluster_oF_` becomes a special case (`F = frame_H`-like root with `call_ == oF_`). The split operation is the same routine restricted to one F's callees, called recursively where it pays off (Vn-style gate analogous to sub+ in `sum2G`).

**Conceptual issues:**

1. **Body's actual role.** If clustering is purely over `call_` and signal is caller_-set overlap, body is unused. But body carries the positional/adjacency information (which callees appear next to which, surrounding AST shape). Two callees adjacent in body are co-called *in this F* — that's a within-F co-occurrence signal cluster_oF_'s global caller_ misses. Should adjacency-in-body contribute, or is it ignorable as redundant with caller_ overlap (which already captures co-occurrence across all Fs)?

2. **Recursion structure.** sub+ in agg_recursion descends one composition level (G → G.N_). Here the descent is F → F.call_, then each clustered sub-group's sub+ would be... over what? The members are CoFs that have their own call_, so the recursion is F.call_ → (cluster) → C.N_ where each C.N_ entry has its own call_, descend into the strongest, repeat. Same pattern. Confirm.

3. **Gate.** sub+ is gated by Vn from Lt stats. Here there's no Lt yet — what's the gate? Sum of caller_-overlap V across pairs in call_, vs ave * (r + c * L)? Need to define V_call_ analog before recursing.

4. **Cross-F vs intra-F partition.** A callee can be called by many Fs. cluster_oF_ groups callees globally. Intra-F cluster_call_ groups them per caller. These can disagree: globally A∼B but within F's call_ they're far apart in body. Which wins, or do both produce separate hierarchies (per-F local subgroups *and* global groups)?

Lean answer to (1) and (4): keep cluster_call_ pure caller_-overlap for now (matches your "more like comp_callers" phrasing), ignore body adjacency, accept that local and global groupings coexist. Confirm.

kwcckw: So the splitting comparison process should be different since original oF.call_ are those calls added from runtime, and their caller should be root instead. 

But all calls under oF should have the same typ and nF, the only difference should be just their caller, or the root.

boris-kz: nF will be different, these are different functions with different bodies:

`typ_= ['root_','root_','root_','comp_','comp_','comp_','comp_','clus_','clus_','clus_','clus_','proj_','proj_','prep_','prep_','prep_']
# need to add oF-specific body:
oF_ = [CoF(nF=i,typ=typ) for i,typ in enumerate(typ_)]`

kwcckw: > Just think about how we split some functions manually

> nF will be different, these are different functions with different bodies

Okay, so this is actually splitting the body instead of calls, which we probably need a separated process to compare and merge different sequential bodies?

boris-kz: merge is cluster_oF_

kwcckw: Yes, i meant after we split the bodies, we need to merge them into sub oF?

Also could you clarify if the `body` here is referring to the primitives? Because other than the primitives in oF.body, i don't see how we can split them like splitting functions manually?

boris-kz: split whatever is called separately, no merging here

PR #157, opened 2026-05-27
5/27/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -236,7 +236,7 @@ def comp_callers(_T, T):
         T.V_[_T] = V
     grp_= []; _T = oF_[0]; grp=[_T]
     for T in oF_[1:]:
-        if T.typ==_T.typ: grp+=[T]
+        if T.typ==_T.typ and _T.V_[T] > ave: grp+=[T]  # include _T.V_[T]? Else the pairwise T.V_ section above is actually redundant
```
kwcckw, line 239: That pairwise `T.V_` added above is actually redundant unless we use it here? Else we should remove them.

frame_2D_alg/meta_code.py
```diff
@@ -258,11 +258,53 @@ def comp_callers(_T, T):
         t_ = [T for j,T in enumerate(oF_) if v__[j].argmax() == i]
         if t_:
             if (gV := v__[:,i].sum()) * ((len(t_)-1)*wL) > ave:
-                nT = sum2O(t_); nT.memb = gV; nT.cmpr = sum(t.fc for t in t_) / t.fc
+                nT = sum2O(t_); nT.memb = gV; nT.cmpr = sum(t.fc for t in t_)  # compression should be just sum of fc?
```
kwcckw, line 261: The `nT.compr` should be just sum of t_'s fc? Or it should be average:

      nT.cmpr = sum(t.fc for t in t_) /len(t_)

frame_2D_alg/meta_code.py
```diff
@@ -304,12 +346,17 @@ def build(node):  # AST → CoF | (type,sub_) | ast_leaf | None
 def sum2O(N_, root=None, w_=None, fdata=0):  # for w,c,r, fw,fc,fr only?
 
     if fdata:
-        c_ = np.array([n.c for n in N_], dtype=float); C = c_.sum(); w_ = c_/C
+        c_ = np.array([n.c for n in N_], dtype=float); C = c_.sum(); w_ = c_/C; body = []
+    else:
+        body = [prim for N in N_ for prim in N.body]  # add body for centroid (not sure but splice oFs' body?)
```
kwcckw, line 351: So looks like we need to add body to the centroid, else we won't have C.body to comp in the iterations?

To add their body, i guess we can just concatenate oF's body?

frame_2D_alg/meta_code.py
```diff
@@ -258,11 +258,53 @@ def comp_callers(_T, T):
...
         if t_:
             if (gV := v__[:,i].sum()) * ((len(t_)-1)*wL) > ave:
-                nT = sum2O(t_); nT.memb = gV; nT.cmpr = sum(t.fc for t in t_) / t.fc
+                nT = sum2O(t_); nT.memb = gV; nT.cmpr = sum(t.fc for t in t_)  # compression should be just sum of fc?
                 new_oF_ += [nT]
             else: new_oF_ += t_  # unpack if weak
     return new_oF_
 
+
+def split_oF_():  # divisive clustering
```
kwcckw, line 267: I added draft of split here, so it's basically follow the structure of cluster_oF_, but without the refinement iterations. 

It mainly grouping the same type's body to sub oFs. Also the comp_prim is tentative since it's just checking their typ.

frame_2D_alg/meta_code.py
boris-kz: that was for average linkage, which is not used now. 
Stlill meaningful, probably good here.
May also combine with centroid match in refinining, it's the same logic?

frame_2D_alg/meta_code.py
boris-kz: it's not a pure gain, we have to create new oF instead, I used / t.fc for that

frame_2D_alg/meta_code.py
boris-kz: this splitting is mostly about separating grouped forks into sub-oFs

kwcckw: > this splitting is mostly about separating grouped forks into sub-oFs

I see, so the original oF_ doesn't have grouped forks, it should be for the merged grp only?

frame_2D_alg/meta_code.py
kwcckw: But why `t.fc`? That `t` is the last `t` in `t_` when we run `sum(t.fc for t in t_) / t.fc`?

frame_2D_alg/meta_code.py
kwcckw: > May also combine with centroid match in refinining, it's the same logic?

Okay, something like this to combine the pairwise v?

    for t in C.call_: v__[j,i] += t.V_[T]

frame_2D_alg/meta_code.py
boris-kz: grouping *is* splitting, and we have if-else forks in every function

frame_2D_alg/meta_code.py
boris-kz: for t in C.N_, why call_?

frame_2D_alg/meta_code.py
boris-kz: it's assuming new oF has the same fc, because we don't have that oF.fc yet.
actually better to get some average-oF.fc

frame_2D_alg/meta_code.py
boris-kz: that's meaningless, we need to compute average body, not sure how?

boris-kz: Chat:

The PR issue is here: concatenating member bodies in `sum2O` is not an “average body”; it just makes a longer body, so `comp_body(C,T)` then compares `T` to a bag of all member operations rather than to a representative code pattern. The discussion already converged on that problem: kwcckw suggested concatenating bodies for centroid comparison, and you rejected it as meaningless because the centroid needs an average body. ([[GitHub](https://github.com/boris-kz/CogAlg/pull/157/changes)][1]) The local `meta_code.py` already frames this as “compare aligned ops between oF_ AST sequences, cluster/merge matches into higher oF typs”, so the right object is an aligned structural centroid, not concatenation. 

Minimal rule:

```python
def sum2O(N_, root=None, w_=None, fdata=0):
    ...
    body = ave_body([N.body for N in N_], w_)
    fc = get_fc(body)  # not sum(fc_)
    oF = CoF(N_=N_, root=root, fc=fc, fw=fw, fr=N_[0].fr+1, body=body)
```

Conceptually, `ave_body` should do recursive weighted alignment:

1. **Align by body position first**, because these are code sequences. Do not globally sort or concatenate.
2. **At each aligned position**, compare primitives with existing `comp_body` / `comp_prim`.
3. **If same primitive kind dominates**, keep that primitive as centroid:

   * same `CoF` type or same `nF` -> average that referenced function/body recursively;
   * same AST tuple type -> keep `(type, ave_body(children))`;
   * same AST leaf type -> keep that leaf/type.
4. **If no dominant match**, form an explicit fork primitive, not a flat concatenation:

   ```python
   CoF(nF='E', N_=alts, fw=w_alt_, fc=fork_fc + weighted_alt_fc)
   ```

   That represents “average body has an if-fork here”.
5. **Drop weak alternatives** if their total weight is below threshold; otherwise they inflate centroid cost and destroy compression value.

So average body is basically **progressive multiple-sequence alignment with forks**, where the fork is the code analogue of fuzzy overlap.

The key distinction:

```python
body = [prim for N in N_ for prim in N.body]   # wrong: union / concatenation
```

should be replaced by something closer to:

```python
body = []
for prim_ in zip_longest(*(N.body for N in N_), fillvalue=None):
    body += [ave_prim([p for p in prim_ if p is not None], w_)]
```

But `zip_longest` is only the first approximation. If insertions/deletions matter, each position should choose between:

* match current primitives,
* fork them,
* skip low-weight unmatched tail.

For now, position-wise is probably good enough.

Also fix the representation split:

```python
CoF(N_=N_, call_=[], body=body, ...)
```

not:

```python
CoF(call_=N_, ...)
```

`call_` should mean actual nested runtime calls. Cluster members should be in `N_`. That also resolves the earlier comment: pairwise average-link addition in centroid refinement should use:

```python
for t in C.N_: v__[j,i] += t.V_[T]
```

not `C.call_`. The PR discussion explicitly hit this distinction. ([[GitHub](https://github.com/boris-kz/CogAlg/pull/157/changes)][1])

For `cmpr`, don’t use either raw sum or average as final value. Use saved cost:

```python
nT.fc = get_fc(nT.body)
nT.cmpr = sum(t.fc for t in t_) / nT.fc
```

or, if you want additive compression:

```python
nT.cmpr = sum(t.fc for t in t_) - nT.fc
```

The old `/ t.fc` is wrong because `t` is just the last loop variable, as kwcckw pointed out. The discussion then says the missing denominator should be some average/new-oF `fc`, which is exactly `get_fc(ave_body(...))`. ([[GitHub](https://github.com/boris-kz/CogAlg/pull/157/changes)][1])

So the clean answer:

```python
average_body = recursively aligned common body + explicit forks for mismatches
new_oF.fc   = cost(average_body)
compression = sum(member.fc) / new_oF.fc   # relative
# or
compression = sum(member.fc) - new_oF.fc   # absolute
```

Do not average `fc` directly. Compute `fc` from the averaged body.

[1]: https://github.com/boris-kz/CogAlg/pull/157/changes "5/27/2026 by kwcckw · Pull Request #157 · boris-kz/CogAlg · GitHub"

boris-kz: Claude, also on this discussion:

Now the question is clear: how to compute an "average body" for a centroid when bodies are heterogeneous trees of CoFs / `(type, sub_)` tuples / leaf types, of varying length. Concatenation isn't an average; it's a multiset.

The natural construction, reusing what already exists:

**Position-wise alignment + recurse-or-gate, weighted by fc.**

Walk member bodies in parallel by index. At each position i, collect the variants `{n.body[i] for n in N_}` with weights `w_[n]`. Three cases:

1. **All variants identical** (same CoF, or same `(type, sub_)` shape, or same leaf type): centroid keeps that element as-is.
2. **All tuples with the same root type**: recurse — centroid element is `(type, avg(sub_))` where avg is applied to the aligned sub_ lists.
3. **Anything else**: emit an E-gate, `CoF(nF='E', call_=[variants], w_=variant_weights)`. This is exactly the merge()-style disjunctive fork, just with weights attached.

Tail handling: longer bodies' overhanging positions become single-variant E-gates with weight = combined fc of the bodies that have something there, plus a "null" variant carrying the weight of bodies that ended early. Or simpler: pad short bodies with a sentinel and treat sentinels as a variant in case (3).

The centroid body is then alignable position-by-position against any member, which is what `comp_body` needs.

**comp_body extension** (minimal): when one side is an E-gate, score is `max_v (comp_body(variant_v, other) * w_v)` or a weighted sum, depending on whether you want hard or soft match. The existing E-gate branch in `merge()` already does max-like logic; just add the weight.

**Weights**: per-member `w_` from sum2O (already `fc/fC` by default). When a position has only k of the n members contributing a variant, normalize the gate's weights by the contributing subset's fc, and store the gate's own w = sum(contributing fc) / total fc — its "presence rate" in the cluster. That's the average's `c` analog at that position.

**Known limitation worth flagging:** index-based alignment is rigid. Two bodies with the same content shifted by one position look maximally different. DP/edit-distance alignment fixes this but is heavier. For v1 the rigid version is consistent with `merge()` and `comp_body`'s existing logic; revisit if it bites.

**Other PR-related comments since they're related:**

- `nT.cmpr = sum(t.fc for t in t_) / t.fc` reading `t.fc` post-loop is just bug-ish (uses last t) but you noted the right fix: divide by *average* t.fc, not a single t.fc. That gives the per-call compression ratio cleanly.
- The added `_T.V_[T] > ave` gate before adding to grp is fine, but as kwcckw flagged, then the pairwise V_ storage block above is only used here — if you also feed it into the EM refinement (`v__[j,i] += sum(t.V_[T] for t in C.N_)`), that's where it earns its keep. With C.N_ (not C.call_): N_ holds the members of the centroid, V_ is per-member pairwise V to other oFs, so summing over C.N_ gives the centroid-to-T value through its members — correct.
- `split_oF_` as drafted is a *single-pass run-grouping by typ-equality* within one body. That's a starting frame, but `comp_prim` reduces "match" to boolean type-equality, which throws away all the structural cost info that comp_body has. Eventually want `split` driven by the same comp_body score, not bare type equality — otherwise split and the merge half of cluster_oF_ use incompatible metrics.

frame_2D_alg/meta_code.py
kwcckw: We pack those oFs in call_ when we use sum2O, i guess we need to change it to N_ now.

frame_2D_alg/meta_code.py
kwcckw: Okay, so we need to merge their body with the same alignment. So in short, we will be reuse the same element in the body if they are the same, else we add gate to fork between both elements? And this gate should add more cost to the merged oF right?

    body = ave_body([N.body for N in N_], w_)
    fc = get_fc(body)  # not sum(fc_)
    oF = CoF(N_=N_, root=root, fc=fc, fw=fw, fr=N_[0].fr+1, body=body)

boris-kz: Yes, I think it's pretty much what we have in merge()?

kwcckw: Actually yes, and instead of call_, we need to use body here:

    if len(f.call_) > len(F.call_): offs = f.call_[i+1:]; call_+=offs; add_+=offs
    elif len(F.call_)>len(f.call_): call_+=F.call_[i+1:]

    if f.fc / (C or 1e-7) > ave:  # high individual_cost / forking_Cost, else keep old F,f
        if add_: add2O(F, sum2O(add_))
        F.call_= call_

boris-kz: yes, and we need to add new typ of oFs: secondary (gated) gate oFs

frame_2D_alg/meta_code.py
boris-kz: so we need to reconcile this with merge(), which is adding the forks you should be splitting by?
that merge wil be forming new body in centroids in sum2O, that's most of your split_oF_?
we need a new typ of gate oFs, it will collect the statistics that determine splitting,

frame_2D_alg/meta_code.py
```diff
@@ -304,12 +346,17 @@ def build(node):  # AST → CoF | (type,sub_) | ast_leaf | None
 def sum2O(N_, root=None, w_=None, fdata=0):  # for w,c,r, fw,fc,fr only?
 
     if fdata:
```
boris-kz, line 348: so it's not if-else, something like

def sum2O(F_, root=None, w_=None, fdata=0):  # for w,c,r, fw,fc,fr only?

    if fdata: c_ = np.array([n.c for n in F_], dtype=float); C = c_.sum(); w_ = c_/C; body = []
    Body = F_[0].body
    for f in F_: merge(Body, f.body, fsel=0)

boris-kz: Several issues:

1. **merge takes CoFs, you're passing lists.** `merge(Body, f.body)` → inside, `F.body` becomes `Body.body` (Body is a list). Should be `merge(F_[0], f, fsel=0)`.

2. **fsel=0 disables the commit.** With fsel=0, the `if fsel and ...` branch never fires, so `F.body = body` and the tuple return never happen. The locally-built `body` list is discarded each call — sum2O does pairwise work that vanishes. Side effects on Subs via `add2O(Sub,sub)` still happen, but the top-level body never accumulates new forks.
   
   You probably want fsel=0 to mean "always commit, skip eval," i.e. `if not fsel or (f.fc/...)` — or just split into two flags.

3. **F_[0] self-merge on first iter.** `for f in F_:` includes index 0. Need `F_[1:]`.

4. **No return.** sum2O has to produce an oF. Currently the function ends after the loop. Are you returning F_[0] (mutated) or constructing a fresh CoF as before (line 309 in old version)?

5. **fdata branch sets `body=[]` but never reads it.** Dead.

6. **w_, fw, fr, caller_ still not handled** — same as last review.

The merge-fold approach is fine conceptually (it's reduce(merge, F_) with side effects), but the "always commit" semantic needs to be explicit in merge's signature, not hidden by reusing fsel.

Want me to draft a fixed shape?

frame_2D_alg/meta_code.py
kwcckw: > so we need to reconcile this with merge(), which is adding the forks you should be splitting by? that merge wil be forming new body in centroids in sum2O, that's most of your split_oF_? we need a new typ of gate oFs, it will collect the statistics that determine splitting,

> yes, and we need to add new typ of oFs: secondary (gated) gate oFs
> 
> if Sub.nF=='E': # previously added gate oF, none yet

So that gate oF should be similar with merge, where their nF and typ == 'E'?  Then we will get 4 types of body:

1. node - primitive
2. (type, sub_) - sub
3. oF - function
4. gate oF - gate

Also the process should be valid on gated forks only? Or it could be from any part of the body? Because some elements such as the node in the body do not have any splitting statistic, so not sure how to correctly eval them for the splitting.

frame_2D_alg/meta_code.py
kwcckw: > so it's not if-else, something like
> 
> def sum2O(F_, root=None, w_=None, fdata=0): # for w,c,r, fw,fc,fr only?
> 
> ```
> if fdata: c_ = np.array([n.c for n in F_], dtype=float); C = c_.sum(); w_ = c_/C; body = []
> Body = F_[0].body
> for f in F_: merge(Body, f.body, fsel=0)
> ```

We may call sum2O on those call tree, so i think it's no relevant to merge their body there since it's only for the typ oF?

boris-kz: ok, so we don't need data when merging bodies, I guess we can use fcall_ instead of fdata, it's clearer?

frame_2D_alg/meta_code.py
boris-kz: I think it's just a 6th typ of oF: root_, comp_, clus_, proj_, prep_, gate_
yes, that's when the gate_ itself follows if-else

frame_2D_alg/meta_code.py
boris-kz, line 164: typ_= ['root_','root_','root_','comp_','comp_','comp_','comp_','clus_','clus_','clus_','clus_','proj_','proj_','prep_','prep_','prep_']
bod_= F_body_()
oF_ = [CoF(nF=i,typ=typ,body=bod) for i,(typ,bod) in enumerate(zip(typ_,bod_))]

kwcckw: We need to init F_body_ after we init oF_?

Because we need to init fc too. Else we need to parse fc along with body so that we can assign them in the init loop.

frame_2D_alg/meta_code.py
kwcckw: Yes, and i think if it's getting complicated, we might need a different sum2call or sum2O for clarity later.

frame_2D_alg/meta_code.py
kwcckw: So in term of splitting statistic, we shouldn't be using callers since they will be the same, and it would be up to their internal complexity fc, process weight fw and probably redundancy fr only?

frame_2D_alg/meta_code.py
boris-kz: right
bod_, fc = F_body_()

frame_2D_alg/meta_code.py
boris-kz: the forks that follow gate_ won't have the same callers, so we can split them out?

kwcckw: Okay, that's valid only if the body are from different oF, so we might need another .gate_ in gate oF so that we can pack their callers in the same order?

boris-kz: not sure, I will check tomorrow

kwcckw: Alright, i will go through the merge with the current cluster_oF_ first.

PR #158, opened 2026-05-28
5/28/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -175,19 +175,27 @@ def build(node):  # AST → CoF | (type,sub_) | ast_leaf | None
...
+            if isinstance(node, ast.FunctionDef) and node.name in iF_:
+                nF_[iF_.get(node.name)] = node
 
 typ_= ['root_','root_','root_','comp_','comp_','comp_','comp_','clus_','clus_','clus_','clus_','proj_','proj_','prep_','prep_','prep_']
-b_c_= F_body_()
-oF_ = [CoF(nF=i, fc=fc,typ=typ,body=bod) for i,(typ,(bod,fc)) in enumerate(zip(typ_,b_c_))]
-iF_ = {n: i for i,n in enumerate(_names)}  # indices name → nF, static
 nF_ = [None] * len(_names)  # FunctionDef s
+iF_ = {n: i for i,n in enumerate(_names)}  # indices name → nF, static
+oF_ = [CoF(nF=i,typ=typ) for i,typ in enumerate(typ_)]
```
kwcckw, line 196: So we need to create oF_ first and assign fc in F_body_ because we need to access the oF in build:

                    if (i:= iF_.get(node.func.id)) is not None and oF_[i].body:
                        return oF_[i], 3

frame_2D_alg/meta_code.py
```diff
@@ -300,40 +309,58 @@ def trace_func(module_dict, module_name=None):
...
-                    continue  # skip if merged:
-            for _sub in Sub.body:
-                if _sub.nF==sub.nF: fin=1; break  # no new fork cost?
-            if not fin:
-                add2O(Sub,sub); C+=sub.fc; add_+=[sub]
-        elif Sub.nF != sub.nF:
-            fork = CoF(nF='E',body=[Sub,sub]); C += cost; add_+=[sub]
-        body += [fork or Sub]
+        if isinstance(Sub, CoF) and isinstance(sub, CoF):  # oF or goF
+            if Sub.typ != sub.typ:  # create goF for different typs
```
kwcckw, line 330: Looks like the actual merge is more complicated when we have 4 different body types.

We need to check for same and different pair of goF, oF, sub and ast node, and create gate whenever the pair has different elements, as below.

frame_2D_alg/meta_code.py
boris-kz: thanks

frame_2D_alg/meta_code.py
boris-kz: this is a just conditional uppack of goF, and will be in split_oF_, not here.

frame_2D_alg/meta_code.py
```diff
@@ -300,40 +309,58 @@ def trace_func(module_dict, module_name=None):
...
-                nF_[iF_.get(node.name)] = node
 # draft
 def merge(F,f, fsel=1):  # combine aligned ops, if-fork per miss, no inline recursion, f can only be added as a whole
 
+    def create_gate(Sub, sub, F, f):
+        goF = CoF(nF='E', typ='gate_', body=[Sub, sub])
+        goF.caller_ = [getattr(F, 'caller_', set()), getattr(f, 'caller_', set())]  # nested and per Sub, sub?
+        return goF
+
+    def get_fc(sub):
```
boris-kz, line 320: just add oF.fcs in costs

frame_2D_alg/meta_code.py
```diff
@@ -300,40 +309,58 @@ def trace_func(module_dict, module_name=None):
...
-    for path in paths:
-        with open(path, encoding="utf-8") as f:
-            tree = ast.parse(f.read(), filename=path)
-        for node in tree.body:
-            if isinstance(node, ast.FunctionDef) and node.name in iF_:
-                nF_[iF_.get(node.name)] = node
 # draft
 def merge(F,f, fsel=1):  # combine aligned ops, if-fork per miss, no inline recursion, f can only be added as a whole
 
+    def create_gate(Sub, sub, F, f):
```
boris-kz, line 315: this is the same as forming new oF in split_

frame_2D_alg/meta_code.py
kwcckw: When we merge multiple Fs' body, we need to check and merge them according to their type? Without this checking, how do we merge body of different types such as oF, gate oF, node or sub?

frame_2D_alg/meta_code.py
kwcckw: So oF.fcs should a flat list packing all nested cs?

frame_2D_alg/meta_code.py
```diff
@@ -300,40 +309,58 @@ def trace_func(module_dict, module_name=None):
...
+
+    def get_fc(sub):  
+        if isinstance(sub, CoF): return sub.fc
+        if isinstance(sub, tuple): return costs.get(sub[0], 0) + sum(get_fc(s) for s in sub[1])
+        return costs.get(type(sub), 0) 
+
     body, add_ = [], []  # replace F.body if eval
-    C = 0; cost = f.fc  # total and fixed-fork costs
+    C = 0
+    i = -1  # for offset when F.Body is empty
```
boris-kz, line 327: it should never be empty?

frame_2D_alg/meta_code.py
kwcckw: Also that `and oF_[i].body` in `if (i:= iF_.get(node.func.id)) is not None and oF_[i].body:` within `F_body_` should be removed since oF's body shouldn't empty.

frame_2D_alg/meta_code.py
boris-kz: no, just append them to AST costs, that will be a single source

frame_2D_alg/meta_code.py
boris-kz: nF, not typ. we will have a fork for different goFs same as for different oFs

frame_2D_alg/meta_code.py
```diff
@@ -300,40 +309,58 @@ def trace_func(module_dict, module_name=None):
...
+                fork = create_gate(Sub, sub, F, f)
+                C += costs.get(type(sub), 0); add_ += [sub]
+        else:  # combinations between CoF, tuple and ast node
+            fork = create_gate(Sub, sub, F, f)
+            C += get_fc(sub); add_ += [sub]        
+        body += [fork]
     if len(f.body) > len(F.body): offs = f.body[i+1:]; body+=offs; add_+=offs
     elif len(F.body)>len(f.body): body+= F.body[i+1:]
 
+    F.body = body  # this should be default anyway?
```
boris-kz, line 361: right

frame_2D_alg/meta_code.py
boris-kz: `costs{}` and `oF.fc` are different kinds of quantity, and folding one into the other has problems:

- `costs` is a static primitive table: ast-node-type → unit op cost, known at module load.
- `oF.fc` is derived — accumulated over a whole body by `F_body_()`. It only exists *after* parsing, and during `build()` itself the callee's `fc` may not be computed yet (parse order / recursion). So you can't put `oF.fc` into a static `costs` dict, and `build` can't look it up there even if you tried.

Also, you don't need a new key space: `fc` keyed by `nF` is already `oF_[nF].fc`. A `costs[nF]` entry would just duplicate that.

The only real motivation I see is uniform lookup — making `build`/`get_fc`/`comp_body` call one `cost(x)` regardless of whether `x` is an ast type or a function ref. If that's the goal, the clean form is a helper, not merging into `costs`:

```python
def cost(x):  # ast type | CoF
    return x.fc if isinstance(x,CoF) else costs.get(x,0)
```

But note `build` currently returns a hardcoded `3` for a resolved Call (`return oF_[i], 3`) — i.e. the *call-site* cost, deliberately separate from the callee body, which `comp_body`/`merge` recurse into structurally. `get_fc` instead returns `P.fc` (the whole body). That inconsistency is the actual question: do you want a function reference to count as a flat `fc` or as a structural subtree?

Which one are you after — uniform lookup, or making `build` use callee `fc`? They pull in opposite directions.

boris-kz: oF_ = [CoF(nF=i,typ=typ) for i,typ in enumerate(typ_)]
parse_funcs(["agg_recursion.py","comp_slice.py","slice_edge.py"])  # populate nF_
fcosts = {F.nF: F.fc for F in oF_}

then:

get_fc: return fcosts.get(P.nF,0) if isinstance(P,CoF) else costs.get(type(P),0)

frame_2D_alg/meta_code.py
kwcckw: But nF won't work for sub or ast node? How do we merge a ast node with oF without using isisntance and create goF here?

frame_2D_alg/meta_code.py
kwcckw: Yes, but this is actually for the accumulation of C when the new C.body is empty. Else we would need zip_longest above.

frame_2D_alg/meta_code.py
boris-kz: def get_fc(P):  # get fc of single prim, similar with build

    if isinstance(P, CoF):
        return fcosts.get(P.nF, P.fc)
    if isinstance(P, tuple):
        return costs.get(P[0], 0) + sum(get_fc(c) for c in P[1])
    return costs.get(type(P), 0)

kwcckw: But this get_fc won't work for sub? We still need to add the condition for sub by checking for tuple and unpack it recursively?

boris-kz: yes, I updated above

frame_2D_alg/meta_code.py
boris-kz: it's not goF, just a simple AST fork

frame_2D_alg/meta_code.py
boris-kz, line 269: def comp_prim(_P,P):
        if isinstance(_P, CoF) or isinstance(P,CoF):
            return P.typ==_P.typ if (isinstance(P,CoF) and isinstance(P,CoF)) else 0
        else: return _P[0]==P[0] if isinstance(P,tuple) else type(_P) == type(P)

frame_2D_alg/meta_code.py
kwcckw: I see, so instead of creating goF, it's replaced by a ast if-forking node? But then we no longer able to retrieve the actual body after being replaced by the forking node.

frame_2D_alg/meta_code.py
boris-kz: ok your version is fine, sorry

frame_2D_alg/meta_code.py
boris-kz: we can, body is AST tree, it upacks all primitives

kwcckw: Okay, so we probably need to pack them into the node.child, not sure how to do that yet.

frame_2D_alg/meta_code.py
kwcckw: So for primitive here, basically we only need to check their type?

boris-kz, line 238: actually  `and _T.V_[T]` is wrong, the eval here is groupwise, not pairwise.
we don't need it here, this is just initial prototypes.

also, 

```python
# replace lines 236-241
grp_ = {}
for T in oF_: grp_.setdefault(T.typ,[]).append(T)
grp_ = list(grp_.values())
```

frame_2D_alg/meta_code.py
```diff
@@ -175,19 +175,27 @@ def build(node):  # AST → CoF | (type,sub_) | ast_leaf | None
...
-b_c_= F_body_()
-oF_ = [CoF(nF=i, fc=fc,typ=typ,body=bod) for i,(typ,(bod,fc)) in enumerate(zip(typ_,b_c_))]
-iF_ = {n: i for i,n in enumerate(_names)}  # indices name → nF, static
 nF_ = [None] * len(_names)  # FunctionDef s
+iF_ = {n: i for i,n in enumerate(_names)}  # indices name → nF, static
+oF_ = [CoF(nF=i,typ=typ) for i,typ in enumerate(typ_)]
+parse_funcs(["agg_recursion.py","comp_slice.py","slice_edge.py"])  # populate nF_ 
+F_body_()
 
 def cluster_oF_():  # cluster Ts if called together, global only
```
boris-kz, line 200: It is possible to refine initial grps by average linkage, instead of centroids. The problem I had last time was using exemplars. But it's purely technical, we don't need to implement it now:

Chat:
You can. It just stops being **standard average-linkage HAC** and becomes a different algorithm: **pairwise-linkage refinement / graph partition refinement**.

Average linkage as usually defined is irreversible because its state is only the dendrogram merge history:

```python
merge(A,B) -> C
```

It does not maintain an operation like:

```python
for n in C:
    test whether n fits C better than neighboring C2
```

But nothing prevents adding that. The problem is cost and bookkeeping.

For a node `n` currently in cluster `A`, pairwise membership value would be something like:

```python
fit(n,A) = sum(sim(n,a) for a in A if a is not n) / (len(A)-1)
```

To test moving `n` from `A` to `B`:

```python
gain =
    fit(n,B) - fit(n,A_without_n)
    - boundary / redundancy / cost terms
```

So yes, it is analogous to centroid reassignment:

```text
centroid clustering:
    compare n to C.centroid

pairwise refinement:
    compare n to all members / rim links of C
```

But pairwise refinement is much more expensive:

```text
centroid update:
    O(features) per n,C

pairwise update:
    O(len(C)) per n,C
    or worse if you recompute cluster cross-links naively
```

You can cache:

```python
S[n,C] = sum(sim(n,c) for c in C)
fit(n,C) = S[n,C] / len(C)
```

Then moving `n` requires updating `S[x,A]` and `S[x,B]` for many affected `x`. That is feasible for sparse rims, but expensive for dense all-pairs linkage.

The deeper issue: average linkage is defined on **cluster pairs**, not **node membership**. Refinement needs a different objective, for example:

```python
C.value = mean internal similarity - mean external similarity
```

or your kind of rim ratio:

```python
in_rim / ex_rim
```

Then node moves/splits become well-defined.

So the correction is:

```text
average linkage has no repair mechanism by definition,
not because pairwise repair is impossible.
```

Pairwise repair is basically:

```text
average-linkage merge criterion
+ divisive / reassignment refinement
+ cached internal/external rim sums
```

In your terms, it is closer to recursive `cluster_N` refinement than to plain HAC. The main missing piece is a stable membership objective, not the ability to re-evaluate nodes.

frame_2D_alg/meta_code.py
kwcckw: > It is possible to refine initial grps by average linkage, instead of centroids. The problem I had last time was using exemplars. But it's purely technical, we don't need to implement it now:

We had that pairwise V for the full linkage purpose, so for average linkage, from your posted LLM discussion, it will be similar with recursive cluster_N,  and we just need to add another eval to make sure the added new node is really belong to the cluster? Something like the existing final eval to remove the overlapping nodes using their max scores?

boris-kz: no, it's like cluster_oF_ but pairwise instead of centroid-based. Don't worry about it

boris-kz: what else can you compare?

kwcckw: If it's pairwise, it will be using V_ and centroid can be removed? Also during the refinement, we will be using only V_ (in_ and ex_), as in our very initial scheme?

frame_2D_alg/meta_code.py
boris-kz: it's the same as with any new fork:

An AST fork in your format is the tuple `(ast_type, (sub1, sub2))` — that's what `build` produces. Replacing two adjacent prims `body[i], body[i+1]`:

```python
body[i:i+2] = [(ast_type, (body[i], body[i+1]))]
```

If they're not adjacent:

```python
fork = (ast_type, (p1, p2))
body = [x for x in body if x is not p1 and x is not p2]
body.insert(i, fork)
```

No `fc` to set — tuples don't carry it; `get_fc` recomputes via `costs[ast_type] + sum(get_fc(c) for c in subs)`.

The actual choice is which `ast_type` represents the pack. `ast.IfExp` / `ast.BoolOp` if disjunction, `ast.Tuple` if just bundling. Which semantics?

frame_2D_alg/meta_code.py
kwcckw: We can compare their complexity along with type, because different types may have the same complexity?

boris-kz: with all pairs precomputed, it's just V__: new V_ per grp. 
then pick the highest sum(V_) and move the node to that grp, if > ave

boris-kz: these are primitive types, the complexity is fixed.

frame_2D_alg/meta_code.py
boris-kz: it's ast.IfExp, but we only need to replace one prim, in the target oF

frame_2D_alg/meta_code.py
boris-kz: def comp_prim(_n,n):

    if isinstance(_n, CoF) or isinstance(n,CoF):
        return n.nF==_n.nF if (isinstance(n,CoF) and isinstance(n,CoF)) else 0
    else: return _n[0]==n[0] if isinstance(n,tuple) else type(_n)==type(n)

frame_2D_alg/meta_code.py
kwcckw: > it's ast.IfExp, but we only need to replace one prim, in the target oF

This packs only `ast.IfExp`, but the actual body is not retrievable from the target, unless we pack it as the suggestion above? 

    fork = (ast.IfExp, (body1, body2))

frame_2D_alg/meta_code.py
kwcckw: > with all pairs precomputed, it's just V__: new V_ per grp. then pick the highest sum(V_) and move the node to that grp, if > ave

Then it's just a single step process since the highest sum(V_) is fixed, so we just need to move them once.

boris-kz: no, every time you move, grps change, and V__ must be recomputed

frame_2D_alg/meta_code.py
boris-kz: for i, (Sub,sub) in enumerate(zip(F.body,f.body)):
            fin=0
            if sub.nF=='E':
                subt = merge(Sub, sub)  # add2F(Sub,ssub_)?
                if isinstance(subt, tuple):
                    sub = subt[1]; C+=sub.fc; body+=[sub]; add_+=[sub]
                    continue  # skip if merged:
            for _sub in Sub.body:
            if not fin: add2O(Sub,sub); C+=sub.fc; add_+=[sub]
            body += [Sub]
        # current default:
        elif not comp_prim(Sub,sub):
            body += [(ast.IfExp, (Sub,sub))]  # was: fork = CoF(nF='E',body=[Sub,sub]); C += cost; add_+=[sub]
        else: body += [Sub]

frame_2D_alg/meta_code.py
boris-kz: ok, what if it's not empty?

frame_2D_alg/meta_code.py
boris-kz: yes,

      elif not comp_prim(Sub,sub):
            body += [(ast.IfExp, (Sub,sub))]  # was: fork = CoF(nF='E',body=[Sub,sub]); C += cost; add_+=[sub]

frame_2D_alg/meta_code.py
kwcckw: Now the V pair is computed based on :

    (comp_callers(_T,T) + comp_body(_T,T)) * min(_T.fc, T.fc)
    
And that's constant between T pair, unless we add grps related info in the comp?

frame_2D_alg/meta_code.py
kwcckw: Then this is not needed. To init non-empty oF, we need to copy it first, similar with the existing Copy_.

frame_2D_alg/meta_code.py
kwcckw: In this section:

      for _sub in Sub.body:
            
Each Sub and sub may not be a CoF and might not have nF? That including the main check of `Sub.nF=='E':` and `sub.nF=='E'` too.

The body could be oF, ast node, sub or (ast.IfExp, (Sub,sub)) now.

frame_2D_alg/meta_code.py
boris-kz: that's for the whole oF_, now it's split across grps, which may change
just forget about it

frame_2D_alg/meta_code.py
boris-kz: this whole section is for goFs, we don't have them yet

kwcckw: I see, and goF should be static right? That's when we getting each function's body node, and we can create goF while unpacking their body?

boris-kz: There is no point in setting static oFs, they exist to collect data

kwcckw: So it should be added in runtime,and added only if there's successful executed if-forking? Or it should be same as the prior goF, added per oF to pack all gate information within?

boris-kz: it's one type of new oFs we can add in split_oF_

PR #159, opened 2026-05-29
5/29/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -247,6 +247,7 @@ def cluster_oF_():  # cluster Ts if called together, global only
                     if t is not T: v__[j,i] += t.V_[T]
         C_ = [sum2O(oF_, w_=v__[:,i]) for i in range(Lc)]  # weighted re-aggr
         V, dV = v__.sum(), np.abs(v__-_v__).sum()
+        # endless loops here, because `v__[j,i] += t.V_[T]`, values getting larger and dV getting larger too
         if V*dV*(wcO*L) <= ave * (Ln+ ccO*L): break  # convergence
```
kwcckw, line 251: When `t.V_[T]` is positive, it increments the same positive values in each iteration. As a result, dV becomes larger in each iteration, and `V*dV*(wcO*L) <= ave * (Ln+ ccO*L)` won't be true, causing endless refinement here.

frame_2D_alg/meta_code.py
```diff
@@ -287,7 +291,7 @@ def split_oF_():  # divisive clustering
         else: out += [oF]
     oF_[:] = out
 
-def merge(F,f, fsel=1):  # combine aligned ops, if-fork per miss, no inline recursion, f can only be added as a whole
+def merge(F,f, fsel=1):  # combine aligned ops, if-fork per miss, no inline recursion, f can only be added as a whole
```
kwcckw, line 294: > this whole section is for goFs, we don't have them yet
> There is no point in setting static oFs, they exist to collect data

This merge should be for the body, which is not limited to goFs? Also if goF should be added in runtime, that's mean we need to add goF to their body during the runtime?

frame_2D_alg/meta_code.py
boris-kz: t.V_ should be reset in each loop

frame_2D_alg/meta_code.py
boris-kz: I meant the if section. add goF in split_oF_.

frame_2D_alg/meta_code.py
```diff
@@ -262,13 +263,16 @@ def cluster_oF_():  # cluster Ts if called together, global only
 
 def comp_prim(_n,n):
     if isinstance(_n, CoF) or isinstance(n,CoF):
-        return n.nF==_n.nF if (isinstance(n,CoF) and isinstance(n,CoF)) else 0
-    else: return _n[0]==n[0] if isinstance(n,tuple) else type(_n)==type(n)
+        return n.nF==_n.nF if (isinstance(_n,CoF) and isinstance(n,CoF)) else 0
+    else: 
+        if isinstance(_n, tuple) and isinstance(n, tuple): return _n[0]==n[0]  # the prior code has issue because the other _n may not be a tuple and _n[0] can't work
```
boris-kz, line 268: thanks

frame_2D_alg/meta_code.py
```diff
@@ -193,8 +193,8 @@ def parse_funcs(paths):
 nF_ = [None] * len(_names)  # FunctionDef s
```
boris-kz, line 193: I am trying out new Opus 4.8 with Max effort, on the conceptual level:

https://claude.ai/share/2190e0d0-5fb3-40bc-b9d0-97f29289009c

pretty good :)

frame_2D_alg/meta_code.py
kwcckw: `t.V_` is per T pair, so in each loop, t.V_[T] should be updated from the values of `v__[j,i]` ?

     v__[j,i] = (comp_callers(C,T) + comp_body(C,T)) * min(C.fc,T.fc)

For each `_T` in C.N_, that `v__[j,i]` should be separated according to each `_T`'s weight in C.N_, and reassign back to `T.V_[_T] and _T.V_[T]`?

boris-kz: this T.V_ should recomputed per centroid. We only need initial V_ if we use average linkage per grp

frame_2D_alg/meta_code.py
kwcckw: I see, so split_oF_ is called after cluster_oF_, and that's for runtime. When we sum C in cluster_oF_, their body needs to be merged differently? Since they do not have any goF assigned yet.

frame_2D_alg/meta_code.py
kwcckw: Yes, but T.V_ is per T pair, so after the centroid is computed, each T.V_ should be mapped to centroids now? And in the next iteration, it will be `v__[j,i] += t.V_[T's centroid in prior loop]`?

boris-kz: yes, remove initial cross-comp, then it's just 
 
`V = (comp_callers(_T,T) + comp_body(_T,T)) * min(_T.fc,T.fc); _T.V_[T] = V; T.V_[_T] = V`

frame_2D_alg/meta_code.py
boris-kz: that's already in else:

        else:  # default
            if comp_prim(Sub,sub): body += [Sub]  # add count?
            else: body += [(ast.IfExp,(Sub,sub))]; C += costs[sub]; add_+=[sub]

but split_oF_ should be called before cluster_oF_

frame_2D_alg/meta_code.py
kwcckw: To clarify, the other `_T` here :

    V = (comp_callers(_T,T) + comp_body(_T,T)) * min(_T.fc,T.fc); _T.V_[T] = V; T.V_[_T] = V

Should be the centroid? But from your code above, we don't use the prior V_? So we should need `T._V_` and `T.V_` for prior and current V_?

frame_2D_alg/meta_code.py
kwcckw: Okay,, that's mean we still need to check if sub is node, goF, oF or sub since this line .nF won't work for node or sub:

Those nodes or sub tuples don't have .nF, we would need multiple `isinstance` to check them.

frame_2D_alg/meta_code.py
boris-kz: sorry I meant remove that, then it's just

                v__[j,i] = (comp_callers(C,T) + comp_body(C,T)) * min(C.fc,T.fc)  # x centroid V

frame_2D_alg/meta_code.py
boris-kz: right, and in comp_prim, if oF, we need to comp callers and data?

kwcckw: Yes, that should be called from split_oF_ right?

I guess we need the same:

    (comp_callers(C,T) + comp_body(C,T)) * min(C.fc,T.fc)

boris-kz: that's not to centroid and we can't generate new oFs with each merge anyway, so no comp_body

kwcckw: Their body is still valid? 

           for n in oF.body[1:]:
                if comp_prim(_n,n): grp+=[n]
                else: grp_+=[grp];  grp=[n]

If their body is oF,  we still need comp_body?

boris-kz: no, if nF is the same then the body is the same

kwcckw: > no, if nF is the same then the body is the same

That's for root oF.nF, but for body , if their nodes are different oFs, they should have different nF?

boris-kz: different oFs are 0 match, they should not be compared

PR #160, opened 2026-05-30
5/30/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -287,27 +286,29 @@ def cluster_oF_():  # cluster Ts if called together, global only
                 nT.memb = gV; fC = sum(t.fc for t in t_); nT.cmpr = fC - fC/len(t_)
                 new_oF_ += [nT]
             else: new_oF_ += t_  # unpack if weak
-    return new_oF_
+    return new_oF_  # after the new oFs, we need recompute iF_, nF_? But their names are different now, not sure how to map them
```
kwcckw, line 289: When we have new oF_ after split and recluster, we need to remap those weights and costs?

      cFrm, cN_,cC_,cN,cF, cE,ccN,ccC,ccP, cAgg, cVct,cTrc,cBac,cPrj,cCS,cSE = (      # function complexity
      wFrm, wN_,wC_,wN,wF, wE,wcN,wcC,wcP, wAgg, wVct,wTrc,wBac,wPrj,wCS,wSE ) = [F.fc+1 for F in oF_]  # ave gain/call, init = cost
      ttFrm, ttN_,ttC_,ttN,ttF, ttE,ttcN,ttcC,ttcP, tAgg,ttVct,ttTrc,ttBac,ttPrj,ttCs,ttSE = [F.dTT+1 for F in oF_]

Or each of them should be updated separately based on feedback only since they should be determined by the data?

frame_2D_alg/meta_code.py
kwcckw, line 244: > if section. add goF in split_oF_.

To add goF in split_oF_, that should be in the else section?

              if comp_prim(_n,n): grp+=[n]
              else: grp_+= [grp]; grp=[n]

When both types are different? Now it's default to new grp, and it should be either new grp or just adding goF?

boris-kz: New grp is converted to oF anyway, so it will be goF if it is nested in IfExp

frame_2D_alg/meta_code.py
boris-kz: Should be in higher order fffeedback

frame_2D_alg/meta_code.py
kwcckw: If it is nested in IfExp, that's in oF.body, so we won't creating new goF here? Or we need to convert n (the tuple of IfExp) to goF when n[0] is IfExp?

boris-kz: Yes, if n[0] is IF, then n[1] is gated ops?

kwcckw: I think yes, so their nF is E, and typ is gate_?

boris-kz: yes, but nF is just and index

kwcckw: > yes, but nF is just and index

Yes, and actually it should be numeric for indexing, but we don't use it in the indexing anyway.

frame_2D_alg/meta_code.py
```diff
@@ -253,13 +253,13 @@ def split_oF_():  # divisive clustering
             for grp in grp_:  # single refinement
                 fc = sum([get_fc(prim) for prim in grp])
                 sub = CoF(root=oF, fc=fc, body=grp)  # add special case: form goF from AST tuple?
+                sub.caller_ = copy(oF.caller_)  # all sub under the same oF should have the same caller_
```
kwcckw, line 256: So this is just temporary. In actual case, each body's caller may be different when we cluster different oFs, so we should have caller per prim in body when we have higher frames?

boris-kz: that would make all prims oFs, too expensive. so that copy is fine

kwcckw: We could have another caller_ to pack body's callers? And it should be aligned with the prim in body?

boris-kz: that's only for IfExp contents?

frame_2D_alg/meta_code.py
boris-kz, line 241: next to work on suggestions:

Grounding this in the current code state rather than the README's prose, since the gap between them is where the conceptual work lives.

What's actually running is feedforward hierarchical clustering plus coordinate-shift attention (`expand_lev` shifts `iy,ix` by projected value via `proj_focus`). What's *stubbed* is the part that would make it predictive-corrective rather than just generative. That's the dividing line that maps onto your "conceptual vs codeable" distinction: a stub with a clear spec is codeable; the items below have no operational spec yet, so no autocoder can touch them.

Ranked by dependency:

1. **The projected-vs-actual match signal (surprise/confirmation).** `ffeedback` returns `rM=rD=0`, and `comp_prj_dH` is commented out. So the system never compares its projection to what actually arrived — it can't be wrong yet. This is the keystone: the objective function ("projected model-to-environment correspondence") only becomes real once projection is scored against outcome. Everything in README part 3 (novelty selection, imagination, planning) hangs off it. The undetermined part isn't the plumbing, it's the rule: how surprise maps to a filter update, and the sign-flip you describe (match laterally, novelty downward) as a *concrete, level-consistent* update rather than prose.

2. **Core+contour comparison operator.** `Bt` exists and contrast-links get correlation-clustered, but the next-level cross-comp *between* composite core+boundary objects has no defined comparand structure. What does `comp_N` mean when each side is a similarity core plus its high-variance separability shell? This is the "inherently discontinuous" frontier you flag — partly specified, genuinely open.

3. **Meta-code weights.** `nw,cw,Nw,Cw,specw` and `cluster_oF_` presuppose a "contribution to projected match" signal per code element — which doesn't exist until (1) does. So this is downstream, not next, despite being the most ambitious.

Cross-cutting flag: the eigenvector reframing is currently `argmax(wTT)` — a single-attr proxy. Real spectral reordering imports eigendecomposition, which collides with your no-stats principle. That tension is unresolved and independent of the above.

Which of these do you read as the current bottleneck — the feedback loop, or do you still consider core+contour structure unsettled enough that (1) would be built on sand?

frame_2D_alg/meta_code.py
kwcckw: Okay yes, then we probably can pack their caller in the node?

    (ast.IfExp,(Sub,sub), (F.caller_, f.caller_))?

boris-kz: this is a primitive, you can't pack anything in it?

frame_2D_alg/meta_code.py
kwcckw: > Which of these do you read as the current bottleneck — the feedback loop, or do you still consider core+contour structure unsettled enough that (1) would be built on sand?

I think probably the feedback loop? But that can't be done before we finalize the meta_code since we need to complete the oF_ structure first as one of feedback process.

frame_2D_alg/meta_code.py
kwcckw: We already have `(ast.IfExp,(Sub,sub))`, and the format isn't restricted to `(ast.IfExp,(Sub,sub))` right? So i don't see why we can't pack additional callers?

frame_2D_alg/meta_code.py
boris-kz: that's in frame_HH fffeedback, we should still fless-out current ffeedback

frame_2D_alg/meta_code.py
boris-kz: ok try that

frame_2D_alg/meta_code.py
kwcckw: Okay, but the current ffeedback is not completed yet? 

          rTT = np.divide(frame.H[0].wTT, frame.H[1].wTT)  # wTT_ is not relevant now
          _wTT = frame.H[1].wTT
          for lev in frame.H[2:]:  # sum ratios between consecutive-level TTs, top-down in frame H, not lev-selective or sub-lev recursive
              rTT += np.divide(_wTT,lev.wTT)
              _wTT = lev.wTT
          rM = rD = 0
          rm, rd = vt_(rTT,wTT)
          return rM+rD, rTT  # add rm,rd?

The existing frame.H[i].wTT is the default wTT anyway, so all levels have the same wTT. In actual case, their wTT should be recomputed based on each level's nested oF's collected data params (dTT or rTT)?

boris-kz: Not sure we need it, last level already represents all lower levels, it's the highest scope

kwcckw: We need each level to compute the level ratio, rTT? With just single last level, that will be compared with prior frame only? But i think we still need per frame or within frame's feedback?

boris-kz: Right now the feedback is from top level only, we can add feedback per level later

PR #161, opened 2026-05-31
5/31/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -307,12 +314,15 @@ def merge_oF(F,f, fsel=1):  # combine aligned ops, if-fork per miss, no inline r
                 C+=get_fc(sub); add_+=[sub]
             body += [Sub]
         else:  # default
-            if comp_prim(Sub,sub): body += [Sub]  # add count?
-            else: body += [(ast.IfExp,(Sub,sub))]; C += get_fc(sub); add_+=[sub]
+            if comp_prim(Sub,sub):
+                body += [Sub]  # add count?
+                if isinstance(sub, tuple) and sub[0] == ast.IfExp:
+                    for caller_ in sub[2]: F.caller_ |= caller_   # loop caller_ from IfExp node, add forks' callers into F using set union
```
kwcckw, line 320: So i think we need to merge prim's caller to F.caller_ when it's a IfExp node? So that F.caller_ carries the additional IfExp forks' caller after the merging.

frame_2D_alg/meta_code.py
```diff
@@ -252,7 +256,10 @@ def split_oF_():  # divisive clustering
             grp_ += [grp]
             for grp in grp_:  # single refinement
                 fc = sum([get_fc(prim) for prim in grp])
-                sub = CoF(root=oF,fc=fc,body=grp); sub.caller_ = copy(oF.caller_)
+                sub = CoF(root=oF,fc=fc,body=grp); sub.caller_ = copy(base_caller_)
+                for prim in grp:
+                    if isinstance(prim, tuple) and prim[0] == ast.IfExp:  # add the IfExp node specific caller_
+                        sub.caller_ += prim[2][0] | prim[2][1]
```
kwcckw, line 262: This looks backwards here, but i'm don't see a better way yet.

So oF.caller_ packs all callers, including IfExp node callers. When we split body, each grp should have different caller_, so we need to find the base caller_ by excluding all callers from IfExp node's callers first. Then, re-add the IfExp node's caller after we form grp if the grp contains those IfExp node.

boris-kz: I don't get it, you inherit that base caller_ from the parent oF anyway, fork caller_ is a specific subset? You can't get it here, it has to be assigned in traced

frame_2D_alg/meta_code.py
```diff
@@ -244,6 +244,10 @@ def split_oF_():  # divisive clustering
     out = []
```
boris-kz, line 244: Trying to generalized cross-comp and clustering across code and meta_code:

Claude:

The embedding is the function's runtime data footprint, and you already collect it. `traced` wraps each call and `sum2O` rolls the data-level TTs that flowed through into `oF.dTT`. That dTT is a vector in the *data* coordinate — ordered, metric, the magnitude axis where min/sub mean something. Make it the function's position.

Then there's nothing to generalize: `vt_`/`sum_vt`/`comp_N` run on `oF.dTT` pairs unmodified, `cluster_N` clusters them, and similarity↔proximity is restored — not in source or call-graph, but in behavioral space, where two functions are near iff they transform data the same way. `comp_body` and `cluster_oF_` become unnecessary; they were the structural-similarity stopgap for a missing coordinate.

The reframe that keeps it from being a hack: this measures *behavioral* similarity, not source similarity, and behavioral is the right target. You don't care that two functions share an AST; you care whether they make the same contribution to prediction. Merging two functions with the same footprint into a higher type is correct compression by your own criterion — active code, compression local. `comp_body` was tracking the wrong thing.

So the modality-specific part shrinks to one thing: the tracer. `traced` is the transducer that senses code as data; above it the data path applies unchanged. That's the actual generalization — not a comparison that swallows code, but sensing code's runtime as a data stream and reusing the existing mechanism.

Where it breaks, in order:

1. **Calibration is gated on feedback — again.** Raw footprint clusters functions by data *throughput*, not predictive *contribution*. Weighting the footprint dimensions by what helped prediction needs the projected-vs-actual signal (`ffeedback`, `comp_prj_dH`) that's still zeroed. The data path literally calibrates the code embedding, so this can't be built before the turn-1 keystone.

2. **Footprint conflates function with input.** Same function on different data lands in different places. Position should probably be the input→output *transform*, not raw output dTT — which means capturing input footprint too. That's exactly what `comp_prj_dH`'s input→projected→actual structure is for.

3. **Granularity is function-level.** This clusters whole functions. Your endgame (`nw,cw`, merging ops and blocks) needs per-block footprints: `traced` at sub-function granularity, attributing TT deltas to AST sub-nodes. The static `oF.body` and the runtime dTT don't meet yet.

4. **A bad code-merge costs more than a bad data-merge.** Clustering data yields a representation; clustering code rewrites the program. The same feedback validates both (did the merged code still predict?), so it's closed — but the stakes differ, so code-merges may need a soundness gate the data path doesn't. That's the one place code stays genuinely different.

So: yes to the branch, but it's not parallel and free. It's the data path applied to a self-sensor, and it switches on only after the data-level feedback exists to calibrate it. Does that ordering match how you're thinking about it, or do you see the footprint as usable uncalibrated?

boris-kz: so basically comparing the data in oF (probably need to add coords) can add detail to comp_caller_

Yes, dTT is a value, not a coordinate — I conflated them. So it's an added comparand alongside caller-overlap and body-structure, not a new mechanism. `comp_body` and the caller relation stay.

The "add coords" is the real point. dTT alone is pure "what" — it has no "where." Two functions with the same TT could be acting on top-left vs bottom-right of the input, or one position-specialized and one generic, and the footprint can't tell them apart. Tagging it with the *data* coordinate it was collected at restores your what×where split, and that completed comparand is what adds detail to `comp_callers`: match iff same behavior in corresponding region, not just shared callers. (Locality isn't the issue at ~16 functions — all-pairs is fine. So this is about completeness of the comparand, not cheap alignment. The coordinate problem only returns at sub-function granularity.)

One line I'd still hold: mechanically it augments `comp_callers`, but semantically it's the only one of the three channels that touches the objective. Caller-overlap and body are both structural proxies for predictive contribution — the same problem as `comp_body` last turn. Two functions with identical callers and identical AST can still differ in what they do to prediction; only footprint+coord sees that. So it should weight-dominate the match, not just decorate it — once the feedback that calibrates the footprint exists.

frame_2D_alg/meta_code.py
boris-kz, line 118: also add _CoF to each called fork in the body here?

frame_2D_alg/meta_code.py
boris-kz: the data here is the contribution of oF to the caller? then we are not just getting overlap of callers between compared callees, it's also comparing call results:  only successful callees should be grouped?

claude — contribution-to-caller is the right definition of the footprint, and it's cheaper than I implied. You don't need a counterfactual: `sum2O` already rolls each callee's dTT into its caller, and the share-weight `n.c/C` in `sum_vt` *is* oF's contribution to its parent. So the signal is already half-computed; it's just not used as a clustering value.

And yes, that turns comp_callers from a pure relation (which callers) into relation × value (what each contributes within them). Shared callers + similar contribution = similar; shared callers + *opposite-sign* contribution = contrast — two callees that play opposite roles in the same context. That's your what×where, and the dual cluster falls out of it.

Which is why I'd push on "only successful should be grouped" — that discards the contour. Cluster *both* signs. A group of consistently-negative callees is the Bt analog: the prune signal, the thing feedback exists to cut. By your own contour principle the negatives are *more* informative than the successes — they tell you what to remove. So success picks which cluster a callee lands in, not whether it's clustered. Success gates legitimately only at the *retention eval* — does the formed cluster's value beat its cost — applied to the cluster, post-formation, not as a membership pre-filter.

One caveat on "successful": it isn't local-positive. A callee that helps a worthless caller isn't successful. Contribution has to be discounted by the caller's own value up the chain — your `brrw` lend, decaying with distance — anchored at the root on actual-vs-projected. So the sign is only real once that root anchor exists. Gated, again, on the same feedback.

frame_2D_alg/meta_code.py
boris-kz, line 303: this is a secondary (gated) gate, it's always packed it in the primary AST gate, below?

frame_2D_alg/meta_code.py
kwcckw: > I don't get it, you inherit that base caller_ from the parent oF anyway, fork caller_ is a specific subset? 

Yes, but IfExp fork is part of the body, so i was thinking to add those fork caller to parent oF.caller_? Or it shouldn't be added to parent oF.caller_ and it should be stayed as specific subset only?

frame_2D_alg/meta_code.py
kwcckw: > The reframe that keeps it from being a hack: this measures behavioral similarity, not source similarity, and behavioral is the right target. You don't care that two functions share an AST; you care whether they make the same contribution to prediction. Merging two functions with the same footprint into a higher type is correct compression by your own criterion — active code, compression local. comp_body was tracking the wrong thing.

So the structural similarity such as ast may not be important compare to data? Since the output data is the prediction, and we need that to measure how similar are them to get the same prediction?

frame_2D_alg/meta_code.py
kwcckw: This line is adding caller to each type oF, so this caller `_CoF` should be added to all oF_[iF_[func.__name__]].body? We would need another param to pack it and align it with body? Since prim is not CoF and we can't pack caller to them.

boris-kz: add a storage prim next to it, still in the body? this is a tuple, is there an AST expression with additional storage?

frame_2D_alg/meta_code.py
boris-kz: I posted above:   it's also comparing call results: only successful callees should be grouped?

frame_2D_alg/meta_code.py
kwcckw: So this Sub is a gated gate, the primary AST gate is referring to `[(ast.IfExp,(Sub,sub)]`?

frame_2D_alg/meta_code.py
boris-kz: Chat: if you want uniformity with your existing tuple format, use tagged tuples instead of dict:

return (ast.IfExp, (

    ('test',  test[0] if test else None),
    ('body',  body[0] if body else None),
    ('else',  els[0]  if els  else None),
)), fc

frame_2D_alg/meta_code.py
boris-kz: yes

frame_2D_alg/meta_code.py
boris-kz: so basically we don't have to mess with the primitives, just make each if and else an expression that contains data in real python

frame_2D_alg/meta_code.py
boris-kz: something like this:

                oF_[iF_[func.__name__]].caller_.add(_CoF)  # caller_ for comp_caller_
                for p in CoF.body:  # add _CoF to each called fork?
                    if isinstance(p,tuple):  # IfExp
                        if p[1][0]: p[1][0].caller_.caller_.add(_CoF)
                        if p[1][1]: p[1][1].caller_.caller_.add(_CoF)
?
That snippet has two likely errors.

First, `CoF.body` is class access, but `body` is an instance field:

```python
for p in oF.body:
```

not:

```python
for p in CoF.body:
```

Second, this is wrong:

```python
p[1][0].caller_.caller_.add(_CoF)
```

`caller_` is a set, not an object with `.caller_`. It should be:

```python
p[1][0].caller_.add(_CoF)
```

But only if the branch item is actually a `CoF`. A primitive AST node has no `caller_`.

Minimal safe form:

```python
for p in oF.body:
    if isinstance(p, tuple) and p[0] is ast.IfExp:
        for q in p[1]:  # test, body, else
            if isinstance(q, CoF):
                q.caller_.add(_CoF)
```

But conceptually: you probably **do not** want to add `_CoF` to every `IfExp` branch just because the enclosing function was called. Only the actually evaluated branch should get runtime caller credit.

Static caller relation:

```python
for q in (test, body, els):
    if isinstance(q, CoF): q.caller_.add(_CoF)
```

Runtime caller relation:

```python
test always gets _CoF
selected branch gets _CoF
unselected branch does not
```

Since `body` is static AST from `F_body_`, you can only do static caller assignment there. Runtime branch-specific caller assignment needs instrumentation at execution time, not just AST scan.

frame_2D_alg/meta_code.py
kwcckw: >  just make each if and else an expression that contains data in real python

To make if and else an expression that contains data, that's mean each if and else section should have data assignment, like `var = 1`?

frame_2D_alg/meta_code.py
kwcckw: > only successful callees should be grouped?

> Which is why I'd push on "only successful should be grouped" — that discards the contour. Cluster both signs. A group of consistently-negative callees is the Bt analog: the prune signal, the thing feedback exists to cut. By your own contour principle the negatives are more informative than the successes — they tell you what to remove. So success picks which cluster a callee lands in, not whether it's clustered. Success gates legitimately only at the retention eval — does the formed cluster's value beat its cost — applied to the cluster, post-formation, not as a membership pre-filter.

I think this is right? Non-successful callees represent part of the code, which tell us what should be remove?

frame_2D_alg/meta_code.py
kwcckw: > so basically we don't have to mess with the primitives

Okay, that's mean we still need to make some changes on top of the primitives right? 

These `p[1][0]` and `p[1][1]` are sets that packing callers, i don't see how we can always pack callers on top of them?

kwcckw: Okay right, so we don't need this check for 'E' now? Since it will be in tuple of IfExp anyway?

boris-kz: we just need to unpack them

frame_2D_alg/meta_code.py
boris-kz: the removal is for individual elements, they won't be forming new oF anyway

frame_2D_alg/meta_code.py
boris-kz: yes, that's in the code itself

frame_2D_alg/meta_code.py
boris-kz: this is confusing because goF is headed by IfExp, so there is no else, it's packed inside goF? 

Claude
A gate-CoF isn't a tuple, so the IfExp-tuple test in the else won't catch it. Sub.nF=='E' is the only way to recognize "this position is a gate" in the CoF form. That's all the check is: the detector for the CoF-shaped gate.
Its branch (ignoring the recursive line) then does the same append-into-gate as the tuple case — test sub against the gate's existing branches (comp_prim over Sub.body), and if it isn't already one, absorb it and keep Sub instead of wrapping a new IfExp.
So the rule is: gate → append inside; non-gate → wrap. Only the detector differs by representation — nF=='E' for CoF gates, "is IfExp tuple" for tuple gates. If both forms can occur, you keep both detectors.

frame_2D_alg/meta_code.py
kwcckw: Okay, i guess they only will be relevant only in the next frame if the fork is taken again.

frame_2D_alg/meta_code.py
kwcckw: Sorry, from your code above:

           for p in CoF.body:  # add _CoF to each called fork?
                if isinstance(p,tuple):  # IfExp
                    if p[1][0]: p[1][0].caller_.caller_.add(_CoF)
                    if p[1][1]: p[1][1].caller_.caller_.add(_CoF)
                    
 And if we pack IfExp fork using `body += [(ast.IfExp,(Sub,sub))]`, `p[1][0]` is Sub, which is a primitive or CoF, so we can't get their caller by unpacking them if the sub is a primitive? I don't think we can have callers per sub unless we pack it separately and align it with them.

frame_2D_alg/meta_code.py
boris-kz: I would stick with oF-only caller_ for now, this is too complex

frame_2D_alg/meta_code.py
kwcckw: > this is confusing because goF is headed by IfExp, so there is no else, it's packed inside goF?

Yes, else information won't be relevant anyway. And goF should be packing those gating information, while ifExp packing the gated information? 

> So the rule is: gate → append inside; non-gate → wrap. Only the detector differs by representation — nF=='E' for CoF gates, "is IfExp tuple" for tuple gates. If both forms can occur, you keep both detectors.

Both forms always occur the same time, so we can pack either one of them into another one? Or just use one to represent both?

boris-kz: yes, let's just leave it out, it's not a priority

boris-kz: ifExp is only packing goF, which will be unpacked later

boris-kz: `contextmanager` is already imported. Code, as minimal diffs:

**1. Add the `'E'` accumulator slot as `oF_[-1]`** (after `F_body_()`, line 196, so the zip in `F_body_` over `nF_` is untouched and the slot gets a real `caller_`):

```python
oF_ += [CoF(nF='E', typ='gate_')]; oF_[-1].caller_ = set()   # gate accumulator, oF_[-1]
```

**2. Gate instrument in `CoF`, sibling of `traced`** (insert after `traced`, before `__bool__`):

```python
    @staticmethod
    @contextmanager
    def gate():  # eval-time goF, mirrors traced but feeds oF_[-1]='E', not iF_
        _CoF = CoF._cur.get(None)
        oF = CoF(nF='E', root=_CoF)
        oF_[-1].call_ += [oF]
        if _CoF is not None:
            _CoF.call_ += [oF]; oF_[-1].caller_.add(_CoF)
        _oF = CoF._cur.set(oF)
        try: yield oF
        finally:
            if oF.call_:
                tree = flat_(oF); sum2O(tree,oF,fcall_=1); wtt = getattr(oF,'rTT',oF.dTT); oF.wTT = cent_TT(wtt,oF.r)
            CoF._cur.reset(_oF)
```

**Use at a conditional site** (the branch must contain traced calls or the goF stays empty — same as `traced` collecting only via `call_`):

```python
with CoF.gate():
    out = a if test else b
```

Two caveats, consistent with the last turn:

- This collects only the goF's *footprint* — the dTT of whatever traced sub-calls ran in the taken branch. It does **not** record the *decision* (which branch, predicate vs threshold); that needs `test` and the branch index passed in, which I left out since you haven't wired the predicate yet.
- All gates fold into the single `oF_[-1]='E'` slot, so `caller_`/`call_` accumulate gate-wide, not per-site. Per-site separation comes later from `split_oF_` dividing that slot by `comp_prim`.

kwcckw: So this is per frame gate oF, which all gate information is packed in this global gate?

But we won't know the exact location of each gate, which we need to sort it out based on the static ast when we need them later?

boris-kz: #2 above is supposed to be a run-time, in CoF

kwcckw: Okay thanks. So we can know their location by the aligned caller.call_ and global goF.caller_ (last line below) to identify this goF.

        oF = CoF(nF='E', root=_CoF)
        oF_[-1].call_ += [oF]
        if _CoF is not None:
            _CoF.call_ += [oF]; oF_[-1].caller_.add(_CoF)

kwcckw: Also in each If forking, we need to add this now?

 > Use at a conditional site (the branch must contain traced calls or the goF stays empty — same as traced collecting only via call_):
  
      with CoF.gate():
          out = a if test else b

boris-kz: probably, please edit

kwcckw: Alright, and do we need to pack this goF to the IfExp body in runtime? The default IfExp node probably should be just packing the gated information, rather than this goF.

PR #162, opened 2026-06-01
6/1/2026

kwcckw:


kwcckw:
So looks like we really need to use the `with CoF.gate()` method if we need to capture the runtime gate info, the alternatives actually loos much more complicated:

From claude:

So we need to place the wrap at each fork, and there is no single central method?

---

Right—for runtime branch capture, **there is no central method**. It has to be at each fork. This isn't a limitation of the design; it's a property of Python.

### Why Centralized Wrapping Fails
The reason is simple: a central wrapper (like a decorator on the function, or anything in `traced`) runs *around* the function call. 

* It sees that the function ran. 
* It **cannot** see which branch the interpreter took inside it.

By the time control returns to the wrapper, the `if` statement has already resolved and left no trace of which way it went. The only place that knows *"this branch, this run"* is the branch itself, at execution time. So the `with` statement has to sit there.

 `gate()` is the central method in the sense that the capture logic lives in one place—you're not rewriting that per site. But the **trigger** can't be centralized. You're placing a one-line marker that says "instrument here"; the machinery behind it is shared.

---

### Alternatives (And Why They Are Heavier)

The only ways to avoid hand-placing the `with` block are all heavier than what you're trying to avoid:

* **AST rewrite / import hook:** Parse each function, auto-wrap every `IfExp`/`if`, and recompile. While this *is* a single central method, it means transforming the source at load time. This is far more invasive than adding `with` lines and is exactly the "changing existing code" you wanted to stay away from (just hidden).
* **`sys.settrace`:** A line-level tracer that fires on every executed line and infers branch-taken from line flow. This is central, but it's a global runtime hook with real overhead. It captures everything and forces you to filter after the fact—the exact opposite of the gated, *only-where-it-matters* approach the PR thread settled on.

frame_2D_alg/agg_recursion.py
```diff
@@ -179,11 +185,12 @@ def comp_C_(C_,_r, _C_=[], fall=1, fC=0):  # simplified for centroids, trans-N_s
         for C in C_: C.compared=set()
         C_ = sorted(C_, key=lambda C: C.dTT[0][np.argmax(ttC_[0])])  # max weight defines eigenvector
         for j in range( len(C_)-1):
-            _C = C_[j]; C = C_[j+1]
-            if _C in C.compared: continue
-            dy_dx = _C.yx-C.yx; dist = np.hypot(*dy_dx); c = min(C.c,_C.c); r = (C.r+_C.r)/2  # tc += c; acc[1]+=c
-            tc += c; tr += r*c
-            comp_N(_C,C,_r, c ,A=dy_dx,span=dist,L_=L_,N_=N_,acc=acc)  # simplified for typ=3
+            with CoF.gate():  # one gate per C pair?
```
kwcckw, line 188: For gates within a loop here, it should create new gate per loop? Since if may or may not run comp_N below. Using a root gate might not capture this per loop information.

boris-kz: We can't create an new function just to store new data.
We have caller_ in all oFs, that should be their gates, and current oF.caller_ will be goF.root?

boris-kz: Confirmed, and it's simpler than the gate() route. The per-call oF that `traced` already builds has `root=_CoF`, so `goF.root = caller` needs no new code — that per-call oF *is* the goF. And `caller_` keeps holding the callers, which are exactly the goF roots, so `comp_callers`' set-overlap is untouched. No new function, no new field — the data (`dTT` via the `if oF.call_` block) is already collected per call.

So the only thing missing is tagging these under the `'E'` slot. Drop the `gate()` method from last turn; keep the `oF_[-1]='E'` slot, and feed it inside `traced`:

```python
oF_[iF_[func.__name__]].call_ += [oF]
oF_[-1].call_ += [oF]                          # 'E': call edge is the gate, goF.root = _CoF
if _CoF is not None:
    _CoF.call_ += [oF]
    oF_[iF_[func.__name__]].caller_.add(_CoF)  # caller_ = gates (= goF roots)
    oF_[-1].caller_.add(_CoF)
```

`oF_[-1]` keeps the slot you added:

```python
F_body_()
oF_ += [CoF(nF='E', typ='gate_')]; oF_[-1].caller_ = set()   # 'E' gate type, oF_[-1]
```

One thing to pin down, since it changes what `'E'` means: this treats *every* call edge as a gate, so `oF_[-1].call_` aggregates all calls — the gate *type*, parallel to the per-function slots. Is that the intent, or is a gate only a call inside a conditional? The latter still needs branch detection at the IfExp; the former is free and falls straight out of `caller_`.

frame_2D_alg/agg_recursion.py
```diff
@@ -85,18 +85,21 @@ def cross_comp(root, rr):  # core function mediating recursive rng+ and der+ cro
...
-        if sum(vt_(TT,ttE)) * (wE*L) > (ave+avd) * (r+cE*L):  # or oF.gF evals?
-            E_ = get_exemplars({N for L in L_ for N in L.nt}, r,c)
-            G_,r = cluster_N(root, E_,r,c)  # cluster_C, _P, eval?
-            if G_:
-                if not root.typ: F2N(root)  # promote @ 1st sub+ or agg+
-                root.H+= [sum2F(L_,root,froot=1)]  # lev: L_+ derivatives
-                root.Nt = sum2F(G_,root,froot=2); L=len(G_)-1  # or C_ s?
-                if vt_(TT,tAgg)[0]*(wAgg*L) > ave* (r+cAgg*L):  # root brrw| rdn?
-                    G_,r = cross_comp(root.Nt,r)  # agg+
+        with CoF.gate():  # both get_exemplars and cluster_n under same gate?
```
boris-kz, line 92: yes, we can actually merge them

frame_2D_alg/agg_recursion.py
kwcckw: > We have caller_ in all oFs, that should be their gates, and current oF.caller_ will be goF.root?

Yes, so all oF callees should be gated right? And since oF.caller_ will be goF.root, the sequence will be caller -> gate -> callee, and callee.root is gate, and gate.root is caller?

frame_2D_alg/agg_recursion.py
kwcckw: Yes, so with caller as gate.root, we should have different gates for both of the functions, and their gate.root shares the same caller?

boris-kz: merging means it's a single function

frame_2D_alg/agg_recursion.py
boris-kz: It should be conditional, not all oFs are gated. The point adding goF is to evaluate costs vs. benefits of gating,  which has it's own costs vs. benefits. So the later is removed if the stored value is negative

boris-kz: So goF should be transparent to the callee, it's only needed for the caller?

Yes. The goF is entirely caller-side: it's the conditional in the caller's body, `root = caller`, and it stores its own value (test cost vs skip savings). The callee carries nothing gate-specific — you only *read* its existing records (`call_`, `root`, `fc`) to count skips and price the gate. It neither knows nor changes whether it's gated, and its dTT is identical either way.

That transparency is required, not just convenient: the same callee can be gated from several callers, each with its own test cost and skip rate. Keeping the goF at the call site gives each a distinct goF, valued and removed independently in its own caller's body. If the value lived on the callee, those sites would conflate into one meaningless aggregate.

So the `oF_[-1]='E'` feed I put in `traced` two turns back is wrong — it tagged every call as a gate. Gates are only the conditional entries in caller bodies; `traced` stays callee-generic, and the gate value/removal is the caller-side pass reading callee counts.

kwcckw: > So goF should be transparent to the callee, it's only needed for the caller?

Okay, so callee.root is still pointing to the caller, but we probably can pack gates to caller.gate_? Where each element in caller.gate_ is mapped and aligned with caller.N_ ?

boris-kz: That means the caller (_CoF in traced) should callect data on callee only, the gate cost is fixed, its signal is binary presence/absence. Then gate +/- value is the net value of the calls it blocks, which is only known if this blocking is hypothetical?

Yes — blocked value is counterfactual. A blocked call produces no oF, so you have its fixed cost (saved) but not its value, which is data-dependent and only exists if you run it. Presence/absence is the whole observable; the magnitude of what absence costs you isn't in the data. And the callee data the caller *does* collect is exactly the passed-call side — never the blocked side.

But you don't need the full counterfactual for the decision. Value is continuous in the test variable, so calls that *just barely* passed approximate the ones that *just barely* failed. The marginal blocked value reads off the boundary passes — which prices whether to tighten or loosen the threshold without ever un-blocking anything. That part is fully observable from the data you already collect, provided the test correlates with value at all (and if it doesn't, the gate is worthless regardless).

What's *not* observable is the gate's total keep-vs-remove value — that needs the whole blocked distribution, not just the margin. That's the projection path (proj_N / comp_prj_dH): predict the blocked callee's value from the input it would have seen, calibrated on the runs where it did fire. Same zeroed keystone as the rest. Without it, your options are cost-only (saved fc − test cost, drop the value term) or sampling — occasionally run the blocked call to measure, i.e. make the block literally hypothetical.

So: gate *tuning* is local and available now; gate *existence* is projection-gated; cost-only is the honest fallback in between.

boris-kz: > > So goF should be transparent to the callee, it's only needed for the caller?
> 
> Okay, so callee.root is still pointing to the caller, but we probably can pack gates to caller.gate_? Where each element in caller.gate_ is mapped and aligned with caller.N_ ?

N_ is data, it has nothing to with anything. Here, we are talking about .body
The caller only needs to know that the gate exists + goF.fc, see above

boris-kz: So it doesn't add anything in base CoF:

- Drop the `gate()` method — no new function.
- Drop the `oF_[-1]='E'` feed in `traced`. Gates aren't every call, so `traced` reverts to exactly what it was: per-call callee oF carrying `dTT`, `root`, and the count. That passed-call data is all the gate reads.

CoF gains no field. The gate is an existing-shape `CoF(nF='E')` living in the *caller's* body at a conditional site (so it has somewhere to hold a value), `root=caller`, gated callee in its `body`. It reuses what's already there:

- `w` = the stored gate value — your "remove if stored value negative."
- `fc` = fixed test cost; `c` = run count (`len(caller.call_)`) if you want it cached, else derived.

And `w` isn't *collected* — it's filled by the eval pass: pass/skip from the callee's `call_` filtered by `root.nF`, times callee `fc`, minus test cost, minus the blocked-value term. That last term is the counterfactual part, supplied by `proj_N` when it exists and zero until then — so `w` is cost-only for now.

So the only new code is outside the class: emit conditional call-sites as gate-CoFs in caller bodies, and the eval pass that fills `w` and un-gates the negatives. The class itself stays put.

kwcckw: Okay, so all of those updates are no longer relevant, and the gate should be added during `F_body_` only now? That's when node is I`ast.ifExp`, and we need to explicitly pack gate as one of the child?

boris-kz: it will be the only child, there is no else?

kwcckw: > it will be the only child, there is no else?

I think there will be else since this is a IfExp instead of If node? And since this is static and before the runtime , we need both forks?

For example:

    comp_C_(N_,rr,fC=1) if fC else comp_N_(combinations(N_,2),rr)


We have comp_N_ in the else fork?

boris-kz: that if is not goF, if it is it will be headed by a lighter if.
and then both fork are called inside goF

boris-kz: That obviates goF altogether, it was only created to store the data, which can be added to IfExp?

Yes — and that closes the loop back to kwcckw's first question. The goF only ever held one computed scalar, `w`. It collects nothing at runtime: `w = skips·fc − runs·test_cost − blocked_value`, all read from the callee's counts and the IfExp's own cost. A CoF buys nothing over a slot on the IfExp.

Where `w` lives splits by case:

- Cost-only, one pass (now): nowhere. `eval_gates` reads counts, computes it, removes the negatives. Plain tuple, no slot.
- With the blocked-value term (proj_N): `w` refines as projection calibrates, so you store it on the IfExp to update across rounds and remove only once it's stably negative — the tagged `(ast.IfExp, (('test',..),('body',..),('else',..)), w)` form. A scalar slot, still not a CoF.

And the merge_oF `nF=='E'` branch goes as a CoF check — but the behavior it was reaching for doesn't fully vanish. `comp_prim` already matches IfExp-to-IfExp; what's left is: on that match, pack sub's branch in and recompute `w` instead of discarding it. Same append-inside logic, keyed on the tuple, not on `nF`.

The callee returns its functional output; its value (dTT/wTT, the quantity w needs) is logged by traced as metadata, not part of the return. The gate evaluates the test, so it owns the half the callee can't see: the fire/skip counts. The callee side only contributes its logged value on the fires. So w = skips·fc − runs·test_cost − blocked_value, assembled at the gate from its own counts + the callee's already-logged value + static costs.

frame_2D_alg/agg_recursion.py
```diff
@@ -179,11 +185,12 @@ def comp_C_(C_,_r, _C_=[], fall=1, fC=0):  # simplified for centroids, trans-N_s
         for C in C_: C.compared=set()
         C_ = sorted(C_, key=lambda C: C.dTT[0][np.argmax(ttC_[0])])  # max weight defines eigenvector
         for j in range( len(C_)-1):
-            _C = C_[j]; C = C_[j+1]
-            if _C in C.compared: continue
-            dy_dx = _C.yx-C.yx; dist = np.hypot(*dy_dx); c = min(C.c,_C.c); r = (C.r+_C.r)/2  # tc += c; acc[1]+=c
-            tc += c; tr += r*c
-            comp_N(_C,C,_r, c ,A=dy_dx,span=dist,L_=L_,N_=N_,acc=acc)  # simplified for typ=3
+            with CoF.gate():  # one gate per C pair?
+                _C = C_[j]; C = C_[j+1]
```
boris-kz, line 189: So the callee should return dTT/wTT, we need to sum them per gate, then sum them them again in oF[caller].body gate, to evaluate the gate in code-reforming cycle of frame_HH?

for the runing code:
if proj_m>ave: M += vt_(oF)
IfExp[0] will be something like (M, sum, m, oF)?

As the code stands, no `(M, sum, m, oF)` — `build` only emits three shapes: a `CoF` for a traced call, `(type(node), sub_)` for a node with children, or the bare `node` for a costed leaf. So for

```python
if proj_m > ave: M += vt_(oF)
```

`build` walks `ast.If` → its children are the test (`ast.Compare`) and the body stmt (`ast.AugAssign`). `vt_` isn't in `iF_` (not traced), so the call isn't an oF — it collapses to costed leaves. You get:

```python
(ast.If, ( (ast.Compare, (...)), (ast.AugAssign, (...)) ))
```

i.e. structure + costs only, no values, no oF reference. That's the current reality.

For your `(M, sum, m, oF)` to exist, `build` has to stop discarding the body and emit a value-carrying entry at the conditional. The four map to the statement as:

- `oF` — the call subject, here the `oF` in `vt_(oF)` (the AugAssign's value side)
- `m` — the test LHS, `proj_m` (`ave` is the threshold, lives in the test)
- `sum` — Σ`vt_(oF)` over fires
- `M` — the AugAssign target

So the body entry would be `(ast.If, test, (M, sum, m, oF))` rather than `(ast.If, (compare, augassign))`. But `sum` is runtime — `build` is static (parse-time), so it can only emit the static slots (`M, m, oF` + a zeroed `sum`); the eval pass fills `sum`/counts later.

One mismatch to flag: `vt_(oF)` is `vt_` applied to `oF`, so statically the "call" is `vt_`, not `oF`. If `oF` is the subject you want, `build` has to pull the argument, not the called name — it doesn't do that today.

frame_2D_alg/agg_recursion.py
kwcckw: I see, thanks. So goF is per fork within the IfExp forking only.

frame_2D_alg/agg_recursion.py
kwcckw: This dTT/wTT should be added in runtime, so we shouldn't pack it to the body?  Since the existing body is static only. It could be packed in each goF under the IfExp node?

frame_2D_alg/agg_recursion.py
boris-kz: Sorry, it seems we don't need goFs at all

frame_2D_alg/agg_recursion.py
boris-kz: yes, those dTTs per oF IfExps can be global, mapping oF_

kwcckw: > Sorry, it seems we don't need goFs at all

I see, so we only need those dTTs, which they will be packed into global oF_'s dTT? Or we need oF.gdTT, so that we can retrieve them in order to compute the runtime `sum` as in ` (M, sum, m, oF)` above?

Also with global dTT in oF, we no longer need this structure `(ast.If, test, (M, sum, m, oF))`?

boris-kz: yes, in traced, something like:

`F_call_T_ = [np.zeros((2,9)) for f in oF_]  # sum dTTs per oF.call`

`i = iF_[func.__name__];
oF_[i].call_ += [oF];
F_call_T_[i][len(oF_[i].call_)] += [oF.dTT]`

kwcckw: This ` vF_[i]` is a global packing each oF.gate's dTT? But why we have another dimension of `[call_tt_] ` here? That's mapping to each call?

boris-kz: yes, it's actually from the body:

F_call_T_ = [np.zeros((2,9)) for f in oF_ for p in f.body if isinstance(p,CoF)]  # sum dTTs per oF.call

kwcckw: I see, thanks. So each of them is mapped to a single gate in the function.

So i guess we can init them as nested array, so that we can use .sum() to retrieve per oF's gate dTT later.

boris-kz: but we probably need to index it by oF.nF, or make it a dict?

F_call_T_[i][oF.nF] += oF.dTT  # call return vals

boris-kz: Intent (sum each callee's returned dTT per call edge) is clear, but the sketch has four problems:

1. **Index spaces don't match, and the caller is absent.** The comprehension is flat — `len(F_call_T_)` = total CoFs across all bodies, not `len(oF_)`. You then index it by `i` (callee), and `[oF.nF]` indexes axis-0 of a (2,9) array (size 2) by a function index — out of range, and `+= oF.dTT` is a shape mismatch on top. Worse, `i` and `oF.nF` are the same value (both the callee), so `_CoF` never enters. Per edge you need both: `F_call_T_[_CoF.nF][i] += oF.dTT`, `F_call_T_` shaped (N, N, 2, 9), inside the `_CoF is not None` guard.

2. **dTT is zero there.** `oF.dTT` is filled only after `out = func(...)` and the post-call `if oF.call_` block. At the line you placed it, oF is fresh zeros — move the `+=` below the collection block.

3. **Leaf callees never accumulate.** Only oFs with `call_` get `sum2O`'d, so a callee making no traced sub-calls keeps dTT = 0 and contributes nothing — and a callee's value (your vt_) isn't its from-subcalls dTT anyway.

4. **Granularity is per-(caller, callee), not per-gate.** traced knows the caller and callee functions but not *which site* in the caller fired, so this merges all sites of the same callee. Gates are per-site — to keep them apart the accumulator has to sit on the body slot and increment at the site, which traced keyed by function index can't do.

boris-kz: You need a (caller, callee)→slot map, because the inner index is body *position*, not callee index — traced has the callee function, not its slot. Build it once beside `F_call_T_`, same filter/order so the j's line up:

```python
F_call_i_ = [{p.nF: j for j,p in enumerate(c for c in F.body if isinstance(c,CoF))} for F in oF_]
```

Then in traced, *after* the call and the dTT-collection block (oF.dTT is zeros before it):

```python
out = func(*a, **kw)
if oF.call_:
    tree = flat_(oF); sum2O(tree,oF,fcall_=1); ...   # fills oF.dTT
if _CoF is not None and (j := F_call_i_[_CoF.nF].get(i)) is not None:
```

`.get` because a traced call reached via an attribute/alias (not a bare Name) won't be in the static body map → otherwise KeyError.

Two caveats, both the site-identity gap resurfacing:

- Same callee at more than one slot: `{p.nF: j}` keeps only the last, so those calls fold into one slot and the rest stay zero. Unique-per-body callees are fine; multi-site still needs the actual site, which traced doesn't have.
- Leaf callees (no `call_`) keep dTT = 0, so they contribute nothing — as before.

kwcckw: > Granularity is per-(caller, callee), not per-gate. 

I think this is a problem because we need to know if the callee is mapped to which index within the same oF's F_call_T_?

Also the index should be oF's index first?

    F_call_T_[oF.nF][i] += oF.dTT

Where `i` is the index of gate within the oF, and we can't identify this `i` from the callee?

boris-kz: probably, please edit

PR #163, opened 2026-06-02
6/2/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -167,38 +203,10 @@ def parse_funcs(paths):
 nF_ = [None] * len(_names)  # FunctionDef s
 iF_ = {n: i for i,n in enumerate(_names)}  # indices name → nF, static
 oF_ = [CoF(nF=i,typ=typ) for i,typ in enumerate(typ_)]
+F_call_T_ = [[] for _ in oF_]
+F_call_i_ = [ {} for _ in oF_ ] 
 parse_funcs(["agg_recursion.py","comp_slice.py","slice_edge.py"])  # populate nF_
 F_body_()  # add F.body from AST
-F_call_T_ = [[np.zeros((2,9)) for p in F.body if isinstance(p,CoF)] for F in oF_]  # sum dTTs per oF.call, via:
```
kwcckw, line 172: So we can't use isinstance CoF here because most of the CoF are nested, either in calling, If or While loop ast. We would need to pack them flat from the F_body_ while we unpacking them, as in my line 170 above:

     F_call_T_[oF_idx].append(np.zeros((2, 9)))

frame_2D_alg/meta_code.py
```diff
@@ -131,21 +162,26 @@ def flat_(oF, call_=None):  # all nested call_ s
...
+    def build(node, oF_idx):  # AST → CoF | (type,sub_) | ast_leaf | None
 
         if isinstance(node, ast.Call) and isinstance(node.func, ast.Name):
-            if (i:= iF_.get(node.func.id)) is not None:
+            if (i:= iF_.get(node.func.id)) is not None:     
+                # init gate dTT, pack as flat
+                F_call_T_[oF_idx].append(np.zeros((2, 9)))
+                # pack gate index
+                gate_idx = len(F_call_T_[oF_idx])
+                F_call_i_[oF_idx][i] = gate_idx   # map from node to their gate idx
```
kwcckw, line 173: So with this `F_call_i_` indexing method, there's problem because it can't differentiate the callees that called from different section of the code. I think we can try the method 3 below:

From claude:

`i` is the index of the callee function (e.g., `comp_N`). If a caller has two different gates that both call `comp_N`, the dictionary uses `i` as the key both times. The second gate will overwrite the first gate.

Right — that's the real limitation of keying by callee nF. A dict keyed `{callee_nF: slot}` is intrinsically one-slot-per-callee, so two gates in the same caller that both call `comp_N` collapse: the second `build` write overwrites the first, and at runtime every `comp_N` call from that caller folds into whichever slot survived. We lose per-site resolution.

The fix is that callee identity is **not** a sufficient key. We need a key that distinguishes the two call sites. There are three honest options, and which one is right depends on whether we can tell the sites apart at runtime.

**Option 1 — key by source position, list the slots, resolve by call order at runtime.** Keep all slots for a callee instead of overwriting:

```python
# build side
F_call_i_[caller].setdefault(i, []).append(idx)   # list, in source order
```

But this only helps if at runtime we can tell *which* occurrence of the `comp_N` call we're in. Across a full execution the same caller may hit gate-A's `comp_N`, then gate-B's `comp_N`, then loop back to gate-A's again — and a plain per-callee counter can't recover source order, because execution order ≠ source order (branches, loops). So a list alone doesn't disambiguate unless we also carry a site tag through the call. Which means Option 1 collapses back into needing Option 3 anyway. It's only correct if each site executes exactly once per caller invocation in source order — too fragile to rely on.

**Option 2 — accept callee-level aggregation as the intended semantics.** If what we ultimately want is "total return-value dTT that `comp_N` contributed to this caller," then collapsing the two sites is *correct*, not a bug — we'd `+=` both into one slot deliberately. This is the cheapest path and may actually match the intent, since `F_call_T_` is summing dTTs for feedback, and per-site distinction may be noise at that level. Worth deciding before we build machinery to preserve a distinction we don't use.

**Option 3 — carry a real per-site id through the call (the only way to truly disambiguate).** Since two sites can't be told apart from callee identity or the ContextVar alone, the site id has to ride along with the call. That's exactly the keyword-injection route from before — but now it has a clear justification: the injected constant *is* the disambiguator. `gate_idx` as an `ast.Constant` baked into each call site is unique per site by construction, because `build` visits each site once and hands out a fresh index. The catch is the one from before: injecting into a re-parsed-but-never-executed tree does nothing. To make per-site work we must compile and exec the rewritten tree so the running calls actually carry their `gate_idx`. There's no shortcut; per-site resolution requires per-site data at the call, and the call is the only place that data can live.

```python
# build: bake the slot into the call, no dict needed
idx = len(F_call_T_[caller]); F_call_T_[caller].append(np.zeros((2,9)))
node.keywords.append(ast.keyword(arg='gate_idx', value=ast.Constant(idx)))
# ... then compile(tree) and exec to get the real functions, wrap THOSE

# inner: pop before the call, use directly
gate_idx = kw.pop('gate_idx', None)
out = func(*a, **kw)
...
if _CoF is not None and gate_idx is not None:
    F_call_T_[_CoF.nF][gate_idx] += oF.dTT
```

So the decision tree is clean: if callee-level aggregation is acceptable → Option 2, we're done, just `+=` into the single slot and stop fighting it. If we genuinely need per-site → Option 3, with the compile/exec step, and the dict becomes unnecessary because the call carries its own slot index directly.

Before building either, the question that decides it: for the feedback `F_call_T_` feeds, do two `comp_N` sites in one caller need separate dTT accumulators, or is their sum what we actually consume? If unsure, I'd lean Option 2 — don't pay for the compile/exec machinery until something downstream reads the per-site split.

frame_2D_alg/meta_code.py
boris-kz: ok thanks

frame_2D_alg/meta_code.py
boris-kz: we never have the same oF in the body twice?
but this is gate index, that gate can have two oF forks, we need index per fork?

frame_2D_alg/meta_code.py
```diff
@@ -114,8 +143,10 @@ def inner(*a, **kw):
             if oF.call_:
                 tree = flat_(oF)  # if len(tree)-1?
                 sum2O(tree,oF,fcall_=1); wtt = getattr(oF,'rTT',oF.dTT); oF.wTT = cent_TT(wtt,oF.r)
-                if _CoF is not None and (j := F_call_i_[_CoF.nF].get(i)) is not None:
-                    F_call_T_[_CoF.nF][j] += oF.dTT  # call return vals, draft
+                if _CoF is not None:
+                    j = F_call_i_[_CoF.nF].get(i)
+                    if j is not None:
```
boris-kz, line 148: how does that j get to be None?

frame_2D_alg/meta_code.py
kwcckw: > we never have the same oF in the body twice? 

We actually have one in proj_V:

        iTT = (_N.dTT + N.dTT) * Dec
        cNt = min(_N.c,N.c); rNt = (_N.r+N.r) / 2
        if abs(vt_(eTT,ttN_)[0])* wPrj* cNt > ave* (cPrj+r+rNt):  # spec per N link:
            eTT+= proj_N(N, dist, dy_dx, r, N.c, dec)  # pTT/ L_,B_,rim, if pV>0
            eTT+= proj_N(_N,dist, -dy_dx, r, _N.c, dec)  # reverse direction
        return iTT+eTT


> but this is gate index, that gate can have two oF forks, we need index per fork?


Yes, that index should be unique to each fork or function of the code, and even we have multiple same functions in the same body, it shouldn't be an issue. So i think we need the ast injection method 3 above so that we can identify each fork's dTT.

Or we can use the runtime packing method to check and pack new and unique fork's dTT in F_call_T_ and F_call_i during runtime?

frame_2D_alg/meta_code.py
kwcckw: Okay right, thanks. It shouldn't be None here since we pack both F_call_i_ indices and F_call_T_ array on the same time.

frame_2D_alg/meta_code.py
```diff
@@ -131,21 +162,26 @@ def flat_(oF, call_=None):  # all nested call_ s
     return call_
 
 def F_body_():  # get function bodies from their AST
-    def build(node):  # AST → CoF | (type,sub_) | ast_leaf | None
+    def build(node, oF_idx):  # AST → CoF | (type,sub_) | ast_leaf | None
```
boris-kz, line 165: where do you get this oF_idx from?

frame_2D_alg/meta_code.py
boris-kz: the way you get that oF_idx doesn't make any sense to me

frame_2D_alg/meta_code.py
kwcckw: > the way you get that oF_idx doesn't make any sense to me

This is actually the caller index, or the function's index in oF_, i parse it when call the function:

      for i, (func,name) in enumerate(zip(oF_,nF_)):
          func.caller_ = set()
          for node in ast.iter_child_nodes(name):  # skip top function definition
              rett = build(node, i)
              if rett:
                  t, fc = rett; func.body += [t]; func.fc += fc
                  
                  
Then we are packing each gated callees' dTT of oF in `F_call_T_`.

boris-kz: function's index in oF_ is oF.nF:

`oF_ = [CoF(nF=i,typ=typ) for i,typ in enumerate(typ_)]`

kwcckw: Thanks, so we just need to parse func.nF:


        for func,name in zip(oF_,nF_):
        func.caller_ = set()
        for node in ast.iter_child_nodes(name):  # skip top function definition
            rett = build(node, func.nF)
            if rett:
                t, fc = rett; func.body += [t]; func.fc += fc

frame_2D_alg/meta_code.py
```diff
@@ -131,21 +162,26 @@ def flat_(oF, call_=None):  # all nested call_ s
...
 
         if isinstance(node, ast.Call) and isinstance(node.func, ast.Name):
-            if (i:= iF_.get(node.func.id)) is not None:
+            if (i:= iF_.get(node.func.id)) is not None:     
+                # init gate dTT, pack as flat
+                F_call_T_[oF_idx].append(np.zeros((2, 9)))
```
boris-kz, line 170: >Thanks, so we just need to parse func.nF:

I have feeling all this is a pile of nonsense

kwcckw: I think the current method with `F_call_T_` and `F_call_i_` is backward because we pre-init it, and then later we need to track it again in runtime.

I would prefer to add them in runtime  since the gated dTT should be only added in runtime, and we don't need those skipped forks anyway.

frame_2D_alg/meta_code.py
boris-kz: >So with this F_call_i_ indexing method, there's problem because it can't differentiate the callees called from different section of the code.

why is that?

frame_2D_alg/meta_code.py
boris-kz: F_call_T_ is supposed to be accumulated across multiple runs

frame_2D_alg/meta_code.py
boris-kz: we just need to flatten the body?

F_call_i_ = [[i for i,p in enumerate(flat_(F.body)) if isinstance(p,CoF)] for F in oF_]

boris-kz: def flat_body_(body, out=None):

    if out is None: out = []
    for p in body:
        out += [p]
        if isinstance(p, tuple):
            flat_body_(p[1], out)
    return out

F_call_i_ = [[i for i,p in enumerate(flat_body_(F.body)) if isinstance(p,CoF)] for F in oF_]

boris-kz: or just pack indices with Ts:

`F_call_T_ = [[[i,np.zeros((2,9))] for i,p in enumerate(flat_body(F.body)) if isinstance(p,CoF)] for F in oF_] ` sum dTTs per oF.call

kwcckw: > why is that?

When we have multiple calls within a same function, this line:

     F_call_i_[oF_idx][i] = gate_idx

We may overwrite the prior gate's index.

For example, if there's 2 proj_N, we may run this first:

    F_call_i_[oF_idx][i] = 1

And then run this again on the second call within the same function , which overwrites the prior gate index:

    F_call_i_[oF_idx][i] = 2

While the `oF_idx` and `i` (callee index) are the same.

kwcckw: > we just need to flatten the body?
> 
> F_call_i_ = [[i for i,p in enumerate(flat_(F.body)) if isinstance(p,CoF)] for F in oF_]

Yes, but we would need to unpack F.body using again, and flat_ doesn't work here since it unpack nested functions, while we only need to unpack the nested ast prims under those if or while node.

boris-kz: def flat_body(body, out=None):

    if out is None: out = []
    for p in body:
        out += [p]
        if isinstance(p, tuple): flat_body(p[1], out)
    return out

F_call_T_, F_call_i_ = [],[]
for F in oF_:

    for i,p in enumerate(flat_body(F.body)):
        if isinstance(p,CoF): F_call_i_ += [i]; F_call_T_ += [np.zeros((2,9))]  # sum dTTs per oF.call

frame_2D_alg/meta_code.py
kwcckw: Yes, but we may append or accumulate at the same time too, we just need to identify if the function within the same position or line is called before or not.

frame_2D_alg/meta_code.py
kwcckw: > def flat_body(body, out=None):
> 
> ```
> if out is None: out = []
> for p in body:
>     out += [p]
>     if isinstance(p, tuple): flat_body(p[1], out)
> return out
> ```
> 
> F_call_T_, F_call_i_ = [],[] for F in oF_:
> 
> ```
> for i,p in enumerate(flat_body(F.body)):
>     if isinstance(p,CoF): F_call_i_ += [i]; F_call_T_ += [np.zeros((2,9))]  # sum dTTs per oF.call 
> ```

But with this approach, how do we know the `i` value in traced?

frame_2D_alg/meta_code.py
boris-kz: there is no append, the body is static

frame_2D_alg/meta_code.py
boris-kz: I think it should be an index of callee CoF in flat `_CoF.body`
we can probably `zip  _CoF.call_ with flat _CoF.body`?

kwcckw: > we can probably `zip _CoF.call_ with flat _CoF.body`?

To do this, the call_ must be aligned with the body, but call_ is added in runtime while body is static?

And we need to pack another flat CoF only body to align their indices?

boris-kz: Yes, that call_ should include skipped oFs.
I will check tomorrow

kwcckw: I'm not sure if this is able to track the gated callees correctly , i will try later.

PR #164, opened 2026-06-03
6/3/2026

kwcckw:


kwcckw:
> Yes, that call_ should include skipped oFs.

Actually it's still the same because we wouldn't know which call is the skipped oF? For example:


      def comp_C_(C_,_r, _C_=[], fall=1, fC=0):  # simplified for centroids, trans-N_s, levels
      
          N_,L_,tc,tr = [],[],0,0; acc = [np.zeros((2,9)),0,0,np.zeros((2,9)),0,0]
          if fall:
              for _C, C in pairs:
                  tc += c; tr += r*c
                  if _C is C:
                  else:
              if fC:
                  # merge similar distant centroids, they are non-local
                  for i, L in enumerate(L_):
                          C0,C1 = L.nt
                          add2F(C0,C1,1); merg_ += [C1]
                          for l in C1.rim: l.nt = [C0 if n is C1 else n for n in l.nt]
                      else: L_ = L_[i:]; break
          else:
              for C in C_: C.compared=set()
              for j in range( len(C_)-1):
                  _C = C_[j]; C = C_[j+1]
                  if _C in C.compared: continue
                  tc += c; tr += r*c
          for N in list(set(N_)):
              if N.rim: sum2F(N.rim, N.Rt)
          if L_: sum2F(L_,CoF.get())
          TTm,cm,rm,TTd,cd,rd = acc
          return L_,TTm,cm, rm/(cm or eps), TTd,cd, rd/(cd or eps)  # L_ is Lm now
          
 There are 2 `comp_N`s in `comp_C_`, we can't know if the call is from the first one or the second one? And we can't pack the non-running fork's call in runtime either, so we can't really track if the call is from the first or the second comp_N in comp_C_ during runtime.

kwcckw:
The flat `F_call_T_` and `F_call_i_` can't work here, especially we have 2 same function callees within the same caller. 

From claude:

is the suggested method using flat `F_call_T_` / `F_call_i_` workable, and if so what edits make it work?

**Short answer: not as a per-fork tracker, because the premise behind it is false in the current code.** The flat method leans on the assumption that a caller never contains the same callee oF twice. I checked this against the actual bodies and it doesn't hold — there are at least four duplicate call sites:

- `proj_N` appears twice in `comp_N_`
- `comp_N` appears twice in `comp_C_`
- `cross_comp` appears twice in `frame_H`
- `proj_N` appears twice in `proj_V`

Wherever a callee appears twice in one caller, a flat index keyed by callee identity collapses both sites into one slot: the second build write overwrites the first, and at runtime both calls accumulate into whichever slot survived. The two forks become indistinguishable.

The deeper reason it can't be patched by keying differently: at runtime the wrapper only knows the **callee** nF (`iF_[func.__name__]`) and the **caller** via the ContextVar. Both duplicate sites share both of those values, so nothing computable inside `traced` from `(func, _CoF)` alone can tell the two `proj_N` calls apart. And the "zip `call_` against the flat body" idea fails too — `call_` is in execution order, the body is in source order, skipped forks produce no `call_` entry, and a looped site produces many. No positional zip survives branches or loops.

**So: the flat structure is fine to keep, but it needs a per-site disambiguator that the static build can't provide.** There are exactly two ways to get one, and which is correct depends on a design question only you can answer:

**If callee-level aggregation is actually what feedback consumes** (i.e. "total dTT `proj_N` contributed to `comp_N_`", summing both sites), then the collapse is *correct, not a bug*. Keep flat `F_call_T_`, key `F_call_i_` as `{caller_nF: {callee_nF: slot}}` built statically, and just `+=` both sites into the one slot deliberately. Done — no further machinery.

**If the two forks need separate accumulators**, the disambiguator has to come from the call site itself, since callee identity isn't enough. Two routes:
- Resolve the site from the live caller frame at runtime — `sys._getframe(1).f_lasti` is distinct per physical call site (and sorted `lasti` reproduces source/body order), so each fork gets its own flat slot with no AST rewriting.
- Or inject a unique `gate_idx` constant per call site, then `compile()`/`exec()` the rewritten module so the running calls actually carry it. (Injecting into a re-parsed-but-unexecuted tree does nothing — that was the reason `gate_idx` kept coming back `None`.)

The frame route is lighter and handles loops/branches for free (each executed site folds into its own persistent slot, skipped forks never create one).

So the one question that decides everything: **for the feedback `F_call_T_` feeds, do the two `proj_N` sites in `comp_N_` need separate dTT accumulators, or is their sum what's consumed?** If summed → the flat method works as-is with deliberate collapse. If separate → the flat method works only once we add a per-site signal (frame `lasti` or injected id); the static pre-init alone cannot.

frame_2D_alg/meta_code.py
```diff
@@ -142,7 +142,10 @@ def inner(*a, **kw):
                 tree = flat_(oF)  # if len(tree)-1?
                 sum2O(tree,oF,fcall_=1); wtt = getattr(oF,'rTT',oF.dTT); oF.wTT = cent_TT(wtt,oF.r)
                 if _CoF is not None:
-                    F_call_T_[_CoF.nF][F_call_i_[CoF.nF]] += oF.dTT
+                    for i, nF in enumerate(F_call_i_[_CoF.nF]):
+                        if nF == oF.nF:  # check for same nF
+                            F_call_T_[_CoF.nF][i] += oF.dTT+1
```
kwcckw, line 147: This method is similar with the prior method, but it tracks with sequence of CoF body.

For example, if caller.body callees' nF = [1,2,3], we check with `callee.nF` to identify the index in order to retrieve the correct dTT in `F_call_T_`.

 However, it still didn't work for the case of 2 same callees under the same caller. We really need the ast node injection method or using the ast code line number to identify the correct callee index.

boris-kz:
> > Yes, that call_ should include skipped oFs.
> 
> Actually it's still the same because we wouldn't know which call is the skipped oF? For example:
> 
> ```
>   def comp_C_(C_,_r, _C_=[], fall=1, fC=0):  # simplified for centroids, trans-N_s, levels
>   
>       N_,L_,tc,tr = [],[],0,0; acc = [np.zeros((2,9)),0,0,np.zeros((2,9)),0,0]
>       if fall:
>           pairs = product(C_,_C_) if _C_ else combinations(C_,r=2)  # comp between | within list
>           for _C, C in pairs:
>               c = min(C.c,_C.c); r = (C.r+_C.r)/2  # comp weight
>               tc += c; tr += r*c
>               if _C is C:
>                   dtt = np.array([C.dTT[1],np.zeros(9)]); acc[0]+=dtt; acc[1]+=c; acc[2]+=1  # overlap=match
>               else:
>                   dy_dx = _C.yx-C.yx; dist = np.hypot(*dy_dx)
>                   comp_N(_C,C,_r,c, A=dy_dx,span=dist,L_=L_,N_=N_,acc=acc)
>           if fC:
>               # merge similar distant centroids, they are non-local
>               L_ = sorted(L_, key=lambda dC: dC.d); _C_= C_; C_, merg_ = [],[]  # from min D
>               for i, L in enumerate(L_):
>                   if vt_(L.dTT,ttF)[0]*wF < avd*(_r+cF):
>                       C0,C1 = L.nt
>                       if C0 is C1 or C1 in merg_ or C0 in merg_: continue  # not merged
>                       add2F(C0,C1,1); merg_ += [C1]
>                       for l in C1.rim: l.nt = [C0 if n is C1 else n for n in l.nt]
>                   else: L_ = L_[i:]; break
>               if merg_: N_ = list(set(N_) - set(merg_))
>       else:
>           # adjacent | distance-constrained cross_comp along eigenvector: sort by max attr, or original yx in sub+, proj C.L_?
>           for C in C_: C.compared=set()
>           C_ = sorted(C_, key=lambda C: C.dTT[0][np.argmax(ttC_[0])])  # max weight defines eigenvector
>           for j in range( len(C_)-1):
>               _C = C_[j]; C = C_[j+1]
>               if _C in C.compared: continue
>               dy_dx = _C.yx-C.yx; dist = np.hypot(*dy_dx); c = min(C.c,_C.c); r = (C.r+_C.r)/2  # tc += c; acc[1]+=c
>               tc += c; tr += r*c
>               comp_N(_C,C,_r, c ,A=dy_dx,span=dist,L_=L_,N_=N_,acc=acc)  # simplified for typ=3
>       for N in list(set(N_)):
>           if N.rim: sum2F(N.rim, N.Rt)
>       if L_: sum2F(L_,CoF.get())
>       TTm,cm,rm,TTd,cd,rd = acc
>       return L_,TTm,cm, rm/(cm or eps), TTd,cd, rd/(cd or eps)  # L_ is Lm now
> ```
> 
> There are 2 `comp_N`s in `comp_C_`, we can't know if the call is from the first one or the second one? And we can't pack the non-running fork's call in runtime either, so we can't really track if the call is from the first or the second comp_N in comp_C_ during runtime.

Ok, that flattening was wrong, we need to have dTT per callee oF in each caller.body?

kwcckw:
> Ok, that flattening was wrong, we need to have dTT per callee oF in each caller.body?

Yes, but that still doesn't solve the problem because we need to track it. The existing problem is we are not able to tell if the callee from traced is which element from the caller' body's CoF if there's more than 1 same callee in the caller's body.

boris-kz:
> > Ok, that flattening was wrong, we need to have dTT per callee oF in each caller.body?
> 
> Yes, but that still doesn't solve the problem because we need to track it. The existing problem is we are not able to tell if the callee from traced is which element from the caller' body's CoF if there's more than 1 same callee in the caller's body.

These global comments are a pain in the neck, let's comment inline only

Claude:
The Python way to inject accumulator without hand-editing every site is an AST transform — which you already have the machinery for. A NodeTransformer wraps each tracked site to push its return into that site's slot and pass it through:
pythonclass Acc(ast.NodeTransformer):

    def __init__(s,k): s.i=0; s.k=k                 # k = caller index
    def visit_Call(s,node):
        s.generic_visit(node)
        if isinstance(node.func,ast.Name) and node.func.id in iF_:
            j=s.i; s.i+=1
            return ast.Call(ast.Name('_acc',ast.Load()),
                            [ast.Constant(s.k),ast.Constant(j),node],[])
        return node
with _acc(k,j,v): F_call_T_[k][j] += <reduce v>; return v, then compile the transformed tree and swap the function in. The site index j is assigned here — that's what finally fixes site identity: repeated callees get distinct slots, no folding.
Two things:

Cost: this means recompiling the instrumented functions and replacing the originals — heavier than the current import + @traced. And F_call_T_'s inner dimension now comes from the transform's site count, not body_oF_.
_acc accumulates the site's return, so the slot's type must match what you keep — the scalar vt_(oF) or, if you want the tensor, the callee's dTT (reach it off the returned oF inside _acc).

Lighter alternative, but only when the per-site value flows through a traced call: read sys._getframe(1).f_lineno in that call's wrapper and map lineno→slot. No recompile — but it can't catch a value that isn't a traced call, e.g. M += oF.val.

frame_2D_alg/meta_code.py
boris-kz: re adding site return accumulator in CoF:

Yes — I overcomplicated it. The AST recompile is only needed if you insist on accumulating the reduced vt_(oF) at the site. If you accumulate the callee's raw dTT (which traced already has) and apply vt_ at eval instead, the one thing traced was missing is the site — and the caller frame's line gives that. So it stays in CoF, no recompile.
In traced, after the collection block:

`if _CoF is not None and (j := F_call_i_[_CoF.nF].get(inspect.currentframe().f_back.f_lineno)) is not None:`

(inspect is already imported; f_back is the caller's frame.) Setup keyed by call-site line, dimensioned by actual sites — so distinct sites get distinct slots and the identity problem is finally solved by source position:
pythondef call_sites(fd): return [n for n in ast.walk(fd) if isinstance(n,ast.Call) and isinstance(n.func,ast.Name) and n.func.id in iF_]
`F_call_T_ = [[np.zeros((2,9)) for _ in call_sites(fd)] for fd in nF_]`
`F_call_i_ = [{n.lineno: j for j,n in enumerate(call_sites(fd))} for fd in nF_]`

Two caveats:
Site = source line; two tracked calls on one line collide — key by (n.lineno, n.col_offset) to split them.
You store the callee dTT per site and reduce with vt_ at gate-eval, not at the site. Fine, since the gate's value is the callee dTT through that site — but a caller expression that isn't a function of that dTT still wouldn't be captured.

kwcckw: Yes, so i think tracking using line is better than the injection only method. 

We need to add their line information and fill in F_call_i_, and then retrieve the line information during the CoF.traced.

boris-kz: so this is good:

def call_sites(fd):  # FunctionDef 

    return [n for n in ast.walk(fd) if isinstance(n,ast.Call) and isinstance(n.func,ast.Name) and n.func.id in iF_]

F_call_T_ = [[np.zeros((2,9)) for _ in call_sites(fd)] for fd in nF_]
F_call_i_ = [{n.lineno: j for j,n in enumerate(call_sites(fd))} for fd in nF_]

kwcckw: Yes, i just perform an initial testing and able to see some mapping now.

frame_2D_alg/meta_code.py
```diff
@@ -204,10 +208,11 @@ def parse_funcs(paths):
...
+F_call_T_, F_call_i_ = [[] for _ in range(len(oF_))],[[] for _ in range(len(oF_))]
 for F in oF_:
-    for i,p in enumerate(flat_body(F.body)):
-        if isinstance(p,CoF): F_call_i_ += [i]; F_call_T_ += [np.zeros((2,9))]  # to sum dTTs per oF.call
+    for p in flat_body(F.body):
+        if isinstance(p,CoF): 
+            F_call_i_[F.nF] += [p.nF]; F_call_T_[F.nF] += [np.zeros((2,9))]  # to sum dTTs per oF.call
 # or F_call_T_ = [[[i, np.zeros((2,9))] for i,p in enumerate(flat_body(F.body)) if isinstance(p,CoF)] for F in oF_]
 
 def comp_body(_n, n, C=0):  # estimated n-merge cost compression, init mean C=3, accum from recursive unpack
```
boris-kz, line 218: so the next step is frame_HH, Chat:

Add this **after `frame_H`**, and import `split_oF_` from `meta_code`.

Change import:

```python
from meta_code import oF_,CF,CL,CC,CN,CoF, wT,wTT, eps_,eps, ave,avd,decay, trace_func,parse_funcs,cluster_oF_,split_oF_,F_body_
```

Draft:

```python
def frame_HH(image, iY,iX, Ly,Lx, Y,X, rV, max_elev=4, max_scope=4):  # higher-scope frame of frame_Hs

    def base_frame(y,x):  # one local frame_H as HH node
        F = frame_H(image, y,x, Ly,Lx, Y,X, rV, max_elev)
        if F:
            F.yx = np.array([y + Ly//2, x + Lx//2])  # actual location in image
            F.box = np.array([y, x, min(y+Ly,Y), min(x+Lx,X)])
            F.span = np.hypot(Ly,Lx) / 2
        return F

    def expand_scope(_iy,_ix, elev, F):  # same logic as frame_H.expand_lev, but inputs are frame_Hs

        frame = np.full((Ly, Lx), None, dtype=object)  # HH coordinate map, in frame_H units?
        iy,ix = _iy,_ix; cy,cx = (Ly-1)//2, (Lx-1)//2; y,x = cy,cx
        F_, PV__,C,R = [],np.zeros([Ly,Lx]),0,0
        while True:
            if not elev: F = base_frame(iy,ix)
            if F and sum(vt_(F.dTT, F.wTT*ttFrm))*(wFrm*(len(F.N_)-1)) > (ave+avd)*(F.r+elev+cFrm*(len(F.N_)-1)):
                frame[y,x] = F; F_ += [F]
                dy_dx = np.array([F.yx[0]-y, F.yx[1]-x])
                pTT = proj_N(F, np.hypot(*dy_dx), dy_dx, elev,F.c)
                if 0 < sum(vt_(pTT,F.wTT*ttFrm))*(elev+cFrm) < ave:
                    proj_focus(PV__,y,x,F)
                    if PV__[y,x] > ave:
                        iy = _iy + (y-cy) * Ly**(elev+1)
                        ix = _ix + (x-cx) * Lx**(elev+1)
                        if elev:
                            F = frame_HH(image, iy,ix, Ly,Lx, Y,X, rV, max_elev, elev)
                    else: break
                else: break
            else: break

        if F_:
            TT,C,R = sum_vt(F_, wTT=ttFrm); R += elev; L = len(F_)-1
            if sum(vt_(TT,ttFrm))*(wFrm*L) <= (ave+avd)*(R+cFrm*L):
                F_=[]; C=0; R=0
        return F_,C,R

    elev = 0
    HH,F = [],[]
    global ave,avd, Fw_,FTT_

    while elev < max_scope:
        F_,C,R = expand_scope(iY,iX, elev, F); oF=CoF.get(); oF.c += C; oF.r += R
        if F_:
            HH = CN(box=np.array([0,0,Y,X]), yx=np.array([Y//2, X//2]))
            HH.H = []; HH.N_ = F_
            if elev: [add_H(HH.H,F.H, HH,fN=1) for F in F_]
            HH.H += [lev := sum2F(F_)]
            if cross_comp(lev, rr=max_elev+elev)[0]:
                elev += 1
                if rV > ave:
                    if elev == max_scope: rV,FTT_ = ffeedback(HH); fffeedback(HH)
                    for i, tF in enumerate(oF_):
                    ave/=rV; avd/=rV; Fw_,FTT_ = np.array(Fw_) / rV, np.array(FTT_) / rV
                F = HH
            else: break
        else: break
    return HH
```

And the meta-code hook:

```python
def fffeedback(HH):  # structural feedback: reform oF_ from accumulated call / body value

    if not HH or not HH.H: return

    if sum(tF.c for tF in oF_) * wcO > ave * (HH.r + ccO*len(oF_)):
        split_oF_()
        oF_[:] = cluster_oF_()
```

Main caveat: `proj_focus()` currently assumes `tile.yx` is in the same small `PV__` coordinate frame, but `frame_HH` sets `F.yx` in image coordinates. So either normalize before `proj_focus`, or make a separate `proj_HH_focus()` that projects over frame-index coordinates. Minimal fix inside `expand_scope`:

```python
_yx = F.yx; F.yx = np.array([y,x])
proj_focus(PV__,y,x,F)
F.yx = _yx
```

Without that, the draft is structurally right but projection coordinates are mixed.

boris-kz: Claude:

Whole function with `lev0` threaded through, your `base_tile` merged in, and `fffeedback` placed:

```python
def frame_H(image, iY,iX, Ly,Lx, Y,X, rV, max_elev=4, lev0=0):  # lev0: inherited scope depth

    def base_tile(y,x, elev):  # pixels at elev=0, lower frame_H above that
        if elev == 0:
            T = frame_blobs_root(comp_pixel(image[y:y+Ly, x:x+Lx]), rV)
            T = vect_edge(T, rV)  # form, trace PP_
            if T: cross_comp(T.Nt,T.r)
        else:
            T = frame_H(image, y,x, Ly,Lx, Y,X, rV, max_elev=elev, lev0=lev0+1)
        if T:
            T.yx = np.array([y + Ly//2, x + Lx//2])
            T.box = np.array([y, x, min(y+Ly,Y), min(x+Lx,X)])
            T.span = np.hypot(Ly,Lx) / 2
        return T

    def expand_lev(_iy, _ix, elev, T):  # seed tile is pixels in 1st lev, or Fg in higher levs

        iy,ix = _iy,_ix; cy,cx = (Ly-1)//2, (Lx-1)//2; y,x = cy,cx  # start=mean
        while True:
            if not elev: T = base_tile(iy, ix, 0)
            if T and sum(vt_(T.dTT, T.wTT*ttFrm))*(wFrm*(len(T.N_)-1)) > (ave+avd)*(T.r+lev0+elev+cFrm*(len(T.N_)-1)):
                frame[y,x] = T; T_ += [T]  # loop adds one tile to level
                if 0 < sum(vt_(pTT,T.wTT*ttFrm))*(lev0+elev+cFrm) < ave:  # extend lev by combined proj T_
                    if PV__[y,x] > ave:
                        iy = _iy + (y-cy) * Ly**(lev0+elev); ix = _ix + (x-cx) * Lx**(lev0+elev)  # shifted coords
                        if elev:
                            T = base_tile(iy, ix, elev)  # lower frame_H, up to current level
                    else: break
                else: break
            else: break
        if T_:
            TT,C,R = sum_vt(T_, wTT=ttFrm); R += lev0+elev; L = len(T_)-1
            if sum(vt_(TT,ttFrm))*(wFrm*L) <= (ave+avd)*(R+cFrm*L): T_=[]; C=0; R=0
        return T_,C,R
    elev = 0
    F,tile = [],[]  # frame, seed lower tile, if any
    global ave,avd, Fw_,FTT_  # update from ffeedback:
    while elev < max_elev:
            F.H =[]; F.N_= tile_
                elev += 1
                if rV > ave:
                    if elev == max_elev: rV,FTT_ = ffeedback(F)  # data: filters, from top lev
                    for i, tF in enumerate(oF_):
            else: break
        else: break
    if lev0 == 0 and F: fffeedback(F)  # code: reform oF_ for next frame_H, outermost only
```

Changed vs your current: signature `+lev0`; `base_tile(iy,ix,0)` at loop top and `base_tile(iy,ix,elev)` replacing the inner `frame_H(...)` call; `Ly**elev`/`Lx**elev` → `**(lev0+elev)`; `+elev`/`R+=elev` → `+lev0+elev` (3 spots); `fffeedback(F)` at function end.

Three flags:

- **`fffeedback` is undefined** — placeholder (we sketched `split_oF_` → `cluster_oF_`). Comment the call out until it exists, or it's a `NameError`. I put it at the end gated on `lev0==0`, *not* inside the `elev==max_elev` block, because reforming `oF_` there changes its length and breaks the `Fw_[i]/FTT_[i]` loop right after. End placement = reform between successive top-level `frame_H` calls, matching your `__main__` note.
- **`base_tile` geometry for `elev>0`**: `yx`-as-cell-center is fine (the inner `F.yx` is image-center, unusable for placement), but `box/span` from a single `Ly×Lx` cell understate the nested frame's real, sprawled extent. Ideally derive `box/span` from its constituent tiles. Left as you wrote it.
- **Reentrant globals**: making `frame_H` recursive means nested calls now also run `ave/=rV; avd/=rV`, compounding the filter shrink across depth. Pre-existing logic, newly exposed.

Still open from last turn: `max_elev=elev` + `lev0+1` don't compose into a single coherent absolute depth, which is exactly what `lev0+elev` in the thresholds now depends on.

kwcckw: From the code above, that `expand_scope` is actually same as `expand_lev`?

It uses new ix and iy in each new iteration and a same frame. Since higher scope is for frame of frame_Hs, we need to somehow combine those frame_Hs and parse it as single higher scope's frame into frame_HH?

boris-kz: it's really the same function with a fork for tiles vs frames, see Claude version above

kwcckw: To clarify, higher scope is actually selecting frame out of past frames, but in a similar process rather than using the the frame_H output?

boris-kz: no, it's all new frames, same as new tiles in expanded frame, but processed by new oFs

boris-kz: def frame_H(image, iY,iX, Ly,Lx, Y,X, rV, max_elev=4, fHH=0):

    def base_tile(y,x):  # pixels at elev=0, lower frame_H above that
        if fHH:
            T = frame_H(image, y,x, Ly,Lx, Y,X, rV, max_elev)  # fHH=0
        else:
            T = frame_blobs_root(comp_pixel(image[y:y+Ly, x:x+Lx]), rV)
            T = vect_edge(T, rV)  # form, trace PP_
            if T: cross_comp(T.Nt,T.r)
        if T:
            T.yx = np.array([y + Ly//2, x + Lx//2]); T.box = np.array([y,x, min(y+Ly,Y),min(x+Lx,X)]); T.span = np.hypot(Ly,Lx) / 2
        return T

    def expand_lev(_iy,_ix, elev, T):  # seed tile is pixels in 1st lev, or Fg in higher levs

        iy,ix = _iy,_ix; cy,cx = (Ly-1)//2, (Lx-1)//2; y,x = cy,cx  # start=mean
        while True:
            if not elev: T = base_tile(iy, ix, 0)
            if T and sum(vt_(T.dTT, T.wTT*ttFrm))*(wFrm*(len(T.N_)-1)) > (ave+avd)*(T.r+elev+cFrm*(len(T.N_)-1)):
                frame[y,x] = T; T_ += [T]  # loop adds one tile to level
                if 0 < sum(vt_(pTT,T.wTT*ttFrm))*(elev+cFrm) < ave:  # extend lev by combined proj T_
                    if PV__[y,x] > ave:
                        iy = _iy + (y-cy) * Ly**(elev); ix = _ix + (x-cx) * Lx**(elev)  # shift coords
                        if elev:
                            T = base_tile(iy, ix, elev)  # lower frame_H, up to current level
                    else: break
                else: break
            else: break
        if T_:
            TT,C,R = sum_vt(T_, wTT=ttFrm); R+= elev; L = len(T_)-1
            if sum(vt_(TT,ttFrm))*(wFrm*L) <= (ave+avd)*(R+cFrm*L): T_=[]; C=0; R=0
        return T_,C,R
    elev = 0
    F,tile = [],[]  # frame, seed lower tile, if any
    global ave,avd, Fw_,FTT_  # update from ffeedback:
    while elev < max_elev:
            F.H =[]; F.N_= tile_
                elev += 1
                if rV > ave:
                    if elev == max_elev: rV,FTT_ = ffeedback(F, fHH)  # data: filters, from top lev
                    for i, tF in enumerate(oF_):
            else: break
        else: break

def ffeedback(frame, fHH=0):  # adjust filters via cross-level wTT ratios; fork: reform oF_ when converged

    rTT = np.divide(frame.H[0].wTT, frame.H[1].wTT); _wTT = frame.H[1].wTT
    for lev in frame.H[2:]:  # sum consecutive-level TT ratios, top-down
    rm,rd = vt_(rTT,wTT)
    if fHH and rm+rd > ave:
        split_oF_()  # speed-up code
        oF_[:] = cluster_oF_()  # compress code
        # also rebuid Fw_,FTT_,etc
    return rm+rd, rTT

kwcckw: I see, and from your code above, fHH is never set, that will be set in higher scope, after we run the first frame_H?

boris-kz: we probably don't need fHH, frame expansion is uniform, just need a new tracking for oF_ reforming cycles:

"Larger units" doesn't remove the scale relationship — it relocates it. Whether you carry `Elev` or pass bigger dimensions, the HH cell still has to span a base-frame footprint (~`Ly**max_elev`); you're only choosing whether that lives in a multiplier or in the numbers.

But the mismatch is self-inflicted by `fHH`. The original depth-matched re-seed — `frame_H(image, iy,ix, ..., elev)` — is scale-consistent for free: the outer stride at `elev` is `Ly**elev`, and a frame built up to `elev` has footprint ~`Ly**elev`, so outer cell spacing equals inner footprint, exactly. No scale param needed; scope is just `max_elev`.

`fHH=1` broke that by re-seeding to *full* depth (`max_elev`, footprint `Ly**max_elev`) while the stride stayed `Ly**elev` — which is the upscale you're now patching around.

So two clean options, pick by what the HH tiles should be:

- **Depth-matched (original):** drop both `fHH` and `Elev`, just raise `max_elev`. Scale is automatic via `Ly**elev`. HH tiles are progressive-depth — frames up to the current level, so higher levels hold deeper units.
- **Uniform full-depth tiles (`fHH`/`Elev`):** you pay the offset `Elev = Ly**max_elev` — stride `Ly**elev * Elev` (a multiplier, not `elev*Elev` in the exponent), `Elev` init 1, ×`Ly**max_elev` per scope-up. "Larger units" is exactly this, reframed.

If you don't specifically need uniform full-depth units at the HH level, the first is simplest — the recursion you already had handles scope and scale together, and `Elev` just re-encodes the `Ly**elev` you have.

One catch if you go "larger units": the larger unit is the HH *grid spacing* only. The inner base frame still needs pixel-sized `Ly,Lx`, so don't pass bigger `Ly,Lx` into the recursive call.

kwcckw: For the oF_ reforming cycle,  right now it's in every elev of frame_H, or that should be sufficient to update oF after each elevation? 

In terms of tracking, so you mean to preserve each elev's values? But why we want to track them since we only need the latest updated oFs?

boris-kz: It's a coarse-grain elev, top level only, and still conditional. Tracking means segmenting frame.H into same-oF_ level_ s

kwcckw: Sorry, oF_ is updated per max level only, i was looking at the wrong code earlier.

So each lev in frame.H should shared a same set of oF_, and to segment frame.H, that's after we split and recluster_oF_?

PR #165, opened 2026-06-04
6/4/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -893,14 +894,16 @@ def expand_lev(_iy,_ix, elev, T):  # seed tile is pixels in 1st lev, or Fg in hi
...
-            F.H += [lev := sum2F(tile_)]  # include top lev, same vals as F?
+            # lev.root should be F?
+            F.H += [lev := sum2F(tile_, root=F, froot=2)]  # include top lev, same vals as F?
             if cross_comp(lev, rr=elev)[0]:  # spec->tN_,tC_,tL_, proj comb N_'L_?
                 elev += 1
                 if rV > ave:
                     if elev == max_elev: rV,FTT_ = ffeedback(F)  # data: filters, from top lev
                     for i, tF in enumerate(oF_):
+                        # need to solve empty fw and c in oF
                         if tF: Fw_[i] = tF.fw/tF.c; FTT_[i] = lev.wTT_[i] = tF.wTT
```
kwcckw, line 907: So one of the problem in current feedback is we do not have any oF.fw and oF.c values. Both of the process weight and c should be accumulated from the data, during runtime?

boris-kz: Of course, it's in CoF

kwcckw: So we need to add the processed data information into oF.c and oF.dTT or oF.rTT. Right now it's done explicitly in some functions, such as clsuter_N：

        oF = CoF.get(); oF.N_=copy(G_); oF.rTT=sel_TT; oF.c = Ft.root.c - Ft.c; oF.r=sum([G.r * G.c for G in G_])/oF.c  # sum_vt?
        
I think we need a more consistent method to do it across all oFs functions? Possibly in traced, with all functions have a fixed output format so that we can accumulate it in traced?

frame_2D_alg/agg_recursion.py
```diff
@@ -847,10 +848,10 @@ def comp_prj_dH(_N, N, ddH, rn, link, angl, span, dec):
...
     add_H(ddH, dddH)    
 '''
-def frame_H(image, iY,iX, Ly,Lx, Y,X, rV, max_elev=4, fHH=0):
+def frame_H(image, iY,iX, Ly,Lx, Y,X, rV, max_elev=4):  # removed fHH
 
-    def base_tile(y,x):  # pixels at elev=0, lower frame_H above that
-        if fHH:
+    def base_tile(y,x, elev):  # pixels at elev=0, lower frame_H above that
+        if elev:
             T = frame_H(image, y,x, Ly,Lx, Y,X, rV, max_elev)  # fHH=0
```
boris-kz, line 855: so nested frame_H will be for Lev = same_filter_levs,
and we need higher nesting for LEV = same_oF_levs?

boris-kz: (1) Yes. `cluster_oF_` already converges the centroid phase internally, and that's the only convergence loop you need *if* split→cluster is a one-pass fixed point — which it is, structurally: `split_oF_` cuts within functions (prim-coherent sub-blocks), `cluster_oF_` groups across functions. Orthogonal axes, so they don't undo each other and won't oscillate across cycles. One assumption: `merge_oF` must preserve the prim boundaries `split_oF_` cuts on; if a merge fuses ops across them, the next split re-cuts differently and idempotence breaks. So no separate cross-cycle convergence tracker — convergence stays in `cluster_oF_`.

(2) Right structure. Two independent forks, each `dV > ave*cost`, with code cost ≫ filter cost → filters update often, `oF_` reforms rarely. One correction: the `oF_` fork's `dV` has to be the *code* value change (`fw/fc` since last reform), not `ffeedback`'s `rm+rd` (a data/wTT signal) — otherwise you trigger a recompile off a data signal.

(3) Yes, and the segmentation is load-bearing, not bookkeeping. Once the two updates fire on their own schedules, levels within a segment share a model and are comparable; across a boundary they aren't. So `ffeedback`'s cross-level `rTT` (and the code `dV`) must be summed *within* a segment — a `wTT` ratio spanning a reform boundary measures code change, not the data convergence you're tracking. That's a latent bug in the current `ffeedback`, which ratios across all of `frame.H` regardless of intervening reforms.

So the tracking that matters isn't convergence — it's a version stamp per level: an `oF_` counter and a filter counter, each incremented on its own update. `same_filter_levs` nest inside `same_oF_levs` (filters change more often, assuming a reform also starts a new filter segment). Feedback then runs per segment.

One thing to pin down: does an `oF_` reform force a filter-segment boundary too (new code ⇒ filters reset), or are the two stampings fully independent? That decides whether the nesting is strict.

frame_2D_alg/agg_recursion.py
boris-kz: Right, that's better

boris-kz: The real problem isn't the boilerplate, it's that it's *incomplete*: not every function sets `c/r/dTT|rTT`, so `traced`'s `getattr(oF,'rTT',oF.dTT)` falls back to default zeros and that oF's `wTT` — hence `fw`, `comp_body` — is meaningless. Uniform population is a prerequisite for the reform value being trustworthy, not just tidiness.

`traced` can do it only partly, and the split matters:

- **Aggregators** (oFs with `call_`): `traced` already rolls up the *code* metrics from the call tree via `sum2O(...,fcall_=1)`. Extend that to also sum `c/r/dTT` from the callee oFs — generic, and it removes the manual aggregation.
- **Leaf / selective** (`comp_*`, `cluster_*`, `proj_*`): the value is function-specific — `comp_*` derive `dTT` from the comparison, `cluster_*` compute a contrastive `sel_TT` relative to `root.Lt`, `proj_*` a projection. `traced` can't compute these; only the function knows its formula. They must emit.

So keep the leaf/selective computation but funnel the write through one helper (`set_oF(c,r, dTT=…|rTT=…)`) replacing the ad-hoc `oF=CoF.get(); oF.x=…` lines — that standardizes the field names and guarantees none are skipped. `traced` then fills only what the function left unset. Gotcha: `cluster_N`'s `oF.c = Ft.root.c - Ft.c` is contrastive, *not* a sum of children — roll-up must not overwrite it, so you need a `None` sentinel to tell "unset" from a real 0.

Also a current gap that this should close: the `wtt = cent_TT(...)` line sits inside `if oF.call_:`, so **leaf oFs never get `wTT`** — yet they're reform targets (`comp_N`, `proj_N` are in `oF_`). Lift that out of the guard (or into the helper) so it runs from whatever `dTT/rTT` the leaf emitted.

A fully uniform *return* format is the heavier alternative: it touches every call site (`G_,r = cross_comp(...)` → unpack from a result object) and still computes the contrastive values in-function, so it only moves plumbing, not computation. The emit-helper reaches the same consistency for far less churn.

Two notes: `N_` doesn't need to be uniform — code feedback only needs `c/r/dTT|rTT` → `wTT/fw`. And fix the convention while you're at it: `rTT` (relative prediction error) is the canonical value field, `dTT` only the raw fallback — which is what `traced` already assumes.

frame_2D_alg/agg_recursion.py
```diff
@@ -893,14 +894,16 @@ def expand_lev(_iy,_ix, elev, T):  # seed tile is pixels in 1st lev, or Fg in hi
         tile_,C,R = expand_lev(iY,iX, elev, tile); oF=CoF.get();oF.c += C; oF.r += R
         if tile_:  # sparse higher-scope tile, if expanded
             F = CN(box=np.array([0,0,Y,X]), yx=np.array([Y//2, X//2]))  # same center on all levels
-            F.H =[]; F.N_= tile_
+            F.H =[]; F.N_= tile_; F.c = sum([tile.c for tile in tile_])  # need c to for bool eval
             if elev: [add_H(F.H,T.H, F,fN=1) for T in tile_]  # lower levels
-            F.H += [lev := sum2F(tile_)]  # include top lev, same vals as F?
+            # lev.root should be F?
```
boris-kz, line 899: right

frame_2D_alg/agg_recursion.py
boris-kz: so frame.H has two segmentation orders, but Y,X, Ly,Lx expansion is still uniform across all levels?

frame_2D_alg/agg_recursion.py
```diff
@@ -847,10 +848,10 @@ def comp_prj_dH(_N, N, ddH, rn, link, angl, span, dec):
     link.m += dddH.m; link.d += dddH.d; link.c += dddH.c; link.dTT += dddH.dTT
     add_H(ddH, dddH)    
 '''
-def frame_H(image, iY,iX, Ly,Lx, Y,X, rV, max_elev=4, fHH=0):
+def frame_H(image, iY,iX, Ly,Lx, Y,X, rV, max_elev=4):  # removed fHH
```
boris-kz, line 851: def frame_H(image, iY,iX, Ly,Lx, Y,X, rV, max_elev=4, fH=0):  # fH=0: tiles, 1:same_filter_levs, 2:same_oF_levs

    def base_tile(y,x,T=0):  # pixels at elev=0, lower frame_H above that
        if fH:
            T = frame_H(image, y,x, Ly,Lx, Y,X, rV, max_elev, fH)
        else:  # flat H
            if not T:  # pixels at elev=0
                T = frame_blobs_root(comp_pixel(image[y:y + Ly, x:x + Lx]), rV)
                T = vect_edge(T, rV)  # form, trace PP_: 
        if T:
            T.yx = np.array([y + Ly//2, x + Lx//2]); T.box = np.array([y,x, min(y+Ly,Y),min(x+Lx,X)]); T.span = np.hypot(Ly,Lx) / 2
        return T

    def expand_lev(_iy,_ix, elev, T):  # seed tile is pixels in 1st lev, or Fg in higher levs

        iy,ix = _iy,_ix; cy,cx = (Ly-1)//2, (Lx-1)//2; y,x = cy,cx  # start=mean
        while True:
            if not elev: T = base_tile(iy, ix)  # T=0
            if T and sum(vt_(T.dTT, T.wTT*ttFrm))*(wFrm*(len(T.N_)-1)) > (ave+avd)*(T.r+elev+cFrm*(len(T.N_)-1)):
                frame[y,x] = T; T_ += [T]  # loop adds one tile to level
                if 0 < sum(vt_(pTT,T.wTT*ttFrm))*(elev+cFrm) < ave:  # extend lev by combined proj T_
                    if PV__[y,x] > ave:
                        iy = _iy + (y-cy) * Ly**(elev); ix = _ix + (x-cx) * Lx**(elev)  # shift coords
                        if elev:
                            T = base_tile(iy, ix, T)  # lower frame_H, up to current level

frame_2D_alg/agg_recursion.py
kwcckw: So for c, dTT or rTT, they must be from the data within the function only, and do we still need to accumulate them from oF.call_?

frame_2D_alg/agg_recursion.py
kwcckw: > so nested frame_H will be for Lev = same_filter_levs,
and we need higher nesting for LEV = same_oF_levs?

Yes, so oF's params should be fixed in nested frame_H, but we only get new oF lev after ffeedback in main max elev, so there's only single LEV for now?

> so frame.H has two segmentation orders, but Y,X, Ly,Lx expansion is still uniform across all levels?

I would think so, since their expansion followed a same process.

kwcckw: > fH=0: tiles, 1:same_filter_levs, 2:same_oF_levs

We have default fH = 0, but i think tiles should be part of same filter levs, while same filter levs should be part of same oF levs? So at this point i'm not sure how exactly should we really set the fH here.

kwcckw:
Sorry, not sure why my comments are pending:

image

Are you able to see them from your end?

frame_2D_alg/agg_recursion.py
boris-kz: if cross_comp(lev, rr=elev)[0]:  # spec->tN_,tC_,tL_, proj comb N_'L_?
                elev += 1
                if rV * wBac *((elev-1)*wL) > ave*cBac:  # terminate old and form new same_filter_levs
                    rV, FTT_, new_oF_ = ffeedback(F, rV, elev)  # replace elev with with len same_oF_levs? 
                    for i, tF in enumerate(oF_):
                    if new_oF_: pass   # terminate old and form new same_oF_levs
            else: break
        else: break

def ffeedback(frame, rV,elev):  # adjust filters via cross-level wTT ratios; fork: reform oF_ when converged

    rTT = np.divide(frame.H[0].wTT, frame.H[1].wTT); _wTT = frame.H[1].wTT
    oF_ = []
    for lev in frame.H[2:]:  # sum consecutive-level TT ratios, top-down
    rm,rd = vt_(rTT,wTT)
    if rV * wBac**2 *((elev-1)*wL) > ave*cBac**2:  # add separate w,c for oF_ reform
        split_oF_()  # speed-up code
        oF_ = cluster_oF_()  # compress code
    return rm+rd, rTT, oF_

frame_2D_alg/agg_recursion.py
```diff
@@ -893,14 +894,16 @@ def expand_lev(_iy,_ix, elev, T):  # seed tile is pixels in 1st lev, or Fg in hi
...
+            # lev.root should be F?
+            F.H += [lev := sum2F(tile_, root=F, froot=2)]  # include top lev, same vals as F?
             if cross_comp(lev, rr=elev)[0]:  # spec->tN_,tC_,tL_, proj comb N_'L_?
                 elev += 1
                 if rV > ave:
                     if elev == max_elev: rV,FTT_ = ffeedback(F)  # data: filters, from top lev
                     for i, tF in enumerate(oF_):
+                        # need to solve empty fw and c in oF
                         if tF: Fw_[i] = tF.fw/tF.c; FTT_[i] = lev.wTT_[i] = tF.wTT
                     ave/=rV; avd/=rV; Fw_,FTT_ = np.array(Fw_) / rV, np.array(FTT_) / rV  # Fc_ is fixed
```
boris-kz, line 908: >Sorry, not sure why my comments are pending:

you clicked on "Start a review" instead of "Comment"

kwcckw: Okay, i think there's new changes now where i need to submit the new comments:

image

But this shouldn't be needed from your end as the repo owner.

frame_2D_alg/agg_recursion.py
boris-kz: 1: terminate same_filter_levs, 2: terminate same_oF_levs

frame_2D_alg/agg_recursion.py
boris-kz: Claude:
Two things break it, plus smaller flags.

**1. No actual segmentation — nothing resets.** Both triggers use absolute `elev`, and `ffeedback` sums `rTT` over all of `frame.H[2:]`. So the window and the counter are global: a segment never closes. Concretely, once `rV*wBac*((elev-1)*wL)` crosses the bar, `ave/=rV` lowers the bar and `(elev-1)` keeps growing, so it fires every level after (segments of length 1), and `rTT` over the full H just accumulates more terms each call. For real `same_filter_levs`/`same_oF_levs` you need the *segment*, not the total: track the last filter-update and last-reform levels (`_filter`, `_oF`), compute `rTT` over the current segment (`frame.H[_filter:]`), trigger on segment length (`elev-_filter`), and set `_filter=elev` on update. Reform is the same with its own counter. This is the version stamping — load-bearing here, not optional.

**2. Reform only half-applies.** `oF_ = cluster_oF_()` assigns a *local* `oF_` (shadowing the global), and `if new_oF_: pass` discards it. So the global `oF_` ends up split (`split_oF_` mutated it in place) but never clustered. Assign the final back — `global oF_; oF_[:] = cluster_oF_()` inside `ffeedback`, or `oF_[:] = new_oF_` in `frame_H`. Also the `for i,tF in enumerate(oF_): Fw_[i]=…` loop runs *before* the reform is applied and over the now-longer split list → `Fw_` index mismatch. It has to run after the reform, on the final `oF_`, with `Fw_/FTT_` resized.

Smaller:

- Rename the local `oF_=[]` — the shadowing is exactly what produced the discard.
- `wBac²/cBac²` only makes reform *rarer* if `cBac/wBac>1`; otherwise it fires more often than the filter update. Use the separate `w,c` you flagged, not the square.
- The reform trigger is still on `rV` (a data signal). Per earlier, the code fork should gate on accumulated code value (`Σfw/Σfc` since the last reform), not `rV`.

frame_2D_alg/agg_recursion.py
kwcckw: Okay, this is much clearer now. In terms of feedback, we update oF params during 2, and update only aves in 1?

frame_2D_alg/agg_recursion.py
kwcckw: > Concretely, once rV*wBac*((elev-1)*wL) crosses the bar, ave/=rV lowers the bar and (elev-1) keeps growing, so it fires every level after (segments of length 1), and rTT over the full H just accumulates more terms each call. 

The oF feedback maybe true for any > 1 elev now?

Also with this eval: 


    if rV * wBac *((elev-1)*wL) > ave*cBac

This is actually only able to be true when elev > 1, so there is no difference between the same filter level and same oF level termination? I think we can remove the -1 in this eval, and the -1 should be needed only in feedback function?

frame_2D_alg/agg_recursion.py
boris-kz: Yes, and every update terminates corresponding order of levs and starts a new one. Each levs is a CN, similar to frame, I think it's not implemented yet.

frame_2D_alg/agg_recursion.py
boris-kz: It's one of the terms in this eval

PR #166, opened 2026-06-05
6/5/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -888,28 +889,35 @@ def expand_lev(_iy,_ix, elev, T):  # seed tile is pixels in 1st lev, or Fg in hi
...
                     ave/=rV; avd/=rV; Fw_,FTT_ = np.array(Fw_) / rV, np.array(FTT_) / rV  # Fc_ is fixed
-                    if new_oF_: pass   # terminate old and form new same_oF_levs
+                    if new_oF_:  # terminate old and form new same_oF_levs
+                        F.H += [sum2F(FoF.H)]; FoF = CN(root=F)
                 tile = F  # lev tile_ is next extension seed
             else: break
         else: break
+    # auto terminates before return?
+    if Fave.N_:
+        FoF.N_ += [sum2F(Fave.N_)]; F.H += [sum2F(FoF.N_)]
```
kwcckw, line 919: So i think the structure could be like this?

      FoF = CN packing same oF levels
      Fave = CN packing same filter levels

F.H packs FoF, while FoF.N_ packs Fave and Fave.N_ packs each tiles level?

So FoF and Fave are terminated and reset before packing to the higher level instance, for example:

      Fave.N_ += [lev := sum2F(tile_)] 
      FoF.N_ += [sum2F(Fave.N_)]; Fave = CN(root=FoF)
      F.H += [sum2F(FoF.H)]; FoF = CN(root=F)

frame_2D_alg/agg_recursion.py
```diff
@@ -868,7 +868,8 @@ def expand_lev(_iy,_ix, elev, T):  # seed tile is pixels in 1st lev, or Fg in hi
         T_, PV__,C,R = [],np.zeros([Ly,Lx]),0,0  # tiles, maps to level frame
         while True:
             if not elev: T = base_tile(iy, ix)  # T=0
-            if T and sum(vt_(T.dTT, T.wTT*ttFrm))*(wFrm*(len(T.N_)-1)) > (ave+avd)*(T.r+elev+cFrm*(len(T.N_)-1)):
+            # if T and sum(vt_(T.dTT, T.wTT*ttFrm))*(wFrm*(len(T.N_)-1)) > (ave+avd)*(T.r+elev+cFrm*(len(T.N_)-1)):
+            if T:
```
kwcckw, line 872: Sorry, this is not needed.

frame_2D_alg/agg_recursion.py
```diff
@@ -888,28 +889,35 @@ def expand_lev(_iy,_ix, elev, T):  # seed tile is pixels in 1st lev, or Fg in hi
...
         tile_,C,R = expand_lev(iY,iX, elev, tile); oF=CoF.get();oF.c += C; oF.r += R
         if tile_: # sparse higher-scope tile
             b__ = np.array([T.box for T in tile_]); box = np.array([*b__[:,:2].min(0),*b__[:,2:].max(0)])  # y,x,Y,X
             F = CN(box=box, span= np.hypot(box[2]-box[0], box[3]-box[1])/2, yx=tile_[0].yx)  # keep seed T center
             F.span = np.hypot(box[2]-box[0], box[3]-box[1]) / 2  # mean extent
-            F.H =[]; F.N_= tile_
-            if elev: [add_H(F.H,T.H, F,fN=1) for T in tile_]  # lower levels
-            F.H += [lev := sum2F(tile_)]  # include top lev, same vals as F?
+            F.N_= tile_
+            if elev: [add_H(Fave.H,T.H, F,fN=1) for T in tile_]  # lower levels
```
kwcckw, line 902: Sorry, this should be Fave.N_ here, instead of Fave.H.

frame_2D_alg/agg_recursion.py
```diff
@@ -49,7 +49,7 @@
 capitalized vars are summed small-case vars
 '''
 from meta_code import (
-    oF_,CF,CL,CC,CN,CoF, wT,wTT, eps_,eps,decay, trace_func,parse_funcs, split_oF_,cluster_oF_,F_body_)
+    oF_,CF,CL,CC,CN,CoF, wT,wTT, eps_,eps,ave, avd,decay, trace_func,parse_funcs, split_oF_,cluster_oF_,F_body_)
```
kwcckw, line 52: So why ave and avd are removed? We still need them?

frame_2D_alg/agg_recursion.py
```diff
@@ -724,7 +724,7 @@ def PP2N(PP):
                     G_,TT,C, R = trace_edge([F2N(n) for n in N_], G_,TT,c,3,tile)  # flatten B_-mediated Gs
     oF = CoF.get(); oF.c = C; oF.r = R
     if G_:
-        Nt = CF(nF='Nt', root=tile); Nt.N_ = G_; Nt.dTT = TT; Nt.c = C; Nt.r = 1; tile.Nt = Nt; Nt.H = tile.H; tile.dTT = TT; tile.c = C; L=len(G_)
+        Nt = CF(nF='Nt', root=tile); Nt.N_ = G_; Nt.dTT = TT; Nt.c = C; Nt.r = 1; Nt.H = tile.H; tile.Nt = Nt; Nt.H = tile.H; tile.dTT = TT; tile.c = C; L=len(G_)
```
kwcckw, line 727: We need to manually assign Nt.H here because CF doesn't have H. The CN.Nt.H is init when we init using CN, while the H will not be added if we init Nt separately using CF.

frame_2D_alg/agg_recursion.py
boris-kz: yes, for some reason they were blanked out

frame_2D_alg/agg_recursion.py
```diff
@@ -888,28 +889,35 @@ def expand_lev(_iy,_ix, elev, T):  # seed tile is pixels in 1st lev, or Fg in hi
             if sum(vt_(TT,ttFrm))*(wFrm*L) <= (ave+avd)*(R+cFrm*L): T_=[]; C=0; R=0
         return T_,C,R
     elev = 0
-    F,tile = [],[]  # frame, seed lower tile, if any
+    tile = []  # seed lower tile, if any
+    F= CN(); FoF = CN(root=F); Fave = CN(root=FoF)  # frame, packs same_filter_levs, packs same_oF_levs
```
boris-kz, line 893: If we are not using H, the levs should be CF?
and they should init with the first lev, not empty

frame_2D_alg/agg_recursion.py
boris-kz: right, just oH and aH, both CFs?

frame_2D_alg/agg_recursion.py
```diff
@@ -49,7 +49,7 @@
 capitalized vars are summed small-case vars
```
boris-kz, line 49: >So for c, dTT or rTT, they must be from the data within the function only, and do we still need to accumulate them from oF.call_?

Yes, that's not added yet?

kwcckw: Sorry, I haven't add that yet. Before that, could you confirm on the output structure? 
 
I'm thinking to have a consistent output structure, where the first is data (G or N_) , then rTT or dTT, c and the rest are function specific outputs. Or do you see a better alternatives?

frame_2D_alg/agg_recursion.py
kwcckw: Yes, actually we just need their N_ now. By init with the first level, we can just copy lev to init oH and aH.

frame_2D_alg/agg_recursion.py
boris-kz: yes, I think N_, dTT, w, c is enough, G would contain them all

frame_2D_alg/agg_recursion.py
boris-kz: this is redundant

frame_2D_alg/agg_recursion.py
kwcckw: Okay right, we need to processed data, so it's N_ here. Also for w, we actually didn't add them in the code now, that will be summed from N.wTT?

boris-kz: ok that's dTT instead of w. we may need fc from nested calls too?

frame_2D_alg/agg_recursion.py
kwcckw: Yes thanks, i can see it's already added.

frame_2D_alg/agg_recursion.py
kwcckw: Yes, that's one of my concern in prior discussion too. I think we need nested fc because that's the actual complexity? The top level fc is just the ast level complexity, which only counts the fc of the calling process.

frame_2D_alg/agg_recursion.py
boris-kz: `if T and vt_(T.dTT,T.wTT*ttFrm) *((len(T.N_)-1)*wL) *wFrm > (ave+avd)*(T.r+elev+cFrm*(len(T.N_)-1)):` # complex gate, make incremental?

frame_2D_alg/agg_recursion.py
boris-kz: if T and vt_(T.dTT,T.wTT*ttFrm) *((len(T.N_)-1)*wL) *wFrm > (ave+avd)*(T.r+elev+cFrm):  # extended frame ( same_oF_levs ( same_filter_levs

but it's probably not needed

kwcckw: Do you mean adding another eval here? But that T shouldn't be relevant to this section since that's in expand_lev.

frame_2D_alg/agg_recursion.py
kwcckw: The existing eval is already incremental when we have elev on the right side?

frame_2D_alg/agg_recursion.py
boris-kz: if elev: [add_H(Fr.H, T.H, Fr, fN=1) for T in tile_]  # lower level tiles
            # add_H should recursively unpack and add nested oH ( aH

frame_2D_alg/agg_recursion.py
boris-kz: no, I meant incrementally detailed eval, nested gating 

old eval was all wrong anyway

frame_2D_alg/agg_recursion.py
boris-kz: I meant init with real levels

frame_2D_alg/agg_recursion.py
```diff
@@ -888,28 +889,35 @@ def expand_lev(_iy,_ix, elev, T):  # seed tile is pixels in 1st lev, or Fg in hi
...
         if tile_: # sparse higher-scope tile
             b__ = np.array([T.box for T in tile_]); box = np.array([*b__[:,:2].min(0),*b__[:,2:].max(0)])  # y,x,Y,X
             F = CN(box=box, span= np.hypot(box[2]-box[0], box[3]-box[1])/2, yx=tile_[0].yx)  # keep seed T center
             F.span = np.hypot(box[2]-box[0], box[3]-box[1]) / 2  # mean extent
-            F.H =[]; F.N_= tile_
-            if elev: [add_H(F.H,T.H, F,fN=1) for T in tile_]  # lower levels
-            F.H += [lev := sum2F(tile_)]  # include top lev, same vals as F?
+            F.N_= tile_
+            if elev: [add_H(Fave.H,T.H, F,fN=1) for T in tile_]  # lower levels
+            Fave.N_ += [lev := sum2F(tile_)]  # include top lev, same vals as F?
```
boris-kz, line 903: this is the whole frame

frame_2D_alg/agg_recursion.py
boris-kz: if tile_: # sparse higher-scope tile
            Fr = sum2F(tile_)
            if elev: [add_H(Fr.H, T.H, Fr, fN=1) for T in tile_]  # lower level tiles, add_H should recursively unpack and add nested oH(aH
            if cross_comp(Fr, rr=elev)[0]:  # spec->tN_,tC_,tL_, proj comb N_'L_?

frame_2D_alg/agg_recursion.py
boris-kz: It's just adding nesting to frame.H, and add_H should unpack that nesting

frame_2D_alg/agg_recursion.py
kwcckw: This is packing lev to Fr.H now, but how about aH.N_? Or the termination is actually done from Fr.H now?

boris-kz: that's in add2F, it handles nested H:

def add2F(F, n, merge=0, fr=0, fo=0):  # unpack for batching in sum2F

    if F.c:
    else:
    if merge <2:
        if fo: F.typ_+= [n]
        else: F.N_ += (n.N_ if merge else [n])
    if hasattr(n,'C_'): F.C_ = getattr(F,'C_',[]) + n.C_  # same for L_?
    return F

def add_H(H,h, root, fN=0):

    for Lev,lev in zip_longest(H, h):  # bottom-up
        if lev:
            if Lev: add2F(Lev,lev,1)  # unpack nested Hs in add2H

boris-kz: Claude:
The recursion path exists, but it only fires under conditions the current construction never produces:

1. `hasattr(F,'H')` — CF has no H (per kwcckw's own comment in the thread, hence the manual `Nt.H = tile.H` in vect_edge). add_H appends levs as Copy_ with cls=CF by default (fN=0), and Copy_ to CF drops the source's H entirely — no H key in its dict. So levs created through the default path can never recurse, and merging a CF Lev with a CN lev silently discards n's nesting. The fN split means one H can hold both classes depending on which call built each lev: cross_comp's add2F→add_H path uses fN=0, frame_H uses fN=1.

2. More fundamental: where do group members live? Every group constructor in play — `sum2F(tile_)`, kwcckw's `FoF.N_ += [sum2F(Fave.N_)]` — puts members in N_, not H. sum2F never aggregates member H into F.H. So when add_H hits two group-levs, the line that executes is `F.N_ += n.N_` (merge=1): flat concat of levs from different frames, no alignment, before the H-recursion line is ever relevant. The recursion unpacks .H nesting; the nesting is being built in .N_. Either the group constructor changes (header = sum_vt(levs), members → F.H, N_ stays empty), or add2F needs a group flag that routes N_-stored members through add_H instead of +=.

3. Even with members in .H, the recursion just pushes positional zip one level down: sub-levs align by index at every depth. That holds only if regime boundaries are synchronized across the Hs being merged — feedback between elevs, top frame only. With sub-frame-local ffeedback, two tiles' groups have different boundaries and index-alignment merges incommensurate levs; recursion doesn't detect that, it propagates it. So this code is consistent with the sync answer to my earlier question, and only that answer.

4. Group-vs-flat mismatch is still undefined: flat lev merging into a group-lev goes into the group *header* (dTT, c, r) and concats with header N_, rather than into the matching sub-lev. If mixed depths can occur — tile that never fired feedback merged with one that did — you need a rule: wrap the flat lev as a singleton group first, or match it to the group's last open sub-lev.

So: keep the add2F recursion as the unpack mechanism, but it needs (a) groups stored in H with a constructor that fills H, (b) uniform lev class (CN, or H added to CF), (c) the sync invariant made explicit — no local feedback in sub-frames — since alignment stays positional. Otherwise the comment describes intent, not behavior.

kwcckw: But this is simply combining all same filter levels and same oF levels, which we don't have termination of same filter levels or same oF levels into a nested level anymore?

boris-kz: that should be added in ffeedback internally

boris-kz: ffeedback with terminate/wrap (lazy form: the open group is the unwrapped tail, "form new" = next levs accumulate after the wrap):

```python
def ffeedback(frame, rV,elev):  # adjust filters via cross-level wTT ratios; fork: reform oF_ when converged

    H = frame.H
    rTT = np.divide(H[0].wTT, H[1].wTT); _wTT = H[1].wTT
    for lev in H[2:]:  # sum consecutive-level TT ratios, bottom-up
    rm,rd = vt_(rTT,wTT)
    def wrap(H, nF, in_=('aH','oH')):  # terminate: group trailing levs formed under old regime
        i = next((j+1 for j in reversed(range(len(H))) if H[j].nF in in_), 0)  # end of last closed group
        if len(H) > i:
            G = CN(root=frame,nF=nF); G.H = H[i:]; G.dTT,G.c,G.r = sum_vt(G.H); G.m,G.d = vt_(G.dTT)
            G.wTT = copy(frame.wTT)  # old-regime stamp: cross-regime rTT, add_H alignment
            return H[:i] + [G]
        return H
    frame.H = wrap(frame.H,'aH')  # terminate old aH: same_filter_levs
    if rV * wBac**2 *((elev-1)*wL) > ave* cBac**2:  # add separate w,c for oF_ reform
        split_oF_()  # speed-up code
        oF_ = cluster_oF_()  # compress code
        frame.H = wrap(frame.H,'oH', in_=('oH',))  # terminate old oH: same_oF_levs = trailing aHs
    return rm+rd, rTT
```

aH wrap is unconditional per call since every ffeedback call is a filter-regime change in frame_H. Singletons get wrapped too: depth then encodes regime history, not lev count, so positional zip in add_H aligns across frames with the same feedback timeline. If you skip singleton wrap, mixed depths reappear and add_H needs the group-vs-flat rule.

For lev as CN, three change points:

1. `sum2F(N_, root=None, m_=[],d_=[], merge=0, froot=0, cls=None)` with `F = (cls or cls_[typ])(dTT=TT, c=C, r=R)` — typ dispatch currently makes cross_comp's lev a CL (links → typ 1), which has no H, so group nesting dies there.
2. cross_comp line 96: `root.H += [sum2F(L_,root,froot=1, cls=CN)]`.
3. add_H: `H.append(Copy_(lev, root, cls=CN))`, drop fN — Copy_ to CF discards H, and the fN split is what made H heterogeneous. Copy_'s CN branch already deep-copies N.H, so appended group levs keep their nesting.

Two caveats. The rTT loop now mixes depths: closed groups (regime-level wTT ratios, the meaningful signal) and flat tail levs (default wTT, ratio ≈ 1 noise). Once wraps exist, the ratios should probably iterate the top grouping only — groups against groups, tail summed as one pseudo-group. Second, `G.wTT = copy(frame.wTT)` stamps a default until something actually writes per-frame wTT at regime close; the caller writes Fr.wTT_ per oF after ffeedback returns, i.e. the new regime, so the old-regime snapshot has to be taken here or the stamp stays vacuous.

kwcckw: I see, so i think we need to redefine `in_=('aH','oH')`? For level added before feedback, lev.nF should be pre-aH, and then this line should be ` frame.H = wrap(frame.H,'aH', ('pre-aH',))`?

Then in the oF termination, since we terminate only lev with nF=aH, `in_` should be `aH`?

    frame.H = wrap(frame.H,'oH', in_=('aH',))

frame_2D_alg/agg_recursion.py
boris-kz: not sure actually, nested vals are not oF-specific

frame_2D_alg/agg_recursion.py
boris-kz: the names should be assigned depending on what's updated, oF_ or just aves

kwcckw: Yes, there's a problem in current naming setup becauswe in this line:

     frame.H = wrap(frame.H,'oH', in_=('oH',))
     
 We are init new G with nF = oH, but looking to merging levels with nF == oH? It should be merging levels == aH instead?

frame_2D_alg/agg_recursion.py
kwcckw: I see, in that case we should skip the nested values .

frame_2D_alg/agg_recursion.py
boris-kz: this wrapping should not be a default anyway

boris-kz: Yes — wrap-per-call was matching the current caller, where every ffeedback return is followed by unconditional filter writes (ave/=rV, Fw_, FTT_[i]=tF.wTT), so call = update. If the update itself becomes conditional, the wrap must follow the same condition, and the two can't live in different functions without desyncing: filters changed but no aH boundary, or boundary with unchanged filters — both corrupt what same_filter_levs means. So move the update into ffeedback and gate both on actual change:

```python
def ffeedback(frame, rV,elev):  # adjust filters via cross-level wTT ratios; fork: reform oF_ when converged

    global ave,avd, Fw_,FTT_
    H = frame.H
    rTT = np.divide(H[0].wTT, H[1].wTT); _wTT = H[1].wTT
    for lev in H[2:]:
    rm,rd = vt_(rTT,wTT); rV = rm+rd
    if abs(rV-1) * wBac > ave * cBac:  # filters actually update: terminate old aH
        frame.H = wrap(frame.H,'aH')   # same_filter_levs
        for i, tF in enumerate(oF_):
            if tF: Fw_[i] = tF.fw/tF.c; FTT_[i] = frame.wTT_[i] = tF.wTT
        ave/=rV; avd/=rV; Fw_ = np.array(Fw_)/rV; FTT_ = np.array(FTT_)/rV
        if rV * wBac**2 *((elev-1)*wL) > ave* cBac**2:  # oF_ reform
            split_oF_(); oF_n = cluster_oF_()
            for i, F in enumerate(oF_n): F.nF = i
            frame.H = wrap(frame.H,'oH', in_=('oH',))  # same_oF_levs = trailing aHs
    else: rV = 1  # no update, no termination, caller no-op
    return rV, rTT
```

Caller shrinks to the outer gate plus `rV, FTT_ = ffeedback(Fr, rV, elev)`.

Two open points in the gate itself. First, the no-change baseline of rTT isn't 1 per cell — it's a sum of len(H)-1 unit ratios, so rm+rd conflates depth with change; significance should be measured against that baseline (rTT normalized by lev count, or accumulate ratio deviations instead of ratios). Second, granularity: FTT_[i] = tF.wTT updates per-oF weights even when the aggregate rV is flat. If per-oF drift below the gate is allowed to accumulate without terminating aH, "same_filter" becomes approximate — fine if intended (aH = same regime within ave tolerance), but then the per-oF writes should also sit inside the gated branch, as above, not run on every call.

kwcckw: Okay, let me go through the updates first.

PR #167, opened 2026-06-06
6/6/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -137,7 +137,10 @@ def inner(*a, **kw):
                 _CoF.call_ += [oF]
                 oF_[iF_[func.__name__]].caller_.add(_CoF)  # for comp_caller_
             _oF = CoF._cur.set(oF)
-            out = func(*a, **kw)
+            out = func(*a, **kw)  # comp_slice and slice_edge is not under traced yet
+            if out:  # not None
+                *out, (N_, rTT, c, r) = out
```
kwcckw, line 142: So instead of (N_, dTT, c, r), i think we can pack the oF params as a tuple in the last output? Then we can unpack them in traced, and the original function output will not be affected.

frame_2D_alg/agg_recursion.py
```diff
@@ -884,13 +882,14 @@ def expand_lev(_iy,_ix, elev, T):  # seed tile is pixels in 1st lev, or Fg in hi
...
+                return TT,C,R
+        return [], 0, 0
     elev = 0
     Fr, tile = [],[]  # seed with lower tile, if any
     global ave,avd, Fw_,FTT_  # update from ffeedback:
     while elev < max_elev:
-        tile_,C,R = expand_lev(iY,iX, elev, tile); oF=CoF.get();oF.c += C; oF.r += R
+        tile_,C,R = expand_lev(iY,iX, elev, tile)
         if tile_: # sparse,2D
             Fr = sum2F(tile_)  # higher-scope tile ( same_oF_levs ( same_filter_levs
```
kwcckw, line 894: The current sum2F sums tiles into Fr, but their H is always empty here?  Or if the prior elev Fr exist, that should be packed into current Fr.H

frame_2D_alg/agg_recursion.py
```diff
@@ -217,7 +217,7 @@ def comp_H(_Nt,Nt, Link):  # tentative pre-comp
         for fb_, nF in zip((L.Nt.fb_,L.Lt.fb_,L.Bt.fb_,L.Ct.fb_), ('Nt','Lt','Bt','Ct')):
             if fb_: # L+= trans-Ls, python-batched bottom-up:
                 sum2F(fb_,getattr(L,nF))
-        sum2F([L.Nt,L.Lt,L.Bt,L.Ct], CoF.get())
+        sum2F([L.Nt,L.Lt,L.Bt,L.Ct], CoF.get())  # data should be L.Fts?
```
kwcckw, line 220: So oF.N_ should be the processed data, which should be N and `_N`  here? And clustering functions such as cluster_N, cluster_C and cluster_P are special cases where their oF.N_ is the resulting clustered data?

frame_2D_alg/agg_recursion.py
```diff
@@ -922,7 +922,7 @@ def ffeedback(frame, rV,elev):  # adjust filters via cross-level wTT ratios; for
             for i, F in enumerate(oF_n): F.nF = i
             frame.H = wrap(frame.H,'oH', in_=('oH',))  # same_oF_levs = trailing aHs
     else: rV = 1  # no update, no termination, caller no-op
-    return rV, rTT
+    return rV, rTT, ([frame.H], rTT, frame.c, frame.r)
```
boris-kz, line 925: return Fr, ([Fr], Fr.dTT, Fr.c, Fr.r) if Fr else ()

frame_2D_alg/agg_recursion.py
boris-kz: Right, L is pairwise
so we need to replace L.nt with L.N_?

frame_2D_alg/agg_recursion.py
boris-kz: yes, `Fr.H += sum2F([N for t in tile_ for N in t.N_], Fr)  # sum2G minus Fts`?

frame_2D_alg/meta_code.py
boris-kz: It's the same
oF.N_ = N_; oF.rTT = rTT; oF.c += c; oF.r += r
only in function body?

frame_2D_alg/agg_recursion.py
```diff
@@ -922,7 +922,7 @@ def ffeedback(frame, rV,elev):  # adjust filters via cross-level wTT ratios; for
             for i, F in enumerate(oF_n): F.nF = i
             frame.H = wrap(frame.H,'oH', in_=('oH',))  # same_oF_levs = trailing aHs
```
boris-kz, line 923: wrap is very similar to sum2F?

Yes — the header math (c-weighted dTT,c,r + vt_) duplicates sum2F; the only substantive differences are member slot (H instead of N_), the tail-boundary search, and the regime stamp. Fold the first into sum2F, keep the other two in ffeedback:

```python
def sum2F(N_, root=None, m_=[],d_=[], merge=0, froot=0, nH=''):  # nH: 'aH'|'oH', group lev with members in H
    ...
    F = (CN if nH else cls_[typ])(dTT=TT, c=C, r=R)  # force CN for group levs, mixed-typ members
    if typ:
        if nH: F.H = n_; F.nF = nH
        else:  F.N_ = n_
        ...
```

ffeedback then:

```python
i = next((j+1 for j in reversed(range(len(H))) if H[j].nF in in_), 0)  # end of last closed group
if len(H) > i:
    G = sum2F(H[i:], nH=nF); G.wTT = copy(frame.wTT); frame.H = H[:i] + [G]
```

CN forcing matters because the aH tail can hold cross_comp levs, which sum2F's typ dispatch makes CL (links → typ 1) — no H slot, group dies. Same reason the earlier point stands: cross_comp's flat lev itself should be CN (cls or nH-less override), or every Copy_/add2F touching mixed CL/CN levs needs hasattr guards.

One check on reuse: sum2F's typ branch also gives the group kern/span/yx as c-weighted member means — harmless for aH/oH, arguably right (group inherits its levs' centroid), and box only on typ 3, which member CNs default anyway. The parts that stay in ffeedback — boundary search and wTT stamp — are regime logic, not summation, so no further collapse available.

kwcckw: So we can actually replace wrap with sum2F?

From claude:

    def ffeedback(frame, rV, elev):
        ...
        i = next((j+1 for j in reversed(range(len(H))) if H[j].nF in in_), 0)
        if len(H) > i:
            G = sum2F(H[i:], nH=nF)  # sum2F now handles CN+H, math, vt_
            G.wTT = copy(frame.wTT)  # regime stamp stays here
            frame.H = H[:i] + [G]

frame_2D_alg/agg_recursion.py
kwcckw: > so we need to replace L.nt with L.N_?

Sorry, i don't get this? Why we need to replace L.nt with L.N_ here when the oF.data should be L.nt?

frame_2D_alg/agg_recursion.py
kwcckw: But that only pack tile_'s N_, but not all the prior elev's Fr.H? I guess we need `_Fr` to pack prior elev's Fr, something like this?

    Fr.H += _Fr.H + [sum2F([N for t in tile_ for N in t.N_], Fr)]

frame_2D_alg/meta_code.py
kwcckw: Okay, so actually it's better to move them into the function body? Then we can skip this additional process in traced.

frame_2D_alg/agg_recursion.py
boris-kz: in sum2F, add:

```
if typ==2:
    if not F.Nt  # frame?
         for N in N_: add_H(F,N, F)  # concat lower levs
         F.H += sum2F([n for N in N_ for n in N.N_], F)  # add top lev
    F.m_,F.d_=m_,d_; F.m,F.d=sum(m_),sum(d_)
else: F.m, F.d = vt_(TT)
```

frame_2D_alg/agg_recursion.py
boris-kz: L.nt is redundant, syntax-wise. We currently only use L.N_ to pack trans_Ls, but that should be nested?

frame_2D_alg/agg_recursion.py
boris-kz: We don't need it, it's all in Fr = sum2F(tile_)

frame_2D_alg/meta_code.py
boris-kz: yes, because this return is redundant

frame_2D_alg/agg_recursion.py
boris-kz: No — flat H isn't the requirement, position-wise uniform nesting is: whatever depth sits at H[i] must be the same depth in every H that meets it in add_H or comp_H. Synchronized wraps give that without flattening — within one frame_H, feedback fires between elevs, so all H's that ever zip share boundaries. comp_H already tolerates nesting since it compares lev headers only, and group dTT is a c-weighted mean, so scale stays commensurate; add2F's H recursion handles group×group. The only way nesting breaks alignment is sub-frame-local feedback — still the open decision from before; if sub-frames inherit the top frame's regime clock, mandatory flatness buys nothing.

Direct H as Nt-alternative: agreed, and `not F.Nt` is the right discriminator — it covers all Ft-less products (frames, centroids, group levs), and once cross_comp or vect_edge installs Nt, the add_Nt path owns H. But the placement makes it unreachable for the case that motivated it: frames don't pass m_, so fC=0 and typ=N.typ — the typ==2 arm never runs for them. Lift the two H lines out of the typ==2 gate into the general `if typ:` block, keep the m_ assignment where it is.

Two traps in the lines themselves:

1. Recursion: the inner `sum2F([n for N in N_ for n in N.N_], F)` product also has no Nt, so it hits the same gate and builds H from grandchildren, and so on to the leaves — every sum2F call rebuilds the entire subtree pyramid. The gate selects who gets direct H; a flag still has to select when, with the inner call plain. So the fH param survives, just demoted to recursion guard.

2. root=F in the inner call: add2F(F, lev, 2) re-adds the children's stats that F's header already contains via the members — F.c roughly doubles. froot=2 for the root link, no header merge.

With those, the wrap reduces to this same path: ffeedback hands sum2F the segment, members land in N_, the direct top lev is their summary, member H's merge below — and aH/oH are ordinary Ft-less products rather than a special class.

boris-kz: def ffeedback(frame, rV,elev):  # adjust filters via cross-level wTT ratios; fork: reform oF_ when converged

    global ave,avd, Fw_,FTT_
    H = frame.H
    rTT = np.divide(H[0].wTT, H[1].wTT); _wTT = H[1].wTT
    for lev in H[2:]:
    rm,rd = vt_(rTT,wTT); rV = rm+rd
    if abs(rV-1) * wBac > ave * cBac:  # filters actually update: terminate old aH
        frame.H += [sum2F(frame.H)]   # lev = same_filter_levs, nF='aH'?
        for i, tF in enumerate(oF_):
            if tF: Fw_[i] = tF.fw/tF.c; FTT_[i] = frame.wTT_[i] = tF.wTT
        ave /= rV; avd /= rV
        Fw_ = np.array(Fw_)/rV; FTT_ = np.array(FTT_)/rV
        if rV * wBac**2 *((elev-1)*wL) > ave* cBac**2:  # oF_ reform
            split_oF_(); oF_n = cluster_oF_()
            for i, F in enumerate(oF_n): F.nF = i
            frame.H += [sum2F(frame.H)]  # lev = same_oF_levs, nF='oH'?
    else: rV = 1  # no update, no termination, caller no-op
    return rV, rTT

frame_2D_alg/meta_code.py
kwcckw: Alright, i guess it's better to assign them internally within the function.

frame_2D_alg/agg_recursion.py
kwcckw: I see, but we don't use L.N_ to pack trans anymore?

    def trans_cluster(G):
        for L in G.L_:
            for tFt in L.Nt,L.Bt,L.Ct:  # Lt doesn't form trans-links, Ct is not root-constrained?
                for tL in tFt.N_:
                
They are packed in L.Nt, Bt and Ct, which we actually can use N_ to replace nt now.

frame_2D_alg/agg_recursion.py
kwcckw: > in sum2F, add:
> 
> ```
> if typ==2:
>     if not F.Nt  # frame?
>          for N in N_: add_H(F,N, F)  # concat lower levs
>          F.H += sum2F([n for N in N_ for n in N.N_], F)  # add top lev
>     F.m_,F.d_=m_,d_; F.m,F.d=sum(m_),sum(d_)
> else: F.m, F.d = vt_(TT)
> ```


I see, but why typ == 2? That's for C, but we need N.typ == 3?

frame_2D_alg/agg_recursion.py
boris-kz: right

frame_2D_alg/agg_recursion.py
boris-kz: yes, N.typ == 3

frame_2D_alg/meta_code.py
boris-kz: I forgot, what's the expression to get oF in the body?

kwcckw: We can use `oF = CoF.get()`, then sum data params into this oF.

frame_2D_alg/agg_recursion.py
kwcckw: So with `frame.H += [sum2F(frame.H)] `, there's no explicit nF = aH or oH, and they will be just nested under H.H..?

boris-kz: this is a tail, it should packed in aH, oH in ffeedback, something like this:

def ffeedback(frame):  # adjust filters via cross-level wTT ratios; fork: reform oF_ when converged

    global ave,avd, Fw_,FTT_
    H = frame.H
    i = next((j+1 for j in reversed(range(len(H))) if H[j].nF in ('aH','oH')), 0)  # open tail start, prior levs are packed
    if tail := H[i:]:
        _wTT = H[i-1].wTT if i else tail[0].wTT  # regime-start stamp | first tail lev, if i==0: len tail>2?
        rTT, dTT, k = np.zeros((2,9)),np.zeros((2,9)),0
        for lev in tail if i else tail[1:]:  # lev.wTT
            r_ = np.divide(_wTT, lev.wTT); rTT+=r_; dTT+=np.abs(r_-1); _wTT=lev.wTT; k+=1
        rV = np.sum(rTT*wTT) / (k*np.sum(wTT))  # scale by mean ratio, 1-centered
        rD = np.sum(dTT*wTT) / np.sum(wTT)  # gate by drift, 0-centered
        if rD * wBac > ave * cBac:  # filters update ->aH:
            aH = CN(nF='aH',root=frame); aH.H= tail; aH.wTT = copy(frame.wTT); aH.wTT_ = copy(frame.wTT_)  # old-regime stamps
            if rD *wBac > ave *(cBac+1):  # 1: nesting cost
                aH.dTT,aH.c,aH.r = sum_vt(tail); aH.m,aH.d = vt_(aH.dTT)
            frame.H = H[:i] + [aH]  # same_filter_levs
            for j, tF in enumerate(oF_):
                if tF: Fw_[j] = tF.fw/tF.c; FTT_[j] = frame.wTT_[j] = tF.wTT
            ave/=rV; avd/=rV; Fw_ = np.array(Fw_)/rV; FTT_ = np.array(FTT_)/rV
            if rD * wBac**2 > ave * cBac**2:  # oF_ reform ->oH:
                split_oF_(); oF_n = cluster_oF_()  # still not rebound
                for j, F in enumerate(oF_n): F.nF = j
                i = next((j+1 for j in reversed(range(len(frame.H))) if frame.H[j].nF=='oH'), 0)
                aH_ = frame.H[i:]  # contiguous aHs, tail was packed above
                if len(aH_) > 1:  # single aH: oH = pure duplicate
                    oH = CN(nF='oH',root=frame); oH.H = aH_; oH.wTT = copy(frame.wTT); oH.wTT_ = copy(frame.wTT_)
                    if all(aH.c for aH in aH_): oH.dTT,oH.c,oH.r = sum_vt(aH_); oH.m,oH.d = vt_(oH.dTT)  # else fill cascades on demand
                    frame.H = frame.H[:i] + [oH]  # same_oF_levs
            return rV
    return 1  # no update, no termination

kwcckw: With this method, we would need to assign lev.nF = aH here? 

    frame.H += [sum2F(frame.H)]

Else we won't get any tail in feedback at all?

boris-kz: Tail is unpacked levs, not aH

PR #168, opened 2026-06-07
6/7/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -926,29 +931,31 @@ def drift(lev_, _wTT):  # ratio chain over wTT stamps
...
+            oH = CN(nF='oH',root=frame); oH.H = aH_; oH.wTT = copy(eTT)
             if all(a.c for a in aH_): oH.dTT,oH.c,oH.r = sum_vt(aH_); oH.m,oH.d = vt_(oH.dTT)
             # else fill cascades on demand
             frame.H = frame.H[:i] + [oH] + frame.H[i+len(aH_):]  # same_oF_levs, open tail follows
+    oF = CoF.get(); oF.N_=H; oF.dTT=frame.dTT; oF.c=frame.c; oF.r=frame.r
     return rV
 
 if __name__ == "__main__":  # './images/toucan_small.jpg' './images/raccoon_eye.jpeg', add larger global image
     trace_func(vars())
+    comp_slice = CoF.traced(comp_slice); slice_edge = CoF.traced(slice_edge)  # add traced to comp_slice and slice_edge
```
kwcckw, line 957: So we need to add `traced` to comp_slice and slice_edge, else we won't get their oF updated.

frame_2D_alg/agg_recursion.py
```diff
@@ -926,29 +931,31 @@ def drift(lev_, _wTT):  # ratio chain over wTT stamps
...
-            ave/=rV; avd/=rV; Fw_ = np.array(Fw_)/rV; FTT_ = np.array(FTT_)/rV
+            # we don't need Fw_, wTT_ and FTT_ now? They should be accessed from oF anyway
+            ave/=rV; avd/=rV
         else: rV = 1  # no update, no termination
     # oF_ reform, any tail gate:
     i = next((j+1 for j in reversed(range(len(frame.H))) if frame.H[j].nF=='oH'), 0)
     if len(aH_ := [l for l in frame.H[i:] if l.nF=='aH']) > 1:  # contiguous from i
         aV,aD,eTT = drift(aH_[1:], aH_[0].wTT)  # cross-regime drift over aH
-        if aD * wBac**2 < ave * cBac**2:  # stable regimes -> reform oH
+        if aD * wBac**2 < ave * cBac**2:  # stable regimes -> reform oH  (why not stable for aH too?)
```
kwcckw, line 945: So here we form oH for stable regime, when D is < ave, but why it's the other way around for aH `rD * wBac > ave * (cBac+1)`?

frame_2D_alg/agg_recursion.py
```diff
@@ -926,29 +931,31 @@ def drift(lev_, _wTT):  # ratio chain over wTT stamps
...
         rV,rD,eTT = drift(tail if i else tail[1:], H[i-1].wTT if i else tail[0].wTT)  # needs lev.wTT stamped at creation
         if rD * wBac > ave * cBac:  # filters update ->aH:
-            aH = CN(nF='aH',root=frame); aH.H = tail; aH.wTT = copy(eTT); aH.wTT_ = copy(frame.wTT_)  # stamps: chain endpoint, old per-oF set
+            aH = CN(nF='aH',root=frame); aH.H = tail; aH.wTT = copy(eTT)  # stamps: chain endpoint, old per-oF set
             if rD * wBac > ave * (cBac+1):  # 1: nesting cost
                 aH.dTT,aH.c,aH.r = sum_vt(tail); aH.m,aH.d = vt_(aH.dTT)
             frame.H = H[:i] + [aH]  # same_filter_levs
-            for j, tF in enumerate(oF_):
-                if tF: Fw_[j] = tF.fw/tF.c; FTT_[j] = frame.wTT_[j] = tF.wTT
-            ave/=rV; avd/=rV; Fw_ = np.array(Fw_)/rV; FTT_ = np.array(FTT_)/rV
```
kwcckw, line 935: i think we don't need FTT_, Fw_ and wTT_ now? They could be accessed from oF_ anyway.

frame_2D_alg/agg_recursion.py
boris-kz: Claude messed up, should be > ave

frame_2D_alg/agg_recursion.py
boris-kz: right

frame_2D_alg/meta_code.py
```diff
@@ -136,7 +136,7 @@ def inner(*a, **kw):
                 _CoF.call_ += [oF]
                 oF_[iF_[func.__name__]].caller_.add(_CoF)  # for comp_caller_
             _oF = CoF._cur.set(oF)
-            out = func(*a, **kw)
+            out = func(*a, **kw)  # comp_slice and slice_edge is not under traced yet
```
boris-kz, line 139: you just added the tracing?

kwcckw: Yes, sorry, this comment is no longer relevant.

frame_2D_alg/agg_recursion.py
```diff
@@ -892,20 +896,20 @@ def expand_lev(_iy,_ix, elev, T):  # seed tile is pixels in 1st lev, or Fg in hi
...
-    global ave,avd, Fw_,FTT_  # update from ffeedback:
+    global ave,avd  # update from ffeedback:
     while elev < max_elev:
         tile_,C,R = expand_lev(iY,iX, elev, tile)
         if tile_: # sparse,2D
             Fr = sum2F(tile_)  # higher-scope tile( oH( aH
-            if cross_comp(Fr, rr=elev)[0]:  # spec->tN_,tC_,tL_, proj comb N_'L_?
+            if cross_comp(Fr.Nt, rr=elev)[0]:  # spec->tN_,tC_,tL_, proj comb N_'L_?
                 elev += 1
                 if rV * wBac *((elev-1)*wL) > ave*cBac:  # flat elev
```
boris-kz, line 911: ((elev-1)*wL) is not relevant

frame_2D_alg/agg_recursion.py
```diff
@@ -914,10 +918,11 @@ def expand_lev(_iy,_ix, elev, T):  # seed tile is pixels in 1st lev, or Fg in hi
 
 def ffeedback(frame):  # adjust filters via cross-level wTT ratios; fork: reform oF_ when aH chain converged
 
-    global ave,avd, Fw_,FTT_
+    global ave,avd
     def drift(lev_, _wTT):  # ratio chain over wTT stamps
         rTT,dTT,k = np.zeros((2,9)),np.zeros((2,9)),0
         for lev in lev_:  # lev.wTT = CoF.get().wTT at creation?
+            # r_ is always 1 because all levs have the same wTT
```
boris-kz, line 925: sorry, this is not needed

kwcckw: But we still need to get the rTT from the ratio of levels here, that should be r? and it should be dTT instead of wTT?

boris-kz: yes:

if (tail := H[i:]) and (i or len(tail) > 1):  # unpacked levs

    _lev = tail[0]; rTT = np.zeros((2,9))
    for lev in tail[1:]: rTT += np.divide(_lev.dTT, eps_(lev.dTT))
    rV,rD = vt_(rTT)

kwcckw: For the prior eTT, that should be computed with cent_TT now , using the rTT above, so that we can assign it to oH.wTT and aH.wTT?

boris-kz: what do you want these eTT, oH.wTT and aH.wTT for?

    if len(tail:= H[i:]) > 1:  # unpacked levs
        _lev = tail[0]; rTT = np.zeros((2,9)); k = 1
        for lev in tail[1:]: rTT += np.divide(_lev.dTT, eps_(lev.dTT)); k += 1
        rV, rD = vt_(rTT /k)

boris-kz: if len(tail := H[i:]) > 1:  # unpacked levs
    rV, rD = vt_(sum(np.divide(tail[0].dTT, eps_(lev.dTT)) for lev in tail[1:]))

rV is accumulated value drift?

boris-kz: No — three ways off.

rV is the match-channel of vt_, rD is the diff-channel. So if anything is "drift" it's rD (the d-share); rV is the match/stability share, which is why you use it for the rescale and rD for the gate.

Not accumulated: vt_ normalizes by m_+|d_|, so the k-summation washes out — rTT scales with k, vt_ cancels it, and the result is count-free. Summing 2 levs or 20 gives the same composition.

Not drift: it's a ratio aggregate, so at zero change every ratio is 1, rTT = k·ones, and vt_ returns a nonzero constant (≈0.5@w split), not 0. It's a level, not a deviation.

For accumulated value drift specifically you'd need the deviation form (|ratio−1|, baseline 0) and to skip vt_'s normalization so magnitude and count survive — which is the v0/extensivity point from earlier. As written, rD is "diff-share of the mean per-step ratio," and the gate compares that constant-baseline level to ave.

kwcckw: > what do you want these eTT, oH.wTT and aH.wTT for?
> 
> ```
> if len(tail:= H[i:]) > 1:  # unpacked levs
>     _lev = tail[0]; rTT = np.zeros((2,9)); k = 1
>     for lev in tail[1:]: rTT += np.divide(_lev.dTT, eps_(lev.dTT)); k += 1
>     rV, rD = vt_(rTT /k)
> ```

Probably not, we only need wTT in root, i guess we can skip them.



> if len(tail := H[i:]) > 1: # unpacked levs rV, rD = vt_(sum(np.divide(tail[0].dTT, eps_(lev.dTT)) for lev in tail[1:]))
> 
> rV is accumulated value drift?

Yes, i think it should accumulate all the drifted values, but why it's always tail[0]? It should be between levs in tails?

boris-kz: Use diffs instead of ratios?

Yes — diffs are 0 at no change and grow with both size and count, so they give you accumulated drift; ratios can't (baseline 1, and vt_ then strips magnitude and count).

But don't pass diffs through vt_ — vt_ normalizes by m_+|d_|, so it re-strips the magnitude and you're back to a count-free composition. Keep the magnitude: sum the diffs against the anchor and take a wTT-weighted norm.

```python
if len(tail := H[i:]) > 1:  # unpacked levs
    dTT = sum(lev.dTT - tail[0].dTT for lev in tail[1:])  # accumulated drift, 0 at no change
    rD = np.sum(dTT * wTT) / np.sum(wTT)
```

One choice to make: signed sum (above) is net displacement — a regime that wobbles and returns reads ~0, so it fires only on real movement. Abs sum (`np.abs(lev.dTT - tail[0].dTT)`) is total path length — accumulates wobble and noise. Net is the one that means "the regime's statistics actually moved." And if you still want the value/drift split, take it from the signed dTT directly (dTT[0] vs dTT[1] weighted), not via vt_.

kwcckw: I think we need to abs drifted vales between each levels so that they won't cancel out each other, which we need abs the sum in this line `dTT = sum(lev.dTT - tail[0].dTT for lev in tail[1:]) `?

boris-kz: Why shouldn't they cancel?

kwcckw: Okay, they should cancel out each other since we need the net displacement here, which measures the final stability.

boris-kz: def pack_seg(frame, nF, w, c):  # drift-gated regime termination + shell, reused for aH and oH

    H = frame.H
    nPacked = ('aH','oH') if nF=='aH' else ('oH',)
    i = next((j+1 for j in reversed(range(len(H))) if H[j].nF in nPacked), 0)  # packed levs
    tail = H[i:] if nF=='aH' else [l for l in H[i:] if l.nF=='aH']  # levs or aH_
    if len(tail) < 2: return None
    D = np.sum(np.abs(sum(tail[0].dTT - t.dTT for t in tail[1:])) * wTT)  # drift
    if (vD := D*w - ave*c) > 0:  # regime stale:
        seg = CN(nF=nF, H=tail, root=frame); seg.dTT = tail[-1].dTT  # regime-end stamp for next level
        if vD > ave: seg.dTT,seg.c,seg.r = sum_vt(tail); seg.m,seg.d = vt_(seg.dTT)  # deep summary, +1 nesting cost
        frame.H = H[:i] + [seg]  # append
        return seg
    
def ffeedback(frame):  # recompute filters from regime drift; fork: reform oF_ on cross-regime drift

    global ave,avd
    if aH := pack_seg(frame,'aH',wBac, cBac):
        ave,avd = vt_(aH.dTT)  # recompute filters
        if pack_seg(frame,'oH', wBac, cBac**2): 

kwcckw: > seg.dTT = tail[-1].dTT
> if vD > ave:  seg.dTT,seg.c,seg.r = sum_vt(tail); seg.m,seg.d = vt_(seg.dTT)

I think the sum_vt could be default here? Since it must have 2 tails level anyway.

boris-kz: It's more expensive, I want this incrementally detailed summary as a general rule

kwcckw: I see, and in this section:

   seg = CN(nF=nF, H=tail, root=frame); seg.dTT = tail[-1].dTT  # regime-end stamp for next level

For shallow segment, we still need their m, d, c and r from tail[-1]?

boris-kz: what do you want them for?

frame_2D_alg/agg_recursion.py
boris-kz: The next direction in agg_recursion is to flesh-out projection mechanisms?

Read the docstring, the proj_* functions, and README parts 2–3. Here's the state and where I'd push.

**What exists:** strictly first-order spatial projection. `proj_N`/`proj_TT` extrapolate a node's link-derived TT to a displaced coordinate with distance decay and one int×ext angle term. Two consumers: `proj_V` (induction gate inside `comp_N_` — prevalues a pair before comparing, and stores `rTT` = relative prediction error), and `proj_focus` (the coordinate filter that picks the next window in `frame_H`). `comp_prj_dH` (projected-vs-actual derH → surprise) is commented out.

**Key gap:** your docstring names projection *orders* (0 summed match → 1 spatial/temporal → 2 projected-vs-actual → recurse). Only order-1-spatial is built. `rTT` is captured but currently only feeds `oF`/code-weight fitting via `ttPrj` — there's no order-2 over *data*. Temporal projection doesn't exist at all.

**Proposal 1 — close the order-2 data loop (lowest risk, testable now).**
Claim: prediction error is structurally a derivative — a miss between projected and actual TT — so it needs no new operations, only a new input fork through existing comp/cluster. You already compute `Link.rTT = abs(pTT - Link.dTT)/dTT`. Make it a comparand: cluster nodes/links by `rTT` (surprise clusters) the same way you cluster by `dTT`, gated by "sufficient newly formed clusters" as the docstring says. Two distinct uses fall out: (a) per-param `rTT` adjusts resolution filters (high error on a param ⇒ model systematically over/under-predicts it ⇒ that's where to spend bits), which is the actual fitness signal "testing accuracy of prior cross_comp"; (b) `rTT` clusters *are* order-2 patterns, recursively projectable. This is the cleanest next step because the data is already on the link and it reuses `comp_N`/`cluster_N` wholesale. Falsifiable: if order-2 needs operations beyond a new fork, the "errors are derivatives" claim is wrong.

**Proposal 2 — co-projection + cancellation (the imagination primitive, richer but ambiguous).**
`proj_focus` currently accumulates scalar V from one tile additively into `PV__`. README's imagination mechanism is multi-source: accumulate projected *TT* per coordinate keyed by source, and where ≥2 sources overlap, combine by `max_fb − alt_fb·match_rate` (match_rate = ave by default, refined by `base_comp` of the co-projected TTs = "cross-comp at the intersection"). The substrate is already there (`PV__` is per-coordinate). But the sign/intent of the cancellation rule is underdetermined in the text — see question below — so I'd nail that before coding.

**Proposal 3 — temporal projection (blocked on input).**
Architecturally absent: everything is yx displacement. Same `proj_N` works if A is a temporal offset and decay uses temporal span, plus the README delay-correction on averages (`ave += dAve·(delay/span)/2`). But there's no frame sequence to test against on a still image, so I'd defer until 1–2 land.

**Two code flags that gate any projection work** (semantics, not nits):

- `proj_TT`: diff is scaled by `cos_d + dec`. At alignment≈1, dec≈1 that's ~2× — a short aligned projection doubles the difference. Your comment says "alignment * decay"; `cos_d * dec` gives ~1 near-source and flips sign on back-projection, which is the physical behavior. Add looks like a typo for multiply.
- Decay base is inconsistent: `proj_V`/`proj_N` use `decay**(dist/span)`, `proj_TT` uses `ave**(1+dist·dec/span)`. Either spatial decay is its own filter (`decay`) everywhere, or you genuinely mean "predictive value decays toward ave-level predictability" — but then say so and apply it consistently. Right now projected magnitudes from the two paths aren't comparable.

**Questions:**
1. Cancellation sign: `filter = max_fb − alt_fb·match_rate` makes the filter *vanish when co-projections agree* (match_rate→1) and *stay strong when they disagree*. If high filter = skip, that attends to agreement and skips disagreement — opposite of "novelty / edge of predictable." Is the filter a skip-value or a novelty-value, and which direction do you actually want attention to peak?
2. Order-2 consumer priority: filter adjustment (a) or recursive order-2 clustering (b) first? They share the `rTT` fork but diverge after.
3. Match (row 0) is never angle-modulated anywhere in projection, only diff is. Intended (isotropic match) or should oriented features' projected match depend on viewing along vs across?

Want me to draft the Proposal-1 fork as a minimal diff?

frame_2D_alg/agg_recursion.py
kwcckw: We may still pack this seg in H, so we need to at least include their c for the next iteration?

boris-kz: it would have to be summed, last c is not representative

frame_2D_alg/agg_recursion.py
```diff
@@ -892,20 +896,20 @@ def expand_lev(_iy,_ix, elev, T):  # seed tile is pixels in 1st lev, or Fg in hi
...
     while elev < max_elev:
         tile_,C,R = expand_lev(iY,iX, elev, tile)
         if tile_: # sparse,2D
             Fr = sum2F(tile_)  # higher-scope tile( oH( aH
-            if cross_comp(Fr, rr=elev)[0]:  # spec->tN_,tC_,tL_, proj comb N_'L_?
+            if cross_comp(Fr.Nt, rr=elev)[0]:  # spec->tN_,tC_,tL_, proj comb N_'L_?
                 elev += 1
                 if rV * wBac *((elev-1)*wL) > ave*cBac:  # flat elev
-                    rV, FTT_ = ffeedback(Fr, rV, elev)  # terminate old, form new oH: same_oF_levs ( aH: same_filter_levs
+                    rV = ffeedback(Fr)  # terminate old, form new oH: same_oF_levs ( aH: same_filter_levs
```
boris-kz, line 912: do we still need this rV?

frame_2D_alg/agg_recursion.py
kwcckw: > Key gap: your docstring names projection orders (0 summed match → 1 spatial/temporal → 2 projected-vs-actual → recurse). Only order-1-spatial is built. 

I guess we can include the temporal approach when it involves multiple frames? The spatial method should be sufficient for the existing single image eval.

> proj_TT: diff is scaled by cos_d + dec. At alignment≈1, dec≈1 that's ~2× — a short aligned projection doubles the difference. Your comment says "alignment * decay"; cos_d * dec gives ~1 near-source and flips sign on back-projection, which is the physical behavior. Add looks like a typo for multiply.

I think the summation is intentional and it shouldn't be a typo, but we need to get their average `(cos_d+dec)/2` instead?

frame_2D_alg/agg_recursion.py
kwcckw: Yes, because it's used in the eval for ffeedback above?

    if rV * wBac *((elev-1)*wL) > ave*cBac:

boris-kz: we need to compute it in expand_lev? right now it dormant

kwcckw: `rV` can be computed with multiple elevs only, but expand_lev is different tiles of a same elev in T_?

frame_2D_alg/agg_recursion.py
boris-kz: I think it's the same as TT[1] * cos_d * dec, just need to unpack it for clarity

frame_2D_alg/agg_recursion.py
boris-kz: seed tile is lower elevation

kwcckw: Okay, but then the rV in the eval above should be just the ratio between tiles?

frame_2D_alg/agg_recursion.py
kwcckw: If we are using sum, it's actually TT[1] *cos_d + TT[1] *Dec, which is different?

PR #169, opened 2026-06-08
6/8/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -136,7 +136,7 @@ def inner(*a, **kw):
                 _CoF.call_ += [oF]
                 oF_[iF_[func.__name__]].caller_.add(_CoF)  # for comp_caller_
             _oF = CoF._cur.set(oF)
-            out = func(*a, **kw)
+            out = func(*a, **kw)  # comp_slice and slice_edge is not under traced yet
```
kwcckw, line 139: Sorry, this is not relevant, it's actually from the prior commit, and it should be fixed when i rebase again in the next update.

frame_2D_alg/agg_recursion.py
```diff
@@ -897,12 +897,16 @@ def expand_lev(_iy,_ix, elev, T):  # seed tile is pixels in 1st lev, or Fg in hi
         if T_:
             TT,C,R = sum_vt(T_, wTT=ttFrm); R+= elev
             if sum(vt_(TT,ttFrm))*(C*wFrm) > (ave+avd)*(R+cFrm):   # cancel weak frame, null T_,C,R
-                return T_,C,R
-        return [], 0, 0
+                if T:
+                    rTT = np.zeros((2,9))
+                    for t in T_: rTT += np.divide(T.dTT, t.dTT)
+                    rV = vt_(rTT /len(T_))[0]  # m of the ratio
```
kwcckw, line 903: > seed tile is lower elevation

So it should be something like this? Where rV is summed from the ratio between the seed tile (T from prior elev ) and the current elev's T_?

frame_2D_alg/agg_recursion.py
```diff
@@ -933,7 +937,7 @@ def pack_seg(frame, nF, w, c):  # drift-gated regime termination for aH and oH
     if len(tail) > 1:
         D = np.sum(np.abs(sum(tail[0].dTT-t.dTT for t in tail[1:])) * wTT)  # drift
```
kwcckw, line 938: So in this case, tail[0] is served as the base lev, and the drifting D is measured from all higher levels against this base lev only, rather than between levs?

frame_2D_alg/agg_recursion.py
boris-kz: elev, _V = 0,eps
    tile, Fr = [],[]
    global ave, avd  # update from ffeedback:
    while elev < max_elev:
        tile_,C,R = expand_lev(iY,iX, elev, tile)  # project from seed tile
        if tile_: # sparse,2D
            if cross_comp(Fr.Nt, rr=elev)[0]:  # spec-> tN_,tC_,tL_, proj comb N_'L_?
                V = Fr.m + Fr.d
                if elev:
                    rV = max(V/_V,_V/V)
                    if (V*rV) * wBac > ave * cBac: 
                        ffeedback(Fr)  # terminate old, form new oH(aH
                        _V = V  # else keep _V 
                elev += 1  # flat
                tile = Fr  # next-extension seed
?

frame_2D_alg/agg_recursion.py
```diff
@@ -872,7 +872,7 @@ def base_tile(y,x,T=0):  # pixels at elev=0, lower frame_H above that
             T.yx = np.array([y+Ly//2, x+Lx//2]); T.box = np.array([y,x, min(y+Ly,Y),min(x+Lx,X)]); T.span = np.hypot(Ly,Lx) / 2
         return T
 
-    def expand_lev(_iy,_ix, elev, T):  # seed tile is pixels in 1st lev, or Fg in higher levs
+    def expand_lev(_iy,_ix, elev, T, rV):  # seed tile is pixels in 1st lev, or Fg in higher levs
```
boris-kz, line 875: so rV is not needed here?

frame_2D_alg/agg_recursion.py
```diff
@@ -933,7 +937,7 @@ def pack_seg(frame, nF, w, c):  # drift-gated regime termination for aH and oH
     if len(tail) > 1:
         D = np.sum(np.abs(sum(tail[0].dTT-t.dTT for t in tail[1:])) * wTT)  # drift
         if (vD := D*w - ave*c) > 0:  # regime stale
-            seg = CN(nF=nF, H=tail, root=frame); seg.dTT = tail[-1].dTT; seg.c = sum(l.c for l in tail)  # default regime summary
+            seg = CN(nF=nF, H=tail, root=frame); seg.dTT = tail[-1].dTT; seg.c = sum([l.c for l in tail])  # default regime summary
```
boris-kz, line 940: you don't need those brackets

frame_2D_alg/agg_recursion.py
boris-kz: it should really be prior seg dTT, vs the whole tail dTTs

frame_2D_alg/agg_recursion.py
```diff
@@ -861,7 +861,7 @@ def comp_prj_dH(_N, N, ddH, rn, link, angl, span, dec):
 '''
 def frame_H(image, iY,iX, Ly,Lx, Y,X, rV, max_elev=4, fH=0):  # fH=0: tiles, 1:same_filter_levs, 2:same_oF_levs
```
boris-kz, line 862: fH is obsolete?

frame_2D_alg/agg_recursion.py
boris-kz: The loop *shape* is fine — expand, summarize, gate on clustering, conditionally feed back, reseed. But three things make it not do what the comments claim, in descending severity.

**1. `V` is identically 0, so feedback never fires (line 911–913).** `sum2F(tile_)` at 908 is called with empty `m_,d_`, so line 522 sets `Fr.m=Fr.d=0`. Then `V=0`; `upV = 0*max(0/_V, _V/0)` is `nan`; `nan*wBac > ave*cBac` is False every time. The only populated value on `Fr` is `dTT` (line 514). So this has to be `m,d = vt_(Fr.dTT); V = m+d`. As written the entire `if elev:` block is dead.

**2. At `fH=0` (your `__main__` call) elev>0 is degenerate.** `base_tile(iy,ix,T)` with `fH=0` returns the seed `T` unchanged and only rewrites its `box/yx/span` (871). So `expand_lev` scatters one *aliased* `Fr` across the grid; `T_=[Fr,Fr,…]`, `sum2F`/`cross_comp` then compare it to itself → no clusters → `[0]` is empty → break at elev=1. So `max_elev=4` effectively runs only elev 0. Either call with `fH≥1`, or at `fH=0` re-derive fresh pixel tiles at the shifted `iy,ix` (drop the `T` arg at 893, and make 881 derive per-cell).

**3. `tile=Fr` reseed is inert.** `base_tile` ignores its `T` argument in *both* branches — `fH` recurses and discards it, non-`fH` uses it only to *skip* re-derivation. So "project from seed tile" buys nothing: the previous level's `Fr` never conditions what the next level derives. If it's meant to bias the search, that has to enter `proj_N`/`expand_lev`'s seed, not `base_tile`.

**Minor / conceptual:** regime change is gated twice — `upV` here (912), then again inside `ffeedback` via `pack_seg`'s drift test (938). This contradicts your earlier "aH-wrap is unconditional per call." Pick one trigger: keep `upV` (once fixed) and make `pack_seg` wrap unconditionally, or drop `upV` and let `ffeedback` decide. Two thresholds on different metrics will disagree (upV fires, drift doesn't → `_V` advances but `ave/avd` don't). `C,R` from `expand_lev` (906) are unused; `global ave,avd` at 904 is dead in this body.

if elev:

                m,d = vt_(Fr.dTT); V = m+d; upV = V * max(V/_V,_V/V)  # regime drift since last fb
                if upV* wBac > ave* cBac:
                    ffeedback(Fr); _V = V   # terminate old, form new oH(aH, else keep _V

kwcckw: > ```
> elev, _V = 0,eps
> tile, Fr = [],[]
> global ave, avd  # update from ffeedback:
> while elev < max_elev:
>     tile_,C,R = expand_lev(iY,iX, elev, tile)  # project from seed tile
>     if tile_: # sparse,2D
>         Fr = sum2F(tile_)  # higher-scope tile( oH( aH
>         if cross_comp(Fr.Nt, rr=elev)[0]:  # spec-> tN_,tC_,tL_, proj comb N_'L_?
>             V = Fr.m + Fr.d
>             if elev:
>                 rV = max(V/_V,_V/V)
>                 if (V*rV) * wBac > ave * cBac: 
>                     ffeedback(Fr)  # terminate old, form new oH(aH
>                     _V = V  # else keep _V 
>             elev += 1  # flat
>             tile = Fr  # next-extension seed
> ```
> 
> ?

I see, and that `_V = V` update should be under every while loop right?

 Also i'm not sure if we should init `_V` with eps？ That's causing a large initial rV value, which could be better with `_V` = 1?

frame_2D_alg/agg_recursion.py
kwcckw: > it should really be prior seg dTT, vs the whole tail dTTs

So the prior seg is from the last ffeedback?

frame_2D_alg/agg_recursion.py
kwcckw: Okay right, thanks.

frame_2D_alg/agg_recursion.py
kwcckw: We probably need another var in base_tile to differentiate between these 2 sections? Maybe parsing the elev to replace fH?

       if fH:
            T = frame_H(image, y,x, Ly,Lx, Y,X, rV, max_elev, fH-1)
        else:  # flat H
            if not T:  # pixels at elev=0
                T = frame_blobs_root(comp_pixel(image[y:y + Ly, x:x + Lx]), rV)
                T = vect_edge(T, rV)  # form, trace PP_:

frame_2D_alg/agg_recursion.py
kwcckw: Yes, since rV can be computed from Fr of tiles, so we not really need it here now.

frame_2D_alg/agg_recursion.py
boris-kz: `_V = V is per ffeedback. _V=1 is wrong`, something else needs to change

frame_2D_alg/agg_recursion.py
boris-kz: yes

frame_2D_alg/agg_recursion.py
boris-kz:
```
if elev:
    V = sum(vt_(Fr.dTT))  # regime drift per fb value
    if abs((_V-V)) * wBac > ave * cBac:
        ffeedback(Fr); _V = V   # term,form oH(aH, | keep _V

```

kwcckw: Okay, but i think we can assign `_V = V` when elev == 0? Since ffeedback runs only when elev>0 and we will get a better `_V` instead of using eps?

frame_2D_alg/agg_recursion.py
kwcckw: We can retrieve last seg from tile.H[-1]? Since we assign last Fr as tile:

    tile = Fr  # next-extension seed

frame_2D_alg/agg_recursion.py
boris-kz: init _V = 0

frame_2D_alg/agg_recursion.py
boris-kz: elev = _V = 0
    aTT = oTT = np.zeros((2,9))  # regime refs, persist across levs
    tile, Fr = [],[]
    global ave, avd  # update from ffeedback:
    while elev < max_elev:
        tile_,C,R = expand_lev(iY,iX, elev, tile)  # project from seed tile
        if tile_: # sparse,2D
            if cross_comp(Fr.Nt, rr=elev)[0]:  # spec-> tN_,tC_,tL_, proj comb N_'L_?
                if elev:
                    V = sum(vt_(Fr.dTT))  # regime drift per fb value
                    if abs((_V-V)) * wBac > ave * cBac:
                        aTT,oTT = ffeedback(Fr, aTT,oTT); _V = V  # term,form oH(aH, else keep _V
                elev += 1; tile = Fr  # next-extension seed
            else: break
        else: break
    if Fr: oF = CoF.get(); oF.N_=[Fr]; oF.dTT=Fr.dTT; oF.c+=Fr.c; oF.r+=Fr.r

def ffeedback(frame, aTT,oTT):  # recompute filters from regime drift; fork: reform oF_ on cross-regime drift

    global ave,avd
        ave,avd = vt_(aH.dTT)  # filters *= ave
        aTT = aH.dTT
            oTT = oH.dTT
    oF = CoF.get(); oF.N_=frame.H; oF.dTT=frame.dTT; oF.c=frame.c; oF.r=frame.r
    return aTT,oTT

def pack_seg(frame, nF, w, c, _dTT):  # drift-gated regime termination for aH and oH

    H = frame.H
    n = ('aH','oH') if nF=='aH' else ('oH',)
    i = next((j+1 for j in reversed(range(len(H))) if H[j].nF in n), 0)  # packed levs
    tail = H[i:] if nF=='aH' else [l for l in H[i:] if l.nF=='aH']  # levs or aH_
    if len(tail) > 1:
        D = sum( np.abs(t.dTT-_dTT) * wTT for t in tail)  # drift
        if (vD := D*w - ave*c) > 0:  # regime stale
            seg = CN(nF=nF, H=tail, root=frame); seg.dTT = tail[-1].dTT; seg.c = sum(l.c for l in tail)  # default regime summary
            if vD > ave: seg.dTT,seg.c,seg.r = sum_vt(tail); seg.m,seg.d = vt_(seg.dTT)  # deep summary
            frame.H = H[:i]+[seg]  # append
            return seg

kwcckw: Thanks, this is actually much better now. Also since the aTT and oTT are init 0s, the first segment packing of aH and oH mostly likely will be true.

boris-kz: actually ffeedback gating is redundant to pack_seg gating, so:

    elev = 0
    aTT = oTT = np.zeros((2,9))  # regime refs, persist across levs
    tile, Fr = [],[]
    global ave, avd  # update from ffeedback:
    while elev < max_elev:
        tile_,C,R = expand_lev(iY,iX, elev, tile)  # project from seed tile
        if tile_: # sparse,2D
            if cross_comp(Fr.Nt, rr=elev)[0]:  # spec-> tN_,tC_,tL_, proj comb N_'L_?
                if elev:
                    aTT,oTT = ffeedback(Fr, aTT,oTT)  # term,form oH(aH, else keep _V
                elev += 1; tile = Fr  # next-extension seed
            else: break
        else: break
    if Fr: oF = CoF.get(); oF.N_=[Fr]; oF.dTT=Fr.dTT; oF.c+=Fr.c; oF.r+=Fr.r

kwcckw: Not sure, in pack_seg, it's eval using drifted value, while for V, it's using both d and m?

    abs((_V-V)) * wBac > ave * cBac:
   
 But i guess we only need one of them.

boris-kz: it's the same thing, just computed more precisely

frame_2D_alg/agg_recursion.py
boris-kz: right,

    def base_tile(y,x, elev):  # pixels at elev=0, lower frame_H above
        if elev:
            T = frame_H(image, y,x, Ly,Lx, Y,X, rV, max_elev=elev)  # smaller-scope unit, keeps own box
        else:
            T = vect_edge(frame_blobs_root(comp_pixel(image[y:y+Ly, x:x+Lx]), rV), rV)
            if T: T.yx=np.array([y+Ly//2,x+Lx//2]); T.box=np.array([y,x,min(y+Ly,Y),min(x+Lx,X)]); T.span=np.hypot(Ly,Lx)/2
        return T

kwcckw:
On the projections discussion earlier, ideally we should have projection to identify and eval the comp operation, to check if it's needed or not, but that's no included in comp_C and trace_edge?

In term of code, we don't need any projection for the oF comp since that's only valid for data?

boris-kz:
> On the projections discussion earlier, ideally we should have projection to identify and eval the comp operation, to check if it's needed or not, but that's no included in comp_C and trace_edge?
> 
> In term of code, we don't need any projection for the oF comp since that's only valid for data?

trace_edge is very basic, and centroid clustering, oF comp are not directional

kwcckw:
> > On the projections discussion earlier, ideally we should have projection to identify and eval the comp operation, to check if it's needed or not, but that's no included in comp_C and trace_edge?
> > In term of code, we don't need any projection for the oF comp since that's only valid for data?
> 
> trace_edge is very basic, and centroid clustering, oF comp are not directional

I see, and right now projection is used to eval for comp, but we could have use the difference between projected and actual comp value for feedback purpose too?

boris-kz:
> > > On the projections discussion earlier, ideally we should have projection to identify and eval the comp operation, to check if it's needed or not, but that's no included in comp_C and trace_edge?
> > > In term of code, we don't need any projection for the oF comp since that's only valid for data?
> > 
> > 
> > trace_edge is very basic, and centroid clustering, oF comp are not directional
> 
> I see, and right now projection is used to eval for comp, but we could have use the difference between projected and actual comp value for feedback purpose too?

feedback is already projected, the filters apply to the next input tile, selected by projection?
BTW, these global comments are awkward as hell, comment inline instead

frame_2D_alg/agg_recursion.py
kwcckw, line 921: > feedback is already projected, the filters apply to the next input tile, selected by projection?

The existing ffeedback updates ave filters with aH.dTT:

    ave,avd = vt_(aH.dTT)

And aH.dTT is simply summed from levs:

    seg.dTT,seg.c,seg.r = sum_vt(tail)

So how is that consider a projected feedback?

boris-kz, line 859: Do not make global comments.

>So how is that consider a projected feedback?

expand_lev is projected:

    while elev < max_elev:
        tile_,C,R = expand_lev(iY,iX, elev, tile)  # project from seed tile
        if tile_: # sparse,2D
            if cross_comp(Fr.Nt, rr=elev)[0]:  # spec-> tN_,tC_,tL_, proj comb N_'L_?
                if elev:
                    aTT,oTT = ffeedback(Fr, aTT,oTT)  # term,form oH(aH

kwcckw: Actually i added the comment in line 921:

image

Not sure if it's due to the comment is in the non-edited code, and it becomes a global comment.

> expand_lev is projected:

I see, thanks. So when we have projected tiles:

    if cross_comp(Fr.Nt, rr=elev)[0]: 
    
Cross comparing those tiles is actually the verification by getting the real comparison values before move to the feedback?

boris-kz: Ok, probably because you quoted global comment

cross_comp is within frame where each tile was selectied by projection

ffeedback results will be applied to the next input tile only, and that tile will be selected by projection too.
but we may call ffeedback  for specific projected tile:

                    if PV__[y,x] > ave:
                        if elev:
                            if PV__[y,x] > ave*cBac**2: ffeedback(frame)
                            T = base_tile(iy,ix, elev)  # lower frame_H, up to current level
?

boris-kz: No — three problems, the first fatal:

**`frame` here is the numpy grid**, not a CN: `frame = np.full((Ly,Lx), None)` (line 877). `ffeedback(frame)` immediately hits `frame.H` → AttributeError. It also drops the `aTT,oTT` threading you just added.

**Granularity mismatch.** ffeedback recomputes filters from drift across `frame.H` — *levels*. Inside `expand_lev` you have intra-level spatial peers (`T_`), and no `H` is accumulating; `Fr = sum2F(tile_)` (with the H) is built only after `expand_lev` returns. So there's no cross-level drift to measure and nothing for `pack_seg` to wrap. ffeedback operates one scope up from where you're calling it.

**It worsens the global-filter leak.** `base_tile(iy,ix,elev)` recurses into `frame_H`, which runs its own elev loop and its own ffeedback on the module globals `ave,avd`. A per-tile ffeedback in the parent mutates those globals mid-expansion, so tiles within a *single* level form under shifting filters → intra-level incomparability — the exact thing `same_filter_levs` exists to prevent, now inside one level.

And `PV__[y,x] > ave*cBac**2` ties filter recompute to one projection step's strength, not accumulated regime drift — orthogonal signal.

Keep ffeedback at the elev boundary. Region-specific adaptation already happens: the recursive `base_tile` sub-frame does its own ffeedback internally. The real defect there is that `ave,avd` are global (siblings stomp each other), which is fixed by making filter state per-frame — not by moving the trigger down here.

kwcckw: > we may call ffeedback for specific projected tile

For specific projected tile feedback, that should be per T here?

But it's still updating the aH and oH levels, but it's per T here?

boris-kz: this is probbaly wrong, see above

kwcckw: > the recursive base_tile sub-frame does its own ffeedback internally

I see, but where do we have the direct or indirect nternal feedback within recursive base_tile?

boris-kz: it's not internal, the ffeedback is per expand_lev, so it's per all base_tile projections

PR #170, opened 2026-06-09
6/9/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -607,7 +608,8 @@ def add_Nt(G, Nt, merge=0):  # in sum2G
         add_H(G.H if merge else G.H[:-1], Nt.H, G)  # else top lev = Nt.N_
     N_ = Nt.N_
     if merge: G.N_ += N_
-    else:     G.H += [sum2F(N_+ (G.H.pop().N_ if G.H else []), G)]  # init or extend top level
+    # i don't think we need this else now. G.Nt.N_ is always = Nt.N_ here, we don't need to add top level into H
+    # else: G.H += [sum2F(N_+ (G.H.pop().N_ if G.H else []), G)]  # init or extend top level
```
kwcckw, line 612: When merge = 0 here, G.N_ is actually Nt.N_ here, so we shouldn't pack it as H? So this else section should be removed.

frame_2D_alg/agg_recursion.py
```diff
@@ -887,7 +889,7 @@ def expand_lev(_iy,_ix, elev, T):  # seed tile is pixels in 1st lev, or Fg in hi
                     if PV__[y,x] > ave:
                         iy = _iy+ (y-cy)*(T.box[2]-T.box[0]); ix = _ix+ (x-cx)*(T.box[3]-T.box[1])  # step by sub-frame footprint
                         if elev:
-                            T = base_tile(iy,ix, elev)  # lower frame_H, up to current level
+                            T = base_tile(iy,ix, elev)  # lower frame_H, up to current level (this may update filters and oF too? So that feedback should be for top layer only?)
```
kwcckw, line 892: This higher elev's base_tile may call frame_H, and that may update global filters of oFs within each higher elev? Should we fix the ffeedback to the top level frame_H only?

boris-kz: Max elev is artificial, ultimately it won't be needed, filter updates should be driven by the diffs

frame_2D_alg/agg_recursion.py
```diff
@@ -140,7 +140,7 @@ def proj_V(_N,N, dist, dy_dx, dec):  # _N x N induction
                 Link = comp_N(_N,N, lr,c,full=not tnF, A=dy_dx, span=dist, rL=root,L_=L_,N_=N_, acc=acc)
                 Link.rTT = np.abs(pTT-Link.dTT) / eps_(Link.dTT)  # relative prediction error to fit oF, direction-agnostic
             else:
-                pL = CN(typ=-1, N_=[_N,N], dTT=pTT,m=m,d=d,c=c,r=lr, angl=[dy_dx,1],span=dist)
+                pL = CN(typ=-1, dTT=pTT,m=m,d=d,c=c,r=lr, angl=[dy_dx,1],span=dist); pL.N_ = [_N,N]  # Nt.N_ needs separate assignment
```
boris-kz, line 143: this is still conflicting with prop_F_?

kwcckw: Yes, because we still can't have N_ as input argument since CN.N_ is actually Nt.N_.

frame_2D_alg/agg_recursion.py
kwcckw: So theoretically the difs in each elev are equally important?

boris-kz: Yes but they are relative to the last same filter span, so they tend to increase until next update

kwcckw: When they relative to the last filters, why the difference is expected to be increased? Suppose the goal is to minimize the difference?

boris-kz: it's only minimized per ffeedback, the diffs increase naturally till they exceed cBac

frame_2D_alg/agg_recursion.py
boris-kz: adding top level into H is sum2F now, but it's only if not Nt, we still need to pack Nt.H?

        if typ == 3:
            if not F.Nt:  # direct H, frame only?
                for N in N_: add_H(F.H, N.H, F, fN = 1 if (N.H and isinstance(N.H[0],CN)) else 0)  # concat lower levs (follow existing H's type)
                F.H += [sum2F([n for N in N_ for n in N.N_], F)]  # add top lev

boris-kz: if typ==3:  # frame?
            H = F.H or F.Nt; h = N.H or N.Nt.H
            for N in N_: add_H(H, h, F, fN = 1 if (H and isinstance(H[0],CN)) else 0)  # concat lower levs
            H += [sum2F([n for N in N_ for n in N.N_], F)]  # add top lev

kwcckw: Yes, so the H packing is sum2F is correct and we shouldn't change them.

boris-kz: I edited that above

kwcckw: > H = F.H or F.Nt; h = N.H or N.Nt.H

Actually F.Nt.H and F.H are the same? Because H is retrieved from Nt.H via prop_F_ now.

frame_2D_alg/agg_recursion.py
```diff
@@ -512,12 +512,13 @@ def sum2F(N_, root=None, m_=[],d_=[], merge=0, froot=0, nF=None):  # -> CF/CL/CC
...
                 if typ==3: box=copy(n.box)
     F = (cls_[typ])(dTT=TT, c=C, r=R, nF=nF)
+    if isinstance(F, CN): F.Nt.dTT = copy(TT); F.Nt.c = C; F.Nt.r = R  # update Nt's params
     if typ:
         F.N_ = n_; F.kern=kern; F.span=span; F.yx=yx
         if angl is not None: F.angl = [angl, np.sign(F.dTT[1] @ ttcP[1])]
         if typ == 3:
             if not F.Nt:  # direct H, frame only?
-                for N in N_: add_H(F.H, N.H, F)  # concat lower levs
+                for N in N_: add_H(F.H, N.H, F, fN = 1 if (N.H and isinstance(N.H[0],CN)) else 0)  # concat lower levs (follow existing H's type)
```
boris-kz, line 521: fN = isinstance(h[0],CN)

frame_2D_alg/agg_recursion.py
boris-kz: not if F is CF?

Read the PR and the touched code (`agg_recursion.py` 108–153, 490–530, 536–620, 862–947; `meta_code.py` 36–130). The five edits are mostly correct, but two threads have real problems and your own proposed unification has bugs. Going by thread:

**L143 — dead `N_` kwarg.** Fix is correct, and yes, this is `prop_F_` shadowing. `CN.N_` is a class property → `Nt.N_`, so by the time `CF.__init__` runs `if not hasattr(f,'N_')`, the getter already returns `Nt.N_` (`[]`), so the guard is True and `kw['N_']` is silently dropped. Post-hoc `pL.N_=[_N,N]` (property setter → `Nt.N_`) is the only thing that lands.

Two follow-ups:
- The *same* bug is unfixed at **L489**: `CN(typ=1, N_=[N,C], ...)` — that centroid link's `Nt.N_` is empty, not `[N,C]`. Latent only if nothing reads it, but it's the identical defect.
- The clean fix is in `CN.__init__`: pop `N_` from kw and assign via the property before `super().__init__`, so no site needs the post-hoc dance. Otherwise every `CN(N_=…)` is a trap.

**L515 — Nt param sync.** Correct *here*, but only because at `sum2F` construction the node has no `Lt/Bt` contribution, so `F.dTT == sum over Nt.N_`. Don't read it as a general invariant: after `comb_Ft`, `F.dTT` legitimately includes `Lt/Bt` (via the `T` accumulator at L595–597) and diverges from `F.Nt.dTT`. So "Nt mirrors F" is birth-only. Readers at L562/L581 want the pure-internal `Nt.dTT`, which stays consistent. OK to merge.

**L612 — removing the `add_Nt` else.** kwcckw is right that `G.N_ == Nt.N_` here (G is built `CN(Nt=Nt,…)`), so packing `sum2F(N_)` into H duplicates the top. But note: in the only caller (`comb_Ft`), `Nt` is a **CF**, so `isinstance(Nt,CN)` at L606 is False and L607 never runs. So today `add_Nt`'s *only* H mutation is that else. Remove it and `G.H` stays `= Nt.H` with top implicit as `G.N_` — self-consistent and matches the "top lev = Nt.N_" comment. Drop the `[:-1]` too; it's stripping a top that won't be there.

**Your unification snippet — don't merge it as written, two bugs:**
- `H = F.H or F.Nt` — `F.Nt` is a CF, not a list; `add_H`/`H += [...]` need a list. You meant `F.Nt.H`. And for a CN, `F.H` *is* `F.Nt.H` (property), so the `or` is a no-op.
- `h = N.H or N.Nt.H` is hoisted *outside* the loop; `N` is the loop variable, so `h` is either a NameError or a stale single value reused for every iteration. The original `for N in N_: add_H(F.H, N.H, F)` is per-N and correct.

But the deeper objection: **don't unify these at all.** They serve different consumers. The frame path (`sum2F` typ==3, no Nt) materializes the top into H at L521 *because `frame.H` is exactly what `pack_seg` drift-scans* (L929–932) — the top level is where drift is most informative, you can't drop it. The graph path keeps the top live because `cross_comp` reads `G.Nt.N_` live (L582/L906), never a summary fork in H. So the asymmetry is load-bearing: frame needs top-in-H, graph needs top-live. Forcing both under one block breaks one consumer. (Before changing the graph side, check `cluster_C` L662 `for lev in N.H` tolerates an implicit top — it copies all levels.)

**L924 — `oF.dTT += frame.dTT`.** This now contradicts its three siblings (L152, L493, L912), which all *assign* `dTT` and accumulate only `c/r`. Either all four accumulate `dTT` or none. Worse, `oF.N_=frame.H` is still overwritten while `dTT` accumulates, so after the elev loop `oF.dTT` is a sum over all levels but `oF.N_` is the last snapshot — value no longer matches membership. Decide whether call-trace `dTT` is cumulative (then fix N_ too, and the other three sites) or a snapshot (then revert this).

**Elev / ffeedback thread.** Three things, in increasing order of how much they bother me:

1. kwcckw's global-filter worry is real and *survives* killing max_elev: nested `base_tile → frame_H` share `global ave,avd` (L917). An inner recursion's diff-driven update mutates the filter the outer scale then reads to gate *its* drift. "Driven by diffs" doesn't answer *which scale's diffs own the global filter*. Filter scope ≠ data scope is unresolved, not artificial.

2. "diffs increase till they exceed cBac" is true but trivially so. `pack_seg`'s `D = Σ|t.dTT − _dTT|` over a growing tail is a non-negative cumulative sum — it can only grow, and it ratchets: a transient excursion permanently raises D even after the signal returns to anchor. So you're firing on *total historical deviation from anchor* (an MDL/compression criterion: "one anchor no longer compresses the regime"), not on *current* filter staleness (a control criterion). You're describing it as control. Pick the framing; if you want control, weight recent levels or measure the latest level vs anchor, not the cumulative sum.

3. The one I'd actually push on: drift is gated on **dTT** deviation, but your stated prediction error is **rTT** (L141, "relative prediction error to fit oF"). Feedback that re-fits filters should plausibly gate on accumulated rTT — prediction *failure* — not dTT drift, which is signal non-stationarity. As written, ffeedback re-fits filters whenever the signal moves *even if projection tracks it perfectly* (rTT≈0). Is the dTT/rTT split intentional — link-level rTT fits oF assignment, frame-level dTT-drift independently re-fits ave/avd — or should rTT propagate up to drive ffeedback? This is the question worth resolving before more ffeedback work.

And to settle kwcckw's "shouldn't we minimize the difference?": in your framework dTT is *signal* (it drives der+), not error. The thing minimized is rTT. Saying "filters minimize prediction error rTT; dTT accumulates" dissolves the confusion better than "minimized per ffeedback" — which is also why point 3 matters: the code currently gates on the thing that accumulates, not the thing that's minimized.

Want diffs for L489 + the `CN.__init__` N_ handling, or are these going in the thread as-is?

frame_2D_alg/agg_recursion.py
boris-kz, line 489: L = CN(typ=1, dTT=N.dTT,c=N.c,r=N.r,m=N.m,d=N.d, span=np.hypot(*(dy_dx:=C.yx-N.yx)), angl=[dy_dx,np.sign(N.dTT[1]@ttcN[1])]) 
                    L.N_ = [N,C]; C.L_ += [L]

boris-kz, line 344: removing merge arg:

                        if rt0 and rt1 and rt0 != rt1:
                            if rt1.H: add_H(rt0.H, rt1.H, rt0)
                            rt0.N_ += rt1.N_; add_Nt(rt0, rt1)

frame_2D_alg/agg_recursion.py
kwcckw: From ` h = N.H or N.Nt.H`, `h` could be empty? Unless we add another eval `if N.H` first.

frame_2D_alg/agg_recursion.py
kwcckw: Actually Fr is CN because we form sum2F with tiles, and tiles are CN, so from sum2F:

    typ = 2 if fC else N.typ
    
Fr class instance will be the same as tiles.

frame_2D_alg/agg_recursion.py
kwcckw: Thanks, i missed the N_ assignment here.

frame_2D_alg/agg_recursion.py
```diff
@@ -140,7 +140,7 @@ def proj_V(_N,N, dist, dy_dx, dec):  # _N x N induction
                 Link = comp_N(_N,N, lr,c,full=not tnF, A=dy_dx, span=dist, rL=root,L_=L_,N_=N_, acc=acc)
```
boris-kz, line 140: Trying Fable with max effort and default to adversarial, the damn thing switched to Opus when I asked for literature search:

https://claude.ai/share/7c4baa11-3f9b-43c7-b72a-ca7aebbc6454

kwcckw: >   The sharpest framing for any of them is the same — same objective (compositional structure from images by compression), no learned parameters, no probabilistic inference, no top-down fitting — because that triad is precisely what separates you from all four.

It's actually looks better in the code analysis since i can see it's already pointing out several bugs. Also on this "no learned parameters", the current work is actually learning mathematically but using filters, weights, costs, etc? Without learning, there shouldn't be any feedback for the improvement.

boris-kz: these days, training is synonimous with learning. they don't think clustering is learning anymore

frame_2D_alg/agg_recursion.py
boris-kz: ok,  fN=N.H and isinstance(N.H[0],CN)

frame_2D_alg/agg_recursion.py
boris-kz: so fN can be passed as arg?
also we need batch summation:

def add_Nt(G):  # in sum2G and trans_cluster

    N_ = G.N_; c_ = np.array([N.c for N in N_]); C = c_.sum()
    G.kern, G.yx = np.zeros(4),np.zeros(2); yx_ = []
    for N in N_:
        if hasattr(N,'m_'):
            if not hasattr(G,'m_'): G.root_,G.m_,G.d_ = [],[],[]
            G.C_ += N.root_; G.m_+= N.m_; G.d_+= N.d_  # Ct || Nt
        G.kern += N.kern*w; yx = N.yx*w; G.yx += yx; yx_ += [yx]
    G.span = (c_ @ np.hypot(*(np.array(yx_)-G.yx).T)) / C if len(N_)>1 else N_[0].span

frame_2D_alg/agg_recursion.py
kwcckw: I see, so i guess the learning is actually referring to the process like gradient descent or a separate training stage + feedback now.

frame_2D_alg/agg_recursion.py
kwcckw: > so fN can be passed as arg?

Yes, but it probably should be default to CF, since we only need fN = 1 when we use sum2F to sum tiles as level.

PR #171, opened 2026-06-10
6/10/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -187,10 +187,10 @@ def comp_C_(C_,_r, _C_=[], fall=1, fC=0):  # simplified for centroids, trans-N_s
             tc += c; tr += r*c
             comp_N(_C,C,_r, c ,A=dy_dx,span=dist,L_=L_,N_=N_,acc=acc)  # simplified for typ=3
     for N in list(set(N_)):
-        if N.rim: sum2F(N.rim, N.Rt)
+        if N.rim: sum2F(N.rim, N.Rt)  # Rt.m is never updated, but we need them in get_exemplar? N.Rt = sum2F(N.rim)?
```
kwcckw, line 190: So looks like the claude fable is significantly better than opus, but it took all the session limit with just one question.

So in this line, we need to reassign `N.Rt = sum2F(N.rim)`? Else we not really get any m, d in N.Rt when we parse it as root.

frame_2D_alg/agg_recursion.py
```diff
@@ -425,11 +425,14 @@ def cluster_C(Ft, E_,_r,_c):  # form centroids by clustering exemplar surround v
...
             r = _r+ R/T  # loop-local, not ave?
-            if M*(_C.m+ wC_*len(n_)+wC_*len(n_)) > Ave*(r+_C.r+ cC_*len(n_)):  # else: Up+= sum(_C._m_)+ sum([abs(d) for d in _C._d_])?
+            # duplicated wC_*len(n_)?
+            if M*(_C.m + wC_*len(n_)) > Ave*(r+_C.r+ cC_*len(n_)):  # else: Up+= sum(_C._m_)+ sum([abs(d) for d in _C._d_])?
                 for n in [_n for n in n_ for l in n.rim for _n in l.N_ if _n is not n]:
                     N__ += [n]  # +|-Ls
                 C = sum2F(n_, Ft,m_,d_)  # C.N_ = n_
                 C._N_ = list(set(N__)- set(n_))  # new frontier
+                for n, m, d in zip(n_, m_, d_):  # now n.m_, d_ and root_ are always empty, we need to assign those after sum2F?
+                    n.root_ += [C]; n.m_ += [m]; n.d_ += [d]
```
kwcckw, line 435: Right now n.m_, d_ and root_ are not filled, i think previously we had them in sum2C, so we need to add it back?

frame_2D_alg/agg_recursion.py
```diff
@@ -520,14 +523,16 @@ def sum2F(N_, root=None, m_=[],d_=[], merge=0, froot=0, nF=None):  # -> CF/CL/CC
         F.N_ = n_; F.kern=kern; F.span=span; F.yx=yx
         if angl is not None: F.angl = [angl, np.sign(F.dTT[1] @ ttcP[1])]
         if typ==3:  # frame?
-            for N in N_: add_H(F.H, N.H, F)  # concat lower levs
+            for N in N_:
+                if N.H: add_H(F.H, N.H, F, fN=isinstance(N.H[0], CN))  # concat lower levs
             F.H += [sum2F([n for N in N_ for n in N.N_], F)]  # top lev
+            # if typ == 3, we shouldn't have m_ and d_ here? This should be in if fC?
             F.m_,F.d_ = m_,d_; F.m, F.d = sum(m_),sum(d_)
```
kwcckw, line 530: When typ == 3 here. we shouldn't get m_ and d_? This should be valid wen fC is true?

frame_2D_alg/agg_recursion.py
```diff
@@ -859,7 +864,7 @@ def comp_prj_dH(_N, N, ddH, rn, link, angl, span, dec):
 def frame_H(image, iY,iX, Ly,Lx, Y,X, rV, max_elev=4):
 
     def base_tile(y,x, elev):  # lower tile if elev else pixels
-        if elev:
+        if elev:  # no direct cross_comp in base_tile now?
```
kwcckw, line 867: I can see we don't have direct cross_comp in base_tile now, and it's only nested in frame_H, when we cross_comp tiles?

        if tile_: # sparse,2D
            if cross_comp(Fr.Nt, rr=elev)[0]:  # spec-> tN_,tC_,tL_, proj comb N_'L_?

frame_2D_alg/agg_recursion.py
```diff
@@ -85,7 +85,7 @@ def cross_comp(root, rr):  # core function mediating recursive rng+ and der+ cro
 
     N_,G_ = root.N_,[]; fC = N_[0].typ==2  # root is Ft, converted below, rc=rdn+olp, comp N_|B_|C_:
     # comp_ s update root oF:
-    L_,TT,c,r,TTd,cd,rd = comp_C_(N_,rr,fC=1) if fC else comp_N_(combinations(N_,2),rr)
+    L_,TT,c,r,TTd,cd,rd = comp_C_(N_,rr,fC=1) if fC else comp_N_(combinations(N_,2),rr)  # TTd, cd and rd are redundant now?
```
boris-kz, line 88: redundant because boundary is computed from rims?

frame_2D_alg/agg_recursion.py
boris-kz: Probably. Session limit?

frame_2D_alg/agg_recursion.py
boris-kz: these would be nodes, not centroids, they are still typ 3

frame_2D_alg/agg_recursion.py
boris-kz: ok, frame_H should be default?

    def base_tile(y,x, elev):  # lower tile if elev else pixels
        if not elev:
            T = vect_edge(frame_blobs_root(comp_pixel(image[y:y+Ly, x:x+Lx]), rV), rV)
            if T: T.yx=np.array([y+Ly//2,x+Lx//2]); T.box=np.array([y,x,min(y+Ly,Y),min(x+Lx,X)]); T.span=np.hypot(Ly,Lx)/2
        T = frame_H(image, y, x, Ly, Lx, Y, X, rV, max_elev=elev)  # tile
        return T

frame_2D_alg/agg_recursion.py
```diff
@@ -622,7 +627,7 @@ def add_Lt(G, Lt,wTT):  # addition to Q2R
     L_ = Lt.N_
     if Lt.m * wN > ave*(cN+Lt.r):  # comp typ -1 pre-links
         L_,pL_ = [],[]; [L_.append(L) if L.typ==1 else pL_.append(L) for L in Lt.N_]
-        if pL_ and sum_vt(pL_,fm=1,wTT=wTT)[0]* ave*wN > ave*(cN * np.mean([L.r for L in pL_])):
+        if pL_ and sum_vt(pL_,fm=1,wTT=wTT)[0]* ave*wN > ave*(cN * np.mean([L.r for L in pL_])):  # * ave on both sides, which cancel out each other?
```
boris-kz, line 630: yes, 1s ave was a typo

frame_2D_alg/agg_recursion.py
boris-kz: n_+=[n]; m_+=[m]; d_+=[d]; c=n.c; T+=c; M+=m*c; D+=abs(d)*c; R+=n.r*c  # scale totals only?

those should be assigned N?

frame_2D_alg/agg_recursion.py
boris-kz: Conceptually yes — but not as pasted. In your snippet the unconditional T = frame_H(...) overwrites the elev=0 vect_edge result, and frame_H(max_elev=0) returns [] (the while elev < max_elev loop never runs, Fr stays []), so the pixel-level case is discarded entirely and the whole hierarchy produces nothing.

The clean version of “frame_H as default” is to move the base case inside frame_H and delete base_tile:

def frame_H(image, iY,iX, Ly,Lx, Y,X, rV, max_elev=4):
    if not max_elev:  # base: pixel-level tile
        T = vect_edge(frame_blobs_root(comp_pixel(image[iY:iY+Ly, iX:iX+Lx]), rV), rV)
        if T: T.yx=np.array([iY+Ly//2,iX+Lx//2]); T.box=np.array([iY,iX,min(iY+Ly,Y),min(iX+Lx,X)]); T.span=np.hypot(Ly,Lx)/2
        return T
    ...


Then fill_frame calls frame_H(image, iy,ix, ..., max_elev=elev) uniformly for both the not elev top-of-loop seed and the elev>0 re-seed. That gives the compositional definition you presumably want: frame_H(e) = frame of frame_H(e-1) tiles, frame_H(0) = vect_edge window. The tile/frame conflation is fine — it’s already inherent in tile = Fr seeding the next level.

Two things to check:

	1.	The fill_frame asymmetry must stay: at elev>0 the initial T is the already-computed lower Fr passed in, not a fresh frame_H call — recomputing it would be wasteful and non-identical under feedback-shifted filters. So you unify base_tile away, but not the if not elev control flow inside the loop.
	2.	Making frame_H the universal path deepens an existing issue: every elev>0 re-seed runs a full sub-hierarchy whose ffeedback mutates global ave/avd mid-parent-loop, so the parent’s gate (ave+avd)*(T.r+cFrm) drifts with each re-seed, order-dependently. Is that intended local adaptation, or should nested frame_H calls restore parent filters on return?

Minor: normalize the empty return — base case yields None, recursive yields []; both falsy so gates work, but pick one.

kwcckw: Not sure if the suggestion is correct, but instead of default frame_H, why we are not adding cross_comp? The default frame_H also cross comparing tiles, and we will never cross_compare any Gs.

frame_2D_alg/agg_recursion.py
kwcckw: Yes, i think this is from prior version, where we have another fork here. Now b-fork should be in sum2G now.

frame_2D_alg/agg_recursion.py
kwcckw: So the ave should be applied to the right side, and left side should be like this?

    if pL_ and sum_vt(pL_,fm=1,wTT=wTT)[0]*wN

frame_2D_alg/agg_recursion.py
kwcckw: Okay, right, thanks.

frame_2D_alg/agg_recursion.py
kwcckw: But those are unconfirmed Ns? We need to add it after sum2F of C, after the eval?

frame_2D_alg/agg_recursion.py
kwcckw: Yes, it used up the current session token with just one question, and need to wait for another 5 hours for reset again.

boris-kz: Is that through API? I never see the token usage

frame_2D_alg/agg_recursion.py
boris-kz: right

frame_2D_alg/agg_recursion.py
boris-kz: yes, it should be in base_tile?

def frame_H(image, iY,iX, Ly,Lx, Y,X, rV, max_elev=4):

    def base_tile(y,x):  # lower tile if elev else pixels
        
        if T := vect_edge( frame_blobs_root( comp_pixel(image[y:y+Ly, x:x+Lx]), rV), rV):
            T.yx = np.array([iY+Ly//2, iX+Lx//2])
            T.box = np.array([iY,iX, min(iY+Ly,Y), min(iX+Lx,X)])
            T.span = np.hypot(Ly,Lx)
            if cross_comp(Fr.Nt, rr=elev)[0]: return T
            
    def fill_frame(_iy,_ix, elev, T):  # seed tile is lower frame if elev else pixels

        cy,cx = (Ly-1)//2, (Lx-1)//2; y,x = cy,cx  # start=mean
        while True:
            if not elev:  # init pixel-level seed tile
                T = base_tile(_iy,_ix) 
            if T and sum(vt_(T.dTT, T.wTT*ttFrm)) * wFrm > (ave+avd)*(T.r+cFrm):  # complex gate, make incremental?
                frame[y,x]=T; T_+=[T]  # loop adds one tile to level
                if 0 < sum(vt_(pTT, T.wTT*ttFrm)) * wFrm < ave * cFrm:  # extend lev by combined proj T_
                    if PV__[y,x] > ave:
                        if elev:
                            T = base_tile(iy,ix)  # lower frame_H, up to current level
                    else: break
                else: break
            else: break

boris-kz: Central question first: `base_tile` no longer recurses — the elev>0 re-seed now gets a pixel-window vect_edge, not `frame_H(..., max_elev=elev)`. The comment "lower frame_H, up to current level" contradicts the code. If intentional, elevation collapses: every level fills frames of same-grain 64×64 tiles, and scope stops expanding — the step `(y-cy)*(T.box[2]-T.box[0])` uses the current T's box, which after the first re-seed is ≤ Ly, so only the inherited seed Fr contributes one big step and the rest of the level scans at level-0 footprint. Higher levels then differ only by `rr=elev` in comparison depth, covering the same image area repeatedly. If that's the redesign — comparison-depth hierarchy instead of composition hierarchy — what replaces spatial scope growth? Scaling Ly,Lx per elev? If it's not the redesign, the elev>0 re-seed must stay `frame_H(image, iy,ix, Ly,Lx, Y,X, rV, max_elev=elev)`.

Concrete breaks:

1. `cross_comp(Fr.Nt, rr=elev)` — wrong referent. `Fr` resolves through closure to frame_H-body's previous-level frame: `[]` on the first tile → AttributeError; later it would re-cross_comp the prior frame per candidate tile. You mean `T.Nt`. (`elev` also resolves by closure to the body's loop variable — works, but implicit.)

2. Even as `T.Nt`: vect_edge already runs comp_slice + trace_edge and gates its return on frame value. Adding a per-tile cross_comp gate means a tile is only accepted if agg-level clustering succeeds inside one window — much stricter, sparse frames — and it duplicates the outer `cross_comp(Fr.Nt, rr=elev)`. In the old design per-tile cross_comp existed only via the frame_H recursion at elev>0. Is this gate the cheap replacement for that recursion, or an addition on top of it?

3. elev=0 window still frozen, half-fixed: base_tile now takes coords, but the loop top calls `base_tile(_iy,_ix)` every iteration, and the projection's `iy,ix` are computed then discarded (`if elev:` guard). Frame fills with copies of the seed tile. Per your own comment "init pixel-level seed tile": seed once before the loop, re-seed at the bottom unconditionally:

```python
T = base_tile(iy,ix) if not elev else frame_H(image, iy,ix, Ly,Lx, Y,X, rV, max_elev=elev)
```

4. Still unfixed from last round: `T.yx`/`T.box` use `iY,iX` (frame_H's seed) instead of base_tile's own `y,x` — every tile claims the same position, corrupting dy_dx and proj_focus. And span is still full diagonal against the radius convention everywhere else (PP: `hypot(dy/2,dx/2)`, decay normalizes by mean span).

frame_2D_alg/agg_recursion.py
boris-kz: yes, pass them in sum2F?

frame_2D_alg/agg_recursion.py
kwcckw: > yes, it should be in base_tile?

Yes, so now elev doesn't matter in base_tile, and it's only used for rr. Also that Fr.Nt should be T.Nt instead.

frame_2D_alg/agg_recursion.py
kwcckw: We actually already have m_ and d_ as input in sum2F, i guess we need to readd this section in sum2F when fC = 1.

frame_2D_alg/agg_recursion.py
kwcckw: It's actually in the settings, under usage, but i'm not sure if the max plan got a different setting:
image

frame_2D_alg/agg_recursion.py
boris-kz: right

frame_2D_alg/agg_recursion.py
boris-kz: so a new tile is always pixel-level base_tile, but seed tile can be a lower frame.
that means we need to expand a new tile to the level of seed tile before the next loop?

kwcckw: Okay, i guess we need to parse the seed tile's elev into base tile, and set it as the max elev of the frame_H within?

boris-kz: def frame_H(image, iY,iX, Ly,Lx, Y,X, rV, max_elev=4):

    def base_tile(y,x):  # pixel-level tile

        if T := vect_edge( frame_blobs_root( comp_pixel(image[y:y+Ly, x:x+Lx]), rV), rV):
            T.yx = np.array([y+Ly//2, x+Lx//2])
            T.box = np.array([y,x, min(y+Ly,Y), min(x+Lx,X)])
            T.span = np.hypot(Ly,Lx)
        if cross_comp(T.Nt, rr=elev)[0]: return T

    def fill_frame(_iy,_ix, elev, T):  # seed tile is lower frame if elev else pixels

        cy,cx = (Ly-1)//2, (Lx-1)//2; y,x = cy,cx  # start=mean
        while True:
            if not elev:  # init pixel-level seed tile
                T = base_tile(_iy,_ix)
            if T and sum(vt_(T.dTT,T.wTT*ttFrm)) * wFrm > (ave+avd)*(T.r+cFrm):  # complex gate, make incremental?
                frame[y,x]=T; T+=[T]  # loop adds one tile to level
                if 0 < sum(vt_(pTT, T.wTT*ttFrm)) * wFrm < ave * cFrm:  # extend lev by combined proj T_
                    if PV__[y,x] > ave:
                        nxt_T = base_tile(iy,ix); d_elev = elev  # lower frame_H, up to current level
                        while d_elev:
                            d_elev-=1; fill_frame(iy,ix, d_elev, nxt_T)
                    else: break
                else: break
            else: break
?

kwcckw: Okay, so the expansion is in fill_frame, but the d_elev should be 0 in the first lev 0, so there will be no expansion in the first elev, and the number of fill_frame (expansion) is proportional to the elev now.

boris-kz: bunch of edits with Fable, still unsettled:

def proj_focus(PV__, y,x, tile):  # radial accum of projected focus value in PV__

    m,d = vt_(tile.dTT, tile.wTT * ttFrm); c = tile.c  # m,d,n = tile.m, tile.d, tile.c  # add r?
    V = (m-ave + d-avd) * c
    Dec = decay ** (c * np.hypot(*(tile.box[2:]-tile.box[:2])) / wYX)  # Dec = decay ** (wYX / np.hypot(*(tile.box[2:]- tile.box[:2])))
    H, W = PV__.shape  # = win__
    n = 1  # radial distance
    while y-n>=0 and x-n>=0 and y+n<H and x+n<W:  # rim is within frame
        dec = Dec * n
        pV__= np.array([
        V * dec * 1.4, V * dec * a, V * dec * 1.4,  # a = aspect = dy/dx, affects axial directions only
        V * dec / a,                V * dec / a,
        V * dec * 1.4, V * dec * a, V * dec * 1.4
        ], dtype=float)
        if np.sum(pV__) < ave * 8:
            break  # < min adjustment
        rim_coords = np.array([
        (y-n,x-n), (y-n,x), (y-n,x+n),
        (y, x-n),           (y, x+n),
        (y+n,x-n), (y+n,x), (y+n,x+n)
        ], dtype=int)
        row,col = rim_coords[:,0], rim_coords[:,1]
        PV__[row,col] += pV__  # in-place accum pV to rim
        n += 1

def proj_TT(L, cos_d, dist, r, pTT, wTT, pc, dec=1, fdec=0, frec=0):  # accumulate link pTT with iTT | eTT internally, L may be N

    dec = dist if fdec else ave ** (1 + (dist * dec) / L.span)  # not fully revised, ave = match decay rate / unit distance
    if not frec: pc[0] += L.c * dec
    cert = abs(sum(vt_(TT,wTT)) * wPrj)  # approximation
    if cert > (ave+avd)*(r+cPrj): # certainty margin
        pTT+=TT; return
    if not frec:  # non-recursive
        for lev in L.Nt.N_:  # refine pTT
            proj_TT(lev, cos_d, dec, r+1, pTT, wTT, fdec=1, frec=1)
    pTT += TT  # L.dTT is redundant to H, neither is redundant to Bt,Ct
    if L.Bt:  # + trans-link tNt, tBt, tCt?
        TT = L.Bt.dTT
        if TT is not None:
            pTT += np.array([TT[0] * dec, TT[1] * cos_d * dec])

def proj_N(N, dist, A, r, _c, dec=1):  # arg rc += N.rc+Nw, recursively specify N projection val, add pN if comp_pN?

    cos_d = (N.angl[0].dot(A) / ((np.hypot(*N.angl[0]) * dist) or eps)) * N.angl[1] if N.angl else 0  # int x ext angle alignment, mean=0
    iTT,eTT = np.zeros((2,9)),np.zeros((2,9))
    pc = np.zeros(1)
    wTT = CoF.get().wTT*ttPrj
    for L in N.L_+ N.B_:
        proj_TT(L, cos_d, dist, L.r+r, iTT, wTT, dec); pc[0] += L.c * dec  # accums TT internally
    if hasattr(N.Ct,'Lt') and N.Ct.Lt: [proj_TT(L,cos_d,dist,L.r+r,iTT,wTT,dec) for L in N.Ct.Lt.N_]  # C-to-N links
    for L in N.rim:
        proj_TT(L, cos_d,dist,L.r+r,eTT,wTT,dec); pc[0] += L.c * dec
    pTT = iTT + eTT  # projected int,ext links, work the same?
    oF = CoF.get(); oF.N_=[N]; oF.dTT=pTT; oF.c+=_c; oF.r+=r

    return pTT, pc[0]  # info_gain = N.m * average link uncertainty, should be separate

def frame_H(image, iY,iX, Ly,Lx, Y,X, rV, max_elev=4, ffb=0):

    def fill_frame(_iy,_ix, elev, T):  # expand level_frame from pixel-level seed tile

        PV__ = np.zeros([Ly,Lx])
        T_ = []
        while T and sum(vt_(T.dTT, T.wTT*ttFrm*T.c)) * wFrm > (ave+avd)*(T.r+(cFrm*T.c)):
            pTT, pc = proj_N(T, np.hypot(*dy_dx), dy_dx, elev, T.c)
            if 0 < sum(vt_(pTT, T.wTT*ttFrm)) * (wFrm*pc) < ave * cFrm:
                proj_focus(PV__,y,x, T)
                y,x = np.unravel_index(pv__.argmax(), PV__.shape)
                if PV__[y,x] > ave:
                    T = frame_H(image, iy,ix, Ly,Lx, Y,X, rV, max_elev=elev)  # expand new tile to current level, no fb
                else: break
            else: break
        if T_:
            TT,C,R = sum_vt(T_, wTT=ttFrm); R+= elev
            if sum(vt_(TT,ttFrm))*(C*wFrm) > (ave+avd)*(R+cFrm):
                return T_,C,R
        return [], 0, 0
    elev = 0
    global ave,avd  # update / ffeedback
    aTT = oTT = np.zeros((2,9))  # regime refs across levs
    Fr = []
    if T := vect_edge(frame_blobs_root(comp_pixel(image[iY:iY + Ly, iX:iX + Lx]), rV), rV):
        T.yx = np.array([iY + Ly // 2, iX + Lx // 2])
        T.box = np.array([iY,iX, min(iY+Ly, Y), min(iX+Lx, X)])
        T.span = np.hypot(Ly, Lx) / 2
        if not cross_comp(T.Nt, rr=0)[0]: T = []

    if not T or not max_elev: return T  # frame_H(0) = pixel tile
    while elev < max_elev:
        if tile_: # sparse,2D
            if cross_comp(Fr.Nt, rr=0)[0]:  # spec-> tN_,tC_,tL_, proj comb N_'L_?
                if elev and ffb:  # ffb =1 in main
                    aTT,oTT = ffeedback(Fr, aTT,oTT)  # term,form oH(aH
                elev +=1; T = Fr  # next-extension seed
            else: break
        else: break
    if Fr: oF = CoF.get(); oF.N_=[Fr]; oF.dTT=Fr.dTT; oF.c+=Fr.c; oF.r+=Fr.r

boris-kz: Three of the four landed half-way; two crash, one kills extension:

1. **proj_TT call sites weren't updated for the new signature.** pc is now the 7th positional, but all three calls still pass `dec` there — `proj_TT(L, cos_d, dist, L.r+r, iTT, wTT, dec)` puts a scalar in the pc slot, and `pc[0]` inside raises TypeError on the first link. And the refine recursion `proj_TT(lev, cos_d, dec, r+1, pTT, wTT, fdec=1, frec=1)` passes no pc at all — missing required positional. Meanwhile the old external `pc[0] += L.c * dec` lines survive at the call sites, which would double-count (undecayed) on top of the internal accumulation once the calls are fixed. Net fix in proj_N:

```python
for L in N.L_+ N.B_:  proj_TT(L, cos_d, dist, L.r+r, iTT, wTT, pc)
if hasattr(N.Ct,'Lt') and N.Ct.Lt: [proj_TT(L, cos_d, dist, L.r+r, iTT, wTT, pc) for L in N.Ct.Lt.N_]
for L in N.rim:  proj_TT(L, cos_d, dist, L.r+r, eTT, wTT, pc)
```

and in proj_TT's refine loop: `proj_TT(lev, cos_d, dec, r+1, pTT, wTT, pc, fdec=1, frec=1)`.

2. **proj_focus: c landed in the decay exponent.** `decay ** (c * f/wYX)` — c is mass, often tens or more, so Dec ≈ 0 and high-mass tiles project nothing: the loop breaks on ring 1, PV__ stays empty, extension dies. Inverted: the strongest tiles should project farthest. Mass already enters once, through V. The n in my formula was the radial loop variable, and `dec = Dec * n` is still linear. Both in one:

```python
Dec = decay ** (np.hypot(*(tile.box[2:]-tile.box[:2])) / wYX)  # per-step
...
dec = Dec ** n  # compounded per ring
```

3. **The proj gate is still the imbalanced one.** You mass-scaled the while gate but left `< ave * cFrm` where pc sits on the LHS — link-rich tiles exceed the constant bound and never extend. `< ave * (cFrm * T.c)`: then pc/T.c is the surviving fraction and the bound is scale-free.

4. The while gate's `T.wTT*ttFrm*T.c` works only because vt_ is linear in its weights — it reads as a weight tweak but is a mass multiplier. Clearer outside: `sum(vt_(T.dTT, T.wTT*ttFrm)) * (wFrm*T.c)`. And note what the edit actually changes: RHS scales only the cFrm term by c, not T.r — uniform c on both sides would cancel exactly (per-unit gate unchanged), so the entire behavioral content of this edit is "redundancy counts less for high-mass tiles." If that's not a position you hold deliberately, revert the gate to per-unit.

kwcckw: > Dec = decay ** (c * np.hypot(*(tile.box[2:]-tile.box[:2])) / wYX

So previously we have a typo? It should be in exponential form of base decay? Also including `c` may scale the `Dec` into a very large value since tile.c accumulates from all nested cs, so it's either 0 or a large tile.c value.

boris-kz: ** because multiplying a ratio makes it bigger, not smaller
but *c is not relevant for decay, it should be in eval only

kwcckw: > ** because multiplying a ratio makes it bigger, not smaller

Both should be the same? But with exponential scale them even larger or smaller?

Because when ratio <1, multiply or exponential scale value lower, while when ratio > 1, they scale it larger?

boris-kz: for ratio <1, exponential will decrease it, try .5**2

frame_2D_alg/agg_recursion.py
```diff
@@ -187,10 +187,10 @@ def comp_C_(C_,_r, _C_=[], fall=1, fC=0):  # simplified for centroids, trans-N_s
             tc += c; tr += r*c
             comp_N(_C,C,_r, c ,A=dy_dx,span=dist,L_=L_,N_=N_,acc=acc)  # simplified for typ=3
     for N in list(set(N_)):
-        if N.rim: sum2F(N.rim, N.Rt)
+        if N.rim: sum2F(N.rim, N.Rt)  # Rt.m is never updated, but we need them in get_exemplar? N.Rt = sum2F(N.rim)?
     if L_: sum2F(L_,CoF.get())
     TTm,cm,rm,TTd,cd,rd = acc
-    oF = CoF.get(); oF.N_ = N_; oF.dTT=TTm; oF.c+=c; oF.r+=r
+    oF = CoF.get(); oF.N_ = N_; oF.dTT=TTm; oF.c+=cm; oF.r+=rm  # cm and rm?
```
boris-kz, line 193: why cm and rm?

frame_2D_alg/agg_recursion.py
kwcckw: Okay thanks. So i misunderstood earlier. Yes, if decay is <1, then we need exponential (decay ** ratio) when ratio is smaller than 1, else (decay * ratio) grows larger when we multiply it.

frame_2D_alg/agg_recursion.py
kwcckw: We don't have accumulated var `c` and `r` here? 

So it's either `cm, rm` or `cm+cd, rm+rd`,  but oF's c and r should be from match only?

boris-kz: cm+cd, rm+rd, it's only selective in clustering

kwcckw: I see, thanks for clarifying. So we need to change it in the same way in comp_N_, and skip the return of ds as well

boris-kz: yes, please edit that and add m_,d_ in comp_C sum2F

PR #172, opened 2026-06-11
6/11/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -219,7 +219,7 @@ def comp_H(_Nt,Nt, Link):  # tentative pre-comp
                     r += (i or 1) -1  # Nt,Lt are not redundant
         for fb_, nF in zip((L.Nt.fb_,L.Lt.fb_,L.Bt.fb_,L.Ct.fb_), ('Nt','Lt','Bt','Ct')):
             if fb_: # L+= trans-Ls, python-batched bottom-up:
-                sum2F(fb_,getattr(L,nF))
+                sum2F(fb_,getattr(L,nF))  # L.Nt.N_ is replace L.N_
```
kwcckw, line 222: Sorry, looks like i overlooked this part. `L.N_` might be replaced by `fb_` here, that's when we sum `fb_` into `L.Nt.N_`, so i think the workaround is to have the separated `N_` and `Nt.N_` structure for trans link here? Where we should add the CN params manually instead of converting it into CN.

frame_2D_alg/agg_recursion.py
```diff
@@ -902,7 +906,7 @@ def fill_frame(_iy,_ix, elev, T):  # expand level_frame from pixel-level seed ti
         if tile_: # sparse,2D
             Fr = sum2F(tile_)  # higher-scope tile( oH( aH
             if cross_comp(Fr.Nt, rr=0)[0]:  # spec-> tN_,tC_,tL_, proj comb N_'L_?
-                if elev and ffb:  # ffb =1 in main
+                if elev and ffb:  # ffb =1 in main  (only top level got ffeedback now? From prior discussion, it could be any level as long there's diff?)
```
boris-kz, line 909: it's any level, but only in the main thread, not independently in each new tile of level expansion

frame_2D_alg/agg_recursion.py
boris-kz: so Ls won't use prop_F? will that conflict downstream?

frame_2D_alg/agg_recursion.py
kwcckw: Yes, thanks for the clarifying.

frame_2D_alg/agg_recursion.py
kwcckw: Yes, Ls won't use prop_F, and we access Nt.N_ via Nt anyway in trans_cluster:

    for tFt in L.Nt,L.Bt,L.Ct
    
`L.nt` is still accessed from `L.N_`.

frame_2D_alg/agg_recursion.py
```diff
@@ -85,7 +85,7 @@ def cross_comp(root, rr):  # core function mediating recursive rng+ and der+ cro
```
boris-kz, line 85: Going over general directions:

Read README, agg_recursion, meta_code. Organized by your three axes, with one cross-cutting thread: projection error (rTT) is your natural feedback currency, and it currently only feeds the top tiers (wTT via cent_TT, code via oF.rTT) while the bottom tiers (comp gate, coord selection) still run on static ave or raw value.

**Cross_comp**

1. The comp gate in comp_N_ is the right shape — comp only marginally predictable pairs, defer confident ones as pLs, break on negative projection — but the margin is static: `abs(m)*wN < ave*(r+cN)`. You already compute Link.rTT = relative prediction error per actual comp, then use it only for oF fitting. The local mean of rTT per root is exactly what the margin should be: where projections have been verified accurate, widen the skip band; where rTT is high, narrow it. This answers your own inline question "ave / proj surprise value?" and makes max comp range self-tuning per region before any ffeedback cycle.

2. Same signal should gate pL retro-comparison. add_Lt currently retro-comps pLs on cluster Lt.m — value of confirmed links. But the decision is really about trust in the deferred projections: low cluster ⟨rTT⟩ means pLs can stand un-comped permanently (pure compression win), high ⟨rTT⟩ means they were deferred on bad projections and should be comped first, not last.

3. Decay is one global scalar over spatial distance, but you have three distance axes: span, derivation depth, composition depth. mdecay already measures empirical -dm/ddist over spans; the identical slope is measurable over H levels (from comp_H's per-level dH) and over sub. Three empirical decay rates instead of one prior — and they're the per-axis range filters that feedback should be adjusting. der+ and agg+ extension would then be gated by their own measured decay, not by r-increments standing in for it.

4. der+ as written in the docstring (link-local, abs_diff × adjacent match) doesn't exist in the code path: comp_N_ accumulates TTd into acc but never spawns link cross-comp from it; der+ only materializes post-clustering via Bt in cluster_N/sum2G. So a high-d region that fails the cluster_N gate never gets derivation increment — yet contour-rich regions are exactly where your README says the information is. Is "no composition → no derivation" a principle, or an accident of routing der+ through Bt?

5. `if _N.sub != N.sub: continue` — but base_comp(C,n) in cluster_C is already cross-composition comp. The principled N-N version is mediated descent: comp the smaller N against the best-matching constituent of the larger one, at composition-distance decay. Cheap to defer as a pL rather than hard-skip.

**Clustering**

6. sum2F weights strictly by c even when m_/d_ are passed: fuzzy memberships are stored but don't weight the centroid's params. A marginal member contributes equal to a core member, so cluster_P's reform isn't actually FCM-converging — the centroid drifts toward evidence mass, not membership mass. If deliberate (c = evidence quantity is the only objective weight), then m_ is purely a prune signal; but then "fuzzy" reduces to soft assignment with hard summation. Which is intended?

7. trans_cluster merges whole roots on any single strong tL. Compression logic says: merge two Gs iff their aggregated trans-match exceeds the Bt contrast separating them — otherwise the merge destroys a boundary that just paid for itself. Per-tL triggering also makes the result order-dependent. Related: the centroid-merge gate in comp_C_ sorts by min d but then merges while `vt_(L.dTT,ttF)[0]*wF < avd*(...)` — gating on *low match*. Inverted, or is low-m-at-low-d meant as "distinction not worth representing"?

8. cluster_P is the one stage that violates your sparsity claim: dense Ln×Lc base_comp per iteration over the union of all C.N_. The fix is the same proj_V gate from comp_N_: candidate Cs per N limited to projected range, md__ sparse. Then "parallel" and "sparse" stop being in tension.

9. cluster_N assignment is strong-exemplar-first with fin — order-dependent leftovers for weak seeds. Parallel label propagation (each N takes strongest claimant) keeps the NMS prioritization but is symmetric and parallelizable; the fuzzy stage only corrects this for nodes that reach cluster_C.

**Feedback**

10. Coords: fill_frame moves to argmax PV__ — max projected value, i.e. max predictability. Your README says selection is (proximity − predictability): edge of predictable. Reconciliation: PV per cell = projected value × projection uncertainty (local ⟨rTT⟩ again) — expected information gain, not expected confirmation. Also proj_focus is rank-1 anisotropic (one mean angle a); mang already measures angle coherence, so it should interpolate between the directional kernel and isotropic spread.

11. Filters: `global ave,avd` mutated mid-run means later tiles inherit filters fitted to earlier regimes, and parallel tiles would race. You already thread wTT through roots; ave/avd belong there too, with pack_seg segments as filter scopes — regime-local filters fall out for free, and ffeedback becomes functional.

12. Code: F_call_T_ is write-only — accumulated per call site, never read. It's the missing third term in oF similarity: cluster_oF_ uses body compression + caller overlap, i.e. syntax + role, but not output statistics. Two oFs with matching accumulated dTT profiles under shared callers are the real merge candidates. Stronger version: give oFs a rim of call-site links carrying those dTTs, define base_comp(oF,oF) = body compression + caller overlap + comp_derT of outputs, distance = call-graph hops — then get_exemplars/cluster_N/cluster_C subsume split_oF_/cluster_oF_ outright. Those two are hand-rolled re-implementations of your general clustering; eliminating them is the test of "singular algorithm."

One inconsistency to settle: rTT means prediction error on Links, selectivity gain in cluster_N/cluster_C, raw comp totals in cross_comp — all funneled into cent_TT as salience. abs() hides the sign mismatch, but error and gain are not the same currency for weighting attrs. If rTT is to be the universal feedback signal (points 1, 2, 10), its semantics need to be fixed as projected-vs-actual deviation everywhere.

kwcckw: > but you have three distance axes: span, derivation depth, composition depth

This is actually very helpful, but for proj_focus, proj_N, proj_TT, the distance is span. For derivation depth and composition depth, it's for the projection from fill_frame?

frame_2D_alg/agg_recursion.py
```diff
@@ -145,12 +145,12 @@ def proj_V(_N,N, dist, dy_dx, dec):  # _N x N induction
...
         else: break  # beyond initial induction range, re-sort by proj_V?
     for N in set(N_):
-        if N.rim: sum2F(N.rim, N.Rt)
+        if N.rim: N.Rt = sum2F(N.rim)
     # call trace:
     if L_ := [n for n in L_ if n.typ>-1]: sum2F(L_,CoF.get())  # skip pLs, oF.call_+=[oF], adds data
     TT,cm,rm, TTd,cd,rd = acc
-    oF = CoF.get(); oF.N_=N_; oF.dTT=TT; oF.c+=c; oF.r+=r
-    return L_,TT,cm,rm/(cm or eps), TTd,cd,rd/(cd or eps)
+    oF = CoF.get(); oF.N_=N_; oF.dTT+=TT+TTd; oF.c+=cm+cd; oF.r+=rm+rd  # normalize r?
```
boris-kz, line 152: r should probably not be summed here

frame_2D_alg/agg_recursion.py
```diff
@@ -219,7 +219,7 @@ def comp_H(_Nt,Nt, Link):  # tentative pre-comp
                     r += (i or 1) -1  # Nt,Lt are not redundant
         for fb_, nF in zip((L.Nt.fb_,L.Lt.fb_,L.Bt.fb_,L.Ct.fb_), ('Nt','Lt','Bt','Ct')):
             if fb_: # L+= trans-Ls, python-batched bottom-up:
-                sum2F(fb_,getattr(L,nF))
+                sum2F(fb_,getattr(L,nF))  # L.Nt.N_ is replace L.N_
         sum2F([L.Nt,L.Lt,L.Bt,L.Ct], CoF.get())  # data should be L.Fts?
```
boris-kz, line 223: also sum L proper, it doesn't include tFts?

        sum2F([L, L.Nt,L.Lt,L.Bt,L.Ct], CoF.get())  # no L += trans-Fts?

frame_2D_alg/agg_recursion.py
boris-kz: It's actually not relevant, derivation and composition add costs, not distance

frame_2D_alg/agg_recursion.py
kwcckw: Those trans-forks are already parsed as L.Nt, L.Lt, L.Bt and L.Ct, i think we shouldn't parse L again.

boris-kz: it's the same oF, we need to combine these:

    sum2F([L.Nt, L.Lt, L.Bt, L.Ct], CoF.get())  # no L += trans-Fts?
    oF = CoF.get(); oF.N_=[_N,N]; oF.dTT=L.dTT; oF.c+=L.c; oF.r+=L.r

frame_2D_alg/agg_recursion.py
kwcckw: It should be normalized `rm/(cm or eps) + rd/(cd or eps)`?

frame_2D_alg/agg_recursion.py
kwcckw: We probably need to check if L has added tFts first? That trans-comp is not default, so it's in the section under `if N.typ and m*wN_ > ave*(r+cN_):`.

boris-kz: yes, but it should all be in in the end.
if L is not CN then its core N_, dTT, etc should not include tFts?

kwcckw: > if L is not CN then its core N_, dTT, etc should not include tFts?

Yes, but if L has added tFts, their Nt.N_ overwrites the nt (L.N_), so we probably shouldn't convert them into CN. We should just need to add Nt, Lt, Ct and Bt into L? Then L.Nt.N_ and L.N_ are different references of trans Nt and nt.

boris-kz: yes,

    oF = CoF.get()
    C = oF.c + L.c + L.Nt.c + L.Lt.c + L.Bt.c + L.Ct.c; dTT,R = np.zeros((2,9)) 
    for Ft in oF, L, L.Nt, L.Lt, L.Bt, L.Ct: rc=Ft.c/C; dTT+=Ft.dTT*rc; R+=Ft.r*rc 
    oF.N_ += [L]; oF.dTT=dTT; oF.c=C; oF.r=R
    return L

kwcckw: Okay, but why the oF.N_ is L? It should be the input N_?

boris-kz: oF is comp_N, the nodes are results: Ls

kwcckw: Okay, previously my understanding was the nodes are the processing data, which is the inputs, so now all oF.N_ should be the results?

boris-kz: yes, that's where the values come from
should be the same for clustering: the nodes are the clusters, not elements?

kwcckw: Yes, but clustering's oF is already using the resulting Gs?

    oF.N_=copy(G_)
    
So there shouldn't be any changes on the clustering's oF.

boris-kz: I think we need to use sum_vt for all oFs?

frame_2D_alg/agg_recursion.py
boris-kz: yes, use sum_vt?

frame_2D_alg/agg_recursion.py
kwcckw: Yes, we can use that for most of the function, but except clustering? Because their TT is based on selectivity?

boris-kz: just use rTT instead?

kwcckw: So clustering oF will be having a separate rTT and dTT?

boris-kz: Just rTT, that's a value param

PR #173, opened 2026-06-12
6/12/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -207,9 +208,10 @@ def comp_H(_Nt,Nt, Link):  # tentative pre-comp
     TT= base_comp(_N,N)[0] if full else comp_derT(_N.dTT[1],N.dTT[1]); m,d = vt_(TT, ttN_)
     L = CL(N_=[_N,N], dTT=TT,m=m,d=d,c=_c,r=r, root=rL)
     if N.typ==3 and _N.typ==3 and m*wF > ave*(r+cF):
-        dH,htt,_,_ = comp_H(_N,N, L); r+=R; m+= vt_(htt,ttN_)[0]  # tentative dH, replace if subcomp?
+        # we not really using this dH?
+        dH,htt,_,_ = comp_H(_N,N, L); m+= vt_(htt,ttN_)[0]  # tentative dH, replace if subcomp?
```
kwcckw, line 212: If this dH is tentative and as replacement for subcomp, it should be under else of this trans section?

    if N.typ and m*wN_ > ave*(r+cN_):

frame_2D_alg/agg_recursion.py
```diff
@@ -233,10 +235,9 @@ def comp_H(_Nt,Nt, Link):  # tentative pre-comp
         if   L.m > ave: acc[0]+=L.dTT*L.c; acc[1]+=L.c; acc[2]+=L.r*L.c; L_ += [L];
         elif L.typ==1 and L.d > avd: acc[3]+=L.dTT*L.c; acc[4]+=L.c; acc[5]+=L.r*L.c
         # no pLs?
-    oF = CoF.get()
-    C = oF.c + L.c + L.Nt.c + L.Lt.c + L.Bt.c + L.Ct.c; dTT, R = np.zeros((2,9))
-    for F in oF, L, L.Nt, L.Lt, L.Bt, L.Ct: rc = F.c/C; dTT+=F.dTT*rc; R+=F.r*rc
-    oF.N_ += [L]; oF.dTT=dTT; oF.c=C; oF.r=R
+    oF = CoF.get(); oF.N_ = [L]; 
+    if hasattr(L, 'Nt'): oF.rTT, oF.c, oF.r = sum_vt([oF, L, L.Nt, L.Lt, L.Bt, L.Ct])
```
kwcckw, line 239: I guess we can use sum_vt here instead of summing and dividing them manually?

frame_2D_alg/agg_recursion.py
boris-kz: dH, m should be used if there is no sub_comp, it's a light-weight estimate

frame_2D_alg/agg_recursion.py
boris-kz: that's what I told you
but it's not just Nt, we need to sum all Fts

frame_2D_alg/agg_recursion.py
```diff
@@ -861,7 +865,7 @@ def proj_N(N, dist, A, r, _c, dec=1):  # arg rc += N.rc+Nw, recursively specify
     for L in N.rim:
         proj_TT(L, cos_d,dist,L.r+r,eTT,wTT,dec); c+=L.c
     pTT = iTT + eTT  # proj int,ext links, work the same?
-    oF = CoF.get(); oF.N_=[N]; oF.dTT=pTT; oF.c+=_c; oF.r+=r
+    oF = CoF.get(); oF.N_=[N]; oF.rTT=pTT; oF.c+=_c; oF.r+=r
```
boris-kz, line 868: the name is still dTT

frame_2D_alg/agg_recursion.py
```diff
@@ -324,9 +325,10 @@ def get_exemplars(N_,_r,_c):  # multi-layer non-maximum suppression -> sparse se
             Inh_.update(set(N.rim))  # extend inhibition zone
         else:
             break  # the rest of N_ is weaker, trace via rims
-    oF = CoF.get(); oF.N_ = N_
+    oF = CoF.get()
     if E_:  sum2F(list(E_),oF)  # or default with wc cost?
```
boris-kz, line 329: oF doesn't need full sum2F

frame_2D_alg/agg_recursion.py
```diff
@@ -97,7 +97,8 @@ def cross_comp(root, rr):  # core function mediating recursive rng+ and der+ cro
                 root.Nt = sum2F(G_,root,froot=2); L=len(G_)-1  # or C_ s?
                 if vt_(TT,tAgg)[0]*(wAgg*L) > ave* (r+cAgg*L):  # root brrw| rdn?
                     G_,r = cross_comp(root.Nt,r)  # agg+
-            oF = CoF.get(); oF.N_ = G_ or N_; oF.rTT=TT; oF.c+=c; oF.r+=r
+            oF = CoF.get(); oF.N_ = G_ or N_;
```
boris-kz, line 100: Each function type in oF_ stores all instances in oF.call_ via traced,
and we also need to accumulate call results, such as:

```
if G_:
    oF = CoF.get(); oF.N_+=G_; oF.dTT,oF.c,oF.r = sum_vt(G_)
    # else costs only: len(oF.call)?

```
can this be done in traced() too?

frame_2D_alg/agg_recursion.py
```diff
@@ -783,7 +786,8 @@ def trace_edge(N_,_G_,_TT,_C, r,root):  # cluster contiguous shapes via PPs in e
             else:
                 for N in n_: N.fin=0; N.root=root
     if sum(vt_(TT,root.wTT*ttTrc))*wTrc > (ave+avd)*(r+1+cTrc): _G_+=G_; _TT+=TT;_C+=C  # eval per edge, concat in tile?
-    oF = CoF.get(); oF.N_= G_ or N_; oF.dTT=_TT; oF.c+=_C; oF.r+=r+R/_C
+    oF = CoF.get(); oF.N_= G_ or N_  # do we really need this or N_ here?
```
boris-kz, line 789: we don't, 

            if G_:
                oF = CoF.get(); oF.N_+=G_; oF.dTT,oF.c,oF.r = sum_vt(G_)
                # else costs only: len(oF.call)?

But then, cross_comp oF is redundant to the final clustering oF?

frame_2D_alg/agg_recursion.py
boris-kz: Mechanically yes — `traced` has `out = func(*a, **kw)` in hand, so it can apply your line as a default, given two things:

1. **Return convention**: first element of return is the result node list (`G_`|`L_`|`E_`|`[N]`). Mostly true already (`cross_comp`, `comp_N_`, `comp_C_`, `cluster_N`, `cluster_C`, `get_exemplars`), but not for `proj_N` (returns `pTT,c`), `trace_edge`, `frame_H`. Duck-type: take it only if it's a list of `CF`s.
2. **Late-bound `sum_vt`**: meta_code can't import agg_recursion. `trace_func` already receives `module_dict` — grab `sum_vt` there and bind it module-level, no circular import.

In `traced.inner`, gated so in-body accumulation wins:

```python
out = func(*a, **kw)
if not oF.c:  # no in-body accum, default from result:
    res = out[0] if isinstance(out,tuple) else out
    G_ = res if isinstance(res,list) and res and isinstance(res[0],CF) else []
    if G_: oF.N_+=G_; oF.dTT,oF.c,oF.r = sum_vt(G_)
    # else costs only: fc is static, len(oF.call_) is runtime count, oF.c=0 keeps __bool__ False
```

The else-branch needs nothing: cost-only calls fall out for free via `CF.__bool__`.

But the conceptual issue is that your existing in-body lines don't compute `sum_vt(result)` — they compute function-specific *net* quantities that the return value alone can't yield:

- `cluster_N`: `oF.rTT = sel_TT` (selectivity vs `Ft.root.Lt`), `oF.c = Ft.root.c - Ft.c` — deltas needing `Ft.root`, not gross `sum_vt(G_)`.
- `comp_N_`: `acc` includes pL contributions (`typ=-1`) that are filtered out of returned `L_`, so `oF.dTT ≠ sum_vt(returned L_)`.

If oF value feeds `Fw_` as realized gain per scope, gross `sum_vt(G_)` is also wrong for non-leaf forks: `G_` is built from sub-call outputs already rolled up via `call_`, so gross summation double-counts children and systematically inflates parent oFs relative to leaves. So `traced` can host only the lowest-common-denominator default for leaf-ish functions without explicit accumulation; the net-gain semantics stay in-body, because "value added" depends on what the function does — that's irreducibly per-function.

Separate flag, same pathway: `sum2O(tree, oF, fcall_=1)` in `traced` passes oF as `root`, builds a *new* CoF, and discards it — currently a no-op on oF (w_ is None, so no fw side effects either). If that was meant to roll child results up into oF, that's the missing half of this accumulation; fix it before adding a third pathway (in-body + sum2O + return-default), or you risk triple-counting.

boris-kz: Four edits, one convention, one open call.

**Convention** (so traced can find in/out without per-function code): input = leading bare-CF positional args, plus a CF-list/set or pair-list only at position 0; output = same extraction over the return tuple. Checked against all 16 traced functions: holds everywhere except `comp_N_`'s consumed generator (edit 3). `image`/`blob`/`edge`/`tile`-first functions duck-type to in=0 (pure generation); functions returning no CFs (`proj_N`, `ffeedback`, `comp_slice`, `slice_edge`) become cost-only oFs — fine until they return node results.

**1. meta_code: move `vt_`,`sum_vt` here** from agg_recursion (their globals — `ave,avd,wTT,eps_` — already live in meta_code); import back into agg_recursion. Avoids circularity, single definition.

**2. meta_code: extractor + traced delta:**

```python
def val_(seq):  # leading-CF convention
    N_ = []
    for j,x in enumerate(seq):
        if isinstance(x,CF): N_ += [x]
        elif j==0 and isinstance(x,(list,set,tuple)) and x:
            x = list(x)
            if   isinstance(x[0],CF): N_ += x
            elif isinstance(x[0],(tuple,list)) and isinstance(x[0][0],CF): N_ += list({n for p in x for n in p})  # pair_
            break
        else: break
    return sum_vt(N_) if N_ else (np.zeros((2,9)),0,0)  # sum_vt allocates TT: snapshot is mutation-safe
```
In `inner`, around the call:
```python
iTT,ic,ir = val_(a)        # pre-call: cross_comp, cluster_N mutate args
out = func(*a, **kw)
oTT,oc,orr = val_(out if isinstance(out,tuple) else (out,))
if oc:  # realized result, else costs only: fc static, len(call_) runtime
    oF.dTT = oTT-iTT; oF.c = oc-ic; oF.r = orr
    oF.N_ = next((x for x in (out if isinstance(out,tuple) else (out,)) if isinstance(x,list) and x and isinstance(x[0],CF)), [])
```
and drop the `rTT` special-case: `oF.wTT = cent_TT(oF.dTT, oF.r)` — the delta is already relative. `oF.r` as output-absolute, not delta: r is a cost rate, a delta can go negative meaninglessly.

**3. agg_recursion, line 88:** materialize pairs — `comp_N_(list(combinations(N_,2)), rr)`. Confirm `Np_` at 254 is a list (it is, if built by comprehension before the branch).

**4. agg_recursion: delete the three in-body oF lines** — `comp_N_` ~152, `cluster_N` ~397, `cluster_C` ~457. The `acc` machinery stays: it feeds the *returns* that callers eval on (`TT` in cross_comp's gate). Return = caller interface, oF = traced delta — now fully decoupled, which is the actual payoff.

**The forced call (pLs):** with out−in, pL value drops from oF — pLs are filtered from returned `L_` and aren't in the input summary. I'd accept that: projection is borrowed value, not realized gain; it shouldn't credit the fork in `Fw_`. If you disagree, the only consistent fix is returning unfiltered `L_` and filtering at call sites — counting value that's in neither in nor out breaks the definition.

**Open semantic point, your choice:** parent deltas are inclusive — `cluster_N`'s `G_` embeds `comp_N`/`comp_F` child gains, so parent oF.dTT ⊇ Σ children. Inclusive-per-scope is defensible for `Fw_` (gain at that scope). If you want leaf-comparable marginals instead, one line in traced after `sum2O`: subtract c-weighted child deltas — children are already in `oF.call_`.

Semantics that change silently: `cluster_N`'s oF.c was `Ft.root.c − Ft.c` (selection savings vs root), now `sum_vt(G_).c − Ft.c` (output vs input scope); `sel_TT` vs `root.Lt` becomes delta vs `Ft`. That's the uniformity you're buying — flagging so it's deliberate.

frame_2D_alg/agg_recursion.py
kwcckw: Yes, so it should be in the else section?

Also we pack trans in L.tFts, and without sub, this dH will be packed in L as well?

frame_2D_alg/agg_recursion.py
kwcckw: Yes, and from your code, we should skip the L? We should include only L.tFts? Since tFts is already summed to the L.

frame_2D_alg/agg_recursion.py
kwcckw: Thanks, actually this is obsolete. We only need sum_vt now.

frame_2D_alg/agg_recursion.py
kwcckw: If this is done in traced, it will be back to prior scheme where we need a pre-defined function output format for unpacking purpose?

> and we also need to accumulate call results, such as:
> else costs only: len(oF.call)?

We can accumulate it from each oF to their root in traced? For the else case above, if G_ is empty, we add `oF.c = len(oF.call_)`?

frame_2D_alg/agg_recursion.py
kwcckw: The final clustering function is cluster_N, and the oF's values are assigned based on selectivity, while here it's just summed from G_, which is different?

frame_2D_alg/agg_recursion.py
boris-kz: It's m is needed to eval for transcomp.
Without it, dH should be from tNt.H

frame_2D_alg/agg_recursion.py
boris-kz: tFts should probably be packed in L.H, but we need separate core L values

frame_2D_alg/agg_recursion.py
boris-kz: We only need to return N_, sum the rest from it

kwcckw: Okay, and from clustering functions, i guess those should be a unique case where their rTT, c and r needs to be summed differently within the function.

frame_2D_alg/agg_recursion.py
kwcckw: I see, so L.H is a unique structure now where each element is tNt, tLt, tBt and tCt? Or all of them should be packed as a t-level's CF in L.H?

frame_2D_alg/agg_recursion.py
boris-kz, line 462: re cluster_P:

Structurally yes — this is multi-seed mean shift over a fixed sample: hold N_ constant, recompute every centroid as a similarity-weighted aggregate of *all* nodes each round, iterate. The kernel is your derived match function instead of a fixed Gaussian, and membership is measured (base_comp → vt_), never optimized — exactly the no-fuzzifier, no-budget scheme. Two load-bearing questions decide how literal the correspondence is: does vt_ return signed m, and do negatives enter sum2F as weights? Negative weights mean repulsion by dissimilar nodes — that's signed-graph dynamics, not kernel density, with different convergence behavior. And does sum2F divide by total weight? Weighted mean vs weighted sum matters: unnormalized, the prototype's magnitude conflates central tendency with total support.

Where you diverge from mean shift, mostly in interesting ways. The d-channel makes your prototype a mixture-component analog — central tendency plus accumulated spread — where a mode is just a location. That's also your exit from the bandwidth knob: per-centroid d is exactly the statistic an adaptive kernel needs, so the falloff could be derived from each cluster's own measured spread rather than a global ave. The stopping rule is the real novelty: value × update-magnitude against iteration rent means refinement compute is allocated proportional to cluster value — weak sets stop mid-transient by design. Mean shift has no economics; you've replaced its convergence proof with a budget. Note the exposure that creates: cnt is incremented and never checked. A high-value oscillating configuration — signed weights make this plausible — passes the threshold forever. Cap it.

Readout: existence pruning (C.m against ave×C.r is AP's preference term) then per-node thresholding, then re-forming the prototype from survivors. That last step is the consistency-over-accuracy choice from before — the output summary matches its member set, but truncating the soft tails reintroduces boundary bias into the prototype. Keeping graded m_ within the survivor set mitigates it; just be aware the converged soft centroid and the shipped centroid differ systematically, pulled toward the cluster interior.

The commented question is the actual gap. Nothing here detects coincidence, and the kernel dynamic *guarantees* seeds in one basin converge to the same fixed point — so duplicates survive pruning, both having high m. Cheapest test uses what you've already computed: two centroids in the same basin have near-parallel weight columns in md__, so a correlation check across columns, merge above threshold, costs nearly nothing at Lc². Pairwise base_comp on the converged C_ works too. Without one of these, the differentiation story we settled on — merge-on-coincidence — is still a TODO, and this is PCM with extra steps.

frame_2D_alg/agg_recursion.py
boris-kz: Most functions are not just the main process, they also gate and process results from other oFs.
So the root function should be summing the data statistics from its callees?
Which will be output - input, not just output as we currently have it?
That will cover the seeming anomaly in agg_recursion and clustering oFs, as well as Ls from comp oFs?

frame_2D_alg/agg_recursion.py
boris-kz: L.H should represent sub-comp levels: N_ remains as the unpacked lev0, with separate pair-wise vals,
initial tNt, tLt, tBt, tCt in H[0],  ttNt, ttLt, ttBt, ttCt in H[1], etc.
The difference from N.H is that these levels are not redundant, base L vals do not represent them

frame_2D_alg/agg_recursion.py
boris-kz: Partly — three claims in there, two hold, one doesn't.

**Yes, parents aggregate callee deltas** — that's the repaired `sum2O(tree, oF, fcall_=1)` (currently a no-op, as flagged). And yes, what's summed are the out−in deltas, not gross outputs. But Σ(children) ≠ parent's own out−in, and the gap is exactly the parent's own contribution: gating, eval rejections, root reformation, summation outside callees. So keep both quantities:

```python
if oF.call_:  # direct children only: each child Δ already includes its subtree, flat_ would double-count
    cTc = sum(f.dTT*f.c for f in oF.call_); cc = sum(f.c for f in oF.call_)
    oF.mTT = oF.dTT*oF.c - cTc; oF.mc = oF.c - cc  # own marginal, extensive units
```

Marginal matters because `Fc_` already counts cost marginally (`ast.Call`=3, callee bodies separate) — pairing inclusive gain with marginal cost would let root_ functions dominate `Fw_` trivially. Note the extensive units: dTT is a c-weighted mean; deltas of means don't add across children, so sum `dTT*c`, divide back where needed.

**No, this doesn't cover the L anomaly** — because that's *sibling* overlap, not parent-child. `cluster_N` is a traced leaf (none of its internals are in `_names`), so callee-summing never touches it. The problem: its input snapshot `Ft` excludes the L-value comp just created — cross_comp sets `root.L_ = L_` but never folds comp's `TT,c` into root's summary. So cluster's Δ = G − Ft re-counts the `lc, bc` embedded in `G.c = nc+lc+bc`, which comp's Δ already claimed. Fix is in data, not in the convention or the summing:

```python
root.L_ = L_; w = root.c/(root.c+c); root.dTT = root.dTT*w + TT*(1-w); root.r = root.r*w + r*(1-w); root.c += c
```

before the cluster call. Then cluster's Δ = G − (N+L): pure reorganization value, typically negative c — compression, which is clustering's value in your own terms (selectivity / loss reduction, the old sel_TT comment). One deliberate side effect: `cluster_N`'s first gate reads `Ft.dTT`, so it now sees link value. Arguably more correct — the clustering decision should be informed by comp results — but it changes gate behavior, your call.

pLs stay consistent throughout: out of returned `L_`, out of the fold, out of every Δ.

Residual imperfection to accept: `get_exemplars` (E−N, negative c) and `cluster_N` both read N, so exact telescoping at cross_comp is unattainable — Σ(children) + marginal won't conserve perfectly across overlapping reads. That's fine: `Fw_` needs comparable per-fork marginals, not a conserved total.

kwcckw: > So the root function should be summing the data statistics from its callees?
Which will be output-input, not just ouptut as we currently have it?

So the problem is the overlapping results across multiple functions, since we have overlapping output in comps and get_exemplars. And to resolve this, we need to update root's oF after each callee?

frame_2D_alg/agg_recursion.py
kwcckw: >  merge-on-coincidence

Actually we shouldn't have merge on coincidence in the current workflow?

frame_2D_alg/agg_recursion.py
boris-kz: yes, update _CoF, but only with input-output delta

frame_2D_alg/agg_recursion.py
boris-kz: we should,

    if out_:
        merge_C_(out_,N_)  # merge similar centroids, needs modification
        dCt = sum2F(list(set(_C_)-set(out_)))  # compress
        return out_

this is similar to the merge sequence in comp_C

frame_2D_alg/agg_recursion.py
kwcckw: Okay, so that should be done in the traced, right after the callee function is completed.

frame_2D_alg/agg_recursion.py
kwcckw: I see, and that ttFts is only possible via Ls, from this section?

       if N.typ==1:
            for _n,n in product(_N.N_,N.N_): L.Nt.fb_ += [comp_N(_n,n,r,_c)]  # link sub-comp

frame_2D_alg/agg_recursion.py
kwcckw: Okay, and the criterion will be the same?

    for i, L in enumerate(L_):
                    C0,C1 = L.N_
                    add2F(C0,C1,1); merg_ += [C1]
                    for l in C1.rim: l.N_ = [C0 if n is C1 else n for n in l.N_]
                else: L_ = L_[i:]; break
            
Which is based on their dTT? Which we should have at least compare the Cs before merge them?

boris-kz: if out_:
        merg_ = []
        for _C, C in combinations(out_,r=2):
            if _C in merg_ or C in merg_: continue
            c = min(C.c,_C.c); r = (C.r+_C.r)/2; dy_dx = _C.yx-C.yx; dist = np.hypot(*dy_dx)
            L = comp_N(_C,C,r,c, A=dy_dx,span=dist)
            if vt_(L.dTT,ttF)[0]*wF < avd*(r+cF): add2F(_C,C,1); merg_ += [C]
        out_= list(set(out_)-set(merg_))

frame_2D_alg/agg_recursion.py
boris-kz: actually no, that H is L.Nt.H, so lets keep it as is

frame_2D_alg/agg_recursion.py
boris-kz: please check:

Four edits, one convention, one open call.

**Convention** (so traced can find in/out without per-function code): input = leading bare-CF positional args, plus a CF-list/set or pair-list only at position 0; output = same extraction over the return tuple. Checked against all 16 traced functions: holds everywhere except `comp_N_`'s consumed generator (edit 3). `image`/`blob`/`edge`/`tile`-first functions duck-type to in=0 (pure generation); functions returning no CFs (`proj_N`, `ffeedback`, `comp_slice`, `slice_edge`) become cost-only oFs — fine until they return node results.

**1. meta_code: move `vt_`,`sum_vt` here** from agg_recursion (their globals — `ave,avd,wTT,eps_` — already live in meta_code); import back into agg_recursion. Avoids circularity, single definition.

**2. meta_code: extractor + traced delta:**

```python
def val_(seq):  # leading-CF convention
    N_ = []
    for j,x in enumerate(seq):
        if isinstance(x,CF): N_ += [x]
        elif j==0 and isinstance(x,(list,set,tuple)) and x:
            x = list(x)
            if   isinstance(x[0],CF): N_ += x
            elif isinstance(x[0],(tuple,list)) and isinstance(x[0][0],CF): N_ += list({n for p in x for n in p})  # pair_
            break
        else: break
    return sum_vt(N_) if N_ else (np.zeros((2,9)),0,0)  # sum_vt allocates TT: snapshot is mutation-safe
```
In `inner`, around the call:
```python
iTT,ic,ir = val_(a)        # pre-call: cross_comp, cluster_N mutate args
out = func(*a, **kw)
oTT,oc,orr = val_(out if isinstance(out,tuple) else (out,))
if oc:  # realized result, else costs only: fc static, len(call_) runtime
    oF.dTT = oTT-iTT; oF.c = oc-ic; oF.r = orr
    oF.N_ = next((x for x in (out if isinstance(out,tuple) else (out,)) if isinstance(x,list) and x and isinstance(x[0],CF)), [])
```
and drop the `rTT` special-case: `oF.wTT = cent_TT(oF.dTT, oF.r)` — the delta is already relative. `oF.r` as output-absolute, not delta: r is a cost rate, a delta can go negative meaninglessly.

**3. agg_recursion, line 88:** materialize pairs — `comp_N_(list(combinations(N_,2)), rr)`. Confirm `Np_` at 254 is a list (it is, if built by comprehension before the branch).

**4. agg_recursion: delete the three in-body oF lines** — `comp_N_` ~152, `cluster_N` ~397, `cluster_C` ~457. The `acc` machinery stays: it feeds the *returns* that callers eval on (`TT` in cross_comp's gate). Return = caller interface, oF = traced delta — now fully decoupled, which is the actual payoff.

**The forced call (pLs):** with out−in, pL value drops from oF — pLs are filtered from returned `L_` and aren't in the input summary. I'd accept that: projection is borrowed value, not realized gain; it shouldn't credit the fork in `Fw_`. If you disagree, the only consistent fix is returning unfiltered `L_` and filtering at call sites — counting value that's in neither in nor out breaks the definition.

**Open semantic point, your choice:** parent deltas are inclusive — `cluster_N`'s `G_` embeds `comp_N`/`comp_F` child gains, so parent oF.dTT ⊇ Σ children. Inclusive-per-scope is defensible for `Fw_` (gain at that scope). If you want leaf-comparable marginals instead, one line in traced after `sum2O`: subtract c-weighted child deltas — children are already in `oF.call_`.

Semantics that change silently: `cluster_N`'s oF.c was `Ft.root.c − Ft.c` (selection savings vs root), now `sum_vt(G_).c − Ft.c` (output vs input scope); `sel_TT` vs `root.Lt` becomes delta vs `Ft`. That's the uniformity you're buying — flagging so it's deliberate.

frame_2D_alg/agg_recursion.py
kwcckw: Alright, I will go through them later.

PR #174, opened 2026-06-13
6/13/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -121,7 +121,8 @@ def proj_V(_N,N, dist, dy_dx, dec):  # _N x N induction
                 Link = comp_N(_N,N, lr,c,full=not tnF, A=dy_dx, span=dist, rL=root,L_=L_,N_=N_, acc=acc)
                 Link.rTT = np.abs(pTT-Link.dTT) / eps_(Link.dTT)  # relative prediction error to fit oF, direction-agnostic
             else:
-                pL = CN(typ=-1, dTT=pTT,m=m,d=d,c=c,r=lr, angl=[dy_dx,1],span=dist); pL.N_ = [_N,N]
+                # why this is CN here?
+                pL = CL(typ=-1, dTT=pTT,m=m,d=d,c=c,r=lr, angl=[dy_dx,1],span=dist); pL.N_ = [_N,N]
```
kwcckw, line 125: This CN is a mistake? We need CL for links?

frame_2D_alg/agg_recursion.py
```diff
@@ -184,7 +185,8 @@ def comp_H(_Nt,Nt, Link):  # tentative pre-comp
     if N.typ==3 and _N.typ==3 and m*wF > ave*(r+cF):
         dH,htt,C,R = comp_H(_N,N, L); c+=C; r+=R; m+= vt_(htt,ttN_)[0]  # tentative dH, replace if subcomp
     if N.typ and m*wN_ > ave*(r+cN_):  # skip PPs
-        F2N(L)  # add tFs for subcomp
+        # F2N(L)  # add tFs for subcomp ( the previous update is not included?)
+        for ft in ('Nt','Lt','Ct','Bt','Xt','Rt'): setattr(L, ft, CF(root=L))
```
kwcckw, line 189: So we need to remove the conversion, and we just need to add Fts to L here.

frame_2D_alg/agg_recursion.py
```diff
@@ -464,14 +460,14 @@ def cluster_P(_C_, _c, root):  # multi-seed mean shift: parallel centroid refine
                 out_ += [C]
     if out_:
         merg_ = []
+        for out in out_: F2N(out)
         for _C, C in combinations(out_,r=2):
             if _C in merg_ or C in merg_: continue
             c = min(C.c,_C.c); r = (C.r+_C.r)/2; dy_dx = _C.yx-C.yx; dist = np.hypot(*dy_dx)
             L = comp_N(_C,C,r,c, A=dy_dx,span=dist)
-            if vt_(L.dTT,ttF)[0]*wF < avd*(r+cF): add2F(_C,C,1); merg_ += [C]
+            if vt_(L.dTT,ttF)[0]*wF < avd*(r+cF): add2F(_C,C,1); merg_ += [C]  # we need to reset their rim for the next cross_comp?
```
kwcckw, line 468: After the comp, i think we need to remove their rim for the next cross_comp again?

frame_2D_alg/agg_recursion.py
```diff
@@ -464,14 +460,14 @@ def cluster_P(_C_, _c, root):  # multi-seed mean shift: parallel centroid refine
                 out_ += [C]
     if out_:
         merg_ = []
+        for out in out_: F2N(out)
```
kwcckw, line 463: Sorry, this shouldn't be here, it should be added to Cs, right before the cross_comp of Ct.

frame_2D_alg/meta_code.py
```diff
@@ -161,6 +162,36 @@ def inner(*a, **kw):
...
+                            TT += np.array([ np.array([mM, mD, mL, mI, mG, mA, mL, mL / 2, 0]),  # extA=0
+                                             np.array([dM, dD, dL, dI, dG, dA, dL, dL / 2, 0])])
+                            c += c
+                    elif not isinstance(out, CF):  # edge, from slice_edge
+                        I, G, Dy, Dx, Y, X = out.latuple
+                        # we only have kern here, not sure how to compute delta TT
+                        out.c = len(out.P_)
+                    else:    
+                        N_ += [out]; TT, c, r = sum_vt(N_)  # CF or edge from slice_edge
+                oF.dTT = (TT*c - _TT*_c) / eps_(_TT*_c)  # why dTT is current - root, while c is root - current?
```
kwcckw, line 193: I added the process to add dTT with delta, where delta is `(current.dTT - root.dTT)/root.dTT`. Also why the original code uses `(current.TT - root.TT)`, but c is other way around and using `root.c - current.c`?

frame_2D_alg/agg_recursion.py
boris-kz: we don't need rim for centroids anyway?

    if N.typ==3:
        for n, _n in (_N,N),(N,_N):

frame_2D_alg/agg_recursion.py
```diff
@@ -309,14 +309,15 @@ def nt_vt(n,_n):
...
         for L in G.L_:
-            for tFt in L.Nt,L.Bt,L.Ct:  # Lt doesn't form trans-links, Ct is not root-constrained?
-                for tL in tFt.N_:
-                    if tL.m*wcN> ave*ccN:  # merge trans_link.N_.roots
-                        rt0 = getattr(tL.N_[0].root,'root',None); rt1 = getattr(tL.N_[1].root,'root',None)  # CNs
-                        if rt0 and rt1 and rt0 != rt1:
-                            if rt1.H: add_H(rt0.H, rt1.H, rt0, fN=1)
-                            rt0.N_ += rt1.N_; add_Nt(rt0)  # recompute Nt attrs / G
-            L.Nt,L.Bt,L.Ct = CF(),CF(),CF()
+            if hasattr(L, 'Nt'):  # has subcomp
```
boris-kz, line 312: what's the point, each Ft is evaluated separately?

frame_2D_alg/agg_recursion.py
boris-kz: right

frame_2D_alg/agg_recursion.py
boris-kz: right

frame_2D_alg/meta_code.py
boris-kz: should be out - input: oF-_oF?

frame_2D_alg/agg_recursion.py
```diff
@@ -121,7 +121,8 @@ def proj_V(_N,N, dist, dy_dx, dec):  # _N x N induction
                 Link = comp_N(_N,N, lr,c,full=not tnF, A=dy_dx, span=dist, rL=root,L_=L_,N_=N_, acc=acc)
                 Link.rTT = np.abs(pTT-Link.dTT) / eps_(Link.dTT)  # relative prediction error to fit oF, direction-agnostic
             else:
-                pL = CN(typ=-1, dTT=pTT,m=m,d=d,c=c,r=lr, angl=[dy_dx,1],span=dist); pL.N_ = [_N,N]
+                # why this is CN here?
+                pL = CL(typ=-1, dTT=pTT,m=m,d=d,c=c,r=lr, angl=[dy_dx,1],span=dist); pL.N_ = [_N,N]
                 L_+= [pL]; N.rim+=[pL]; _N.rim += [pL]; N_+=pL.N_; acc[0]+=pTT*pL.c; acc[1]+=pL.c; acc[2]+=pL.r*pL.c  # all +ve
```
boris-kz, line 126: oF needs  +- Ls

frame_2D_alg/agg_recursion.py
boris-kz: this doesn't depend on cross_comp, why do Cts need to be CNs?

frame_2D_alg/agg_recursion.py
```diff
@@ -464,14 +460,14 @@ def cluster_P(_C_, _c, root):  # multi-seed mean shift: parallel centroid refine
...
         merg_ = []
+        for out in out_: F2N(out)
         for _C, C in combinations(out_,r=2):
             if _C in merg_ or C in merg_: continue
             c = min(C.c,_C.c); r = (C.r+_C.r)/2; dy_dx = _C.yx-C.yx; dist = np.hypot(*dy_dx)
             L = comp_N(_C,C,r,c, A=dy_dx,span=dist)
-            if vt_(L.dTT,ttF)[0]*wF < avd*(r+cF): add2F(_C,C,1); merg_ += [C]
+            if vt_(L.dTT,ttF)[0]*wF < avd*(r+cF): add2F(_C,C,1); merg_ += [C]  # we need to reset their rim for the next cross_comp?
         out_= list(set(out_)-set(merg_))
         dCt = sum2F(list(set(_C_)-set(out_)))  # compress
```
boris-kz, line 470: dCt should be the out_ for CoF?

kwcckw: Okay right, i guess we need another separate section to get dCt from the function output in the traced later.

frame_2D_alg/agg_recursion.py
kwcckw: This pos L could be identified with their typ? Why we need to use CN? Because this is conflicting with the L since L should separate their N_ and Nt.N_.

frame_2D_alg/agg_recursion.py
kwcckw: Yes, but we add empty tFts only if the sub eval is true, so this `for tFt in L.Nt,L.Bt,L.C` gets syntax error when there's no sub section.

frame_2D_alg/meta_code.py
kwcckw: I see, so that was a mistake. So  it should be `oF.c   = c - _c`.

Also for slice_edge, we only have kern, so they shouldn't have this delta dTT?

frame_2D_alg/agg_recursion.py
kwcckw: Okay right. Sorry i was confused earlier.

It should be a CC to enable fC comp and clustering.

frame_2D_alg/agg_recursion.py
kwcckw: We need it in comp_C_?

      def comp_C_(C_,_r, _C_=[], fall=1, fC=0):  # simplified for centroids, trans-N_s, levels
      
          N_,L_,tc,tr = [],[],0,0; acc = [np.zeros((2,9)),0,0,np.zeros((2,9)),0,0]
          if fall:
              for _C, C in pairs:
                  tc += c; tr += r*c
                  if _C is C:
                  else:
              if fC: # merge similar distant centroids
                  for i, L in enumerate(L_):
                          C0,C1 = L.N_
                          add2F(C0,C1,1); merg_ += [C1]
                          for l in C1.rim: l.N_ = [C0 if n is C1 else n for n in l.N_]
                      else: L_ = L_[i:]; break

We add their rim in comp_N above, and rim is used to pack l.N_. I guess we need another flag to skip their rim if it's called from cluster_P?

frame_2D_alg/agg_recursion.py
boris-kz: the data in oF must include Ls of both signs: all results

frame_2D_alg/agg_recursion.py
boris-kz: this is trans_cluster(G), all Ls should have Fts here

frame_2D_alg/agg_recursion.py
boris-kz: it should be C1.N_ instead of C1.rim, get the Ns directly

kwcckw: Here we are merging C1 to C0, so all rim's nt should be replacing C1 with C0?

frame_2D_alg/agg_recursion.py
kwcckw: This L is from G.L_, since trans_cluster is default on all Gs, but not all L in G has subcomp?

    def trans_cluster(G):
        for L in G.L_:
        
 So if L doesn't have sub comp:
 
     if N.typ and m*wN_ > ave*(r+cN_):  # skip PPs
        for ft in ('Nt','Lt','Ct','Bt','Xt','Rt'): setattr(L, ft, CF(root=L))
        
 They won't have added Fts, as above.

frame_2D_alg/agg_recursion.py
kwcckw: Yes, my initial question was on the CN, but it should be CL instead.

frame_2D_alg/agg_recursion.py
boris-kz: > it should be C1.N_ instead of C1.rim, get the Ns directly

frame_2D_alg/agg_recursion.py
boris-kz: ok right

frame_2D_alg/agg_recursion.py
boris-kz: Going back to my prior scheme, L Fts are L.N_ - mediated so they should be summed in L.H[0], etc.
So we need default L.H, not empty if sub_comp

boris-kz: if N.typ and m* wN_*c > ave*(r+cN_):  # skip PPs
        dn_ = []  # cross_comp N_|Ft_ -> top dLev
        if N.typ == 1:
        else:
        if dn_:
            L.H = sum2F(dn_,L)  # top dLev, dnH s added internally
            if len(L.H)==1 and _N.H and N.H: comp_H(_N.H,N.H,L)  # cheap version if no deeper comp_N

boris-kz: sum2H, then the two lines it replaces:

```python
def sum2H(N_, root, fmemb=0, ffold=0):  # build level-H: concat children's H, top lev from N_ or members'
    H = []
    for n in N_:
        if getattr(n,'H',None): add_H(H, n.H, root)
    top = [n for N in N_ for n in N.N_] if fmemb else N_
    H += [sum2F(top, root if ffold else None)]
    return H
```

In sum2F typ==3, replace lines 524-525:

```python
        if typ==3:  # frame?
            F.H = sum2H(N_, F, fmemb=1, ffold=1)  # was: add_H loop + F.H += [sum2F(members)]
```

In comp_N, the sub_ block:

```python
    if sub_:
        L.H = sum2H(sub_, L)  # ffold=0: top lev off L.dTT, tracing intact
```

Behavior-identical in sum2F (fmemb=1 → members'-members top lev, ffold=1 → folds into F as the old `sum2F(...,F)` did). The recursion is benign — sum2H's inner sum2F gets a flat node list, hits typ-based branches, won't re-enter typ==3 unless those members are themselves typ-3 (frames-of-frames), which is the same recursion the old inline had.

One thing the swap exposes that the inline hid: old `F.H +=` *appended* to F.H, assuming F.H was `[]` (CL/CF init). sum2H *assigns* `F.H =`. If F ever reaches this with a pre-existing H (it shouldn't at construction, but Copy_ paths set H), assignment drops it. At sum2F's construction point F is fresh from `cls_[typ](...)` → H=[], so identical. Just don't reuse sum2H on an already-populated F expecting append — it replaces.

frame_2D_alg/meta_code.py
```diff
@@ -161,6 +162,36 @@ def inner(*a, **kw):
...
+                    else:                              
+                        N_ = out[0]
+                        if N_: TT, c, r = sum_vt(N_) 
+                elif isinstance(out, set):  # get_exemplar
+                    N_ = list(out)
+                    if N_: TT, c, r = sum_vt(N_) 
+                else:    
+                    if isinstance(out, list):  # PPt_, from comp_slice
+                        N_ = out
+                        for PPt in out:
```
boris-kz, line 181: this is a monstrosity

frame_2D_alg/agg_recursion.py
kwcckw: So if we pack them as H here, the top should skip those tFts and sum from their tF.N_?

frame_2D_alg/agg_recursion.py
kwcckw: You are referring to this line right? 

    for l in C1.rim: l.N_ = [C0 if n is C1 else n for n in l.N_]

This line is actually removing C1 from l.nt after the merging, so we can't use the same C1.N_ across all links. This l.N_ is actually the nt instead of their N_.

frame_2D_alg/meta_code.py
kwcckw: This section is actually same as PP2N, where we need the TT conversion first. It's the special case for comp_slice, unless we edit comp_slice to output a more consistent format. Or do you see a better method here?

boris-kz: anything is better than putting it here

frame_2D_alg/agg_recursion.py
boris-kz: I think tFts in H[0] are fine, F.N_ are mostly empty anyway

frame_2D_alg/agg_recursion.py
boris-kz: Forget links, this will be removing C and m,d from N.root_,m_,d_
and we don't need to go through C.N_, it's just C index in C_, they all map

frame_2D_alg/meta_code.py
boris-kz: they all have vals.

dc = _c-c; oF.dTT = (_TT*_c-TT*c) / dc or eps

boris-kz: this is on the version I just pushed:

Read the snippet in context (traced, lines 164–179) plus `sum_vt` and `sum2O`. The mechanics mostly work; the conceptual problems are about *what `oF.dTT`/`oF.c` are supposed to mean*.

**1. Time inversion — the baseline is read before it exists.** `_TT,_c = _CoF.dTT,_CoF.c` is the *caller's* aggregate. But `_CoF.dTT` is only assigned when `_CoF`'s own frame returns (this same line 173), which hasn't happened yet — the caller is still mid-execution, you're inside one of its children. So `_TT,_c` is the caller's stale value from a *previous* invocation or the init zeros. The whole subtraction `(_TT*_c − TT*c)/dc` is therefore against an undefined baseline. You can't compute a parent-minus-child decomposition from inside the child; the parent total isn't known until all its children return.

**2. Wrong baseline even if timing worked.** The caller's running total is not this function's *input*. If `dTT` is meant to be marginal gain (output − input), the baseline is this frame's args captured at entry (line 162–163), not the caller's accumulated dTT. "Caller" and "input" are being conflated.

**3. `oF.c = dc` is a type confusion.** Everywhere else `c` is a positive extent/normalizer (`w = n.c/C` in sum_vt). Here you redefine it as a count *difference*, which:
- goes negative whenever output extent > caller's current extent (normal early), giving negative cost,
- poisons every downstream weighted sum that includes this oF,
- `or eps` only guards exact zero, not small/negative.

**4. `r` is extracted but never stored.** Lines 168/169/171 unpack `r`, but `oF.r` is never assigned — yet line 176 uses `cent_TT(wtt, oF.r)`. The output's redundancy is silently dropped.

**5. Unhandled output shapes fail silently.** A bare returned list (not tuple/set) matches neither branch, so `TT,c=0`, and `oF.dTT = _TT*_c/_c = _TT` — it just copies the (stale) parent. No error.

The real decision your `# all this should be computed for the output?` comment is circling: **decide whether `oF.dTT` is *produced value* or *marginal gain*.**

- Produced value → drop the subtraction entirely: `oF.dTT, oF.c, oF.r = sum_vt(N_)`. Clean, and the marginal-vs-parent comparison belongs in the *caller* after it aggregates children (where it actually has both totals).
- Marginal gain → capture the input baseline from args at entry, subtract that, and keep `c` as a real extent (don't overwrite it with `dc`).

Either way the caller-read baseline has to go. Which semantics do you intend for `dTT` here — the value the function emitted, or its increment over input?

kwcckw: > The whole subtraction (_TT*_c − TT*c)/dc is therefore against an undefined baseline. 
You can't compute a parent-minus-child decomposition from inside the child; the parent total isn't known until all its children return.

I think we shouldn't use the parent total since we need to get the delta at the line we call the children?

frame_2D_alg/agg_recursion.py
kwcckw: Okay, but why we need fmemb? They could be sum and merged the same way?

frame_2D_alg/agg_recursion.py
kwcckw: Yes, so those are still missing above. It's just summing C1 to C0, and we need to update their N's m_ , d_ and root_ too.

PR #175, opened 2026-06-14
6/14/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -146,7 +146,10 @@ def comp_C_(C_,_r, _C_=[], fall=1, fC=0):  # simplified for centroids, trans-N_s
                     C0,C1 = L.N_
                     if C0 is C1 or C1 in merg_ or C0 in merg_: continue  # not merged
                     add2F(C0,C1,1); merg_ += [C1]
-                    for l in C1.rim: l.N_ = [C0 if n is C1 else n for n in l.N_]
+                    for l in C1.rim: 
+                        l.N_ = [C0 if n is C1 else n for n in l.N_]  # replace nt
+                        if l not in C0.rim: C0.rim += [l]  # merge rim
+                    for n in C1.N_: n.root_[n.root_.index(C1)] = C0  # the position of m and d should remain the same, and only root_ needs to be updated
```
kwcckw, line 152: So for the C merging, we only need to update n.root_ to the new C, and their m_, and d_ should be remained the same since the position of n's root C doesn't change.

For rim, other than replacing the merged C's rim.nt, we should merge their rim too? Because we will need it in cluster_C later:

            if M*(_C.m+wC_*len(n_)) > Ave*(r+_C.r+ cC_*len(n_)):  # else: Up+= sum(_C._m_)+ sum([abs(d) for d in _C._d_])?
                    N__ += [n]  # +|-Ls

frame_2D_alg/meta_code.py
```diff
@@ -162,15 +163,20 @@ def inner(*a, **kw):
...
+                    elif isinstance(out[0], np.ndarray): TT,_,N_,c,r = out  # for proj_N
                     elif N_ := out[0]: TT,c,r = sum_vt(N_)  # default
                 elif isinstance(out, set):
                     if N_:= list(out): TT,c,r = sum_vt(N_)  # get_exemplar
+                elif isinstance(out, CF):  # frame_H, vect_edge, comp_N, comp_F
+                    if func.__name__ == 'vect_edge': N_, TT, c, r = out.N_, out.Nt.dTT, out.Nt.c, out.Nt.r  # get from tile.Nt instead
+                    else: N_, TT, c, r  = [out], out.dTT, out.c, out.r
                 # all this should be computed for the output?
-                dc = (_c-c) or eps; oF.c = dc; oF.dTT = (_TT*_c-TT*c) / dc
+                dc = (c-_c) or eps; oF.c = dc; oF.r = r; oF.dTT = (TT*c-_TT*_c) / dc  # should be out - root?
```
kwcckw, line 179: > they all have vals.
> dc = _c-c; oF.dTT = (_TT*_c-TT*c) / dc or eps

Actually the delta should be output - input? Why it's the other way around here?

frame_2D_alg/agg_recursion.py
boris-kz: >the position of n's root C doesn't change

m_, d_, root_ all map 1/1 and change together

frame_2D_alg/meta_code.py
boris-kz: this is compression, output is compressed / selected input

frame_2D_alg/agg_recursion.py
kwcckw: Yes, but we don't have the new m or d value of merged C0 in the C1's N.m_ and N.d_? That m or d should be recompute using base comp again?

frame_2D_alg/meta_code.py
kwcckw: Please correct me if I misunderstood this. So if output is compressed, input - output (compressed) = what was compressed, and we need to pack this to oF?

boris-kz: yes, that's in terms of c, which represents raw volume.
m, d, r don't actually need to be subtracted from the input, they are the computed values, *= dc

frame_2D_alg/agg_recursion.py
boris-kz: yes, but only if merged C is significantly different from prior C0

frame_2D_alg/meta_code.py
```diff
@@ -162,15 +163,20 @@ def inner(*a, **kw):
             _oF = CoF._cur.set(oF)
             out = func(*a, **kw)
             if out:
+                # comp_slice and slice_edge are added from the function body?
```
boris-kz, line 166: it shouldn't matter, the top return should be converted to the same TT,c,r in the body of each function

frame_2D_alg/agg_recursion.py
kwcckw: To identify the difference, it should be using vt_ of dTT?

    _TT = C0.dTT

Then after the merge, using d from vt_?

     vt_(C0.dTT - _TT)[1] > ave

frame_2D_alg/agg_recursion.py
```diff
@@ -207,7 +210,7 @@ def comp_H(_Nt,Nt, Link):  # tentative pre-comp
 def comp_F(_F, F, ir=0, rL=None):
 
     ddTT = comp_derT(_F.dTT[1],F.dTT[1]); r=(_F.r+F.r)/2; m,d = vt_(ddTT,ttF); r+=ir
-    dF = CF(dTT=ddTT, m=m,d=d,r=r,c=min(_F.c,F.c),nF=F.nF)
+    dF = CF(dTT=ddTT, m=m,d=d,r=r,c=min(_F.c,F.c),nF=F.nF); dF.H = []
```
boris-kz, line 213: I guess we can put H in CF too

frame_2D_alg/agg_recursion.py
boris-kz: that's only after C is recomputed from C.N_.
anyway, I think it's a lot simpler:

        for _C, C in pairs:
            tc += c; tr += r*c
            if _C is C:
            else:
                L = comp_N(_C,C,_r,c, A=dy_dx,span=dist,L_=L_,N_=N_,acc=acc)
                if fC and L.m*wF > ave*(L.r+cF): 
                    add2F(_C,C,1); mrg_ += [C]  # may not already be merged
                    for n in C.N_: n.root_[n.root_.index(C)] = _C  # keep m,d positions

frame_2D_alg/agg_recursion.py
```diff
@@ -185,7 +188,7 @@ def comp_H(_Nt,Nt, Link):  # tentative pre-comp
             for i,(_Ft,Ft, tnF) in enumerate(zip((_N.Nt,_N.Lt,_N.Bt,_N.Ct),(N.Nt,N.Lt,N.Bt,N.Ct),('Nt','Lt','Bt','Ct'))):
                 if _Ft and Ft: dn_ += [comp_F(_Ft,Ft,r,L)]; r+=(i or 1)-1  # unique Nt,Lt, rL spec in comp_F
         if dn_:
-            [add_H(L.H, d.H, L) for d in dn_ if d.H]
+            [add_H(L.H, d.H, L) for d in dn_ if d.H]  # mostly empty here, except agg+++
```
boris-kz, line 191: [add_H(L.H, d.H, L) for d in dn_ if d.H]  # lower levs
            if not L.H and _N.H and N.H: comp_H(_N.H, N.H, L)  # cheap version if no deeper comp_N?
            L.H += [sum2F(dn_,None)]  # top lev

frame_2D_alg/agg_recursion.py
```diff
@@ -483,6 +486,7 @@ def sum2F(N_, root=None, m_=[],d_=[], merge=0, froot=0, nF=None):  # -> CF/CL/CC
                 if typ==3: box=copy(n.box)
     F = (cls_[typ])(c=C, r=R, nF=nF)
     if isinstance(F, CN): F.Nt.dTT = copy(TT); F.Nt.c = C; F.Nt.r = R  # update Nt's params
+    else: F.dTT = copy(TT); F.c = C; F.r = R  # this is missed out?
```
boris-kz, line 489: F attrs are set above, Nt is actually redundant here:

    if typ==3: F.Nt.dTT = copy(TT); F.Nt.c = C; F.Nt.r = R  # redundant?

frame_2D_alg/meta_code.py
boris-kz, line 94: we don't need fb_ anymore, it's just tFt.N_?

frame_2D_alg/meta_code.py
```diff
@@ -162,15 +163,20 @@ def inner(*a, **kw):
...
                 _TT,_c = (_CoF.dTT,_CoF.c) if _CoF is not None else (np.zeros((2,9)),0)
                 N_,TT,c,r = [],np.zeros((2,9)),0,1
                 if isinstance(out, tuple):
```
boris-kz, line 169: so none of that is relevant, N_,TT,c,r should be computed and returned by the function

            if out := func(*a, **kw): oF.N_, oF.TT, oF.c, oF.r = out

frame_2D_alg/agg_recursion.py
boris-kz, line 443: We need to add eval for centroid split and merge per refine cycle. Merge is  partially implemented, but that means _C_ is no longer fixed, we need to revise the code accordingly

boris-kz: Read enough. Let me look at the merge test in `comp_C_` to confirm the intended sign, since the sketch in `cluster_P` has it inverted.Confirmed — merge fires on **high** match (`>`), so the sketch's `<` was inverted. Applying the revision:The revised region:Done. The conceptual shape of the revision:

**Why variable `_C_` propagates.** `md__` is positionally indexed `(N, C)`. The moment count changes, `Lc`, `L`, the matrix shape, and the convergence subtraction can't be loop-invariant — all moved inside the cycle.

**Split/merge as re-seeding, not column surgery.** The eval runs on the freshly mean-shifted `C_` and emits `nC_` = next cycle's seeds. Within a cycle `md__ ↔ C_` stays 1:1, so the final pruning block is untouched and correct; the next cycle recomputes membership fresh against the new seed set. Splicing columns mid-matrix would tangle indices for no gain — mean-shift re-derives everything anyway.

**Split/merge are complementary and self-settling.** Split bisects a high-`d` centroid by the sign of each member's diff (`md__[:,i,1]`) — side of the principal difference axis, deterministic, no random seeds (consistent with no-random-init). The two offset seeds either separate next cycle (real substructure) or the merge test folds them back (spurious). So the centroid count converges on its own; you don't preset it.

**Convergence under changing count.** Element-wise membership-shift is only meaningful at fixed shape — guarded, so a shape change forces another cycle. Once split/merge stop firing, shapes match and your original convergence resumes. Break requires `cnv and not chg`: membership-stable *and* no pending split/merge. Without the second clause, a redundant-but-stable state would falsely converge.

**Overlap `O`** = per-N (total membership − best membership), summed. It's the global-overlap signal that justifies the fully-fuzzy mode (docstring L26), now actually computed — your `if O:` was dead, `O` was never set.

**Three bugs in the sketch, fixed:**
- merge sign was inverted (`<` merged *dissimilar* Cs); corrected to `>` per `comp_C_` L143.
- `L = comp_N(...)` shadowed `L = Lc*Ln`; link renamed `lk`.
- `O` gate was unreachable.

**Left alone (flagging, not fixing):** caller L397 `out_ = cluster_P(...)` discards the returned `(out_, dCt)` tuple — it'll bind the tuple, not the list. And `merge`-then-`split` on the same C in one pass uses pre-merge `d`; harmless since the next cycle recomputes.

One open question for you: split currently fires on `C.d > avd*(...)*2` independent of `O` magnitude (only the pass is O-gated). If you want split strictly proportional to where the overlap concentrates, the criterion should read per-centroid ambiguity (column sum of non-max memberships) rather than `C.d`. Which do you intend — internal spread, or overlap mass?

frame_2D_alg/agg_recursion.py
kwcckw: Okay, but i think we need to merge their rim too? We still need the rim to get the frontier n in cluster_C later?

frame_2D_alg/agg_recursion.py
kwcckw: So the root of L.H level should be L?

frame_2D_alg/meta_code.py
kwcckw: Yes, actually it's nested in dn_'s N_ now before we sum it into H lev's N_'s N_.

frame_2D_alg/meta_code.py
kwcckw: That requires us to change every function's output format?

frame_2D_alg/agg_recursion.py
kwcckw: > We need to add eval for centroid split and merge per refine cycle. Merge is partially implemented, but that means _C_ is no longer fixed, we need to revise the code accordingly

Okay, that's similar with oF right? So after the merging, each C will be eval to be split into multiple Cs again?

frame_2D_alg/agg_recursion.py
boris-kz: the Cs don't need rim

frame_2D_alg/agg_recursion.py
boris-kz: right

frame_2D_alg/agg_recursion.py
boris-kz: no, first split then merge. something like this:

    cnt = 0; Ln = len(N_:= list(set([N for C in _C_ for N in C.N_])))  # all Ns are in all Cs
    _md__ = np.zeros((Ln, len(_C_), 2))  # NxC, cols aligned to _C_
    for j, N in enumerate(N_):
        for c,m,d in zip(N.root_,N.m_,N.d_):
            if c in _C_: _md__[j,_C_.index(c)] = m,d
    while True:
        Lc = len(_C_); L = Lc*Ln  # Lc,L,md__ per cycle since _C_ varies
        md__ = np.zeros((Ln,Lc,2)); O = 0
        for j,N in enumerate(N_):
            for i,C in enumerate(_C_): md__[j,i] = vt_(base_comp(C,N)[0], ttcP)
        conv = md__.shape==_md__.shape and Mt* np.abs(md__-_md__).sum()* (wcP*L) <= ave*(root.r+ccP*L)  # convergence
        nC_, updt = C_,0
        if O*wcP > ave*(root.r+ccP*L):  # redundancy|structure: revise count
            mrg_,new_ = [],[]
            for i,C in enumerate(C_):
                if C in mrg_: continue
                for _C in C_[i+1:]:  # merge redundant (similar) Cs
                    if _C in mrg_: continue
                    lk = comp_N(C,_C,(C.r+_C.r)/2, min(C.c,_C.c), A=(a:=_C.yx-C.yx), span=np.hypot(*a))
                    if lk.m*wF > ave*(lk.r+cF): add2F(C,_C,1); mrg_+=[_C]
                if C.d * wF > avd * (C.r+cF)*2:  # split C spanning 2 diff-modes, higher filter
                    mpos = md__[:,i,1]>0  # member diff sign = side of principal axis
                    for mask in (mpos,~mpos):
                        sub = [N_[k] for k in np.where(mask)[0]]
                        if sub: new_+= [sum2F(sub,root, md__[mask,i,0],md__[mask,i,1])]
                    mrg_+=[C]
            updt = bool(mrg_ or new_); nC_ = list(set(C_)-set(mrg_))+ new_
        cnt += 1
        if conv and not updt: break  # stable memberships & count: C_,md__ used below
        _C_,_md__ = nC_,md__
    out_ = []
    for i, _C in enumerate(C_):
            N_,m_,d_ = [],[],[]
            if N_:
                C = sum2F(N_,root, m_,d_)
                for N in N_:
                    L = CN(typ=1, dTT=N.dTT,c=N.c,r=N.r,m=N.m,d=N.d, span=np.hypot(*(dy_dx:=C.yx-N.yx)), angl=[dy_dx,np.sign(N.dTT[1]@ttcN[1])])
                    L.N_ = [N,C]; C.L_ += [L]
                out_ += [C]
    if out_:
        dCt = sum2F(list(set(_C_)-set(out_)))  # compress, out_ for CoF?
        return out_, dCt

frame_2D_alg/meta_code.py
boris-kz: yes, could you do that?

frame_2D_alg/agg_recursion.py
kwcckw: Previously we use rim to get the frontier:

            if M*(_C.m+wC_*len(n_)) > Ave*(r+_C.r+ cC_*len(n_)):  # else: Up+= sum(_C._m_)+ sum([abs(d) for d in _C._d_])?
                    N__ += [n]  # +|-Ls
                C = sum2F(n_, Ft,m_,d_)  # C.N_ = n_
                if D<Avd: out_+=[C]  # output if stable
                else:     C_ += [C]  # reform
                DTT+=dTT; mat+=M; dif+=D; cnt+=T; rdn+=R; Up+=up
                
Without rim, how do we really get the new frontier for the next iteration?

boris-kz: this is per N, not C, and Ns are not affected?

frame_2D_alg/agg_recursion.py
kwcckw: From the code:

            for _C in C_[i+1:]:  # merge redundant (similar) Cs
                if _C in mrg_: continue
                lk = comp_N(C,_C,(C.r+_C.r)/2, min(C.c,_C.c), A=(a:=_C.yx-C.yx), span=np.hypot(*a))
                if lk.m*wF > ave*(lk.r+cF): add2F(C,_C,1); mrg_+=[_C]
            if C.d * wF > avd * (C.r+cF)*2:  # split C spanning 2 diff-modes, higher filter
                mpos = md__[:,i,1]>0  # member diff sign = side of principal axis
                for mask in (mpos,~mpos):
                    sub = [N_[k] for k in np.where(mask)[0]]
                    if sub: new_+= [sum2F(sub,root, md__[mask,i,0],md__[mask,i,1])]
                mrg_+=[C]
                
This is actually merge and split within the same step, but from this line:

    list(set(C_)-set(mrg_))+ new_

This is packing both the merged and split Cs?

frame_2D_alg/meta_code.py
kwcckw: Alright, i will change that next.

frame_2D_alg/agg_recursion.py
kwcckw: But `C._N_` is retrieved from N.rim:

          N__ += [n]  # +|-Ls
      
At this point i don't see any other method yet, other than retrieve it from rim.

boris-kz: N.rim is not affected?

frame_2D_alg/agg_recursion.py
boris-kz: Yes, but there is no splitting of merged Cs

kwcckw: I see, the code is quite difficult to read now. Let me step through them first.

frame_2D_alg/agg_recursion.py
kwcckw: When we cross_comp(Ct): comp_C -> cluster_N -> sum2G-> cluster_C, it's still using the same C.rim from Ct's C_, added from comp_C?

boris-kz: that's a new rim, it has nothing to do with cluster_C

PR #176, opened 2026-06-15
6/15/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -154,7 +154,7 @@ def inner(*a, **kw):
                 _CoF.call_ += [oF]
                 oF_[iF_[func.__name__]].caller_.add(_CoF)  # for comp_caller_
             _oF = CoF._cur.set(oF)
-            if out := func(*a, **kw): oF.N_, oF.TT, oF.c, oF.r = out
+            if out := func(*a, **kw): *_, oF.N_, oF.dTT, oF.c, oF.r = out  # should be dTT?
```
kwcckw: So i added *_ here because we might still need to output some additional params in some functions, such as cluster_P:

    return out_, [dCt], dCt.dTT, dCt.c, dCt.r

frame_2D_alg/agg_recursion.py
```diff
@@ -433,11 +434,11 @@ def cluster_P(_C_, _c, root):  # multi-seed mean shift: parallel centroid refine
                 if C in mrg_: continue
                 for _C in C_[i+1:]:  # merge redundant (similar) Cs
                     if _C in mrg_: continue
-                    lk = comp_N(C,_C,(C.r+_C.r)/2, min(C.c,_C.c), A=(a:=_C.yx-C.yx), span=np.hypot(*a))
+                    lk = comp_N(C,_C,(C.r+_C.r)/2, min(C.c,_C.c), A=(a:=_C.yx-C.yx), span=np.hypot(*a))[0][0]
                     if lk.m*wF > ave*(lk.r+cF): add2F(C,_C,1); mrg_+=[_C]
                 if C.d * wF > avd * (C.r+cF)*2:  # split C spanning 2 diff-modes, higher filter
                     mpos = md__[:,i,1]>0  # member diff sign = side of principal axis
-                    for mask in (mpos,~mpos):
+                    for mask in (mpos,~mpos):  # max 2 split C per iteration
```
kwcckw: So in this scheme, each C is split into 2 subs here. But if one of the subs doesn't have positive md__, there's just single new sub, which mean we can remain the original C?

boris-kz: I am not really sure we need this splitting, the centroids are refined anyway

frame_2D_alg/meta_code.py
boris-kz: ok thanks

frame_2D_alg/agg_recursion.py
```diff
@@ -880,7 +881,7 @@ def ffeedback(frame, aTT,oTT):  # recompute filters from regime drift; fork: ref
         if oH := pack_seg(frame,'oH', wBac, cBac**2, oTT):
             split_oF_(); cluster_oF_()  # reform oF_
             oTT = oH.dTT
-    return frame, aTT,oTT  # frame in oF.N_?
+    return aTT,oTT, [frame], aTT+oTT, frame.c, frame.r  # frame in oF.N_? (not sure, sum Tts here?)
```
boris-kz: right, but c and r should be from updates only too

frame_2D_alg/agg_recursion.py
kwcckw: Actually the splitting should be similar with sub+, but it's involves only the weight reassignment?

frame_2D_alg/agg_recursion.py
kwcckw: So the c and r should be from oH and aH?

boris-kz: only updated oFs, not sure how to account for filters

frame_2D_alg/agg_recursion.py
boris-kz: It has to start from new seeds: Ns that are stradling multiple centroids with about equal high ms in N.m_
but then it's not really splitting, just starting new global Cs

boris-kz: Claude:
A seed placed at a straddling N is grounded in a real node, so it can correspond to density the seed set actually missed. Variance-bisection fabricated a centroid from an existing one's spread and never corresponded to new structure. Those aren't the same operation and I was wrong to lump them — variance-split stays rejected, node re-seeding doesn't.

Reconciling with my "count only goes down": merge (down) is always needed since duplicate seeds are a fixed point. Up is needed only when initial seeding under-covers — which is exactly cluster_P's regime. It runs *because* overlap is high, and high overlap is precisely where exemplar selection suppresses adjacent peaks. So node re-seeding here isn't a new operator; it's exemplar selection reapplied to the region cluster_P exists for — recovering exemplars overlap suppressed.

The catch is your criterion. Balanced-high `N.m_` is necessary but not sufficient. A node between two centroids has balanced-high m whether it sits on a missed peak (seed it) or in the valley between two real modes (don't — it drifts to a mode and merges, or sits in the valley and prunes: churn). Balanced-m identifies "between," it can't separate peak from valley. The discriminator is local density: is N a maximum that its straddled centroids don't sit on.

Merge+prune make "seed every straddler" *correctness*-safe — saddle seeds die. But in high overlap straddlers line whole ridges, so you'd seed a ridge and then mass-merge: an efficiency hole, not a correctness one. So you want the local-max straddler per overlap region, not all of them.

One concrete wrinkle if you reach for `n.exe` as the peak signal: it's set in cluster_C *after* cluster_P returns (`n.exe = ... + np.sum(n.m_) - ave`), so mid-refine it's stale or unset. Worse, it *adds* `Σn.m_`, so it rewards straddling — it would favor exactly the nodes you can't yet distinguish. The density signal has to come from N's own match against its rim's match to the straddled Cs, computed inside the loop, or you accept the churn.

frame_2D_alg/agg_recursion.py
```diff
@@ -175,14 +175,15 @@ def comp_H(_Nt,Nt, Link):  # tentative pre-comp
     if N.typ and m* wN_*c > ave*(r+cN_):  # skip PPs
         dn_ = []  # cross_comp N_|Ft_ -> top dLev
         if N.typ == 1:
-            for _n,n in product(_N.N_,N.N_): dn_ += [comp_N(_n,n,r,c, rL=L)]  # CN L.nt, rL spec in comp.N
-        else:
+            for _n,n in product(_N.N_,N.N_): dn_ += comp_N(_n,n,r,c, rL=L)[0]  # CN L.nt, rL spec in comp.N
+        elif N.typ == 3:
+            # skip Cs here? Cs don't have Fts?
```
boris-kz: For Cs it's N_, so it actually works the same as typ=1?

frame_2D_alg/agg_recursion.py
boris-kz: This clustering is not local, N positions don't matter, so local peak or saddle makes no difference?

Claude
Let me check whether `comp_N` (merge) actually uses position, or only `base_comp` features.Correct, and I overcomplicated it. Verified in the code: `base_comp` compares M, D, C, kern (I, G, Dy, Dx), L, span, derT, angle — no `yx`. And `comp_N`'s match is the same thing (line 173, `vt_(base_comp(...))`); its `span`/`A` arguments only populate the resulting link's geometry (lines 186–191), they never enter `m`. So both membership *and* merge are feature-space and position-free. There's no neighborhood metric anywhere in cluster_P, so peak/saddle/ridge are category errors — my discriminator and the rim signal were both wrong, since the rim is spatial adjacency, an unrelated space.

So straddling is the entire signal. Balanced high m to ≥2 Cs means N's feature vector isn't strongly claimed by any centroid — a feature region the set under-covers. That alone warrants a new global C seeded at N; there's no density to filter on.

What I was reaching for with peak-vs-saddle — is there real structure at N or is it a lone point between clusters — is answered by refinement itself, globally: seed C at N, next cycle its column is every node's feature-match to N. Shared structure → C gathers members and survives; lone outlier → little membership, prunes. The test is global gathering, not local inspection. That's your point exactly.

And redundant seeds dissolve on their own: straddlers of the same pair share a feature profile, so their seeds are feature-similar and merge collapses them regardless of `yx`. Merge is the count-control, re-seeding the hypothesis generator, prune the filter — and "ridge" was never in this space to begin with.

frame_2D_alg/agg_recursion.py
kwcckw: > only updated oFs

The updated oFs' params are packed in F_call_T_?

So c and r can be done the same way?

frame_2D_alg/agg_recursion.py
kwcckw: Yes, so i guess this `if N.typ == 1:` should be `if N.typ == 1 or N.typ == 2` .

frame_2D_alg/agg_recursion.py
kwcckw: > It has to start from new seeds: Ns that are stradling multiple centroids with about equal high ms in N.m_
but then it's not really splitting, just starting new global Cs

Okay, so this is actually adding seed in `_C_`, which mean all Ns need to add new element in root_, m_ and d_ for this new seed?

kwcckw: And i think we can have this eval as criterion to prune those Cs? If seed C doesn't have high m on multiple Ns, we should remove them too?

boris-kz: yes, weak Cs. we currently prune them in the end, now it can be between cycles

frame_2D_alg/agg_recursion.py
```diff
@@ -407,8 +408,8 @@ def cluster_C(Ft, E_,_r,_c):  # form centroids by clustering exemplar surround v
             n.exe = (n.d if n.typ==1 else n.m) + np.sum(n.m_) - ave
         if vt_(DTT,Ft.root.wTT*ttcC)[0]*wcC*(len(out_)-1) > ave*(r+ccC*(len(out_)-1)):
             Ct = sum2F(out_); Ft.root.Ct = Ct; Ct.root = Ft.root
-            _,r = cross_comp(Ct,r)  # all distant Cs, seq C_ in eigenvector = argmax(root.wTT)?
-    return out_, r
+            *_,r = cross_comp(Ct,r)  # all distant Cs, seq C_ in eigenvector = argmax(root.wTT)?
+    return out_, *sum_vt(out_)[:-1], r  # only r is from deeper cross_comp?
```
boris-kz: dTT, c, r should be summed from callee returns, not G_? 
Actually it's a return gating value: projected skipped negative returns
In general, oF return = [own_return, gating_vals]?

kwcckw: Those dTT , c and r should be summed from the result, which is G_? 

The callee returns are gating value, and it should be packed separately into gate?

boris-kz: in general, oF return = [own_return, gating_vals]

but even own_return vals are not directly from G_, it's either L_ from comp or compression from clust
cross_comp doesn't compute any vals on it's own, it's the comp_N_ vals because they are not gated

frame_2D_alg/agg_recursion.py
boris-kz: That's after the update, so these are prior oH | aH vals. 
current ffeedback vals should be projected from them

frame_2D_alg/agg_recursion.py
kwcckw: Yes, so there will be overlapping between the vals returned from multiple functions. I guess we can shift to the " oF return = [own_return, gating_vals]" format?

frame_2D_alg/agg_recursion.py
kwcckw: The prior oH or aH values should be values before feedback?

boris-kz: yes, but we only compute them at termination, if ffeedback

kwcckw: That will be using  the input oTT and aTT ?

frame_2D_alg/agg_recursion.py
boris-kz: yes try that, and it's all in return[-1], this is secondary to primary code

kwcckw: The gating vals can be retrieved via oF.calls, which we can retrieve via traced but not the function body right?

boris-kz: it's actually just average negative val that maps to negative filter val that blocks the call.
so we need to return negative evals blocking oF, if any

frame_2D_alg/agg_recursion.py

frame_2D_alg/agg_recursion.py
kwcckw: Alright, thanks. Let me try it later.

frame_2D_alg/agg_recursion.py
```diff
@@ -407,8 +408,8 @@ def cluster_C(Ft, E_,_r,_c):  # form centroids by clustering exemplar surround v
...
 
 def cluster_P(_C_, _c, root):  # multi-seed mean shift: parallel centroid refine, _C_ varies via split/merge
```
boris-kz, line 438: this is on my new push just now:

Two breaking bugs, both structural.

**`new_` is dead.** You build it, never use it. `_C_` is set from the pruned `C_` only; the seeds are discarded every cycle. Reseeding does nothing until `new_` is concatenated into `_C_`.

**Merge mutates `C_`, which desyncs the output.** The output block reads `C_` and `md__` positionally (`zip(_C.N_, md__[:,i,0], md__[:,i,1])`). Two things in this draft break that alignment:
- `C_ = list(set(C_)-set(removed))` — `set()` discards order, so `C_[i]` no longer matches column `md__[:,i]`. Even with order preserved, dropping merged elements shifts indices while `md__` keeps all `Lc` columns.
- `if conv: break` with no guard lets you break on a cycle where merge just shrank/reordered `C_`. Output then runs on a misaligned, set-reordered `C_`.

The principle you keep hitting: **`C_` must stay column-aligned to `md__` for the whole cycle, because the output consumes it positionally. So never reassign `C_` in the loop — all revision (merge-survivors, prune, new seeds) goes into `_C_`, the next-cycle seeds — and break only on a cycle where nothing was revised**, so `_C_ == C_` and the output's `C_`/`md__` are still aligned.

Corrected tail (your names, minimal change):

```python
    removed = []
    if O*wcP > ave*(root.r+ccP*L):  # merge redundant Cs
        for i,_C in enumerate(C_):
            if _C in removed: continue
            for C in C_[i+1:]:
                if C in removed: continue
                if l.m*wF > ave*(l.r+cF): add2F(_C,C,1); removed += [C]
    new_ = [Copy_(N,root,init=1,cls=CC) for j,N in enumerate(N_)
            if np.sort(md__[j,:,0])[-2:].min()*wcP > ave*(N.r+ccP)]  # seed overlap Ns
    _C_ = [c for c in C_ if c not in removed and c.m*wcP > ave*c.r*ccP] + new_  # survive+prune, +seeds
    cnt += 1
    if conv and not (removed or new_) and len(_C_)==Lc: break  # nothing revised; C_,md__ aligned
    _md__ = md__
```

Note `C_` is never reassigned — merge only fills `removed`; survivor/prune/seed all land in `_C_`. The `len(_C_)==Lc` clause catches a prune-only count drop (which leaves `removed`/`new_` empty but still desyncs). On break, `removed` and `new_` empty ⇒ `C_` untouched and aligned ⇒ output valid.

Still unfixed from last turn: `comp_N(...)[0][0]` — `comp_N` returns the link, not a nested sequence; it's `l = comp_N(...)`, `l.m` (corrected above).

And now that `new_` is actually wired, the reseed caveat is live: a seed that merges straight back keeps `removed`/`new_` nonempty on alternating cycles, so `not (removed or new_)` may never hold — bound it with `or cnt > lim`.

PR #177, opened 2026-06-16
6/16/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -65,25 +66,31 @@
...
                 if not root.typ: F2N(root)  # promote @ 1st sub+ or agg+
                 root.H+= [sum2F(L_,root,froot=1)]  # lev: L_+ derivatives
                 root.Nt = sum2F(G_,root,froot=2); L=len(G_)-1  # or C_ s?
-                if vt_(TT,tAgg)[0]*(wAgg*L) > ave* (r+cAgg*L):
+                gAgg = vt_(TT,tAgg)[0]*(wAgg*L) - ave* (r+cAgg*L)
+                if gAgg > 0:
                     G_, *_ = cross_comp(root.Nt,r) # agg+
-    return G_, *sum_vt(G_)  # -> traced, + gating val?
+                else: gates['cross_comp'] = gAgg
+        else: gates['get_exemplars'] = gE
```
kwcckw, line 87: So looks like we need to add gating val in the function body if we need the negative vals that blocks the call because it's not retrieveable within `traced`. From `traced`, we only able to identify which call is blocked via the matching from function body.

frame_2D_alg/agg_recursion.py
```diff
@@ -139,6 +149,7 @@ def comp_C_(C_,_r, _C_=[], fall=1, fC=0):  # simplified for centroids, trans-N_s
                 dtt = np.array([C.dTT[1],np.zeros(9)]); acc[0]+=dtt; acc[1]+=c; acc[2]+=1  # overlap=match
             else:
                 dy_dx = _C.yx-C.yx; dist = np.hypot(*dy_dx)
+                # no gated values here when the func is default?
                 L = comp_N(_C,C,_r,c, A=dy_dx,span=dist,L_=L_,N_=N_,acc=acc)[0][0]
```
kwcckw, line 153: If the function is default, there will be no gating values?

frame_2D_alg/agg_recursion.py
```diff
@@ -806,10 +844,12 @@ def proj_TT(L, cos_d, dist, r, pTT, wTT, dec=1, fdec=0, frec=0):  # accumulate l
         for lev in L.Nt.N_:  # refine pTT
             proj_TT(lev, cos_d, dec, r+1, pTT, wTT, fdec=1, frec=1)
     pTT+=TT  # L.dTT is redundant to H, neither is redundant to Bt,Ct
-    if L.Bt: # + trans-link tNt, tBt, tCt?
-        TT = L.Bt.dTT
-        if TT is not None:
-            pTT += np.array([TT[0] * dec, TT[1] * cos_d * dec])
+    if L.H: # + trans-link tNt, tBt, tCt?
+        for dFt in L.H[-1].N_:
```
kwcckw, line 848: Bt is packed in L.H's N_ now, so we need extra step to retrieve it.

frame_2D_alg/agg_recursion.py
```diff
@@ -868,18 +909,25 @@ def fill_frame(_iy,_ix, elev, T):  # expand level_frame from pixel-level seed ti
...
+    return [Fr], Fr.dTT, Fr.c, Fr.r, {}  # intra-lev feedback
 
 def ffeedback(frame, aTT,oTT):  # recompute filters from regime drift; fork: reform oF_ on cross-regime drift
 
     global ave,avd
+    _aTT, _oTT = copy(aTT), copy(oTT)
     if aH := pack_seg(frame,'aH',wBac, cBac, aTT):
         ave, avd = vt_(aH.dTT)  # filters *= ave
+        '''
+        proj = proj(_aTT)
```
kwcckw, line 921: > That's after the update, so these are prior oH | aH vals.
current ffeedback vals should be projected from them

So the current ffeedback val should be projected from prior vals, and it should be similar with proj_TT? But the prior TT is non directional here, and we don't have span or dist, so how do we really project them here?

boris-kz: terminated vals are fine as oF data, it doesn't have to be current

frame_2D_alg/agg_recursion.py
boris-kz: it's not just Bt, we need to add all tFts?

frame_2D_alg/agg_recursion.py
boris-kz: right, we just need to add that oF return vals, from L here

frame_2D_alg/meta_code.py
```diff
@@ -154,7 +154,7 @@ def inner(*a, **kw):
                 _CoF.call_ += [oF]
                 oF_[iF_[func.__name__]].caller_.add(_CoF)  # for comp_caller_
             _oF = CoF._cur.set(oF)
-            if out := func(*a, **kw): *_, oF.N_, oF.dTT, oF.c, oF.r = out
+            if out := func(*a, **kw): *out, gates = out; *_, oF.N_, oF.dTT, oF.c, oF.r = out
```
boris-kz, line 157: out here should be [computed_vals, gated_vals], why gates?

frame_2D_alg/agg_recursion.py
kwcckw: So this projection should be applied to all tFts here? Not sure if there's redundancy from tNt?

frame_2D_alg/meta_code.py
kwcckw: Sorry, could you clarify again if the computed vals are those dTT, c and r? While gated vals are the negative vals that block the call?

boris-kz: yes, for example:

def ffeedback(frame, _aTT,_oTT, _aH=[],_oH=[]):  # recompute filters from regime drift; fork: reform oF_ on cross-regime drift

    global ave,avd
    aTT = oTT = aH = oH = dTT = dc = dr = 0
    _ac,_ar = (_aH.c,_aH.r) if _aH else (0,0)
    _oc,_or = (_oH.c,_oH.r) if _oH else (0,0)
    # init empty aH,oH
        dTT = aH.dTT-_aTT; _aTT=aTT; dc= aH.c-_ac; dr= aH.r-_ar
        ave, avd = vt_(aH.dTT)  
        # filters *= ave
            dTT += oH.dTT-_oTT; _oTT=oTT; dc+=oH.c-_oc; dr+=oH.r-_or 
            split_oF_(); cluster_oF_()  
            # reform oF_
    # out = active attrs, own vals, gating vals:  
    return frame, aTT or _aTT, oTT or _oTT, aH or _aH, oH or _oH, [dTT,dc,dr], {}

frame_2D_alg/agg_recursion.py
boris-kz: so we need _aH,_oH here to compute the diffs, I posted below

frame_2D_alg/meta_code.py
kwcckw: I see, thanks, this is much clearer now. Also for gated vals, eventually we need to pack it as var in oF.gate?

boris-kz: why do you need separate oF.gate?  It sounds like the gate of the oF, which is local only

kwcckw: Those negative blocking vals will be retrieved in traced, and we only need it locally in traced? But how do we need it in traced? To pack it as negative val to the root oF?

boris-kz: add it as posive (abs) to computed vals, that's the combined contribution of oF

frame_2D_alg/agg_recursion.py
boris-kz: how is that redundant, tFt are not summed in L vals?

frame_2D_alg/agg_recursion.py
```diff
@@ -65,25 +66,31 @@
 '''
 def cross_comp(root, rr):  # core function mediating recursive rng+ and der+ cross-comp and clustering
 
+    gates = {}
```
boris-kz, line 69: so this gates is not relevant:

    N_,G_, gV = root.N_,[],0  # root is Ft, converted below, rc=rdn+olp, comp N_| B_| C_:
    # cLt: (sum_vt(+-L_), gv) for oF:
    L_,TT,c,r, cLt = comp_C_(N_,rr,fC=1) if fC else comp_N_(combinations(N_,2),rr)
    if L_:  # +ve only
        if gv := sum(vt_(TT,ttE))* (wE*L) - (ave+avd) * (r+cE*L) > 0:  # gate value
            E_,_ = get_exemplars({N for L in L_ for N in L.N_}, r,c)
            G_,r,_ = cluster_N(root, E_,r,c)  # cluster_C, _P, eval?
            if G_:
                if gv := vt_(TT,tAgg)[0]* (wAgg*L) - ave* (r+cAgg*L) > 0:
                    G_,_ = cross_comp(root.Nt,r)  # agg+
                else: gV-= gv
        else: gV-= gv  # invert sign to sum, include deep forks?
    return G_, (cLt[:-1],gV)  # last term for oF, skip nested gv

frame_2D_alg/agg_recursion.py
boris-kz: just put it in the return

frame_2D_alg/agg_recursion.py
```diff
@@ -65,25 +66,31 @@
...
+        gE = sum(vt_(TT,ttE)) * (wE*L) - (ave+avd) * (r+cE*L)
+        if gE > 0:  # or oF.gF evals?
             E_, *_ = get_exemplars({N for L in L_ for N in L.N_}, r,c)
             G_,*_,_r = cluster_N(root, E_,r,c)  # cluster_C, _P, eval?
             if G_:
                 if not root.typ: F2N(root)  # promote @ 1st sub+ or agg+
                 root.H+= [sum2F(L_,root,froot=1)]  # lev: L_+ derivatives
                 root.Nt = sum2F(G_,root,froot=2); L=len(G_)-1  # or C_ s?
-                if vt_(TT,tAgg)[0]*(wAgg*L) > ave* (r+cAgg*L):
+                gAgg = vt_(TT,tAgg)[0]*(wAgg*L) - ave* (r+cAgg*L)
```
boris-kz, line 83: this is ridiculous, get used to walrus expression

frame_2D_alg/agg_recursion.py
kwcckw: Okay, so we only need the summed gV from all gates. Also for cross_comp, the computed vals from nested cross_comp will be ignored?

frame_2D_alg/agg_recursion.py
kwcckw: When we form lev from dn_:

      if dn_:
          [add_H(L.H, d.H, L) for d in dn_ if d.H]  # lower levs
          if not L.H and _N.H and N.H: comp_H(_N.H, N.H, L)  # cheap scan if no deeper comp_N?
          L.H += [sum2F(dn_,L)]  # top lev
          
From sum2F:
  
      if root is not None:
          F.wTT = root.wTT
          if typ==3 and getattr(root,'C_', None): root.Ct = sum2F(root.C_)

Actually all tFts are already summed to L vals since typ != 2 here.

frame_2D_alg/agg_recursion.py
boris-kz: yes, that's because it's a gated process, we have the gating val instead

frame_2D_alg/agg_recursion.py
boris-kz: ok, we probably need to pack the base L vals in H[0], so it's recoverable

frame_2D_alg/agg_recursion.py
```diff
@@ -440,7 +464,10 @@ def cluster_P(_C_, _c, root):  # multi-seed mean shift: parallel centroid refine
         _C_ = [c for c in C_ if c.m * wcP > ave * c.r * ccP]  # prune
         new_ = []  # draft seed Ns straddling >=2 Cs with strong ~equal m?
         for j, N in enumerate(N_):
-            if np.sort(md__[j, :, 0])[-2:].min() * wcP > ave * (N.r + ccP): new_ += [Copy_(N, root, init=1, cls=CC)]
+            if np.sort(md__[j, :, 0])[-2:].min() * wcP > ave * (N.r + ccP):
```
boris-kz, line 467: removed = []
        if O*wcP > ave*(root.r+ccP*L):  # merge redundant Cs
            for i,_C in enumerate(C_):
                if _C in removed: continue
                for C in C_[i+1:]:
                    if C in removed: continue
                    if l.m*wF > ave*(l.r+cF):
                        add2F(_C,C,1); removed += [C]
        new_ = [Copy_(N,root,init=1,cls=CC) for j,N in enumerate(N_)
                if np.sort(md__[j,:,0])[-2:].min()*wcP > ave*(N.r+ccP)]  # seed overlap Ns
        _C_ = [c for c in C_ if c not in removed and c.m*wcP > ave*c.r*ccP] + new_  # survive+prune, +seeds
        if conv and not (removed or new_) and len(_C_)==Lc:
            break
        _md__ = md__; cnt += 1

frame_2D_alg/agg_recursion.py
boris-kz: most of it is obsolete, just

def proj_TT(L, cos_d, dist, r, pTT, wTT, dec=1, fdec=0):  # accumulate link pTT with iTT | eTT internally, L may be N

    dec = dist if fdec else ave ** (1+ dist*dec / L.span)  # not fully revised, ave = match decay rate / unit distance
    cert = abs(sum(vt_(TT,wTT)) * wPrj)  # approximation
    if cert > (ave+avd)*(r+cPrj):  # certainty margin
        pTT+=TT
?

frame_2D_alg/agg_recursion.py
kwcckw: So after adding new C and removing Cs from the section below:

    new_ = [Copy_(N,root,init=1,cls=CC) for j,N in enumerate(N_)
            if np.sort(md__[j,:,0])[-2:].min()*wcP > ave*(N.r+ccP)]  # seed overlap Ns
    _C_ = [c for c in C_ if c not in removed and c.m*wcP > ave*c.r*ccP] + new_  # survive
    
I think we need to reset N's root_, m_ and d_ again since that will be added(append) in the next iteration via sum2F of root and md__.

frame_2D_alg/agg_recursion.py
kwcckw: I see, and the gV above is added on both if and else section of eval, but with -= gv in else section. Wouldn't be that adding redundancy since we already have the callee's val, and the same positive gV?

frame_2D_alg/agg_recursion.py
kwcckw: Okay, then H[0].N_ is actually same as the L.N_, packing L's nt only? Or we probably can copy L into L.H[0] instead.

kwcckw: > most of it is obsolete, just
> 
> def proj_TT(L, cos_d, dist, r, pTT, wTT, dec=1, fdec=0): # accumulate link pTT with iTT | eTT internally, L may be N
> 
> ```
> dec = dist if fdec else ave ** (1+ dist*dec / L.span)  # not fully revised, ave = match decay rate / unit distance
> TT = np.array([L.dTT[0] * dec, L.dTT[1] * cos_d * dec])  # IxE angle alignment * decay?
> cert = abs(sum(vt_(TT,wTT)) * wPrj)  # approximation
> if cert > (ave+avd)*(r+cPrj):  # certainty margin
>     pTT+=TT
> ```
> 
> ?

Okay, i guess the projection via tFts can be skipped now since L.dTT included all the subs vals.

boris-kz: copy just before adding the first tLev

frame_2D_alg/meta_code.py
boris-kz: then

            if out := func(*a, **kw): oF.N_, oF.dTT, oF.c, oF.r, oF.m = out[-1]  # m is gating val, accum from vt_(dTT) later

boris-kz: Just food thor thought:

Right — I was wrong to lump those together. cross_comp's `out[-1] = (cLt[:-1], gV)` proves the point: `gV` accumulates the *rejected* value too (every `gV -= gv` branch), so it can't be recovered from `G_` or from any returned node. It's a measure of the function's decision quality — how much value its gates found vs. discarded — which is exactly the auto-refactor signal and is invisible to an external observer. So a separate channel is correct.

The weak part isn't the separation, it's the *carrier*. Braiding the stat into the return tuple means every call site has to slice around it — you already pay this with `_` discards, `cLt[:-1]`, `[0][0]`, `*sum_vt(...)[:-1]`. Functional and meta channels can't evolve independently, and the unpack arity in traced() has to track whatever shape each function packs.

You already built the right carrier: `CoF._cur`. Write the stat to the current oF at each gate instead of accumulating a local and packing it. traced() reads it off oF afterward; the functional return goes back to clean `return G_`. Recursion handles itself — inner cross_comp's gates land on the inner oF because `_cur` is already reset to it.

More important for refactoring: don't collapse to a single `gV`. Refactoring acts on *sites* (split this gate, drop this branch, inline this callee), so the stat should be a per-site map keyed by lineno — the generalization of `F_call_T_`/`F_call_i_` from call-sites to gate-sites. A net scalar can't tell you *which* gate is unproductive; a site map can.

Concretely, one helper using the lineno trick you already use at line 162:

```python
def gate(v):  # record decision value at this site, test it
    oF = CoF.get(); l = inspect.currentframe().f_back.f_lineno
    oF.gateV_[l] = oF.gateV_.get(l,0) + v; return v > 0
```

Then `if gate(vt_(TT,tAgg)[0]*(wAgg*L) - ave*(r+cAgg*L)):` — no local gV, no `out[-1]`, and you get per-gate resolution. Call-site value (`F_call_T_`) and gate-site value then live in the same shape: function performance = map from source location to value outcome.

Cost to be honest about: ambient writes need the correct `_cur` scope, but traced() already guarantees that deterministically. The only thing you lose is the locality of "stats travel with the value" — which you're already breaking anyway by discarding `out[-1]` at most call sites.

kwcckw: The gating val should be different with oF.dTT, so i think that shouldn't be redundant in 1 above.

Also gating val is only needed in traced, we can pop it or remove it in traced, so that function body doesn't read it?

boris-kz: I posted a new reply above, it suggested using CoF._cur?

kwcckw: Since gV is summed, we no longer able to retrieve per gate Val. Using cur able to remove the last gV output, but I think we still need to call this gate function in every gating eval.

boris-kz: we can buffer them in gv_
I will check tomorrow

frame_2D_alg/agg_recursion.py
boris-kz: callee's val is separate, not summed in own vals

PR #178, opened 2026-06-17
6/17/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -70,22 +70,24 @@ def cross_comp(root, rr, fC=0):  # core function mediating recursive rng+ and de
     L_,TT,c,r, cLt = comp_C_(N_,rr,fC=1) if fC else comp_N_(combinations(N_,2),rr)
     if L_:  # +ve only
         root.L_ = L_; L= len(L_)-1  # val=m+d /clust, m/comp
-        if gv := sum(vt_(TT,ttE))* (wE*L) - (ave+avd) * (r+cE*L) > 0:  # gate value
-            E_,_ = get_exemplars({N for L in L_ for N in L.N_}, r,c)
+        # bracket to prevent gv = val>0
+        if (gv := sum(vt_(TT,ttE))* (wE*L) - (ave+avd) * (r+cE*L)) > 0:  # gate value
```
kwcckw, line 74: So we need bracket here to prevent assigning gv as val > 0.

frame_2D_alg/agg_recursion.py
```diff
@@ -408,19 +414,21 @@ def cluster_C(Ft, E_,_r,_c):  # form centroids by clustering exemplar surround v
...
 def cluster_P(_C_, _c, root):  # multi-seed mean shift: parallel centroid refine, _C_ varies via split/merge
 
-    cnt = 0; Ln = len(N_:= list(set([N for C in _C_ for N in C.N_])))  # all Ns are in all Cs
+    cnt, gV = 0,0; Ln = len(N_:= list(set([N for C in _C_ for N in C.N_])))  # all Ns are in all Cs
     _md__ = np.zeros((Ln, len(_C_), 2))  # NxC, cols aligned to _C_
     for j, N in enumerate(N_):
         for c,m,d in zip(N.root_,N.m_,N.d_):
             if c in _C_: _md__[j,_C_.index(c)] = m,d
     while True:
+        for N in N_: N.root_, N.m_, N.d_ = [], [], []  # reset, append in sum2F
```
kwcckw, line 431: We should reset root, m_ and d_ here because they will be appended in the sum2F below, and this should be done in each iteration since the number of seed C may differ.

frame_2D_alg/meta_code.py
```diff
@@ -148,14 +148,15 @@ def traced(func):
...
-            oF = CoF(nF=iF_[func.__name__], root=_CoF)
+            oF = CoF(nF=iF_[func.__name__], root=_CoF); oF.gV_ = {}
             i = iF_[func.__name__]; oF_[i].call_ += [oF]  # F_call_T_[i][oF.nF] += oF.dTT
             if _CoF is not None:
                 _CoF.call_ += [oF]
                 oF_[iF_[func.__name__]].caller_.add(_CoF)  # for comp_caller_
             _oF = CoF._cur.set(oF)
             if out := func(*a, **kw):
                 oF.N_, oF.dTT, oF.c, oF.r, oF.w = out[-1]  # w is gating val, then += vt_(dTT)[0]
+                # oF.w = sum(oF.gV_.values())  # gate method
```
kwcckw, line 159: Actually i think the gate method is better, we just need to wrap the eval with `gate`, for example:

    if gate(sum(vt_(TT, ttE))*(wE*L) - (ave+avd)*(r + cE*L)):

And we can skip the init of gv and gV.

boris-kz: Your gate is per callee, we need it per caller?

kwcckw: It should be the same here because we need to add all gates gV after the func ops (after all callees are done), which can be per oF here?

frame_2D_alg/agg_recursion.py
```diff
@@ -115,18 +117,19 @@ def proj_V(_N,N, dist, dy_dx, dec):  # _N x N induction
         lr = r+ (N.r+_N.r)/2; m,d = vt_(pTT,ttN_)  # +|-match certainty
         if m > 0:
```
boris-kz, line 118: this is another gate

frame_2D_alg/agg_recursion.py
```diff
@@ -172,20 +175,21 @@ def comp_H(_Nt,Nt, Link):  # tentative pre-comp
             tt += ltt*lc; C+=lc; R+=lr*lc
         return dH,tt,C, (r* Link.c+R)/ (Link.c+C)  # same norm for tt?
 
-    TT = base_comp(_N,N)[0] if full else comp_derT(_N.dTT[1],N.dTT[1]); m,d = vt_(TT, ttN_)
+    gV=0; TT = base_comp(_N,N)[0] if full else comp_derT(_N.dTT[1],N.dTT[1]); m,d = vt_(TT, ttN_)
     L = CL(N_=[_N,N], dTT=TT,m=m,d=d,c=c,r=r, root=rL)
-    if N.typ and m* wN_*c > ave*(r+cN_):  # skip PPs, may comp Nts?
+    if (gv:=N.typ and m* wN_*c - ave*(r+cN_))>0:  # skip PPs, may comp Nts? (may not be direct, but gv here is for comp_F?)
```
boris-kz, line 180: it's still in function body, direct for our purposes

frame_2D_alg/agg_recursion.py
boris-kz: do we actually use these N params within a loop?

frame_2D_alg/agg_recursion.py
kwcckw: Thanks, so gate is only valid for numeric values only right?

For example, for negative blob in vect_edge:

    if not blob.sign:

This shouldn't be added ?

frame_2D_alg/meta_code.py
```diff
@@ -148,14 +148,15 @@ def traced(func):
         @wraps(func)
         def inner(*a, **kw):
             _CoF = CoF._cur.get(None)
-            oF = CoF(nF=iF_[func.__name__], root=_CoF)
+            oF = CoF(nF=iF_[func.__name__], root=_CoF); oF.gV_ = {}
```
boris-kz, line 151: why is gV_ a set, it has to be a specific call site sequence?

frame_2D_alg/agg_recursion.py
kwcckw: Actually no, it's just for recording purpose to indicate the m and d values of each node in C. We use `md__` and `_md__` instead.

frame_2D_alg/meta_code.py
kwcckw: Yes, it's specific because it's based on the line number, added in gate function:

      def gate(v):  
          oF = CoF.get()
          if oF is not None:
              l = inspect.currentframe().f_back.f_lineno
              oF.gV_[l] = oF.gV_.get(l, 0) + v
          return v > 0

frame_2D_alg/agg_recursion.py
boris-kz: yes, this is logic, not an eval. 
Actually we only need collect stats on complex evals, where the eval cost justifies secondary eval

frame_2D_alg/agg_recursion.py
boris-kz: so we only need this assignment after convergence

frame_2D_alg/meta_code.py
boris-kz: we don't need that, the call sites in .body already uniquely map to gv_

frame_2D_alg/meta_code.py
boris-kz: ok then it's just
def gv_(v):

    oF = CoF.get()
    if oF is not None and v<0: oF.gv_ += [v]; oF.w += v

?

kwcckw: This `gV_` should be called per function, and `v` in `oF.gv_ += [v]` is the current `gv`?

So in the else section, it will be `gV_(-gv)`?

boris-kz: it's only for neg vals, else the gate is passive
actually we need to pass site index here:

def gv_(v, i):  # called same as gate()

oF = CoF.get()
if oF is not None and v<0: oF.gv_[i] += -v; oF.w += -v
return v

boris-kz: if gv_(sum(vt_(TT,ttE))*(wE*L) - (ave+avd)*(r+cE*L), 0):  # store neg gate values in oF.gv_[0]

kwcckw: If it's in the else section, `v` must be negative, so `v<0` is redundant? 

Also `oF is not None ` is redundant too since we only add gate's v in those oF specific functions?

kwcckw: > ```
>     if gv_(sum(vt_(TT,ttE))*(wE*L) - (ave+avd)*(r+cE*L), 0):  # store neg gate values in oF.gv_[0]
> ```

Okay, this site index should be init with number of oF body, and some funcs are default, so they always will be empty?

boris-kz: there is no else, this eval is internal.
also this is per callee, so it can replace all meta-return if we can get callee oF too? 

we only need it for non-defaults

boris-kz: Claude
The per-site `gv_[i]` is the right shape, and manual `i` is more edit-stable than keying by lineno. Two issues:

**Bug:** `return v` makes `if gv_(...)` fire on negative v too — negative numbers are truthy. You want `return v > 0` so the gate still tests sign; the site stays `if gv_(expr, 0):`.

**Redundancy:** `oF.w += -v` duplicates what's already in `gv_[i]`. The rejected component of `w` is just `sum(gv_)`, so set `w = sum(gv_) + vt_(dTT)[0]` once (in traced, after dTT lands) and drop the `+= -v` — single source of truth, no drift.

With that, both gates collapse and the `else: gV -= gv` accounting disappears (it's the side-effect now), so `gV` drops from the return:

```python
if gv_(sum(vt_(TT,ttE))*(wE*L) - (ave+avd)*(r+cE*L), 0):
    E_,_ = get_exemplars({N for L in L_ for N in L.N_}, r,c)
    G_,r,_ = cluster_N(root, E_,r,c)
    if G_:
        if not root.typ: F2N(root)
        root.H += [sum2F(L_,root,froot=1)]
        root.Nt = sum2F(G_,root,froot=2); L=len(G_)-1
        if gv_(vt_(TT,tAgg)[0]*(wAgg*L) - ave*(r+cAgg*L), 1):
            G_,_ = cross_comp(root.Nt, r)  # agg+
return G_, cLt[:-1]  # gate stats on oF.gv_/oF.w; out[-1] = data summary
```

Init: `gv_` isn't on CoF yet — add `f.gv_ = kw.get('gv_', defaultdict(float))` (or a zeros array if you keep integer indices and want fixed size). `w` already exists.

boris-kz:
```
def gv_(v, i):
    oF = CoF.get()
    if v > 0: return v
    else: oF.gv_[i] -= v  # then oF.w = vt_(oF.dTT)[0] + sum(oF.gv_)
```
Gate logic's correct now: `None` on reject, `v` on pass, and `v == 0` falls through to falsy → fails, so it matches `> 0` without the extra comparison. The implicit-None return is the load-bearing part and it works.

The thing to pin down is what `w` means, because the `+` decides it. `sum(oF.gv_)` is accumulated *deficit* — how far rejected candidates fell below threshold — and you're adding it to delivered value `vt_(dTT)[0]`. Those aren't the same quantity: one is value produced, the other is the margin by which gated paths failed. As written, a gate that rejects lots of far-below-bar candidates *raises* w, so the more wasteful a function is, the more it weighs in comp_callers (`c.w * c.c`).

Two coherent readings, pick one:
- **w = value** (for caller-weighting / clustering): then it should be `vt_(dTT)[0]` alone, with `sum(gv_)` kept separate as the per-site refactor signal — or `vt_(dTT)[0] - sum(gv_)` if you want net, penalizing waste.
- **w = throughput** (total value evaluated, weighting callers by activity not efficiency): then `+` is right as-is.

The deficit is a clean *diagnostic* either way; the only question is whether it belongs inside the scalar you weight by, or beside it.

boris-kz: oF.w is a combination of internal gain:  vt_(dTT)[0] 
and filtering gain: sum(gv_): here we invert the sign because filtering saves on wastefull calls

...
Coherent, and the two are disjoint so there's no double count: internal gain is value realized from work done (flows into dTT), filtering gain is value-loss avoided on work *not* done (the rejected paths). Same units, both positive, `+` correct.

One thing to check, not a flaw: "filtering saves wasteful calls" is exact only if the gate threshold prices the *avoided downstream* work — the get_exemplars/cluster_N/agg+ the gate declines. In cross_comp the cost term `(ave+avd)*(r+cE*L)` scales with `L = len(L_)-1`, the links *already computed*. If `cE` is calibrated as downstream cost-per-link (clustering cost does scale with link/node count, so this can be legitimate), then `-v` is the real avoided cost and the magnitude is right. If `cE` instead prices the comparison just done, then `-v` is the comparison's own deficit, not the cost of the call you skipped — sign and units still hold, but filtering gets over/under-credited against internal gain, which then skews `c.w` in comp_callers.

So the load-bearing assumption is: the cost subtracted in the gate is the cost of what the gate *prevents*, not what it already spent. Worth confirming `cE` is on the right side of that.

kwcckw: > so it can replace all meta-return if we can get callee oF too?

    def gv_(v, i):
        oF = CoF.get()
        if v > 0: return v
        else: oF.gv_[i] -= v  # then oF.w = vt_(oF.dTT)[0] + sum(oF.gv_)
        
Yes, so this is actually better an cleaner without returning the gV.

> and filtering gain: sum(gv_): here we invert the sign because filtering saves on wastefull calls

This gV_ is the avoided cost, which is actually the gain from saving? So why it should be negative here?

boris-kz: it's a double negative, which is positive
so we need a similar function to add own vals?

kwcckw: Okay right, it's a subtraction of negative vals. 

> so we need a similar function to add own vals?

Yes, i think we can have another function to pack N_, TT, c and r into oF, then we can clean up the existing messy return structure. It should be more straightforward since it's just need to assigning N_, TT, c and r into oF.

frame_2D_alg/meta_code.py
```diff
@@ -148,14 +148,15 @@ def traced(func):
...
             i = iF_[func.__name__]; oF_[i].call_ += [oF]  # F_call_T_[i][oF.nF] += oF.dTT
             if _CoF is not None:
                 _CoF.call_ += [oF]
                 oF_[iF_[func.__name__]].caller_.add(_CoF)  # for comp_caller_
             _oF = CoF._cur.set(oF)
             if out := func(*a, **kw):
                 oF.N_, oF.dTT, oF.c, oF.r, oF.w = out[-1]  # w is gating val, then += vt_(dTT)[0]
+                # oF.w = sum(oF.gV_.values())  # gate method
             if oF.call_:
```
boris-kz, line 160: sometimes N_,TT,c,r is the primary return, as in comp_ oFs, but separate Fvt() is still better for consistency

kwcckw: Yes, but we can have both? So the output may not necessary in the format of (N_, TT, c, r), it's just we must parse those information into Fvt()?

boris-kz: right

PR #179, opened 2026-06-18
6/18/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -173,7 +172,13 @@ def __bool__(f): return bool(f.call_)
 def gv_(v, i):
     oF = CoF.get()
     if v > 0: return v
-    else: oF.gv_[i] -= v  # then oF.w = vt_(oF.dTT)[0] + sum(oF.gv_)
+    else: 
+        while (len(oF.gv_)<=i): oF.gv_ += [0]  # init empty gv_, then we can skip the step to init from checking oF.body
```
kwcckw, line 176: So we need to init gv_ based on the max number of i, and i think we can do it using while loop here instead of checking the oF.body?

frame_2D_alg/agg_recursion.py
```diff
@@ -66,24 +66,24 @@
...
+            G_,r = cluster_N(root, E_,r,c)  # cluster_C, _P, eval?
             if G_:
                 if not root.typ: F2N(root)  # promote at 1st sub+ or agg+
                 root.H += [sum2F(L_,root,froot=1)]  # dLev per L_
                 root.Nt = sum2F(G_,root,froot=2); L=len(G_)-1  #| C_?
                 if gv_(vt_(TT,tAgg)[0]* (wAgg*L) - ave* (r+cAgg*L), 1):  # store in oF.gv_[1]
-                    G_,_ = cross_comp(root.Nt,r)  # agg+
-    return G_, cLt  # last term for oF
+                    G_ = cross_comp(root.Nt,r)  # agg+     
+    Fvt(L_, *sum_vt(L_))  # this is exactly the same comp' output, include vals from G_ too?
```
kwcckw, line 81: For the cross_comp values, if we use only L_, it's exactly the same as the comp's values? I think we need to add values from G_ as well?

boris-kz: that's a default result, clustering is gated, that's why it's called cross_comp

kwcckw: Okay, thanks for the clarification.

frame_2D_alg/meta_code.py
boris-kz: this is per caller, you are initializing the whole gv_ from a single call, it doesn't make any sense
oF.gv_ must be initialized manually in function def and call sites mapped to oF.body primitives

frame_2D_alg/meta_code.py
```diff
@@ -392,6 +398,10 @@ def cent_TT(dTT, r):  # EM-like weight attr matches | diffs by their match to th
         if np.sum(np.abs(wTT-_wTT)) < ave * r:  # if np.linalg.norm(wT - _wT, 1) < r?
             break
         _wTT = np.array(wTT); wTT = []
+    
+        aa+=1
```
boris-kz, line 402: what is that for?

frame_2D_alg/agg_recursion.py
```diff
@@ -66,24 +66,24 @@
 def cross_comp(root, rr, fC=0):  # core function mediating recursive rng+ and der+ cross-comp and clustering
 
     N_,G_, gV = root.N_,[],0  # root is Ft, converted below, rc=rdn+olp, comp N_| B_| C_:
-    # cLt: (sum_vt(+-L_), gv) for oF:
-    L_,TT,c,r, cLt = comp_C_(N_,rr,fC=1) if fC else comp_N_(combinations(N_,2),rr)
-    if L_:  # +ve only
+    L_,TT,c,r = comp_C_(N_,rr,fC=1) if fC else comp_N_(combinations(N_,2),rr)
+    if L_:  # +ve only (this L includes +=L?)
```
boris-kz, line 70: only +ves, while oF accumulates results of both signs

boris-kz: L_,TT,c,r, cLt = comp_C_(N_,rr,fC=1) if fC else comp_N_(combinations(N_,2),rr)
    Fvt(cLt)  # copy default results from comp_

frame_2D_alg/agg_recursion.py
boris-kz: this is the difference between return: includes results from the callees, 
and ambient summary: contribution of current oF only

frame_2D_alg/meta_code.py
kwcckw: Yes, this is actually a hack since it may not add all gv_ if the max i is not run. 

I think we need to initialize it manually because it's actually not mapping 1-to-1 to each oF body?

For example, in cluster_N, this is not mapping to any of the oF function:

        if G_:
            for G in G_: trans_cluster(G)  # splice trans_links, merge L.N_.roots
            C = sum([g[1] for g in Gt_]); TT=np.zeros((2,9)); R=0; r+=1  # + wC?
            for tt,c,gr in Gt_: w=c/C; TT+=tt*w; R+=gr*w
            L = len(G_)-1
            if gv_(sum(vt_(TT, Ft.root.wTT*ttcN))*(wcN*L) - (ave+avd)*(r+R+ccN*L), 1):  # reform root,Nt, no other forks yet:
                rG = Ft.root; Nt=rG.Nt; Nt.N_=G_; Nt.dTT=TT; Nt.r=R
                rG.dTT=TT; rG.c=C; rG.r=R

frame_2D_alg/agg_recursion.py
kwcckw: I see, but this is exclusively for cross_comp only since the most of the other function have the same results and contributions (comp_N_, comp_C, comp_N, etc).

frame_2D_alg/meta_code.py
kwcckw: Sorry, this is redundant, it was added by mistake.

frame_2D_alg/meta_code.py
```diff
@@ -173,7 +172,13 @@ def __bool__(f): return bool(f.call_)
 def gv_(v, i):
     oF = CoF.get()
     if v > 0: return v
-    else: oF.gv_[i] -= v  # then oF.w = vt_(oF.dTT)[0] + sum(oF.gv_)
+    else: 
+        while (len(oF.gv_)<=i): oF.gv_ += [0]  # init empty gv_, then we can skip the step to init from checking oF.body
+        oF.gv_[i] -= v  # then oF.w = vt_(oF.dTT)[0] + sum(oF.gv_)
+
+def Fvt(N_, TT, c, r):
```
boris-kz, line 179: if out := func(*a, **kw):
                C = oF.c; TT=np.zeros((2,9)); R=0
                for tt,c,r in oF.vt_: w= c/C; TT+=tt*w; R+=r*w
                oF.dTT,oF.r = TT,R  # this is replacing sum2O?
            if oF.call_:
                if _CoF is not None:
                    if (j := F_call_i_[_CoF.nF].get(inspect.currentframe().f_back.f_lineno)) is not None:  # get callee site
                        F_call_T_[_CoF.nF][j] += oF.dTT  # add,c,r: results per callee to refine the code
            CoF._cur.reset(_oF)
            return out
        inner.wrapped = True
        return inner

def Fvt_(N_,TT,c,r):
    oF = CoF.get(); oF.N_+=N_; oF.c+=c; oF.vt_ += [[TT,c,r]]  # data per call

frame_2D_alg/agg_recursion.py
kwcckw: Okay, so oF.N_ is actually packing only positive Ls since only positive L is appended to L_:

        elif L.typ==1 and L.d > avd: acc[3]+=L.dTT*L.c; acc[4]+=L.c; acc[5]+=L.r*L.c

frame_2D_alg/meta_code.py
boris-kz: we need to set indices of those gates (ifExp) in oF body?

frame_2D_alg/agg_recursion.py
boris-kz: No, it should be all Ls, that's why I have separate cLt

frame_2D_alg/agg_recursion.py
boris-kz: No, most of oFs have callee oFs, their resuts are not oF' own contribution

frame_2D_alg/agg_recursion.py
```diff
@@ -543,10 +551,10 @@ def sum2G(ft_, fTT, root=None, init=1):  # core clustering function
         Lt = G.Lt; L_,lm,ld,lr = Lt.N_,Lt.m,Lt.d,Lt.r; L=len(L_)
         Vn = (lm+ld)* (wcN*L) - (ave+avd)* (lr+1+ccN*L)
         if Vn > 0:
-            c = G.Lt.c; E_ = get_exemplars({N for L in L_ for N in L.N_}, r,c)[0]
-            if Vn *(mdecay(L_)-decay) > (ave+avd)*(lr+1+ccC*L):
-                r+=1; G_,r,_ = cluster_C(G.Nt,E_,r,c)  # higher V, low decay, eval cluster_P
-            else:     G_,r,_ = cluster_N(G.Nt,E_,r,c)  # updates G
+            c = G.Lt.c; E_ = get_exemplars({N for L in L_ for N in L.N_}, r,c)
+            if Vn *(mdecay(L_)-decay) > (ave+avd)*(lr+1+ccC*L):  # parse oF here to sum gV?
```
boris-kz, line 555: yes

kwcckw: Okay, we probably need have an additional oF input argument in Fvt, and use `oF = CoF.get()` only if input oF is not given.

frame_2D_alg/agg_recursion.py
kwcckw: But we didn't add neg Ls now?

    elif L.typ==1 and L.d > avd: acc[3]+=L.dTT*L.c; acc[4]+=L.c; acc[5]+=L.r*L.c
    
So we need to pack negative L into cLt's L_ in the line above.

frame_2D_alg/meta_code.py
kwcckw: > we need to set indices of those gates (ifExp) in oF body?

Theoretically it should be like that, but we can't assign differently to each ifExp since there's some special cases where we may have multiple ifExp for single oF, or there's a gate for non oF ops?

frame_2D_alg/agg_recursion.py
boris-kz: right.
claude' PR discussion review:

Read both files. Five threads; two are dead (the "yes" and the redundant-line). The rest:

**gv_ init (cluster_N thread).** You're right that per-call sizing is unsound; kwcckw is right that gates don't map to body primitives. Both resolve the same way and neither needs a runtime loop — the gate count is static: number of `gv_(...)` Call nodes per FunctionDef. I checked all of them: source order of gv_ calls equals the manual `i` in every function (cross_comp 73→0, 80→1; comp_N_ 114→0, 116→1; cluster_N 348→0, 356→1; cluster_C 402→0, 415→1; …). So the literal `i` is redundant. Mirror the existing F_call_ machinery exactly:

```python
def gv_sites(fd): return [n for n in ast.walk(fd) if isinstance(n,ast.Call) and isinstance(n.func,ast.Name) and n.func.id=='gv_']
gv_i_ = [{n.lineno:i for i,n in enumerate(gv_sites(fd))} for fd in nF_]
```

Size `oF.gv_=[0]*len(gv_sites(nF_[nF]))` at construction; in `gv_`, derive i from `f_back.f_lineno` via `gv_i_[oF.nF]`, drop the i arg. Branch-gates are to control flow what call-sites are to calls — same indexing.

On your "set indices of those gates (ifExp) in body": key on the gv_ Call nodes, not If/IfExp. `if G_:` in cluster_N is a plain If with no gv_ that wraps `gv_(...,1)` — that's exactly kwcckw's "not mapping to any oF function." Indexing by If-nodes makes `if G_:` eat a slot and misaligns the inner gate. Gate ≡ gv_ call, not the conditional around it.

Concrete bug independent of the design: `gv_` defaults to `[]`, and the else branch does `oF.gv_[i] -= v` → IndexError on the first -ve gate in any function. Nothing executes a negative gate today. Static pre-sizing is mandatory regardless.

**cross_comp own-vs-aggregate (add-G_ thread).** You're right; both of kwcckw's claims are wrong. "Add G_": G_ comes from cluster_N, a callee — already in the call tree via `flat_`→`sum2O`→`F_call_T_`. Folding it into `cross_comp.dTT` double-counts cluster_N. `dTT` is own contribution by construction (summed from `oF.vt_`, i.e. only what `Fvt_` appended in this frame); the callee aggregate is the separate `sum2O(tree)` path. "Unique to cross_comp": false — comp_N_ delegates the link comp to comp_N (a traced callee), so `comp_N_.dTT ≠ its return` for the same reason. kwcckw's observation that `cross_comp.dTT ≈ comp.dTT` is correct but expected: with clustering gated off, the level's default contribution *is* the comp output. Same value, different node (level vs operation), not redundancy.

**cLt = all Ls (comp_ return thread).** Design is coherent (L_ = +m for clustering, cLt = both signs for oF.N_/dTT) but unimplemented, and the interface is broken three ways:
- cross_comp unpacks 5 (`L_,TT,c,r,cLt`); comp_N_/comp_C_ `return` 4 → ValueError.
- `Fvt_(cLt)` passes one arg; `Fvt_` needs `(N_,TT,c,r)` → TypeError. Should be `Fvt_(cLt,*sum_vt(cLt))`.
- The -ve Ls get summed into `acc[3:6]` in comp_N but are never retained as objects, so cLt can't be assembled. comp_N's per-L `Fvt_([L],…)` already puts all-signs into comp_N's own node; the +ve-only leak is comp_N_/comp_C_'s `Fvt_(L_,*sum_vt(L_))` — switch those to cLt, or drop them and Fvt_ only in cross_comp.

**Your two inline TODOs in the decorator:**
- `oF.dTT,oF.r = TT,R  # replacing sum2O?` — for own data, yes: the inline vt_-weighted loop over `oF.vt_` replaces sum2O. But sum2O stays for the callee tree (line 166, `fcall_=1`), different input (vt_ data vs `F.call_` fc/c). It's a split of roles, not a replacement — same own/aggregate boundary as the thread above.
- `tree = flat_(oF)  # if len(tree)-1?` — `flat_` returns a flat callee list, not pairs, so `len-1` is "callees beyond the first." Use -1 only if you treat the call tree as a compositional cluster (the `len(X_)-1` = edge-count convention everywhere); for a plain tally, no -1. Since everything else clusters on -1, -1 is consistent *if* `sum2O(tree)` is a compression rather than a sum.

frame_2D_alg/agg_recursion.py
boris-kz: how is that not given?

frame_2D_alg/meta_code.py
boris-kz: we need to pick single if per oF only

frame_2D_alg/meta_code.py
```diff
@@ -156,7 +156,6 @@ def inner(*a, **kw):
                 oF_[iF_[func.__name__]].caller_.add(_CoF)  # for comp_caller_
             _oF = CoF._cur.set(oF)
             if out := func(*a, **kw):
-                oF.N_, oF.dTT, oF.c, oF.r = out[-1]  # replace with Fvt()?
                 oF.w = vt_(oF.dTT)[0] + sum(oF.gv_)
             if oF.call_:
                 tree = flat_(oF)  # if len(tree)-1?
```
boris-kz, line 161: Two lines in `meta_code.py`.

**166** — drop the dead sum2O (its return was discarded, `oF` unmutated) and the own-dTT fallback; flow the subtree tensor into `wTT`:

```python
# old
# new
                tt,_,r = sum_vt(tree); oF.wTT = cent_TT(tt, r)  # subtree dTT/r, not own
```

Note: don't route this through `oF.rTT` — `rTT` means relative prediction error (agg_recursion 118, consumed by `sum_vt(fr=1)` at line 27). Storing a subtree dTT-sum there corrupts any `fr=1` parent sum. Use the local.

**361** — docstring overclaims `w,c,r`, which it never computes (those now come from the inline `vt_` loop):

```python
# old
def sum2O(F_, root=None, w_=None, fcall_=0):  # for w,c,r, fw,fc,fr only?
# new
def sum2O(F_, root=None, w_=None, fcall_=0):  # fw,fc,fr,body only; dTT,r,w from vt_
```

sum2O stays as-is for 305/313/322/357. `len(tree)-1` is your separate TODO — orthogonal to this.

frame_2D_alg/agg_recursion.py
kwcckw: > we need to pick single if per oF only

But how about this non oF ops?

      if gv_(sum(vt_(TT, Ft.root.wTT*ttcN))*(wcN*L) - (ave+avd)*(r+R+ccN*L), 1):  # reform root,Nt, no other forks yet:
            rG = Ft.root; Nt=rG.Nt; Nt.N_=G_; Nt.dTT=TT; Nt.r=R
            rG.dTT=TT; rG.c=C; rG.r=R

This reform is a unique case where there's no gated oF functions, and gv_sites above is tracking only oF functions.

frame_2D_alg/agg_recursion.py
kwcckw: This is sum2G, it's not included in traced, so we need to parse the function's oF that calls this sum2G?

frame_2D_alg/agg_recursion.py
boris-kz: we don't need to track non-oF

frame_2D_alg/agg_recursion.py
boris-kz: right

frame_2D_alg/agg_recursion.py
```diff
@@ -111,10 +111,10 @@ def proj_V(_N,N, dist, dy_dx, dec):  # _N x N induction
         dist, dy_dx, _N,N, c = pL  # rim angl is not canonic
         pTT = proj_V(_N,N, dist, dy_dx, root.m if root!=2 else decay** (dist/((_N.span+N.span)/2)))  # based on current rim
         lr = r+ (N.r+_N.r)/2; m,d = vt_(pTT,ttN_)  # +|-match certainty
-        if m > 0:
+        if gv_(m, 0) > 0:
```
boris-kz, line 114: so this is not needed

frame_2D_alg/meta_code.py
kwcckw: > 166 — drop the dead sum2O (its return was discarded, oF unmutated) and the own-dTT fallback; flow the subtree tensor into wTT:

I think the suggestion is better, since we actually don't need sum2O here in line 166 right? We only need to compute the wTT, which could be done with sum_vt and cent_TT.

frame_2D_alg/agg_recursion.py
kwcckw: Okay, so with this approach, the gv_ must be used inside oF functions and tracking gated oF functions. For the case of cluster_N and cluster_C in sum2G, that can't be tracked now, so we may not need it anymore? Since cluster_N and cluster_C's oF won't be in the caller function's body anyway.

boris-kz: I think we need sum2G an oF too, it's not gated but it has internal gates

frame_2D_alg/meta_code.py
boris-kz: I guess so

frame_2D_alg/agg_recursion.py
kwcckw: Okay, sum2G should be included in clus_ cateogry?

boris-kz: right

frame_2D_alg/agg_recursion.py
```diff
@@ -156,8 +157,9 @@ def comp_C_(C_,_r, _C_=[], fall=1, fC=0):  # simplified for centroids, trans-N_s
             comp_N(_C,C,_r, c ,A=dy_dx,span=dist,L_=L_,N_=N_,acc=acc)  # simplified for typ=3
     for N in list(set(N_)):
         if N.rim: N.Rt = sum2F(N.rim)
-    TTm,cm,rm,TTd,cd,rd = acc
-    return L_,TTm,cm, rm/(cm or eps), (L_, *sum_vt(L_), 0)  # oF.N_ should be +-Ls
+    TTm,cm,rm,*_ = acc
+    Fvt(L_, *sum_vt(L_))
+    return L_,TTm,cm, rm/(cm or eps)  # oF.N_ should be +-Ls
 
 def comp_N(_N,N, r,c, full=1, A=np.zeros(2),span=None, rL=None, L_=None, N_=None, acc=None):
```
boris-kz, line 164: revised, returning +-Ls:

        iTT = (_N.dTT + N.dTT) * Dec
        cNt = min(_N.c,N.c); rNt = (_N.r+N.r) / 2
        if abs(vt_(eTT,ttN_)[0])* wPrj* cNt > ave* (cPrj+r+rNt):  # not oF, spec / link:
            eTT+= proj_N(N, dist, dy_dx, r, N.c, dec)[0]  # pTT/ L_,B_,rim, if pV >0
            eTT+= proj_N(_N,dist, -dy_dx, r, _N.c, dec)[0]  # reverse direction
        return iTT+eTT

    N_, L_ = [],[]
        dist, dy_dx, _N,N, c = pL  # rim angl is not canonic
        if m > 0:
            if gv_(abs(m)*wN - ave*(r+cN),1):  # comp if marginally predictable, update N.Rt pair eval, ave / proj surprise value?
                Link = comp_N(_N,N, lr, c, full=not tnF, A=dy_dx, span=dist, rL=root,L_=L_,N_=N_)
            else:
                pL = CL(typ=-1, N_=[_N,N],dTT=pTT,m=m,d=d,c=c,r=lr, angl=[dy_dx,1],span=dist)
                L_+= [pL]; N.rim+=[pL]; _N.rim += [pL]; N_+=pL.N_  # add neg C?
                # oF.N_ is +-Ls
    for N in set(N_):
        if N.rim: N.Rt = sum2F(N.rim)
    vT = sum_vt(L_)
    Fvt_(L_, *vT)  # same here
    return L_,*vT  # oF.N_=L_, !pLs

kwcckw: This line:

     Fvt_(L_, *vT)  # same here

It's still packing positive L_ only? We need another set of pL_ and L_ to pack pos and all Ls?

boris-kz: This is positive pLs, before actual Ls are evaluated. 
Could you edit comp_C_ accordingly?

kwcckw: Okay, and the positive or negative Ls should be eval in comp_N?

        elif L.typ==1 and L.d > avd: acc[3]+=L.dTT*L.c; acc[4]+=L.c; acc[5]+=L.r*L.c
        
 The negative Ls should be added in the elif section?
 
 Those pLs shouldn't be added in comp_N since they are added in the else section.

boris-kz: I think that separation should be in cross comp

kwcckw: Alright, so this should be moved out instead.

boris-kz: Actually, cross comp doesn't need neg Ls? So yes, we can filter them out in comp_N_

kwcckw: Yes, we only need neg Ls in comp_N_'s oF, which will be parsed to cross_comp anyway.

PR #180, opened 2026-06-19
6/19/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -115,24 +116,25 @@ def proj_V(_N,N, dist, dy_dx, dec):  # _N x N induction
         else: break  # beyond initial induction range, re-sort by proj_V?
     for N in set(N_):
         if N.rim: N.Rt = sum2F(N.rim)
-    vT = sum_vt(L_)
-    Fvt_(L_, *vT)  # same here
-    return L_,*vT  # oF.N_=L_, !pLs
```
kwcckw, line 120: We actually shouldn't return *vT here? It should be vals from positive Ls, which we need to compute with pLs?

frame_2D_alg/agg_recursion.py
```diff
@@ -146,14 +148,14 @@ def comp_C_(C_,_r, _C_=[], fall=1, fC=0):  # simplified for centroids, trans-N_s
             if _C in C.compared: continue
             dy_dx = _C.yx-C.yx; dist = np.hypot(*dy_dx); c = min(C.c,_C.c); r = (C.r+_C.r)/2  # tc += c; acc[1]+=c
             tc += c; tr += r*c
-            comp_N(_C,C,_r, c ,A=dy_dx,span=dist,L_=L_,N_=N_,acc=acc)  # simplified for typ=3
```
kwcckw, line 149: I removed the acc, N_ and L_ in comp_N since we not really need TTd, cd, and rd now. Those values can be summed with Ls again using sum_vt.

frame_2D_alg/agg_recursion.py
boris-kz: right

frame_2D_alg/agg_recursion.py
```diff
@@ -542,18 +539,19 @@ def sum2G(ft_, fTT, root=None, init=1):  # core clustering function
     if G.Lt:  # sub+
         Lt = G.Lt; L_,lm,ld,lr = Lt.N_,Lt.m,Lt.d,Lt.r; L=len(L_)
         Vn = (lm+ld)* (wcN*L) - (ave+avd)* (lr+1+ccN*L)
-        if Vn > 0:
+        if gv_(Vn, 0):  # this gate for both below? Or we need a separate gv_ below again?
```
boris-kz, line 542: probably both, only one of them can be taken.

kwcckw: Okay, but if we have another gate below, the else section will be considered as the blocking neg value? But it's actually running cluster_N too?

boris-kz: It's blocking one of the forks, I guess the gating gain here is val_selected_fork - val_blocked_fork
actually it's proj_gain - oF.fc, but right now we only have the fc

frame_2D_alg/agg_recursion.py
boris-kz: not directly related, but we seem to be missing summation of compared scalars in dTT? 
I have that "I" on a pixel level, and c across cluster levels, but no sum weighted scalars across dTT? 
match/miss ratio needs scaling by that [sum( sum(dTT*wTT) for dTT in dTT_] *  c?

Claude:
This is a code-specific question about `dTT`/`wTT`/`I`/`c` and the actual match/miss computation — I shouldn't reason about it from the abstract thread. Let me read the actual files.Read the code. Your read is right, and the locus is `vt_` (meta_code 17–21):

`t_ = m_+|d_|` (≈ max comparand, line 19); `m = (m_/t_) @ wTT[0]`. So `vt_` returns *relative* match — min/max per param, weighted — magnitude-blind by construction. The eval gates (agg 316, 348, 356, 559…) then multiply that ratio only by element count, `len(N_)`, or in a few (854, 840) by `C=Σc` — never by the magnitude of the compared params. Two clusters with equal count and equal match-ratio therefore score equally whether they match on large I,G or tiny ones. That absolute param-magnitude is the missing term, and it contradicts your own docstring, which sets fitness as *summed* match, not relative. So I and c exist as accumulators on the nodes, and even enter dTT as params — but `vt_` relativizes every one of them, so the absolute scale is divided out at eval, not at storage. That's the actual gap.

Three corrections to your formula:

1. Don't sum `dTT*wTT`. `dTT[1]` is signed (`comp_derT`: `d_ = _i_ - i_`), so opposite diffs cancel — and summing the match row into the signed-diff row is incoherent. The comparand magnitude you want is `m_+|d_|` per param, which is `t_`, already computed at line 19 and divided out. Re-inject `t_`; don't rebuild it externally.

2. `×c` is mostly redundant. c is element-mass, and `sum_vt` (meta_code 23) already stores a c-weighted *average* dTT (`w=n.c/C`), not a sum — so absolute mass is average × C. That's what your `×c` gropes toward, but it's `×C`, not per-element c, and gates already carrying C or len shouldn't also `×c` or you double the count.

3. Decide what you want, because they diverge. Absolute summed match — your docstring's fitness — is just un-normalized `vt_`: `Σ m_·w`, i.e. drop the `/t_` on the value path. "Ratio × mass" is a *different* quantity (a product of two sums, not `Σ m_·w`); your formula approximates that one, the docstring says the former.

Minimal diff: give `vt_` a flag to return absolute match (skip `/t_`) alongside the relative pair; use absolute × count in the value gates, keep relative for membership.

boris-kz: Initial c=P.L is a simplification. I doesn't have a mass on its own,  that would be m_.
So it's c *= Σ m_ * w: c is summed across all composition levels anyway, not level-specific

...
Agree on m_ — match is the mass. That's more than a fix, it's the right identification: predictive mass = shared quantity = match, never I. L was the placeholder, m_ is principled. And for the *value/gate* path, `c = Σ m_·w` is exactly right — it re-injects the magnitude the gates were missing. So folding it in helps there, and my blanket "don't" last turn was too strong.

The one place it breaks is `rn = C/_C` in base_comp, and it's not about levels. rn rescales the prior node's *non-match* params (`_I,_G,_L`) by the mass ratio to make the two commensurable — that only works if C is *size*, not match-mass. Counterexample: two nodes, same size, same I, but match-mass 50 vs 5. With size, rn=1 and the I's compare equal — correct. With match-mass, rn=0.1 and you'd read one node's intensity as 10× the other's though they're identical. Same size, different match quality → rn distorts every non-match comparison.

So it's not "don't change c," it's *split*: `c = Σ m_·w` (predictive mass, cumulative, for value) and a separate count/extent — your current L — for rn's size-normalization. One variable can't carry both; they diverge exactly when match-per-element varies, which is the interesting case.

kwcckw: > Initial c=P.L is a simplification. I doesn't have a mass on its own, that would be m_. So it's c *= Σ m_ * w: c is summed across all composition levels anyway, not level-specific

I see, and `c *= Σ m_ * w` , this w is from wTT? 

Also this is applicable to sum_vt only?

frame_2D_alg/agg_recursion.py
kwcckw: In this eval, the projected gain is from G? Which we can compute it from proj_N?

boris-kz: projected gain for cluster_N is

we need a separate eval for cluster_C, that if >ave 
becomes if Vn * (wcC-wcN) > ave * (ccC-ccN)

kwcckw: I see, then it's really 2 gv_ for different gates here, but gv_ for cluster_C is nested, which it should use `i = 1`.

boris-kz: right:

    if G.Lt:  # sub+
        if gv_(Vn := (lm+ld)* (wcN*L) - (ave+avd)* (lr+1+ccN*L), 0):  # default cluster_N
            c = G.Lt.c; E_ = get_exemplars({N for L in L_ for N in L.N_}, r,c)
            if gv_(Vn * (wcC-wcN) *(mdecay(L_)-decay) - (ave+avd)*(lr+1+(ccC-ccN)*L), 1):
                r+=1; G_,r = cluster_C(G.Nt,E_,r,c)  # higher V, low decay, eval cluster_P
            else:     G_,r = cluster_N(G.Nt,E_,r,c)  # updates G
            L= (len(G_)-1) **2  # if full cross_comp?
            if G_ and gv_(vt_(G.Nt.dTT,G.wTT*tAgg)[0]*(wAgg*L) - ave*(r+cAgg*L),2):
                cross_comp(G.Nt,r)

frame_2D_alg/agg_recursion.py
boris-kz: It might be just for comp_latuple, then c already represents base match?

kwcckw: From the existing comp_latuple:

      def comp_latT(_latT, latT, _n,n):  # 0der params, add align?
      
          _I,_G,_Dy,_Dx,_M,_D,_L = _latT
          I, G, Dy, Dx, M, D, L = latT
          rn = _n / n
          _pars = np.array([_M,_D,_I,_G, np.array([_Dy,_Dx]),_L], dtype=object)
          pars  = np.array([ M, D, I, G, np.array([ Dy, Dx]), L], dtype=object) * rn
          pars[2] = [pars[2],aI]  # no avd*rn: d/=t
          m_,d_ = comp(_pars,pars)
          return (np.array([m_,d_]),  # verT
                  m_@ w_t[0],         # m
                  np.abs(d_)@ w_t[1], # d
                  min(L,_L))          # c


`c` will be just*sum(m_) * w? Then yes, it's actually same as the m now.

boris-kz: right, so we don't need min(L,_L))          # c

frame_2D_alg/meta_code.py
```diff
@@ -152,13 +152,16 @@ def inner(*a, **kw):
             _CoF = CoF._cur.get(None)
             oF = CoF(nF=iF_[func.__name__], root=_CoF)
             i = iF_[func.__name__]; oF_[i].call_ += [oF]  # F_call_T_[i][oF.nF] += oF.dTT
+            oF.gv_ = np.zeros(oF_[i].gi)
```
boris-kz, line 155: so we don't need the actual call sites, just their sequence?

kwcckw: Yes, and that can be tracked by checking for the gv_ function when we build oF body.

frame_2D_alg/agg_recursion.py
boris-kz: So I removed all cs from comp_slice, we have M and L instead, please check how it works downstream

kwcckw: Okay, so we use M in computing rc instead of c:

    rc = M/_M; minn = min(_M,M)

and this is only applicable to comp_slice, specifically in latuple only?

boris-kz: Actually I am not sure. c is a normalizing factor because it's a comparison count, not a mass.
M can't be normalized by itself. So hold off on changing downstream.

frame_2D_alg/meta_code.py
```diff
@@ -152,13 +152,16 @@ def inner(*a, **kw):
...
             if _CoF is not None:
                 _CoF.call_ += [oF]
                 oF_[iF_[func.__name__]].caller_.add(_CoF)  # for comp_caller_
             _oF = CoF._cur.set(oF)
             if out := func(*a, **kw):
                 C = oF.c; TT=np.zeros((2,9)); R=0
-                for tt,c,r in oF.vt_: w= c/C; TT+=tt*w; R+=r*w
+                for tt,c,r in oF.vt_: 
+                    if func.__name__ in ('comp_slice', 'slice_edge'): continue   # skip for now since incompatible TT
```
boris-kz, line 163: Just exclude 'comp_slice', 'slice_edge' from oF_, they are non-core anyway?

kwcckw: Alright, I  will remove them later.

frame_2D_alg/agg_recursion.py
boris-kz: claude
Yes — that's the category error, and it's more fundamental than the rn and link/node-site issues I raised. c is a denominator. `M/c` is the rate (comp_dP_: `rM = M/c`); c is the count that produces it. Set c = M and M/c = 1 — a normalizer can't be its own numerator. So c stays a count, necessarily.

And it generalizes: match is always the numerator, with two denominators — magnitude (`t_ = m_+|d_|`, what vt_ divides by) and count (c, what M/c divides by). Neither can be the match. So the absolute substance you were trying to load into c is just the numerator un-normalized: M, the summed match, already there. You don't move magnitude into c; you read M directly where you want absolute, M/c or vt_ where you want relative. `c = Σ m_·w` was loading the numerator into the denominator — drop it.

kwcckw: > So the absolute substance you were trying to load into c is just the numerator un-normalized: M, the summed match,

We don't always have `m_` to get the un-normalized M?

boris-kz: m_ is dTT[0], * c

PR #181, opened 2026-06-20
6/20/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -698,6 +699,8 @@ def PP2N(PP):
     if G_:
         Nt = CF(nF='Nt', root=tile); Nt.N_=G_; Nt.dTT=TT; Nt.c=C; Nt.r=1; Nt.H=tile.H; tile.Nt=Nt; tile.dTT=TT; tile.c=C; L=len(G_)
         if vt_(TT,ttVct)[0]*(wFrm*L) > ave * (cFrm*L):  # L for trans-comp only?
+            A_ = [G.angl[0] for G in G_ if G.angl]
+            tile.angl = [np.sum(A_, axis=0) if A_ else np.zeros(2), np.sign(tile.dTT[1] @ ttVct[1])]
```
kwcckw, line 703: Looks like we need tile.angl for proj_focus:

      def proj_focus(PV__, y,x, tile):  # radial accum of projected focus value in PV__
      
          m,d = vt_(tile.dTT, tile.wTT * ttFrm); c = tile.c  # m,d,n = tile.m, tile.d, tile.c  # add r?
          V = (m-ave + d-avd) * c
          Dec = decay ** (np.hypot(*(tile.box[2:]-tile.box[:2])) / wYX)
          H,W = PV__.shape  # = win__
          n = 1  # radial distance

frame_2D_alg/agg_recursion.py
```diff
@@ -659,13 +659,14 @@ def vect_edge(tile, rV=1):  # PP_ cross_comp and floodfill to init focal frame g
...
     def PP2N(PP):
-        P_,L_,B_,verT,latT,A,S,box,yx, m,d,c = PP
+        P_,L_,B_,verT,latT,A,S,box,yx, m,d = PP
         kern = np.array(latT[:4])
         [mM, mD, mI, mG, mA, mL], [dM, dD, dI, dG, dA, dL] = verT  # re-pack in dTT:
         dTT = np.array([ np.array([mM, mD, mL, mI, mG, mA, mL, mL / 2, 0]),  # extA=0
                          np.array([dM, dD, dL, dI, dG, dA, dL, dL / 2, 0])])
         y,x,Y,X = box; dy,dx = Y+1-y, X+1-x
         A = [np.array(A), np.sign(dTT[1] @ ttVct[1])]  # append sign
+        c = sum(P.latT[-1] for P in P_) if hasattr(P_[0], 'latT') else sum(P.latT[-1] for dP in P_ for P in dP.nt)
```
kwcckw, line 669: Without the prior c, and without m_ here, i'm not sure if the initial c should be init with P.L? Or just 1?

boris-kz: I reverted on c, so this change is not needed

frame_2D_alg/agg_recursion.py
```diff
@@ -681,9 +682,9 @@ def PP2N(PP):
     blob_ = tile.N_; G_,TT,C,R = [],np.zeros((2,9)),0,0
     for blob in blob_:
         if not blob.sign:
-            if gv_(blob.G * wSE - ave * cSE, 0):
+            if gv_(blob.G * wVct - ave * cVct, 0):  # with removed comp_slice and slice_edge, using w and c from vect_edge here?
```
boris-kz, line 685: right

frame_2D_alg/agg_recursion.py
```diff
@@ -115,9 +115,9 @@ def proj_V(_N,N, dist, dy_dx, dec):  # _N x N induction
         else: break  # beyond initial induction range, re-sort by proj_V?
     for N in set(N_):
         if N.rim: N.Rt = sum2F(N.rim)
-    Fvt_(L_, *sum_vt(L_))  # pos + neg vals for oF
+    Lt = sum_vt(L_); Fvt_(L_,*Lt)  # pos + neg vals for oF
     pL_ = [L for L in L_ if (L.m>ave or L.typ==-1)]
-    return pL_, *sum_vt(pL_)  # oF.N_=L_, !pLs  (return only pLs and pL's values? )
+    return pL_, *sum_vt(pL_), (L_, *Lt)  # oF.N_=L_, !pLs  (return only pLs and pL's values? )
```
kwcckw, line 120: So cross_comp's oF should be using the same N_, TT, c and r here, which we need to return the same L_, and sum_vt(L_)?

boris-kz: Not the same because cross_comp is combining from comp_N_ anf comp_C_, but yes same return form

kwcckw: But both comp and cross_comp's oF should be both positive and negative Ls?

boris-kz: yes, it's just not fully redundant

kwcckw: Okay, so it should parse the Lt above to cross_comp? Since that Lt is actually summed from Ls instead of just pLs.

frame_2D_alg/agg_recursion.py
```diff
@@ -50,9 +50,9 @@
 capitalized vars are summed small-case vars
```
boris-kz, line 50: I am trying to figure how to make LLMs more thorough in tracing the code, it seems that we need external scaffold for that:

This is a code-file task, so I'll write a runnable scaffold rather than describe one. It implements the loop exactly as designed: deterministic control flow (the loop decides when to stop, not the model), ripgrep as the recall floor, optional jedi unioned in, model-proposed couplings folded back through the same search machinery, and a fresh per-site model call doing both the precision filter and the edit.Run it with:

```
export ANTHROPIC_API_KEY=...
python trace_implications.py --files <dir-or-files> --symbol cluster_P \
  --change "now returns signed vertical residuals" [--jedi] [--model claude-opus-4-8]
```

Three design points are load-bearing in the code. The stop decision is the `while frontier` loop, never the model — it halts when nothing is unexplored, with `closed` (keyed on site + upstream symbol) as the real terminator and depth/call caps as backstops. Enumeration is ripgrep at the floor; model-proposed couplings become *additional grep terms* (semantic feed folded into the same machinery); jedi only unions in. The per-site `judge` call is fresh and does double duty — `related` filters ripgrep's over-inclusion (model = precision), `needs_change`/`edit` produces the change, and `induced_symbol`/`induced_change` is the recursion edge that pushes the next ripple onto the frontier.

Two things to know. It *proposes* edits, doesn't apply them — so all edits are computed against the original snapshot and aren't sequenced; it's a planner, and the JSON output is what you'd feed to whatever applies them. And the coupling feed is the one place coverage isn't guaranteed (the ripgrep floor is exhaustive over textual references; behavioral couplings depend on the model thinking of them) — which is exactly the residue you'd harden over time by freezing found couplings into assertions. Default model is sonnet for throughput; switch to opus where edit fidelity to your idiom matters.

#!/usr/bin/env python3
"""
trace_implications.py - frontier-driven consequence tracer for entangled code.

Given a change to one symbol, exhaustively trace and propose every downstream
edit it forces, by treating propagation as an EXPLICIT SEARCH whose stopping
decision lives in the loop, not in the model.

Per frontier item (an upstream change to `symbol`):
  1. Enumerate candidate sites:
        ripgrep(symbol)                         <- recall floor, exhaustive
      U ripgrep(model-proposed coupling terms)  <- semantic feed, folded into grep
      U jedi references(symbol)   [optional]    <- additive scope-resolved hits
     The smart tools only ADD candidates; they never prune the ripgrep set.
  2. For each candidate, a FRESH model call judges:
        - is this site genuinely affected? (precision: filters grep noise)
        - does it need to change given the upstream change?
        - if yes: the edit, plus the new (symbol, change) the edit itself induces
  3. Dedup on (candidate site, upstream symbol); push induced changes; loop until
     the frontier is empty. The model never says "done" - the loop does.

You get the full propagation tree for free via provenance links.

Requires: ripgrep (`rg`) on PATH, `anthropic` SDK, ANTHROPIC_API_KEY in env.
Optional: `jedi` (pip install jedi), enabled with --jedi.

Usage:
  export ANTHROPIC_API_KEY=...
  python trace_implications.py \
      --files src/ \
      --symbol cluster_P \
      --change "now returns signed vertical residuals instead of unsigned magnitudes"
"""

import argparse
import ast
import json
import os
import shutil
import subprocess
import sys
from collections import deque
from dataclasses import dataclass, field, asdict
from glob import glob
from typing import Optional

from anthropic import Anthropic

# Judge/edit/coupling all run here. These are LOCAL, bounded calls, so a mid model
# is usually fine; switch to claude-opus-4-8 if edit fidelity to your idiom matters.
MODEL = "claude-sonnet-4-6"

# Safety bounds (cost + termination backstops; the closed-set is the real terminator).
DEFAULT_MAX_DEPTH = 6
DEFAULT_MAX_CALLS = 500
COUPLING_MAX_DEPTH = 2      # only mine semantic couplings near the root
MAX_COUPLINGS = 5           # cap coupling search-terms per item

client = Anthropic()


# --------------------------------------------------------------------------- #
# Data
# --------------------------------------------------------------------------- #
@dataclass
class Match:
    path: str
    line: int
    snippet: str
    via: str = "direct"     # why this candidate surfaced (direct / coupling / jedi)


@dataclass
class Item:
    symbol: str             # identifier whose semantics changed
    change: str             # natural-language description of the property change
    origin: str             # "root" or "<path>:<line>" that spawned this change
    parent: Optional[str]   # parent symbol, for the provenance tree
    depth: int = 0


@dataclass
class RecordedEdit:
    upstream_symbol: str
    upstream_change: str
    path: str
    line: int
    via: str
    reason: str
    edit: str
    induced_symbol: str
    induced_change: str
    depth: int
    parent: Optional[str]


# --------------------------------------------------------------------------- #
# Model plumbing
# --------------------------------------------------------------------------- #
def _extract_json(text: str):
    text = text.strip()
    if text.startswith("```"):
        text = text[3:]
        if text[:4].lower() == "json":
            text = text[4:]
        text = text.rsplit("```", 1)[0]
    text = text.strip()
    try:
        return json.loads(text)
    except json.JSONDecodeError:
        # salvage: slice from first bracket to its last matching closer
        starts = [i for i in (text.find("{"), text.find("[")) if i != -1]
        if not starts:
            raise
        s = min(starts)
        closer = "}" if text[s] == "{" else "]"
        e = text.rfind(closer)
        return json.loads(text[s:e + 1])


def _call_json(system: str, user: str, max_tokens: int = 2000):
    resp = client.messages.create(
        model=MODEL, max_tokens=max_tokens, system=system,
        messages=[{"role": "user", "content": user}],
    )
    text = "".join(b.text for b in resp.content if b.type == "text")
    return _extract_json(text)


# --------------------------------------------------------------------------- #
# Candidate enumeration
# --------------------------------------------------------------------------- #
def run_ripgrep(term: str, files: list[str], via: str) -> list[Match]:
    """Exhaustive word-boundary search. Over-includes by design; the model filters."""
    cmd = ["rg", "--json", "-w", "--", term, *files]
    try:
        out = subprocess.run(cmd, capture_output=True, text=True, timeout=120).stdout
    except (subprocess.TimeoutExpired, FileNotFoundError):
        return []
    matches = []
    for raw in out.splitlines():
        try:
            obj = json.loads(raw)
        except json.JSONDecodeError:
            continue
        if obj.get("type") != "match":
            continue
        d = obj["data"]
        matches.append(Match(
            path=d["path"]["text"],
            line=d["line_number"],
            snippet=d["lines"]["text"].rstrip("\n"),
            via=via,
        ))
    return matches


def jedi_references(term: str, files: list[str]) -> list[Match]:
    """Best-effort scope-resolved references, unioned in. Adds only; never prunes."""
    try:
        import jedi
    except ImportError:
        return []
    out: list[Match] = []
    root = os.path.commonpath([os.path.abspath(f) for f in files]) if files else "."
    try:
        project = jedi.Project(root)
    except Exception:
        project = None
    for f in files:
        try:
            src = open(f).read()
        except OSError:
            continue
        # find a position of `term` in this file to anchor jedi, then resolve refs
        for ln, line in enumerate(src.splitlines(), start=1):
            col = line.find(term)
            if col == -1:
                continue
            try:
                script = jedi.Script(code=src, path=f, project=project)
                for ref in script.get_references(line=ln, column=col + 1,
                                                 include_builtins=False):
                    if ref.module_path:
                        out.append(Match(str(ref.module_path), ref.line,
                                         (ref.get_line_code() or "").rstrip(),
                                         via="jedi"))
            except Exception:
                pass
            break  # one anchor per file is enough; refs are project-wide
    return out


def enclosing_context(path: str, line: int, window: int = 25) -> str:
    """Innermost enclosing def/class around `line`, else a line window."""
    try:
        src = open(path).read()
    except OSError:
        return ""
    lines = src.splitlines()
    try:
        tree = ast.parse(src)
        best = None
        for node in ast.walk(tree):
            if isinstance(node, (ast.FunctionDef, ast.AsyncFunctionDef, ast.ClassDef)):
                end = node.end_lineno or node.lineno
                if node.lineno <= line <= end:
                    if best is None or (end - node.lineno) < (best[2] - best[1]):
                        best = (node, node.lineno, end)
        if best:
            node, lo, hi = best
            body = "\n".join(lines[lo - 1:hi])
            return (f"# {path}: enclosing {type(node).__name__} "
                    f"`{node.name}` (lines {lo}-{hi})\n{body}")
    except SyntaxError:
        pass
    lo = max(0, line - window)
    hi = min(len(lines), line + window)
    return f"# {path}: lines {lo + 1}-{hi}\n" + "\n".join(lines[lo:hi])


# --------------------------------------------------------------------------- #
# Model judgments
# --------------------------------------------------------------------------- #
COUPLING_SYS = (
    "You analyze a code change for BEHAVIORAL couplings that a textual symbol "
    "search would miss: code that depends on an implicit contract (ordering, "
    "shape, units, mutation timing, invariants) of the changed thing without "
    "referencing its name directly. Output ONLY a JSON array of objects "
    '{"symbol": "<identifier to grep for>", "reason": "<one line>"}. '
    "Identifiers must be real names likely present in this codebase. "
    "Empty array if none. No prose, no code fences."
)


def propose_couplings(item: Item) -> list[tuple[str, str]]:
    user = (f"Changed symbol: {item.symbol}\n"
            f"What changed: {item.change}\n\n"
            "List identifiers elsewhere whose behavior may break from this, "
            "via an implicit contract rather than a direct reference.")
    try:
        data = _call_json(COUPLING_SYS, user, max_tokens=600)
    except Exception:
        return []
    pairs = []
    for d in data if isinstance(data, list) else []:
        sym = (d.get("symbol") or "").strip()
        if sym and sym != item.symbol:
            pairs.append((sym, (d.get("reason") or "").strip()))
    return pairs[:MAX_COUPLINGS]


JUDGE_SYS = (
    "You judge whether ONE code site must change in response to an upstream "
    "change elsewhere. Be precise: many candidate sites are coincidental name "
    "matches and are NOT affected. Output ONLY this JSON object, no prose, no "
    "fences:\n"
    "{\n"
    '  "related": <bool, is this site genuinely coupled to the upstream change>,\n'
    '  "needs_change": <bool>,\n'
    '  "reason": "<one or two sentences>",\n'
    '  "edit": "<the concrete code change, or empty string if none>",\n'
    '  "induced_symbol": "<identifier at THIS site whose semantics the edit now '
    'changes, or empty>",\n'
    '  "induced_change": "<what that new property change is, or empty>"\n'
    "}\n"
    "induced_symbol/induced_change describe the NEXT ripple this edit creates; "
    "leave them empty if the change stops here."
)


def judge(item: Item, m: Match, context: str) -> dict:
    user = (
        f"UPSTREAM CHANGE\n  symbol: {item.symbol}\n  what changed: {item.change}\n"
        f"  origin: {item.origin}\n\n"
        f"CANDIDATE SITE (surfaced via: {m.via})\n  {m.path}:{m.line}\n\n"
        f"CANDIDATE CONTEXT\n{context}\n\n"
        "Does this site need to change given the upstream change?"
    )
    try:
        v = _call_json(JUDGE_SYS, user, max_tokens=2000)
    except Exception as e:
        return {"related": False, "needs_change": False,
                "reason": f"judge error: {e}", "edit": "",
                "induced_symbol": "", "induced_change": ""}
    return v


# --------------------------------------------------------------------------- #
# The loop
# --------------------------------------------------------------------------- #
def trace(symbol: str, change: str, files: list[str], use_jedi: bool,
          max_depth: int, max_calls: int):
    root = Item(symbol=symbol, change=change, origin="root", parent=None, depth=0)
    frontier: deque[Item] = deque([root])
    closed: set[tuple] = set()          # (path, line, upstream_symbol) already judged
    edits: list[RecordedEdit] = []
    stats = {"iterations": 0, "calls": 0, "candidates": 0, "capped": False}

    while frontier:
        item = frontier.popleft()
        stats["iterations"] += 1

        # build search terms: the symbol itself + (near root) model-proposed couplings
        terms = [(item.symbol, f"direct:{item.symbol}")]
        if item.depth <= COUPLING_MAX_DEPTH:
            for sym, reason in propose_couplings(item):
                stats["calls"] += 0  # coupling call counted separately below if you wish
                terms.append((sym, f"coupling:{sym} ({reason})"))

        # enumerate candidates (recall floor + optional jedi), dedup within this item
        seen_here: set[tuple] = set()
        candidates: list[Match] = []
        for term, via in terms:
            hits = run_ripgrep(term, files, via)
            if use_jedi:
                hits += jedi_references(term, files)
            for m in hits:
                sig = (m.path, m.line)
                if sig in seen_here:
                    continue
                seen_here.add(sig)
                candidates.append(m)

        for m in candidates:
            key = (m.path, m.line, item.symbol)
            if key in closed:
                continue
            closed.add(key)

            if stats["calls"] >= max_calls:
                stats["capped"] = True
                frontier.clear()
                break

            context = enclosing_context(m.path, m.line)
            v = judge(item, m, context)
            stats["calls"] += 1
            stats["candidates"] += 1

            if v.get("related") and v.get("needs_change"):
                induced_sym = (v.get("induced_symbol") or "").strip()
                induced_chg = (v.get("induced_change") or "").strip()
                edits.append(RecordedEdit(
                    upstream_symbol=item.symbol, upstream_change=item.change,
                    path=m.path, line=m.line, via=m.via,
                    reason=v.get("reason", ""), edit=v.get("edit", ""),
                    induced_symbol=induced_sym, induced_change=induced_chg,
                    depth=item.depth, parent=item.parent,
                ))
                if induced_sym and induced_chg and item.depth < max_depth:
                    frontier.append(Item(
                        symbol=induced_sym, change=induced_chg,
                        origin=f"{m.path}:{m.line}", parent=item.symbol,
                        depth=item.depth + 1,
                    ))
    return edits, stats


# --------------------------------------------------------------------------- #
# Reporting
# --------------------------------------------------------------------------- #
def print_report(symbol, change, edits, stats):
    print(f"\nROOT CHANGE: `{symbol}` - {change}\n")
    if not edits:
        print("No downstream edits proposed.")
    for e in edits:
        pad = "  " * (e.depth + 1)
        print(f"{pad}- {e.path}:{e.line}  [{e.via}]")
        print(f"{pad}  reason: {e.reason}")
        if e.edit:
            first = e.edit.strip().splitlines()[0] if e.edit.strip() else ""
            extra = "" if e.edit.strip().count("\n") == 0 else "  ..."
            print(f"{pad}  edit:   {first}{extra}")
        if e.induced_symbol:
            print(f"{pad}  -> induces `{e.induced_symbol}`: {e.induced_change}")
    print(f"\nfrontier closed in {stats['iterations']} iterations | "
          f"{stats['candidates']} sites judged | {stats['calls']} model calls | "
          f"{len(edits)} edits proposed"
          + ("  [CALL CAP HIT - result is partial]" if stats["capped"] else ""))


def expand_files(patterns: list[str]) -> list[str]:
    out: list[str] = []
    for p in patterns:
        if os.path.isdir(p):
            out += glob(os.path.join(p, "**", "*.py"), recursive=True)
        else:
            out += [g for g in glob(p, recursive=True)]
    return sorted(set(f for f in out if f.endswith(".py")))


def main():
    ap = argparse.ArgumentParser(description=__doc__,
                                 formatter_class=argparse.RawDescriptionHelpFormatter)
    ap.add_argument("--files", nargs="+", required=True,
                    help="files, globs, or directories (recursed for *.py)")
    ap.add_argument("--symbol", required=True, help="changed identifier")
    ap.add_argument("--change", required=True, help="what changed about it")
    ap.add_argument("--jedi", action="store_true", help="union jedi references in")
    ap.add_argument("--model", default=MODEL, help=f"override model (default {MODEL})")
    ap.add_argument("--max-depth", type=int, default=DEFAULT_MAX_DEPTH)
    ap.add_argument("--max-calls", type=int, default=DEFAULT_MAX_CALLS)
    ap.add_argument("--out", default="propagation.json")
    args = ap.parse_args()

    global MODEL
    MODEL = args.model

    if not shutil.which("rg"):
        sys.exit("error: ripgrep (`rg`) not found on PATH.")
    if not os.environ.get("ANTHROPIC_API_KEY"):
        sys.exit("error: ANTHROPIC_API_KEY not set.")

    files = expand_files(args.files)
    if not files:
        sys.exit("error: no .py files matched --files.")
    print(f"searching {len(files)} file(s) with model {MODEL}"
          + (" + jedi" if args.jedi else ""))

    edits, stats = trace(args.symbol, args.change, files,
                         args.jedi, args.max_depth, args.max_calls)
    print_report(args.symbol, args.change, edits, stats)

    with open(args.out, "w") as fh:
        json.dump({"root": {"symbol": args.symbol, "change": args.change},
                   "edits": [asdict(e) for e in edits], "stats": stats}, fh, indent=2)
    print(f"\nfull trace written to {args.out}")


if __name__ == "__main__":
    main()

frame_2D_alg/agg_recursion.py
boris-kz: yes, thanks

frame_2D_alg/agg_recursion.py
boris-kz: Chat:
Yes. What you sketched is basically a **worklist / fixed-point change-impact analysis** loop.

Standard names:

- Change impact analysis: find all code locations potentially affected by a change.

- Program slicing: compute code that can affect, or be affected by, a value / variable / field.

- Data-flow analysis: track how values propagate through assignments, calls, returns, containers, attributes. CodeQL describes this directly as computing how possible values propagate through a program. ([[CodeQL](https://codeql.github.com/docs/writing-codeql-queries/about-data-flow-analysis/?utm_source=chatgpt.com)][1])

- Taint analysis: same general mechanism, but phrased as source → propagator → sink. Semgrep’s taint mode is a practical version of this: assignments and function calls propagate a marked property until it reaches a sink. ([[Semgrep](https://semgrep.dev/docs/writing-rules/data-flow/taint-mode/overview?utm_source=chatgpt.com)][2])

- Static analysis over AST / CST: parse code, query structural references, track syntax-preserving edits. Tree-sitter is commonly used for incremental concrete syntax trees, which is useful for repeated edit/trace loops. ([[Tree-sitter](https://tree-sitter.github.io/?utm_source=chatgpt.com)][3])

- Gradual typing / contract checking: encode part of the expected field/function contract so propagation failures become type/check failures. Mypy is the standard Python example: static type checking without running the program. ([[mypy](https://mypy.readthedocs.io/?utm_source=chatgpt.com)][4])

So your harness is not exotic. The nonstandard part is using an LLM for the **semantic coupling expansion** step.

The standard architecture would be:

```python
frontier = changed facts
while frontier:
    fact = frontier.pop()
    affected = static_dependency_query(fact)
    affected += semantic_dependency_query(fact)  # LLM or custom rules
    for site in affected:
        new_facts = transfer(site, fact)
        frontier += new_facts - known_facts
```

That is a data-flow fixed point.

For CogAlg, the useful standard pieces are:

1. **AST / CST reference index**

   Use Python `ast`, maybe Tree-sitter if you want more robust source-preserving edits. This finds textual/structural uses: `N.rim`, `.Nt`, `.N_`, `sum2F(...)`, `Copy_(...)`, etc.

2. **Call graph**

   Direct calls are easy. Dynamic calls and method/property aliases need approximations.

3. **Def-use / use-def graph**

   Tracks assignment and read chains. Example: `CN.rim` is not just a field; in `meta_code`, it is a property alias backed by `Rt.N_`, so changing `rim` implies checking `Rt`, `CF.N_`, `sum2F`, `Copy_`, and `add2F`.

4. **Program slicing**

   Given a changed property, compute forward slice: what reads it downstream. Also backward slice: what constructs it upstream.

5. **Type/contract layer**

   Even lightweight annotations help. Not necessarily full mypy compliance. You can define runtime assertions for “rim is list of CL” vs. “rim is fork-indexed container” and let failures reveal missed propagation.

6. **Semantic grep / pattern rules**

   Semgrep-style rules can catch recurring idioms: “function copies `N_` but not `L_`”, “temporary flags not cleared”, “fork alias read without corresponding fork update”.

7. **LLM as missing-edge proposer**

   The LLM should not replace static analysis. It proposes couplings the static graph cannot know: “changing `rim` also affects exemplar inhibition because `get_exemplars` treats rims as inhibition zones.”

The closest off-the-shelf tools are:

* **[CodeQL](https://codeql.github.com/?utm_source=chatgpt.com)**: best fit for custom data-flow queries over a codebase. It treats code as queryable data and has Python data-flow libraries. ([CodeQL][5])
* **[Semgrep](https://semgrep.dev/docs/writing-rules/data-flow/taint-mode/overview?utm_source=chatgpt.com)**: easier rule-writing for structural and taint-like patterns, less suitable for deep whole-program custom semantics. ([Semgrep][2])
* **mypy / pyright**: useful only after you encode contracts in types.
* **Tree-sitter / LibCST**: useful for reliable parsing and rewriting, not semantic closure by itself.

But there is probably no standard tool that will understand CogAlg’s real invariants. Your fields are semantically overloaded: `N_`, `L_`, `rim`, `root`, `root_`, `Nt`, `Lt`, `Ct`, `Bt`, temporary flags, and fork aliases. Static tools will find references; they will not know which references are conceptually coupled.

So the practical answer is:

Use standard **change impact analysis + program slicing + dataflow worklist** as the harness.

Use custom CogAlg contracts for object lifecycle.

Use the model only for:

```text
static graph missed this because the coupling is semantic
```

not for:

```text
find all references
```

The harness name I would use in your proposal is:

**hybrid change-impact fixed-point loop**

or more concretely:

**static slicing + semantic worklist propagation**.

[1]: https://codeql.github.com/docs/writing-codeql-queries/about-data-flow-analysis/?utm_source=chatgpt.com "About data flow analysis - CodeQL - GitHub"
[2]: https://semgrep.dev/docs/writing-rules/data-flow/taint-mode/overview?utm_source=chatgpt.com "Taint analysis overview"
[3]: https://tree-sitter.github.io/?utm_source=chatgpt.com "Tree-sitter: Introduction"
[4]: https://mypy.readthedocs.io/?utm_source=chatgpt.com "mypy 2.1.0 documentation"
[5]: https://codeql.github.com/?utm_source=chatgpt.com "CodeQL"

kwcckw: I see, sometimes LLM may missed certain changes when tracing the code, so this should be helpful to tell them where to look first.

But this is actually partly using LLM in the model judgement section, and that part prone to error too?

boris-kz: LLMs now use all kinds of coding tools, that track the code precisely. I think it's just a matter making them work harder, not sure how, I already use max effort.

kwcckw: I see, that probably could be improved by adding the relevant skills, under customize?

boris-kz: It seems that we do need Claude Code, it's specifically trained to maintain long reasoning chains via partial execution:

...
You're right, and it revises what I said. The split I drew — reasoning is the core and harness-neutral, execution is peripheral — is wrong for *stateful* tracing, and that's most of your code. Correcting it properly:

Why focus is lost: deep tracing of mutation-entangled code is symbolic execution done in the head. The model has to hold the evolving state — what's aliased to what, what's been mutated, in what order — across the whole chain. That state-model is the single hardest thing to keep in working memory, and it's exactly what decays. The drift you see isn't the inference failing; it's the *state representation* rotting over the length of the trace, after which the model is reasoning about a state it has begun to confabulate.

How that squares with "same model, same ceiling": the ceiling is the same, but unaided the model *spends* that ceiling holding decaying state, and drifts. Running the code externalizes the state — the model observes it instead of maintaining it — so the fixed budget goes to inference rather than bookkeeping. Same ceiling, reallocated from remembering to thinking. So execution isn't an add-on here; it's the substrate that frees the reasoning. I under-weighted that.

Your phrase names the mechanism, and there's a fidelity spectrum to it. Pure abstract trace: state in the head, drifts. A written hand-simulation — the model forced to spell out the evolving state step by step — externalizes it into tokens and helps, but the model can write down a *wrong* state; it's a confabulatable simulacrum. Real execution externalizes *and* grounds: it can't confabulate what actually got mutated. The realer the simulation, the more focus and the less invention — and the limit is just running it. So the fidelity of the simulacrum is what determines how much it buys you.

Two bounds keep it from being unconditional. Execution anchors only the path you run — and obscure couplings cluster on rare, unexecuted branches (part of why they're obscure), so those still drift; the skill becomes driving inputs to light up the branches you care about. And the intentional-semantic couplings — signed-versus-unsigned where the consumer mishandles it — are simulation-blind and worse: the run produces a plausible value, giving *false* focus that masks the intent-violation. That residue needs abstract reasoning against a *stated* contract, which no simulation supplies.

So yes — major, and it does upgrade the harness case: for the stateful majority of CogAlg, the ability to run mid-trace is closer to an enabling condition than a convenience. But it's the *capability* — execute and observe inside the trace — that matters, which any agent loop gives you, including your own scaffold with a run tool, not Claude Code specifically. The core abstract ceiling, for the un-runnable intent-level residue, is unchanged.

boris-kz: This is all wrong again, 
we just need plain Claude, plus instructions to check thoroughly, plus these invariants:

# agg_recursion / meta_code — structural invariants

The contracts the aliased-mutation couplings depend on, extracted from the class
definitions and the sum/merge/copy primitives. A change to one of these objects is
safe iff it preserves these. Each is anchored to the code that establishes it.
Where your own comments flag a point as unsettled, it's marked **(unsettled)**.

## 1. Storage lives in the forks, not on the node

CN's `N_, C_, L_, B_, X_, rim, H` are **property views** onto fork objects, not stored
attributes (`prop_F_`, class body of `CN`):

- `n.rim` ≡ `n.Rt.N_`
- `n.N_` ≡ `n.Nt.N_`  ·  `n.L_` ≡ `n.Lt.N_`  ·  `n.B_` ≡ `n.Bt.N_`  ·  `n.C_` ≡ `n.Ct.N_`  ·  `n.X_` ≡ `n.Xt.N_`
- `n.H` ≡ `n.Nt.H`

**Invariant:** the six forks `Nt, Lt, Bt, Ct, Xt, Rt` are the actual storage; the CN
attributes are aliases. A write to `n.rim += [L]` mutates `n.Rt.N_`; `n.Nt.N_ = …` and
`n.N_ = …` are the same store. Any analysis of `.rim` must also catch `.Rt.N_`, and vice-versa.

## 2. Aggregation: `c` is the weight, `dTT` is primary

`sum2F` and `add2F` are the only summation primitives:

- `F.c = Σ n.c` — **additive**.
- `F.dTT = Σ n.dTT·(n.c/F.c)` — **c-weighted mean**. Same form for `r, kern, span, yx`.
- `F.box = extend_box(…)` — **union** over members (typ==3), not a mean.
- `add2F` maintains the same means incrementally: `F.dTT = F.dTT·_w + n.dTT·w`, `_w,w = F.c/C, n.c/C`.

**Invariant:** `c` is additive and is the weight for every other aggregate; `dTT/r/kern/span/yx`
are c-weighted means; `box` is a union. Any change to what `c` means, or to how dTT
aggregates, must change `sum2F`, `add2F`, **and** `sum_vt` (which repeats the c-weighting)
together. `__bool__ = bool(c)`, so a zero-c object is falsy/empty and `if N:` tests `c`.

## 3. `m, d` are cached projections of `dTT`

`vt_(TT)` defines them: `m = (m_/(m_+|d_|)) @ wTT[0]`, symmetric for `d`. `Fvt_` writes them back.

**Invariant:** `m, d` are never independent state — always `vt_(dTT, wTT)`. Every mutation of
`dTT` must be followed by `Fvt_`/`vt_`, or `m/d` desync. `dTT` is `(2,9)`; the 9 params are fixed
(`wM, wD, wi, wG, wI, wa, wL, wS, wA`).

## 4. Link ↔ rim symmetry

`comp_N` builds `L = CL(N_=[_N,N])`, then for typ==3 nodes:
`_N.rim += [L]; N.rim += [L]; n.compared.add(_n)`.

**Invariant:** for any link `L` with `L.N_=[a,b]`: `L ∈ a.rim` **and** `L ∈ b.rim`, and
`a, b ∈ L.N_` (rim is symmetric, stored in `Rt.N_`). `_n ∈ n.compared` ⟺ they were linked.
Holds only for `typ==3`; lower-typ nodes (PPs) don't accumulate rim.

## 5. Centroid membership: `root_ / m_ / d_` stay index-aligned

`sum2F` with `m_,d_` (centroid formation) appends per member in one step:
`N.m_ += [m]; N.d_ += [d]; N.root_ += [F]`.

**Invariant:** for every node `N`, `root_[i], m_[i], d_[i]` describe membership in the *same*
centroid, fixed at append time. The double buffer `_root_/_m_/_d_` holds the prior iteration,
equally aligned (`k = n._root_.index(_C); n._m_[k], n._d_[k]`). Promotion is a **triple** move
(`n._root_=n.root_; n._m_=n.m_; n._d_=n.d_`); `cluster_C` resets all six together.

**Merge** (`comp_C_`): `n.root_[n.root_.index(C)] = _C` replaces in place to "keep m,d positions" —
merging two centroids must preserve each member's index alignment, never reorder `root_`
independently of `m_/d_`.

**(unsettled)** whether `root_` ever holds B-roots as well as centroids (`# Cs not Bs?`), and
whether a promoted centroid's own `m_/d_` collides with its per-member `m_/d_`
(`# may conflict with promoted C m_,d_?`). Treat both as fragile, not law.

## 6. Fork ownership

`comb_Ft`: `G = CN(Nt,Lt,Bt,Ct); Nt.root = Lt.root = Bt.root = Ct.root = G`.
`add_Nt`: `N.root = G` for members.

**Invariant:** a G's four forks point back to G via `.root`; G's member nodes carry `.root = G`.
`root` (singular) is the one connectivity parent; `root_` (plural) is the set of centroids —
distinct fields, distinct meaning, do not conflate.

## 7. `Copy_` is shallow at the leaves, deep in the forks

`Copy_`: `N_=copy(N.N_)` (shallow — new list, **shared elements**); forks `Nt..Rt` recursively
`Copy_`'d (fresh objects); `angl` deep-copied; `root_=list(N.root_)` (new list, shared centroids).

**Invariant:** a copied CN shares its member nodes/links with the original but owns independent
fork objects. Mutating a member *through* the copy is visible in the original; mutating a fork is
not. Any change that relies on copy-isolation must check which side of this line it sits on.

## 8. `typ` ↔ class

`sum2F` promotes via `cls_[typ]`, `cls_=[CF, CL, CC, CN]` (typ 0/1/2/3). rim updates gate on
`typ==3`; comp depth gates on `typ`. Class chain is `CF → CL → CC` and `CF → CL → CN`
(CN extends CL, **not** CC, though both carry `m_/d_`).

**Invariant:** `typ` selects the class on summation and gates behavior; a node's `typ` must match
its actual class and role (0 fork, 1 link, 2 centroid, 3 node).

## 9. Cross-module param-count boundary

meta_code's CN uses `dTT (2,9)`; frame_blobs' CN uses `derTT (2,8)`; comp_slice uses `verT (2,6)`.

**Invariant:** the comparand-vector width changes across the pipeline boundary. Anything crossing
frame_blobs / comp_slice → agg_recursion must map `(2,6)`/`(2,8)` → `(2,9)`; a change to the
9-param layout is a boundary change, not a local one.

boris-kz: Yes for the engine — plain Claude (the strongest reasoning tier), files in context, no fine-tuning, no critic, no agent, no execution, no synthetic input. You correctly stripped all of that.

But "check thoroughly" is carrying the part that doesn't work by being asked nicely. The invariants fix **judgment** — given a candidate site, does this change break the contract. They do nothing for **enumeration** — finding every site — which is exactly where a plain model satisfices and silently drops a few, worst on the shared base attrs (`dTT, N_, m, d, c, r`, on `CF`, accessed through generic vars everywhere) and the property aliases (`rim` ≡ `Rt.N_`, which it has to remember are one store). "Check thoroughly" is a *request* for completeness, not a *mechanism* for it. The only mechanism is the dumb-exhaustive site-finder — your impact-script. So: invariants for judgment, script for coverage; the three-part recipe alone leaves coverage best-effort.

Two smaller precisions. The invariants are the *plumbing* contract — aliasing, aggregation, membership alignment. Necessary for those couplings, not sufficient for a change to the algorithmic logic (gate values, eigenvector ordering, projection orders), which needs its own stated principle. And "check thoroughly" in practice is a closure-forcing prompt **plus a re-scan pass** plus the change stated as a precise contract — iterated, not one-shot.

Strip those and yes: strong plain Claude, the change stated as a contract, the invariants, and an exhaustive site list. That's the whole thing — and the only non-model piece in it is the site list.

Impact script:

```text
tools/impact.py CN.rim

step 0 — resolve the field (the original skipped this):
  parse target class body.
  if field is prop_F_(...)  → expand to backing store:
      rim → {rim, Rt.N_, Rt}                  # view, list store, fork object
      N_ → {N_, Nt.N_, Nt}   H → {H, Nt.H}   L_/B_/C_/X_ likewise
  else (stored base attr: dTT,m,d,c,r,root_…) → flag SHARED across CF/CL/CC/CN,
      site set over-includes, model disambiguates by object type

scope: every FunctionDef in agg_recursion.py + meta_code.py
  (reuse meta_code's ast parse, but NOT its iF_ whitelist —
   writers live in sum2F / Copy_ / add2F, outside it)

outputs:
  reads   : Attribute-load of .rim OR .Rt.N_
  writes  : two syntactically unrelated forms —
      list mutation    : AugAssign / .append on .rim | .Rt.N_     (rim += [L])
      fork replacement : Assign to .Rt                            (N.Rt = sum2F(N.rim))
                         ← contains no `rim` token; name-search is blind to it
  enclosing function per site
  lifecycle touched (resolved, not guessed):
      construct : CN.__init__ (Rt = CF(root=n)), CL(N_=[…]) links entering rims
      copy      : Copy_ Rt branch (fresh fork, shared link elements)
      sum       : sum2F(N.rim) → produces new Rt; add2F if Rt merged
  low-confidence flag: fork bound to a local before access
      (Rt = n.Rt; … Rt.N_ …) — qualifier gone, ownership unrecoverable

NOT an output: obligations. rim symmetry, root_/m_/d_ alignment
  live in the invariants doc, judged by the model on the site set above.

hand to model: site set + changed contract + relevant invariants
  → classify each site: no_change | needs_change | needs_decision
```

The two load-bearing fixes versus your original: a CN field name doesn't denote storage, so the script resolves `prop_F_` → fork before anything; and the write surface for any fork-backed field includes `.Fork = …` replacements, which are invisible to a search on the field name. Everything below step 0 is the dumb-exhaustive half — obligations stay with the invariants and the model.

kwcckw: I see, so "Check thoroughly" is not the correct prompt here.

Also i'm not sure the 9 invariants above are really useful, but we probably can try it first.

boris-kz: I thinks it's redundant too, but claude says it will not keep them in context on it's own. 
"Check thoroughly" is supposed to be replaced by impact script that I added above.
These invariants + impact script should be in project instructions, 
or project files, with each promp referencing them?

boris-kz: Split them by what consumes them, and watch one trap in each location.

**Method → project instructions.** The stable, always-on piece is the *procedure*, not the content: "trace from the supplied site list, check each site against the supplied invariants, classify `no_change | needs_change | needs_decision`, then re-scan for missed sites." That's small, it doesn't drift (it's procedure, not code-derived), and it's harmless on non-tracing prompts. Instructions are exactly for this. The invariants and the site list do *not* go here.

**impact.py → repo / project files, as a tool you run, not text the model reads.** The model consumes its *output* — the site set — not the script. So it lives in the repo like any code; its spec is for you, not the prompt.

**Invariants → stored as a project file, but injected by full paste, not retrieval.** Don't lean on project-file retrieval for them: project files come back *chunked* by semantic search — that's literally how I just read your code, in fragments — so a change to `root_` would surface invariant #5 and silently miss #1/#2 that it also violates. Partial retrieval reopens the exact coverage hole the invariants exist to close. Keep the canonical copy in the repo for versioning; paste the *whole* doc into each tracing prompt.

The discipline that outweighs location: the invariants are a snapshot of current plumbing (`prop_F_`, the forks, `sum2F`). When that changes, regenerate them. Buried in static instructions they go stale unnoticed, and checking against a stale contract is worse than none — false confidence. Anything code-derived wants project-file-plus-regenerate, never set-and-forget instructions.

So: instructions = method; repo = impact.py + the canonical invariants; each tracing prompt = the change + impact.py's output + the full invariants pasted.

kwcckw: > These invariants + impact script should be in project instructions,
or project files, with each promp referencing them?

Yes, i think that's the correct location rather than paste them each time. Could you share me the same impact script and updated 9 variants too? I'm thinking to try from my end as well.

boris-kz: I will commit them tonight.

this should be "trace from agg_recustion and meta_code, check each site against the invariants file, classify no_change | needs_change | needs_decision, then re-scan for missed sites, run impact.py as a tool."?

As a *plain-Claude* instruction, no — it reverses two things we just settled and adds one the model can't do:

- **"trace from agg_recursion and meta_code"** is the by-eye enumeration impact.py exists to replace. The model should trace from impact.py's *output*; the raw modules are what impact.py reads, not what the model walks.
- **"the project_invariants file"** reintroduces chunked retrieval. Last turn's conclusion: referencing it as a project file pulls fragments and drops the invariants that don't match the query — the exact hole the invariants close. It has to be pasted whole, hence "supplied," not "the file."
- **"run impact.py as a tool"** — from a project chat the model can't run it against your real repo. Project files are retrieval-only; they're not on any execution filesystem (we hit that earlier — bash couldn't reach `/mnt/project`, only search could). So nothing runs your script. And even if it could, it goes *first* (it produces the site list you trace from), not last.

But notice what your version actually is: it's exactly what **Claude Code** would do — run impact.py, read the whole invariants file off disk, trace, classify, re-scan. And the two operations you added — executing the tool and reading the file *whole* — are precisely what close the two coverage holes plain Claude leaves open: enumeration, and chunked retrieval of the invariants. Claude Code runs on your machine with the actual repo and files, so both work. Your rewrite is a well-formed *agent* instruction, and it's a real case for Claude Code: those holes don't close without tool-execution and filesystem reads.

So it's the fork again. Claude Code → your line works, with two fixes: run impact.py *first*, and "trace from impact.py's output." Plain Claude → keep "supplied site list / supplied invariants," and you do the running and pasting by hand. Your rewrite is you picking the agent path.

boris-kz: three fixes:

1. **Trace from the site list, check against the invariants** — two roles, don't merge them. impact.py's output is what you enumerate over; the invariants are what you test each site against. "Trace from the invariants" collapses both, and the invariants don't give you the sites.

2. **Attach impact.py's *output*, not the script.** The model consumes the site list, can't run the script, and reading the code itself is the weak enumeration the script exists to replace. And attach by *uploading to the prompt*, not as project knowledge: a per-message upload lands whole in context — that's how your earlier excerpts came in — while project files come back chunked, which is how I read your code, and would silently drop the invariants that don't match the query. Same for the invariants doc: upload or paste it whole, never rely on project retrieval for it.

3. **State the change** — it's missing. The model can't classify `no_change`/`needs_change` without the new contract.

So:

```
attach to the prompt (whole, as uploads): invariants doc + impact.py output for the field

prompt: "Change: <what's changing>. Walk every site in the supplied list,
         check each against the supplied invariants, classify
         no_change | needs_change | needs_decision, then re-scan
         agg_recursion and meta_code for any site the list missed."
```

The re-scan is a backstop — impact.py's output stays the primary enumeration; the re-scan only catches what its heuristics flagged low-confidence or dropped.

PR #183, opened 2026-06-22
6/22/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -284,19 +284,24 @@ def get_fc(n):
...
+    if isinstance(n,tuple): return sum(get_gi(c) for c in n[1])
+    if isinstance(n,ast.Call) and n.function.id == 'gv_': return 1
+    return 0
+
 def split_oF_():  # divisive clustering
     out = []
     for oF in oF_:
         if (len(oF.body)-1) * wL > ave:  # * split w,c
-            grp_=[]; _n = oF.body[0]; grp = [_n]
+            grp_=[]; _n = oF.body[0]; grp = [_n]; gi_ = []; gi = get_gi(_n)
```
kwcckw, line 296: So the split and merged oF should have the updated `oF.gi` to init call.gv_, which we need to init them here.

frame_2D_alg/agg_recursion.py
```diff
@@ -872,11 +873,11 @@ def ffeedback(frame, aTT,oTT, aL,oL):  # recompute filters from regime drift; fo
     _ac,_ar = (aL.c,aL.r) if aL else (0,0); _oc,_or = (oL.c,oL.r) if oL else (0,0)
     # H init @ 1st term:
     if aL := pack_seg(frame,'aH',wBac, cBac, aTT):  # L: new level
-        dTT = aL.dTT-aTT; aTT=aL.dTT; dc= aL.c-_ac; dr= aL.r-_ar
+        dTT = aL.dTT-aTT; aTT=aL.H[-1].dTT; dc= aL.c-_ac; dr= aL.r-_ar  # should be aL.H[-1] for the last terminated level?
```
boris-kz, line 876: aL is the last terminated level

frame_2D_alg/agg_recursion.py
```diff
@@ -537,7 +537,8 @@ def sum2G(ft_, fTT, root=None, init=1):  # core clustering function
     N_ = G.N_; N=N_[0]; G.sub = N.sub+1 if G.L_ else N.sub; r=G.r
     if G.Lt:  # sub+
         Lt = G.Lt; L_,lm,ld,lr = Lt.N_,Lt.m,Lt.d,Lt.r; L=len(L_); Av = ave+avd
-        if gv_(Vn := (lm+ld)*(wcN*L) - Av* (lr+1+ccN*L), 0):  # default cluster_N
+        # we need at least 2 Ls for mdecay below when using np.diff, so eval with L-1?
```
boris-kz, line 540: I guess so

frame_2D_alg/agg_recursion.py
kwcckw: Thanks, i misunderstood earlier.

frame_2D_alg/meta_code.py
```diff
@@ -284,19 +284,24 @@ def get_fc(n):
     if isinstance(n, tuple): return costs.get(n[0],0) + sum(get_fc(c) for c in n[1])
     return costs.get(type(n),0)
 
+def get_gi(n):
```
boris-kz, line 287: return sum(get_gi(c) for c in n[1]) if isinstance(n,tuple) else isinstance(n,ast.Call) and n.function.id == 'gv_'

kwcckw: Actually the `i` in `gv_` should be dynamic?

Because after the merging or split of oF, the `i` index will no longer be correct, and they need to be reassigned again.

boris-kz: we just need to set new i first? where do you do that?

kwcckw: Yes, and that `i` should be set in oF and retrieved from the function during runtime. But the current problem is the mapping since after merge or split, I don't see how we can map the `gv_` in certain section to the correct `i` index yet.

boris-kz: the same way gv_ was initialized? this is after merge/split anyway?

kwcckw: Yes, this is after the merge/split, which may not be relevant now.

But the real problem is the value of `i` because right now it's static in the code, where we need to set it manually. After the split or merge, this `i` should be dynamic, and I don't see a method to map gv_'s i to oF.gv_ yet.

boris-kz: we need to mark these gates in the way AST tracing would recognize it?

frame_2D_alg/agg_recursion.py
```diff
@@ -320,10 +320,10 @@ def trans_cluster(G):
                             span = np.sqrt(len(N_))  # approx span
```
boris-kz, line 320: not directly related, I was thinking about the latest trends in LLMs, a Chat summary in the end:
https://chatgpt.com/share/6a39da2f-1000-83ea-ba09-35658280c8c5

frame_2D_alg/meta_code.py
kwcckw: > we need to mark these gates in the way AST tracing would recognize it?

Yes, and we need another part so that it can be mapped from the prior static ast tracing to the `i` value in runtime.

boris-kz: it doesn't need to be at runtime, this oF updating ffeedback is much more coarse

frame_2D_alg/agg_recursion.py
boris-kz: final chat:

The precise version is: **frontier quality will migrate from model scale to coordination quality.** The best system will not be the largest general model. It will be the system that best decomposes tasks, selects specialists, allocates compute, verifies outputs, manages disagreement, and synthesizes the result.

That means the frontier becomes **organizational**, not merely parametric.

Sakana is already explicitly framing this direction: Fugu is presented as “a multi-agent system, delivered as one model,” where a language-model controller coordinates a pool of powerful models and learns agent assembly, model selection, delegation, verification, and synthesis rather than relying on one monolithic solver. ([[Sakana AI](https://sakana.ai/fugu/?utm_source=chatgpt.com)][1]) Its beta post is even closer to your claim: Sakana says the most capable systems will not be monolithic models scaled in isolation, but collections of specialized agents working together. ([[Sakana AI](https://sakana.ai/fugu-beta/?utm_source=chatgpt.com)][2])

The quality argument is stronger than the cost argument because a monolith has a structural limitation: it must compress all competences into one shared parameter field and one active context. A specialist ecosystem does not. It can keep capabilities separated, updated, benchmarked, swapped, adversarially checked, and composed at inference time. That is a strictly richer design space.

A frontier general model has to be good at everything internally. A specialist ecosystem only has to know how to call the best available component for each subproblem and preserve enough state across the handoffs. Once the handoff problem is solved, the ecosystem has more degrees of freedom.

The key phrase is **once the handoff problem is solved**. That is the only serious caveat. Current multi-agent systems still lose information through bad decomposition, redundant work, shallow summaries, and coordination overhead. A 2026 Stanford paper argues that when reasoning-token budgets are equalized, single-agent systems can match or outperform multi-agent systems on multi-hop reasoning, largely because decomposition creates information bottlenecks. ([[arXiv](https://arxiv.org/abs/2604.02460?utm_source=chatgpt.com)][3]) But that does not refute your thesis. It identifies the bottleneck the new frontier has to solve.

The solution is not “better chat between agents.” It is better shared artifacts: tests, traces, schemas, formal constraints, proof states, execution logs, citations, typed memory, AST diffs, simulators, calibrated confidence, and adversarial verifier outputs. Once specialists coordinate through these artifacts, the monolith’s advantage of holding everything in one latent context becomes less important.

Anthropic’s research-agent work points in that direction. Their multi-agent research system reportedly outperformed a single-agent Claude Opus 4 setup by 90.2% on an internal research eval, especially for breadth-first tasks where subagents can explore independent directions in parallel. ([[Anthropic](https://www.anthropic.com/engineering/multi-agent-research-system?utm_source=chatgpt.com)][4]) That is not yet proof of universal superiority, but it shows the quality mechanism: parallel search, separate context windows, specialized subgoals, and synthesis.

So I would revise the model hierarchy this way:

The old frontier was: bigger general model.

The transitional frontier is: general model plus tools and some routing.

The next frontier is: learned router plus specialist ecosystem.

The later frontier is: self-improving specialist economy, where models are trained for coalition value rather than standalone chat performance.

That last part matters. In the mature version, specialists are not merely “small models that save money.” They are the atoms of intelligence. Some will be cheap and narrow. Some will be large and narrow. Some will be verifiers. Some will be searchers. Some will be simulators. Some will be synthesis engines. Some will exist mainly to generate training data for other models. Their value is not “can this model answer the whole user request?” but “does this model improve the system’s final answer when inserted into the coalition?”

This is why benchmark panels become central. RouterBench frames the problem as evaluating routing among multiple LLMs because no single model is optimal across all applications under cost-performance tradeoffs, and it provides hundreds of thousands of model-output records for studying routing strategies. ([[arXiv](https://arxiv.org/html/2403.12031v1?utm_source=chatgpt.com)][5]) But the deeper implication is quality-market formation: benchmarks identify niches, routers allocate demand to niche winners, and model builders optimize for those niches.

Your “will not stop halfway” point is important. The halfway architecture says: use small specialists for easy tasks, then escalate to frontier generalists for hard ones. That probably is not the endpoint. The endpoint is: hard tasks are routed into deeper specialist societies. Escalation means more coordination, more verification, more search, more tools, more adversarial review, and more compute allocation—not necessarily one bigger general model.

The strongest objection is that some tasks are tightly integrated and resist decomposition. But even there, the mature ecosystem can respond by deploying an integrated specialist as one component. That component may be large, but it is no longer a frontier generalist in the old sense. It is just one executable inside the system.

So yes: I think your thesis is directionally right. The terminal frontier generalist is likely a historical phase. The durable frontier is the **generalist-router plus specialist ecosystem**.

The cleanest formulation:

**Monolithic general models lose because they internalize what a routed ecosystem can externalize, specialize, verify, update, and compose.**

That is not merely cheaper. It should ultimately be better.

[1]: https://sakana.ai/fugu/?utm_source=chatgpt.com "Sakana Fugu — Multi-Agent System as a Model"
[2]: https://sakana.ai/fugu-beta/?utm_source=chatgpt.com "Sakana Fugu: A Multi-Agent Orchestration System as a ..."
[3]: https://arxiv.org/abs/2604.02460?utm_source=chatgpt.com "Single-Agent LLMs Outperform Multi-Agent Systems on Multi-Hop Reasoning Under Equal Thinking Token Budgets"
[4]: https://www.anthropic.com/engineering/multi-agent-research-system?utm_source=chatgpt.com "How we built our multi-agent research system"
[5]: https://arxiv.org/html/2403.12031v1?utm_source=chatgpt.com "RouterBench: A Benchmark for Multi-LLM Routing System"

kwcckw: Thanks, this is very informative. For me, right now the token usage is still a problem since it's get cap very fast. I think it should be helpful to have much more smaller refined specialist model so that to let the right model do the right job with reduced token usage. 

Also this router within the models is a new thing to me since i think multi-agents or models is only become a trend in the recent years.

boris-kz: Right. We have MoEs, but they are just differentially trained copies of the same model, with router trained at the same time. This scheme disaggregates everything. Although not nearly as much as CogAlg :)

boris-kz: https://chatgpt.com/s/t_6a39e484ce8881918611dd76fd4fc883

kwcckw: > So yes, this is fundamentally a clustering problem. But the clustering criterion is not “similar topics.” It is optimal modular decomposition of intelligence under quality, cost, latency, verification, and context-transfer constraints.

So this would back to the clustering problem again.

> This scheme disaggregates everything. Although not nearly as much as CogAlg :)

Yes, but for CogAlg, it doesn't need router and ultimately it should be applicable to all kinds of data?

boris-kz: The router here is the top generalist, which is a global cluster.

The difference in cogalg is that it's strictly bottom up, seeded by connectivity clustering. all these specialist models are still trained top-down

kwcckw: Without training top-down, we actually doesn't have the "correct label", and there's chance the feedback is actually "correcting" in the wrong direction although the whole system is stable?

boris-kz: Read the readme, it was never about fitting. Cogalg is supposed to maximize projected match to new input

PR #184, opened 2026-06-23
6/23/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -177,8 +177,9 @@ def __bool__(f): return bool(f.call_)
 def Fvt_(N_,TT,c,r):
     oF = CoF.get(); oF.N_+=N_; oF.c+=c; oF.vt_ += [[TT,c,r]]  # data per call
 
-def gv_(v,i):
+def gv_(v):
     oF = CoF.get()
+    i = oF_[oF.nF].gv_map[inspect.currentframe().f_back.f_lineno]
```
kwcckw, line 182: I removed the oF.gi here and replace it with oF.gv_map, so that we can skip the manual `i` value from the code and it will be determined by iterating through the oF ast body. The mapping is done by the line number now, so we shouldn't use 2 `gv_` s in a single line.

frame_2D_alg/agg_recursion.py
```diff
@@ -51,7 +51,7 @@
 wM,wD,wi, wG,wI,wa, wL,wS,wA = wT
 cFrm,cAgg,cTrc, cN_,cC_,cN,cF, cE,ccN,ccC,ccP,csG, cBac,cPrj, cVct = (      # function complexity
 wFrm,wAgg,wTrc, wN_,wC_,wN,wF, wE,wcN,wcC,wcP,wsG, wBac,wPrj, wVct ) = [F.fc for F in oF_]  # ave gain/call, init = cost
-ttFrm,tAgg,ttTrc, ttN_,ttC_,ttN,ttF, ttE,ttcN,ttcC,ttcP,ttsG, ttBac,ttPrj, ttVct = [F.dTT for F in oF_]
+ttFrm,tAgg,ttTrc, ttN_,ttC_,ttN,ttF, ttE,ttcN,ttcC,ttcP,ttsG, ttBac,ttPrj, ttVct = [F.wTT for F in oF_]  # this should be oF.wTT?
```
boris-kz, line 54: right

frame_2D_alg/agg_recursion.py
```diff
@@ -694,7 +696,7 @@ def PP2N(PP):
                             N.Bt.N_ = PPd_; [setattr(B,'root',N.Bt) for B in PPd_]
                     L = len(PPm_)-1
                     if gv_(sum(vt_(sum_vt(N_,wTT=ttVct)[0])) * (wVct*L) - (ave+avd) * (3+cVct*L),2):
-                        G_,TT,C,R = trace_edge([F2N(n) for n in N_], G_,TT,c,3,tile)  # flatten B_-mediated Gs
+                        G_,TT,c,R = trace_edge([F2N(n) for n in N_], G_,TT,c,3,tile); C += c  # flatten B_-mediated Gs  (sum to the main C instead of replace it?)
```
boris-kz, line 699: thanks

frame_2D_alg/meta_code.py
boris-kz: ok right, I forgot the gate is a function now

frame_2D_alg/meta_code.py
```diff
@@ -278,22 +278,28 @@ def comp_prim(_n,n):
...
+        gv_map[n.lineno] = gi
+        return gi + 1
+    return gi
 
 def split_oF_():  # divisive clustering
     out = []
     for oF in oF_:
         if (len(oF.body)-1) * wL > ave:  # * split w,c
-            grp_=[]; _n = oF.body[0]; grp = [_n]; gi_ = []; gi = get_gi(_n)
+            grp_=[]; _n = oF.body[0]; grp = [_n] ; gv_map = {}; gi=get_gvmap(gv_map, _n); gv_map_ = []
```
boris-kz, line 294: gv_map key is lineno and value is a gate?

kwcckw: Yes, key is lineno and the value is actually the gate index, gi, which is the index of oF.gv_.

boris-kz: so it should not be in split_oF, we just need to run F_body_ after all splits and merges?
with some refactoring:

`get_gmap` is intended to rebuild the `gv_` gate-index map for a static code fragment after `split_oF_` splits a `CoF.body` into smaller contiguous groups.

Conceptually:

```python
def get_gmap(gmap, n, gi=0):
    if isinstance(n, tuple):
        for c in n[1]:
            gi = get_gmap(gmap, c, gi)
        return gi
    if isinstance(n, ast.Call) and n.function.id == 'gv_':
        gmap[n.lineno] = gi
        return gi + 1
    return gi
```

It recursively walks one body primitive `n`. If `n` is an AST fork tuple, it descends into its children. If it finds a `gv_` call, it records `{source_lineno: local_gate_index}` and increments `gi`. That map is needed because current `gv_` no longer uses the explicit gate index argument; it looks up the caller line number in `oF_[oF.nF].gv_map`, then stores negative gate value in `oF.gv_[i]`. `CoF.traced` allocates `oF.gv_` from `len(oF_[i].gv_map)`, and the final function value becomes `vt_(oF.dTT)[0] + sum(oF.gv_)`. So the final consumer is the runtime function-value estimate, not `split_oF_` itself. 

The intended implication chain is:

`F_body_()` builds `oF.body` and the original `func.gv_map` from AST line numbers; `trace_func()` wraps real functions; each runtime `gv_` call maps source line → gate slot; `CoF.traced` sums those gate penalties into `oF.w`; later `ffeedback()` may call `split_oF_()` and `cluster_oF_()` to modify code-level `CoF` structure, so `get_gmap` is supposed to preserve gate indexing for the newly split sub-`CoF`s. `ffeedback` is the path that triggers this code-reform stage.    

Main problems in the current implementation:

1. `n.function.id` is wrong. Python AST call nodes use `n.func`, not `n.function`. So as soon as `get_gmap` reaches an `ast.Call` leaf, it raises `AttributeError`. The guard should be roughly:

```python
isinstance(n, ast.Call) and isinstance(n.func, ast.Name) and n.func.id == 'gv_'
```

2. Even with that typo fixed, it will miss most `gv_` calls. `F_body_.build` converts non-leaf AST nodes into `(type(node), sub_)`. For a normal `gv_(expr, i)` call, the `ast.Call` wrapper has children, so it becomes `(ast.Call, sub_)`; the original `ast.Call` object, its `lineno`, and its `func.id` are lost. Then `get_gmap` only sees the tuple and recurses into children, where it can no longer identify that the parent call was `gv_`.  

3. `split_oF_` stores the result as `sub.gmap`, but runtime uses `gv_map`. So even if `get_gmap` returned the right map, it is assigned to the wrong attribute and is not consumed by `gv_` / `CoF.traced`. 

4. `sum2O` calls `get_gvmap`, which is undefined. That looks like a naming drift from `get_gmap`; this will crash in the code-clustering path after `cluster_oF_()` calls `sum2O()`. 

5. There is a larger structural mismatch: `CoF.traced` uses `iF_[func.__name__]` as a stable index into `oF_`. But `split_oF_()` replaces `oF_[:]` with a different list of sub-`CoF`s. After that, the original function-name index no longer reliably points to the original function `CoF`. So a fixed `get_gmap` is still insufficient unless the original `oF_` indexing is preserved or the runtime tracer is updated to reference a separate stable table.  

There is also an adjacent mismatch: current `agg_recursion.py` callsites still pass explicit indices, for example `gv_(..., 0)`, `gv_(..., 1)`, `gv_(..., 2)`, while current `meta_code.gv_` is defined as `gv_(v)` and derives the index from line number. That will fail before `get_gmap` matters unless `gv_` is changed to accept the explicit index again or the callsites are updated.  

Minimal direction: don’t patch `get_gmap` alone. First decide whether gates are indexed explicitly or by source line. If line-based, preserve the original AST node in `F_body_.build` tuples, assign `sub.gv_map = gmap`, rename `get_gvmap` to `get_gmap`, and keep the original function-indexed `oF_` table separate from any split/clustered code-level `CoF_` list.

boris-kz: this gv_map is a pile of useless mess, we just need to get right order in oF.gv_?

boris-kz: in build:

            if node.func.id == 'gv_': func.gv_ +=[0]
            if (i := iF_.get(node.func.id)) is not None: return oF_[i], 3

...
That works as the per-function gate count, but watch what `func` is at that point and what it implies downstream.

In `build`, `func` is the *static* `oF_[i]` (the parsed CoF), not a runtime frame. So `func.gv_ += [0]` accumulates the gate list on the static template, once at parse. That's the right place to count. But then the contract splits cleanly:

- **Static `oF_[i].gv_`** = the parse-time template, length = gate count. Init `func.gv_ = []` at 207 (replacing `gv_map = {}`).
- **Runtime `oF.gv_`** = a *fresh* per-call accumulator. Line 155 must clone, not alias: `oF.gv_ = [0] * len(oF_[i].gv_`)` (or `copy`). If you let the runtime CoF share the template list, every call mutates the parsed template and they cross-contaminate.

So `+= [0]` on the template is fine for sizing, but the value `0` is decorative — only `len()` is used. `func.n_gv += 1` carries identical information. The list form is justified only if a slot ever holds a non-zero *template default* (none does today).

Second: this still leaves the index assignment unsolved. You're counting sites, but `gv_(eval, i)` needs each site's `i`. Parse-order counting gives you the *order* — site k is the k-th `gv_` `build` visits. If you want indices auto-assigned by that order rather than hand-written literals:

```python
            if node.func.id == 'gv_': func.gv_ += [0]   # site index = len before append
```

then the literal `i` at the call site must match parse-visit order — which `build` traverses via `ast.iter_child_nodes`, *not* source-line order. Branch/expr nesting means visit order ≠ lineno order. So a hand-written `gv_(expr, 3)` can desync from build's count. Either:

- keep explicit literals and have `build` only *validate* count (`assert max matches`), or
- drop literals entirely and have the runtime hook recover order — but that reintroduces per-call site identity, i.e. the map you just deleted.

The stable resolution: literals are the source of truth, `build` just sizes. Then `+= [0]` is correct but `len(oF_[i].gv_)` must be `max(literal)+1`, not the visit count — those differ if any index is skipped. Use `func.gv_` length only if you guarantee dense, parse-order literals; otherwise size from the literals themselves.

kwcckw: > this gv_map is a pile of useless mess, we don't need even need gv_, just oF.gv += all gv_()s?

Yes, actually if we don't need to identify each individual gv, oF.gv += all gate's v is already sufficient, then we can skip all of this mess.

> The stable resolution: literals are the source of truth, build just sizes. Then += [0] is correct but len(oF_[i].gv_) must be max(literal)+1, not the visit count — those differ if any index is skipped. Use func.gv_ length only if you guarantee dense, parse-order literals; otherwise size from the literals themselves.

As above, this append method is only working for sizing, but in term of retrieval, we need the mapping, else we won't know which index is map to which line of gv_.

boris-kz: >but in term of retrieval, we need the mapping, else we won't know which index is map to which line of gv_.

we can get mapping gates the same as in build:

    def build(func, node):  # AST → CoF | (type,sub_) | ast_leaf | None
            if node.func.id=='gv_': func.gv_ +=[0]

only here we need to get gv_ index

boris-kz: You've hit the exact wall from three turns ago, restated. The reason build *can't* hand back the index is structural, not a missing line:

`gv_` is not in `iF_`, so it has no early return. It falls through to `(type(node), sub_)` and lands in the body as `(ast.Call, sub_)` — `func.id` gone, position-in-`gv_` gone. So when you later walk `oF.body` to map a `gv_` slot back to its node, **every** non-callee call is `(ast.Call, sub_)`; you can't tell `gv_` from `len`/`copy`, and you have no counter telling you *which* gv_ this is. The body simply doesn't carry the identity.

So "same as build but get the index" fails because build's own gate detection only works on **raw AST** (`node.func.id`), and the body is post-AST. There are exactly three places the gv_↔slot correspondence can live:

**(a) Stamp the index into the body node at build time.** This is the only way to make `oF.body` self-describing. Make gv_ return early like callees, carrying its slot:

```python
            if node.func.id == 'gv_':
                gv_i = len(func.gv_); func.gv_ += [0]
                return ('gv_', gv_i), 0   # body node IS the index
            if (i := iF_.get(node.func.id)) is not None: return oF_[i], 3
```

Now walking the body, `('gv_', k)` tells you slot `k` maps here, no ambiguity, no second pass. Cost: gv_'s arg subtree is dropped from the body (you stop recursing). If you need the gated expression represented for cost/comp, keep it: `return (ast.Call, [build(func, a) for a in node.args]) + carry the index` — but a tuple can't cleanly hold both, so use a 3-element marker `('gv_', gv_i, sub_)` and teach the tuple-consumers to skip element 1.

**(b) Parallel list, index = position.** `func.gv_node_ += [node]` in build, aligned with `func.gv_` by append order. Body stays unchanged; the map is the sidecar list. Slot `k` ↔ `gv_node_[k]`. Simpler, but the body itself remains non-self-describing — fine if only the sidecar ever needs the mapping.

**(c) Don't map back at all.** The honest question first: *who consumes the back-map?* As of the current code, the only reader of `oF.gv_` is `sum(oF.gv_)` (164) — a reduction that needs no per-slot identity. If nothing walks the body asking "which node is slot k," (a) and (b) are both unnecessary. You'd only build the back-map for a consumer that doesn't exist yet.

So before editing build again: name the consumer. If it's just the `sum`, you need neither index nor back-map — a scalar (two turns ago) suffices. If a future pass attributes gate penalties to specific body sites, use (a) — it's the only option that survives the AST→body erasure without a fragile order-alignment assumption. Restating the `+= [0]` line a fourth time won't produce the index because the index has to be *returned into the body*, and the current fall-through throws it away.

kwcckw: But to map them, we need the unique identifier, which we need the line number lineno?

    func.gv_map[node.lineno] = gi
    
> only here we need to get gv_ index

We need the `gv_` index when we call `gv_` in runtime, and that needs to be dynamic after the split or merge.

> (a) Stamp the index into the body node at build time. This is the only way to make oF.body self-describing. Make gv_ return early like callees, carrying its slot:
> 
> python if isinstance(node, ast.Call) and isinstance(node.func, ast.Name): if node.func.id == 'gv_': gv_i = len(func.gv_); func.gv_ += [0] return ('gv_', gv_i), 0 # body node IS the index if (i := iF_.get(node.func.id)) is not None: return oF_[i], 3

Not sure if this is the correct method, how do we map the gv_ during runtime if we use this method?

boris-kz: we only need this mapping very rarely, in batches.
so I can get the index the same way oF.gv_ was formed, but gi += 1 instead of oF.gv_+=[0]?

...
Right instinct for something batch and rare — recompute over store. But "the same way `oF.gv_` was formed" hides two quirks that, if not replicated, desync the recovered index from the slot sizing/runtime actually used. There is exactly one formation order; recovery must re-derive it from the same source, not a convenient substitute.

**Quirk 1 — raw AST, not the body.** `oF.gv_` was formed in `build` walking `nF_[oF.nF]` (the FunctionDef), where `node.func.id=='gv_'` still resolves. Walking `oF.body` can't work — gv_ sites are `(ast.Call, sub_)` with `func.id` erased, indistinguishable from `len`/`copy`. So the batch pass re-walks the retained FunctionDef, not the body.

**Quirk 2 — stop at callees, and match traversal shape.** Formation order is `build`'s pre-order DFS via `ast.iter_child_nodes`, and `build` returns at a callee Call (`return oF_[i],3`) *before* descending its args. So a `gv_` sitting inside a callee's argument — `callee(gv_(x))` — is never counted at formation. A naive `ast.walk` (BFS) or an unconditional recursion would count it, and every slot after it shifts by one. The recovery walk must replicate both: pre-order, `iter_child_nodes`, and prune at callee calls.

```python
def gv_sites(fd):  # (gv_ node, slot) in formation order
    out, i = [], 0
    def rec(node):
        nonlocal i
            if node.func.id == 'gv_': out.append((node, i)); i += 1
            if node.func.id in iF_: return  # build substitutes callee, no descent
        for c in ast.iter_child_nodes(node): rec(c)
    rec(fd); return out
```

Now `for node, i in gv_sites(nF_[oF.nF]): ... oF.gv_[i] ...` is your batch eval, indices guaranteed identical to the ones `func.gv_ += [0]` assigned, because it's the same predicate, same order, same source.

Two flags:

**`oF.gi += 1` as persistent state is a trap.** Make it a local counter reset per walk (as above). A counter living on `oF` goes stale across batches or under re-entrancy and silently drifts the indices — exactly the failure this whole thread has been circling.

**This recovers the back-map only; it does not give runtime its write slot.** Branch/loop nesting means execution order ≠ static order, so runtime `gv_` still can't recover `i` by counting at call time — it needs the per-site identity (explicit literal `i`, or the lineno map you dropped). The recompute-on-demand trick is sound *because* the batch pass runs over static AST where order is deterministic; the runtime hook isn't, and remains the separate unsolved contract. Same single index, three users (sizing, runtime write, batch recovery) — all three must derive from this one order or they disagree.

kwcckw: > we only need this mapping very rarely, in batches. so I can get the index the same way oF.gv_ was formed, but gi += 1 instead of oF.gv_+=[0]?

Yes, but with this index, we can't really map it during runtime since it can't identify which `gv_` call is referring to which index, so i'm not sure why we need it here? 

And if we need oF.gv only, we not really need this index.

boris-kz: right, no index, just g_:

def gv_sites(fdef):  # g_[i] <-> oF.gv_[i], same order as build

    g_ = []
    def rec(node):
            if node.func.id == 'gv_': g_.append(node)
            if node.func.id in iF_: return   # build substitutes callee, no descent
        for child in ast.iter_child_nodes(node): rec(child)
    rec(fdef); return g_

kwcckw: So this `g_` serves as the reference to the gate's nodes only right?

boris-kz: right, it maps to gv_

kwcckw: > right, it maps to gv_

Could you clarify if we still need to use gv_ instead of gv?

During runtime, each `gv_` function call will be matched to the g_'s node.lineno to get the `oF.gv_` index?

boris-kz: I just pushed. 
each gv_ value will be matched to g_ gate, but not at runtime, it's to evaluate the gates for potential removal. So we need to add this eval vs the cost of the gate in ffeedback. If negative, the gate is removed or simplified, the gated function may be called directly

kwcckw: But this still requires us to fill each element in gv_, which needs the tracing of gv_ in runtime?

boris-kz: each gate in the body is assigned an index:

if gv_(sum(vt_(TT,ttE))*(wE*L) - (ave+avd)*(r+cE*L), 0):  # if +ve, store neg gate values

we need to use that index (0) to sum neg vals in gv_:

def gv_(v, i ):

    oF = CoF.get()
    if v > 0: return v
    else: oF.gv_[i] -= v  # then oF.w = vt_(oF.dTT)[0] + sum(oF.gv_)

boris-kz: actually, we can use that index for final eval too, so  gv_sites(fdef) is not needed?

Yes — the explicit literal collapses all three users to one source, and `gv_sites` was only ever needserved by the *implicit* order. Trace it:

The literal `i` at the call site is now the single source of truth. Runtime write (`oF.gv_[i] -= v`) keys on it. Final eval keys on it. Sizing keys on `max(i)+1`. None of these needs to *recover* the index from a walk, because the index is written into the call. `gv_sites` existed solely to reconstruct the implicit append-order mapping — that mapping is gone, so the function is gone. Delete it.

What this kills along with it — the two desync hazards this thread kept hitting both vanish, because both were artifacts of order-dependence:

- The callee-prune requirement (`if ... in iF_: return`) is irrelevant now. A `gv_` buried in a callee arg gets whatever literal you wrote; position in traversal no longer assigns anything.
- The build/recovery traversal-shape match is irrelevant. Nothing re-walks to count.

But the literal moves the burden, doesn't erase it. Three things now ride on *you* writing indices correctly, none enforced:

**Sizing.** `build` still has to size `oF.gv_`. With literals it's no longer "count gv_ sites" — it's `max(i over sites) + 1`, and it must read the literal arg:

```python
        if node.func.id == 'gv_':
            i = node.args[1].value          # the literal index arg
            func.gv_n = max(func.gv_n, i+1) # init func.gv_n = 0 at 207
```

then `oF.gv_ = [0]*oF_[i].gv_n` at 155. Counting (`+= [0]`) gives the wrong length the instant indices are non-dense (you skip a number) or repeated.

**Density.** `sum(oF.gv_)` (164) sums all slots. A skipped index = a permanent `0` slot — harmless to the sum, so gaps don't corrupt eval, only waste a slot. Not worth policing.

**Collision is the real exposure.** Two sites sharing index `i` both write `oF.gv_[i]` — their penalties merge, and final eval can't separate them. Nothing catches this; it's a silent semantic merge, not a crash. If you ever want per-gate attribution at eval, uniqueness is the invariant you're now hand-maintaining. If eval only ever consumes `sum(oF.gv_)`, collisions are as harmless as gaps and you needn't care.

So: drop `gv_sites`, keep `gv_(v, i)` as written, change sizing from count to `max+1`. The only thing you've traded is automatic index assignment for manual literals — fine given the gates are few and hand-written, as long as the eval consumer stays a reduction. The moment eval addresses a *specific* gate, manual uniqueness becomes load-bearing and unchecked.

kwcckw: If we preserve the `i` input argument during the `gv_` function call, that site method is redundant, but that will not work after we merge or split oF, which is why we need the gv_sites similar method in the first place?

boris-kz:
Yes, so gsites can assign the Is in the body at initialization too, instead
of manually.
That assigned i will still be used at runtime

PR #185, opened 2026-06-24
6/24/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -178,18 +180,13 @@ def Fvt_(N_,TT,c,r):
 
 def gv_(v):
     oF = CoF.get()
-    i = oF_[oF.nF].gv_map[inspect.currentframe().f_back.f_lineno]
```
kwcckw, line 181: > actually, we can use that index for final eval too, so gv_sites(fdef) is not needed?

> Sizing. build still has to size oF.gv_. With literals it's no longer "count gv_ sites" — it's max(i over sites) + 1, and it must read the literal arg:
>
>        if node.func.id == 'gv_':
>            i = node.args[1].value          # the literal index arg
>            func.gv_n = max(func.gv_n, i+1) # init func.gv_n = 0 at 207
>
> then oF.gv_ = [0]*oF_[i].gv_n at 155. Counting (+= [0]) gives the wrong length the instant indices are non-dense (you skip a number) or repeated.

I don't think we need gv_sites, and the suggestion above is mainly using the static `i` from the `gv_` function, which is something we need to remove later?

If we pack node in oF.g_, we actually can use that to map their lineno (between g_'s ast node and gv_'s ast node), then we can totally skip the static `i` within the code.

frame_2D_alg/meta_code.py
```diff
@@ -178,18 +180,13 @@ def Fvt_(N_,TT,c,r):
...
+    n = executing.Source.executing(inspect.currentframe().f_back).node  # get current gv_ ast node
+    toF = oF_[oF.nF]
+    for i, g in enumerate(toF.g_): 
+        if g.lineno == n.lineno: break
     if v > 0: return v
-    else: oF.gv_[i] -= v  # then oF.w = vt_(oF.dTT)[0] + sum(oF.gv_)
+    else:     oF.gv_[i] -=v; toF.gv_[i] -= v  # then oF.w = vt_(oF.dTT)[0] + sum(oF.gv_)  (should be accumulating to the typ oF in oF_?)
```
kwcckw, line 188: For the gate values, i think other than callee's gv_, we need to update it into the main type oF.gv_?

frame_2D_alg/meta_code.py
```diff
@@ -297,6 +294,7 @@ def split_oF_():  # divisive clustering
                 out += [sub]
         else: out += [oF]
     oF_[:] = out
+    # after split here, nF and iF will be no longer relevant
```
kwcckw, line 297: After the split or merge, the `nF_` and `iF_` is no longer matching with the new oF_. In this case, we no longer able to use F_body_ the same way to those new oFs. Also `F_body_` uses function name to retrieve the nodes:

    for func,name in zip(oF_,nF_):
        func.caller_ = set(); func.g_ = []  # only need g_ in type oF
        for node in ast.iter_child_nodes(name):  # skip top function definition
            rett = build(func, node)
            if rett:
                t, fc = rett; func.body += [t]; func.fc += fc

Which is no longer applicable since we don't have the matching function name with the static code after the merge or split?

boris-kz:
Reassign all that in ffeedback

frame_2D_alg/meta_code.py
```diff
@@ -178,18 +180,13 @@ def Fvt_(N_,TT,c,r):
...
-    i = oF_[oF.nF].gv_map[inspect.currentframe().f_back.f_lineno]
+    n = executing.Source.executing(inspect.currentframe().f_back).node  # get current gv_ ast node
+    toF = oF_[oF.nF]
+    for i, g in enumerate(toF.g_): 
+        if g.lineno == n.lineno: break
     if v > 0: return v
-    else: oF.gv_[i] -= v  # then oF.w = vt_(oF.dTT)[0] + sum(oF.gv_)
+    else:     oF.gv_[i] -=v; toF.gv_[i] -= v  # then oF.w = vt_(oF.dTT)[0] + sum(oF.gv_)  (should be accumulating to the typ oF in oF_?)
 
-def gv_sites(fdef):  # g_[i] <-> oF.gv_[i], same order as build
```
kwcckw, line 185: > Yes, so gsites can assign the Is in the body at initialization too, instead
of manually.
That assigned i will still be used at runtime

I see, this sounds better. So during runtime, we still need to get the calling gv_ function's ast node so that we can retrieve their `i` in oF.gv_, and i think this could be done with lineno, same as the approach in my current commit?

    n = executing.Source.executing(inspect.currentframe().f_back).node  # get current gv_ ast node
    toF = oF_[oF.nF]
    for i, g in enumerate(toF.g_): 
        if g.lineno == n.lineno: break

boris-kz: It should be assigned by gv_sites in ffeedback, not at runtime

frame_2D_alg/meta_code.py
boris-kz: I should be static, gv_sites needs to assign it at init and in ffeedback

frame_2D_alg/meta_code.py
boris-kz: that will be in ffeedback too, summed from toF.call_?

frame_2D_alg/meta_code.py
boris-kz: update all those in ffeedback

frame_2D_alg/meta_code.py
```diff
@@ -203,15 +200,15 @@ def F_body_():  # form function body by AST tracing
 
     def build(func, node):  # AST → CoF | (type,sub_) | ast_leaf | None
         if isinstance(node, ast.Call) and isinstance(node.func, ast.Name):
-            if node.func.id=='gv_': func.gv_ += [0]
+            if node.func.id=='gv_': func.gv_ += [0]; func.g_ += [node]
```
boris-kz, line 203: that g_ can be local, we don't need it at runtime?

kwcckw: We need it during init and ffeedback right? Which mean we need to pack it during init, and after the runtime, access it again during ffeedback?

frame_2D_alg/meta_code.py
kwcckw: Okay, during runtime, when we call gv_ function, the `i` should be retrieved from the pre-computed values via gv_sites? But my concern is how do we really match the pre-computed `i` values with the calling gv_ without using node.lineno?

frame_2D_alg/meta_code.py
kwcckw: Okay right, we can do it per batch in ffeedback using toF call_.

frame_2D_alg/meta_code.py
boris-kz: no, it's only needed to be formed in ffeedback, not as a permanent oF attr

frame_2D_alg/meta_code.py
boris-kz: all we need is gv_() i matching oF.gv_[i], never mind lineno

...
This is the clean resolution — it works *because* it splits the index into two disjoint spaces that never have to agree:

- **Original functions:** manual literal `gv_(expr, 0)`, frozen in bytecode, written at runtime by the `gv_(v,i)` hook. Never renumbered — split/merge don't touch original source.
- **Derived bodies (split/merge):** index reassigned by *position in the rebuilt body*, automatically, at rebuild time. Position ⇒ unique and dense for free, so the collision/gap worry that's haunted the last few turns evaporates exactly where it had teeth.

The desync fear from two turns ago is now void: ffeedback reassigns only derived bodies; the bytecode literal of the original is never the thing being rewritten. No contradiction.

But the precondition is the same wall we keep hitting: **rebuild operates on `oF.body`, which is post-AST** — `gv_` sites are `(ast.Call, sub_)` with `func.id` gone. Renumber can't find gates unless build stamps gate identity *into the body*. So the marker is mandatory (uniform arity, last turn's form):

```python
            if node.func.id == 'gv_':
                sub_ = [r for t in ast.iter_child_nodes(node) if (r := build(func,t)) is not None]
                sub_, fc_ = zip(*sub_) if sub_ else ((),())
                return (('gv_', node.args[-1].value), sub_), sum(fc_)+costs.get(ast.Call,0)
            if (i := iF_.get(node.func.id)) is not None: return oF_[i], 3
```

Renumber rides on the rebuild that already happens — `split_oF_` builds `grp` lists, merge assembles a new body — so it's a counter threaded through assembly, not a separate mutating pass (body nodes are immutable tuples; you emit fresh markers, you don't edit in place):

```python
def renum_gv(body, gi=0):  # reassign derived-body gate slots by position
    out = []
    for n in body:
        if isinstance(n, tuple) and isinstance(n[0], tuple) and n[0][0] == 'gv_':
            out.append((('gv_', gi), n[1])); gi += 1
            gi = _descend(n[1], ...)   # recurse if gates nest in sub_
        elif isinstance(n, tuple):
            sub_, gi = renum_gv(list(n[1]), gi); out.append((n[0], tuple(sub_)))
        else: out.append(n)
    return out, gi
```

Closure — the one half that's still missing. Naming the consumer of the *derived* index:

- If it feeds **regenerated source** (`gv_(expr, i)` literals re-emitted for a refactored function), this is complete and correct by construction.
- If it feeds a **reconstructed `oF.gv_` on the derived CoF**, renumber is necessary but *not sufficient*: nothing populates that vector. `gv_` is runtime-only — allocated at call entry (155), written during the call (184), summed into `oF.w` at 164, then gone. It is never aggregated across split/merge. So a derived CoF has no gv_ accumulator and no aggregation path into one. You'd be assigning slots to a vector with no filler and no reader.

So: the marker + position-renumber gives you automatic, collision-free indexing on rebuild — but it only closes if the derived index lands in regenerated literals. If you intend derived CoFs to *carry* gate values, the gv_ aggregation pipeline (child→parent sum, mirroring the `dTT`/`fc` summing in `sum2O`/`add2O`) is the unbuilt half. Which target is the index for — regenerated code, or a live derived `oF.gv_`?

kwcckw: > all we need is gv_() i matching oF.gv_[i], never mind lineno

> no, it's only needed to be formed in ffeedback, not as a permanent oF attr

Okay, so that's mean we still need to define it in each `gv_` input argument, and `g_` actually could be retrieved via `gv_sites` and `nF_` during the ffeedback.

The existing `func.gv_ += [0]` in the `build` should be already good enough, we don't even need that input `i` value from `i = node.args[1].value ` too.

boris-kz: def gv_sites(fdef, fi=1):  # g_[i] <-> oF.gv_[i], same order as build

    g_ = []                # eval gate / oF.gv_[i] after accumulation
    def rec(node):
            if node.func.id == 'gv_':
                if fi: node.args[-1].value = len(g_)   # overwrite gv_ index
                g_.append(node)
            if node.func.id in iF_: return   # build substitutes callee, no descent
        for child in ast.iter_child_nodes(node): rec(child)
    rec(fdef); return g_

...
`gv_(expr, 0)` parses to `args = [expr, Constant(0)]`, so `args[-1]` is the second positional, which binds to param `i`. Overwrite `Constant.value`, compile the fdef, and the running call passes the new int as `i`; `oF.gv_[i] -= v` already consumes it. The param *is* the slot — `gv_(v, i)` needs nothing added.

Two conditions, both already met on your path:

- **Positional.** `args[-1]` is the slot only if the index is positional. `gv_(expr, 0)` ✓. `gv_(expr, i=0)` would put it in `node.keywords` and `args[-1]` would be `expr` → you'd corrupt the gated value. Keep it positional; that's the unenforced invariant.
- **Edit-before-compile.** The overwrite must precede `compile`/`exec`. Since `oF_split`/`oF_merge` emit the callable, the order is rebuild fdef → `gv_sites(fdef, fi=1)` → compile → exec → `traced`. Run assign *inside* that pipeline, not after.

So the derived path closes: position ⇒ dense `0..N-1` ⇒ no collisions, no gaps, automatic. The manual-literal original path is untouched and separate — both feed the same `i` param.

Remaining contract — sizing. `gv_sites` in assign-mode must also yield `N` for allocation. `len(g_)` after the walk is it. Line 155 currently reads `len(oF_[i].gv_map)`; for the derived callable, store the count instead:

```python
            if node.func.id == 'gv_':
                if fi: node.args[-1].value = len(g_)
                g_.append(node)
    rec(fdef)
    if fi: oF_[fi_index].gv_n = len(g_)   # or return len(g_) to the caller to stash
    return g_
```

and 155 → `oF.gv_ = np.zeros(oF_[i].gv_n)`. Same number that assign just wrote as the max slot, so sizing and indices can't disagree — they're the same counter.

That's closure for B-via-literal: `gv_` unchanged, `gmap`/`get_gmap`/lineno-lookup all deleted, original gates manual, derived gates auto-assigned by `gv_sites(fi=1)` at rebuild, sizing from the same `len(g_)`. The only thing still riding on you is the positional invariant — nothing checks that `args[-1]` is the slot and not an operand.

frame_2D_alg/meta_code.py
boris-kz: Yes — `gv_sites` is a redundant second traversal. `build` already shares its exact predicate (`node.func.id=='gv_'`), its callee prune (the `iF_` early return), and its order (pre-order `iter_child_nodes`). The only things `build` doesn't currently emit are `gv_sites`'s three products: the node back-map, the position index, the count. Fold all three into build's gv_ branch and the separate walk disappears:

```python
            if node.func.id == 'gv_':
                i = len(func.g_)                # position = slot
                node.args[-1].value = i         # assign literal (read by recompiled derived fdef)
                func.g_ += [node]               # back-map: func.g_[i] <-> oF.gv_[i]
                sub_ = [r for t in ast.iter_child_nodes(node) if (r := build(func,t)) is not None]
                sub_,fc_ = zip(*sub_) if sub_ else ((),())
                return (('gv_', i), sub_), sum(fc_)+costs.get(ast.Call,0)
            if (i := iF_.get(node.func.id)) is not None: return oF_[i], 3
```

init (207): `func.g_ = []` (replaces `gv_map = {}`). sizing (155): `np.zeros(len(oF_[i].g_))`. Delete `gv_sites`.

Naming: I used `func.g_` for the node list to match your own `g_[i] <-> oF.gv_[i]` comment — keep `gv_` for the runtime value array (164 `sum(oF.gv_)`). Template-`g_` (nodes) and runtime-`gv_` (zeros) are distinct objects so no aliasing, but reusing `gv_` for both would overload it.

Closure — fold covers originals only, because `F_body_` runs build over `nF_` (originals) at init. Two consequences:

**Derived fdefs must also pass through build.** `split_oF_` currently assembles `CoF(...body=grp)` by hand (301) — never calls build — so the fold doesn't reach them. For `gv_sites` to vanish *entirely*, split/merge must regenerate an `ast.FunctionDef` and run it through build (then compile), instead of stitching body tuples. If you keep hand-assembly there, you still need one gate pass on the derived body — but it can't be this build-fold (no AST), so you'd re-walk the tuple markers. The consolidation only completes if rebuild produces fdefs and feeds build.

**`node.args[-1].value = i` is inert on originals, load-bearing on derived.** F_body_ over `nF_` mutates the mirror; originals run from their own bytecode, so the overwrite there just confirms the manual literal (and silently corrects a mismatched one). On a *recompiled* derived fdef it's the actual assignment. Same line, two regimes — consistent only because position-index equals the manual literal you write for originals. If they ever differ, the recompile path wins for derived, the bytecode wins for originals; nothing reconciles them, but they never share a function, so no conflict.

So: fold into build, drop `gv_sites`, route derived fdefs through build at rebuild. That makes build the single gate authority — assign, back-map, marker, count, all one walk.

kwcckw: > node.args[-1].value = i  

So this is actually replacing the `gv_` function `i` value before the runtime? And the defined input `i` value will be ignored?

If this is working, it's actually can resolve both the gi init and the i tracing issues now.

frame_2D_alg/meta_code.py
boris-kz: def ffeedback(frame, aTT,oTT, aL,oL):  # recompute filters from regime drift; fork: reform oF_ on cross-regime drift

    global ave, avd
    dTT = dc = dr = 0
    # H init @ 1st term:
        dTT = aL.dTT-aTT; aTT=aL.dTT; dc= aL.c-_ac; dr= aL.r-_ar
        ave, avd = vt_(aTT)
        # filters *= ave
            dTT += oL.dTT-oTT; oTT=oL.dTT; dc+=oL.c-_oc; dr+=oL.r-_or
            split_oF_(); cluster_oF_()  # add eval?
            for oF in oF_:  # from split/merge
                oF.body=[]; oF.fc=0; oF.gv_=[]; oF.g_=[]
                    if (r := build(oF, node)): t,fc = r; oF.body += [t]; oF.fc += fc  # compile oF.fdef
    Fvt_([frame],dTT,dc,dr)
    return frame, aTT, oTT, aL, oL

frame_2D_alg/meta_code.py
boris-kz: yes, this is supposed to assign i:

return (('gv_', i), sub_), sum(fc_)+costs.get(ast.Call,0)

kwcckw: Okay, this is really something new to me, i will need to step through it and check again first.

frame_2D_alg/meta_code.py
kwcckw: In this section:

      for oF in oF_:  # from split/merge
            oF.body=[]; oF.fc=0; oF.gv_=[]; oF.g_=[]
            
 After me merge or split oF, `oF.fdef` might not be referencing any of the prior function now? Their body may contain partly of the functions of consists of multiple prior oFs' body? So i think `ast.iter_child_nodes(oF.fdef)` shouldn't be working here?

boris-kz: This is what I got:

This is a recompile-ordering question, and it splits cleanly on one fact: a caller's fdef references a callee **by name** (`ast.Name.id`), not by object. So whether a caller needs touching depends entirely on whether the *name binding* moved.

**Two distinct things a merge/split changes:**

1. **A callee's body** (split A into A′, or merge A+B into A′). Callers that call `A(...)` are unaffected at the *fdef* level — the name `A` still resolves, just to recompiled bytecode. Build's view of the caller is also unchanged: `build` substitutes the callee as `oF_[iF_['A']]` (the early return), an *index* lookup, not the callee's body. So the caller's body-tuples don't embed the callee's internals; they embed a CoF reference resolved at build time. Recompiling A′ under the same name → callers need nothing.

2. **The name set itself** — split creates a *new* name (A → A′, A″), or merge *removes* one (A,B → A′, B gone). Now any caller of A or B has a dangling `ast.Name.id` that no longer exists in `iF_`. Those callers' fdefs must be rewritten (redirect the `Name.id`) and recompiled. This is the only case that propagates.

So the answer to "update in each, or after all": **after all**, and it's forced, not just convenient. Within a batch of merges/splits, an intermediate name may be created then immediately consumed by a later merge (A′ formed, then A′+C merged away). If you rewrite callers eagerly per-operation, you redirect them to A′, then have to redirect again — and worse, a caller of A *and* of C gets touched twice with a stale intermediate in between. Defer: run the whole batch, producing the final name set and a single `old_name → final_name` map, then one pass over all fdefs to apply redirects + a single recompile per changed fdef.

The ordering contract that makes deferral correct:

- **Phase 1** — all merges/splits mutate bodies and the `iF_`/`oF_` name set, recording renames. No compile yet.
- **Phase 2** — resolve renames transitively (A→A′→A″ collapses to A→A″), so callers skip intermediates.
- **Phase 3** — for every fdef whose body calls a renamed name, rewrite the `Name.id`; then `gv_sites`/build-fold to re-stamp gate slots (slots are position-local, unaffected by the rename, but a rewritten fdef recompiles anyway); then compile.

Caller detection is already in the code: `call_sites(fd)` (235) returns the `ast.Call` nodes with `func.id in iF_`. Walk it per fdef, redirect any `id` in your rename map.

One sharper point — whether a caller even needs recompiling when only a *callee's body* changed (case 1) depends on your execution model. If callers hold a Python reference to the *callee function object*, rebinding the name in the module namespace updates all callers for free (Python resolves the global at call time). If callers were compiled to inline or capture the old object, you'd need to recompile them. Given `traced` wraps by `__name__` and dispatch goes through the name, case 1 almost certainly needs **no caller recompile** — only the callee. So the expensive propagation is purely the rename case (2), and batching collapses it to one pass.

Net: per-operation, update only the oF being formed. Defer all caller-fdef redirects to a single post-batch phase keyed on the transitive rename map. The thing that determines blast radius is name churn, not body churn.

frame_2D_alg/meta_code.py
```diff
@@ -152,6 +153,7 @@ def inner(*a, **kw):
             _CoF = CoF._cur.get(None)
             oF = CoF(nF=iF_[func.__name__], root=_CoF)
             i = iF_[func.__name__]; oF_[i].call_ += [oF]  # F_call_T_[i][oF.nF] += oF.dTT
+            oF.gv_ = copy(oF_[i].gv_)
```
boris-kz, line 156: so we only need gv_ in toF? actually it's the same for all oF attrs, they are not instance-specific?

kwcckw: But we need it to update w in callees?

frame_2D_alg/meta_code.py
kwcckw: Okay, so this involves the new fdef names. So theoretically, when we call a specific function such as cross_comp, it should resolves to a new fdef after the split or merge, and that fdef should packs the new body, which the code will run?

frame_2D_alg/meta_code.py
boris-kz: that can still be per ffeedback, which works on toF_

frame_2D_alg/meta_code.py
boris-kz: right. even if oF was not modified, its callees may be replaced.
another draft:

    global ave, avd
    dTT = dc = dr = 0
    # H init @ 1st term:
        dTT = aL.dTT-aTT; aTT=aL.dTT; dc= aL.c-_ac; dr= aL.r-_ar
        ave, avd = vt_(aTT)
        # filters *= ave
            sF_, rF_ = split_oF_()  # splits, remaining oF_
            smF_,rF_ = cluster_oF_(sF_,rF_)  # splits+merges, remaining
            # for total dict oF_, not verified:
            for F in oF_: oF_[F] = oF_.get(oF_[F], oF_[F])   # collapse A->A'->A''
            for oF in oF_:  # update all fDefs in new + remaining oFs
                cs = call_sites(oF.fdef)
                if not oF.body or any(n.func.id in oF_ for n in cs):
                    for n in cs:
                        if n.func.id in oF_: n.func.id = oF_[n.func.id]
                    oF.body=[]; oF.fc=0; oF.g_=[]
                        if (r := build(oF,node)): t,fc=r; oF.body += [t]; oF.fc += fc
                    compile_oF(oF)    Fvt_([frame],dTT,dc,dr)
    return frame, aTT, oTT, aL, oL

kwcckw: The compile_oF above is needed to resolve the function name, so that it can point to this new oF.body when we run the related code? Could you share the code too?

frame_2D_alg/meta_code.py
kwcckw: So we don't need gv_ in those callees, and we only need gv_ in toF, and that will be assigned in feedback only?

frame_2D_alg/meta_code.py
boris-kz: def compile_oF(oF):

    exec(compile(ast.Module(body=[oF.fdef], type_ignores=[]), '<oF>','exec'), oF.g)

probably still messed-up

frame_2D_alg/meta_code.py
boris-kz: I think all oF attrs are only needed fot the typ, not instances

frame_2D_alg/meta_code.py
kwcckw: I see, so that compile function should be the function to resolve the function name in the code. 

But in this line: 

Suppose it's adding the new oF to traced? I guess that g is a typo?

boris-kz: no idea if that makes sense:

...
`g` is one dict: the global namespace of the module the function belongs to (agg_recursion's). When Python runs `def foo(): ... ave ...`, the function stores a pointer to the dict it was defined in, and at call time every free name (`ave`, `vt_`, `gv_`) is looked up in that dict. That dict is what `g` holds.

It does two jobs in `compile_oF`, both just "use that namespace":

`exec(code, g)` — run the recompiled def *with `g` as its globals*, so (1) the fork's free names resolve against the real agg_recursion variables, and (2) the new function object gets stored at `g['foo']`.

`g['foo'] = CoF.traced(g['foo'])` — read that just-created function back out, wrap it, write the wrapped one back under the same key, so callers in that module now hit the instrumented version.

Where it comes from: it's the live `__dict__` of the running agg_recursion module, captured at parse time and stashed on the oF (`oF.g = _src.__dict__`) so you don't have to re-resolve it by name later. Concretely it's the same object as `agg_recursion.__dict__` / `vars(agg_recursion)` — the module's variable table.

The reason it must be *that* dict and not meta_code's `globals()`: the fork is agg_recursion code referencing agg_recursion names. Exec it against the wrong dict and `ave`/`vt_`/`gv_` are undefined at call time → `NameError`. `g` is just "the right namespace to compile into," named short because it's used twice on one line.

frame_2D_alg/meta_code.py
kwcckw: That's including called dTT, c and r and N_?

frame_2D_alg/meta_code.py
kwcckw: I see, so that's a dict that store the function pointer, which the code will run later. But this maybe complicated to run and debug later. Not sure if there's a better method?

frame_2D_alg/meta_code.py
boris-kz: do we ever use them at runtime?
oF data is only needed to eval for refactoring?

frame_2D_alg/meta_code.py
boris-kz: don't know, look into it

frame_2D_alg/meta_code.py
kwcckw: Not at runtime, they mostly will be used in ffeedback, during split or megre, but that's also involves only c and w. So all callees params are actually redundant, we just need them for the tree structure tracing? Those params should be packed in toF?

frame_2D_alg/meta_code.py
kwcckw: Yes, I will step through them later to verify if it's can work as intended.

frame_2D_alg/meta_code.py
boris-kz: Right, it's a meta code thing

PR #186, opened 2026-06-25
6/25/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -908,8 +914,10 @@ def pack_seg(frame, nF, w, c, _dTT):  # drift-gated regime termination for aH an
             return seg
 
 if __name__ == "__main__":  # './images/toucan_small.jpg' './images/raccoon_eye.jpeg', add larger global image
-    trace_func(vars())
-    parse_funcs(["agg_recursion.py"])  # populate nF_
+    trace_func(vars()); g = vars()   
+    parse_funcs(["agg_recursion.py"])   # populate nF_
+    for oF in oF_: oF.fdef = nF_[oF.nF]  # update fdef with ast node
+    inject_oF_(oF_, g)
```
kwcckw, line 941: So by using the `node.args` approach in `build` to insert the `i` value, we need to run `compile_oF` to inject the updated node to the code first, which we need to run  `compile_oF`  here, and i packed it in the `inject_oF_`.

frame_2D_alg/agg_recursion.py
```diff
@@ -868,7 +868,15 @@ def fill_frame(_iy,_ix, elev, T):  # expand level_frame from pixel-level seed ti
...
 
-    global ave, avd
+    global ave, avd, oF_
+    # this should be here to update toF only? Right before the ffeedback.
+    for oF in oF_:
+        C = oF.c; TT=np.zeros((2,9)); R=0
+        for tt,c,r in oF.vt_: w= c/(C or eps); TT+=tt*w; R+=r*w
+        oF.dTT,oF.r = TT,R
+        oF.w = vt_(oF.dTT)[0] + sum(oF.gv_)
+        oF.wTT = cent_TT(tt, r)
```
kwcckw: So params should be packed to toF only, and we just need to update them right before ffeedback? The code to update callee's params in the `traced` can be removed now.

frame_2D_alg/meta_code.py
```diff
@@ -193,8 +199,8 @@ def build(func, node):  # AST → CoF | (type,sub_) | ast_leaf | None
 
     if isinstance(node, ast.Call) and isinstance(node.func, ast.Name):
         if node.func.id=='gv_':
-            func.g_+=[node]; l=len(func.g_)  # func.g_[i] <-> oF.gv_[i]
-            node.args[-1].value = l  # read by recompiled fdef
+            func.gv_+=[0]; func.g_+=[node]; l=len(func.g_)  # func.g_[i] <-> oF.gv_[i]
+            node.args.append(ast.Constant(value=l-1))  # gv_(v): add the i index (we must not parse i to the gv_)
```
kwcckw: So we should use `I-1` here since the index starts from 0.

frame_2D_alg/agg_recursion.py
boris-kz: it should be sum_vt(oF.call_) + sum(oF.gv_)?
or sum(call.gv_) for call in oF-call_ too?

frame_2D_alg/meta_code.py
boris-kz: ok thanks

frame_2D_alg/agg_recursion.py
kwcckw: I think we need both ` sum_vt(oF.call_) + sum(oF.gv_)?` to sum same type's oF params, and `sum(call.gv_) for call in oF-call_ ` to sum gate params? The gate params didn't sum to the oF's param in traced so there shouldn't be any redundancy.

boris-kz: each call should be summing gv_ locally, then sum_vt(oF.call_) is enough?

kwcckw: From my updated Fvt_:

      def Fvt_(N_,TT,c,r):
          oF = CoF.get(); toF = oF_[oF.nF]  
          toF.N_+=N_; toF.c+=c; toF.vt_ += [[TT,c,r]]  # data per call

We actually pack those gate params into `toF.vt_` first, so those g's params should be added separately into toF after sum_vt(oF.call_)?

boris-kz: I don't think it needed, just add fg here:

def sum_vt(N_, fg=0, fr=0, fm=0, wTT=wTT, fdiv=1):  # basic weighted sum of CN|CF list

    C = sum(n.c for n in N_); R = gV = 0; TT = np.zeros((2,9))
    for n in N_:
        w = n.c / C; TT += (n.rTT if fr else n.dTT)*w; R += n.r*w
        if fg: gV += sum(n.gv_) * w
    if fm:
        m,d = vt_(TT, wTT)
        if fdiv: m/= ave*R; d/= avd*R  # in 0-inf for summation
        else:    m-= ave*R; d-= avd*R  # in -1:1 without r
        if fg:   m+= gV
        return   m,d,TT,C,R
    else: return TT,C,R

boris-kz: This folds last turn's aggregation into the helper. It fixes the genexpr, names it `gV` not `gv_`, and weights it `n.c/C` — same frame as TT/R. That part's right; the c-weighting resolves the raw-sum inconsistency I flagged.

Three things remain, one live, two structural.

**Live bug — `m += gV` is a scale mismatch.** `m` out of `vt_` is *relative* (`(m_/t_) @ wTT`, normalized by `t_`), then `fdiv` divides it again by `ave*R`. `gV` is **raw value units** — `gv_[i] -= v` accumulates unnormalized magnitudes. So `m += gV` adds a normalized rate to an absolute magnitude, and the offset *between* them depends on `fdiv`: under `fdiv=1` they're apart by a factor of `ave*R`; under `fdiv=0` `m` is in ~-1:1 and `gV` is unbounded. The gate term's weight relative to membership is then set by an accident of `R`, not design. To add at all, `gV` must ride the same normalization: `m += gV/(ave*R)` under fdiv, scaled to the -1:1 frame otherwise. (Line 164 has the same latent mismatch — `vt_(dTT)[0] + sum(gv_)` — but there it's at least a single fixed frame, not toggled by a flag.)

**`gV` is computed then dropped when `fm=0`.** The loop accumulates it unconditionally under `fg`, but it's only consumed inside `if fm:`. `sum_vt(N_, fg=1)` with `fm=0` does the work and returns `(TT,C,R)` — gV gone. So this gives no path to the gate value *without* the m/d eval. And it's a **scalar** (`sum(n.gv_)` collapses per-site), so even via `m` the literal-`i` index is lost. → This helper serves the line-164 `w`-update use and **cannot** serve the ffeedback consumer that needs the per-site vector. Those are now provably two aggregations: feedback's is a separate per-function (homogeneous), `axis=0` pass. Don't try to make `fg` do both — the scalar collapse is what makes this one heterogeneity-safe and simultaneously useless for indexing. You can have the index or mixed-`N_` tolerance, not both in one function.

**`fg` types the helper.** `gv_` exists only on CoF (line 143); base CF/CN/CL don't carry it (fields at 87-88). So `fg=1` presupposes `N_ : list[CoF]` — on a data-node `N_` it's `AttributeError`, not a benign 0. But every other `sum_vt` caller passes CN/CL data nodes for clustering. So `fg` is a CoF-only mode bolted onto the generic data-node summer, and folding gate residual into the clustering `m` is a category error regardless of scale: gate residual is a property of the *op* (does this gate suppress value → is the code worth refining, which is the `w` frame), while `m` here is a node's *fit to a cluster*. A node that suppressed value getting higher cluster-membership is meaningless. It's coherent at line 164 because `oF` *is* a code object in the `w` frame; it's incoherent in generic `sum_vt(fm=1)`.

Closure: `fg` buys one line at 164 and costs a type precondition the signature can't state plus a normalization the function can't honor. The gate add belongs at the one site that owns `CoF.w`, where `oF.gv_` and the right frame are both in hand. Keep `sum_vt` pure value/membership. If you still want it here, the minimum to not ship a scale bug is normalize `gV` by `ave*R` before the add and document `fg ⟹ N_:CoF` — but it still won't advance the feedback contract, which needs the separate axis=0 pass.

kwcckw: > I don't think it needed, just add fg here:
> 
> def sum_vt(N_, fg=0, fr=0, fm=0, wTT=wTT, fdiv=1): # basic weighted sum of CN|CF list
> 
> ```
> C = sum(n.c for n in N_); R = gV = 0; TT = np.zeros((2,9))
> for n in N_:
>     w = n.c / C; TT += (n.rTT if fr else n.dTT)*w; R += n.r*w
>     if fg: gV += sum(n.gv_) * w
> if fm:
>     m,d = vt_(TT, wTT)
>     if fdiv: m/= ave*R; d/= avd*R  # in 0-inf for summation
>     else:    m-= ave*R; d-= avd*R  # in -1:1 without r
>     if fg:   m+= gV
>     return   m,d,TT,C,R
> else: return TT,C,R
> ```

Okay right, thanks. We actually can fold gv_ related accumulation into the sum_vt. But in this section:

      for n in N_:
          w = n.c / C; TT += (n.rTT if fr else n.dTT)*w; R += n.r*w
          if fg: gV += sum(n.gv_) * w

We don't need gV when fm is false since we sum gV into m only?

boris-kz: it should really be separate:

    global ave, avd, oF_
    for oF in oF_:
        oF.m, oF.d, oF.dTT,oF.c,oF.r = sum_vt(oF.call_, fm=1, fnorm=0)
        oF.gv_ = np.sum([c.gv_ * (c.c / oF.c) for c in oF.call_], axis=0)  # for gate eval
        oF.w = oF.m*oF.d + sum(oF.gv_)
...

def sum_vt(N_, fr=0, fm=0, wTT=wTT, fnorm=2):  # basic weighted sum of CN|CF list

    for n in N_:
        w = n.c/C; TT += (n.rTT if fr else n.dTT)*w; R += n.r*w
    if fm:
        m,d = vt_(TT, wTT)
        if fnorm:
            if fnorm==2: m/= ave*R; d/= avd*R  # in 0-inf for summation
            else:        m-= ave*R; d-= avd*R  # in -1:1 without r
        return   m,d,TT,C,R
    else: return TT,C,R

boris-kz: fnorm=0 fixes the normalization split: `oF.m` is now raw, so it agrees with `oF.dTT` (both un-normalized), and `vt_(oF.dTT)[0]==oF.m`. Single c-weighted `gv_` aggregation, reused via `sum(oF.gv_)`. Those three are clean now.

`oF.w = oF.m*oF.d + sum(oF.gv_)` introduces one new problem and leaves the standing one.

**`m*d` doesn't fix the additive mismatch — it sharpens it.** `m,d` from `vt_` are dimensionless rates (`(·/t_)@wTT`), so `m*d` is rate². `sum(oF.gv_)` is raw comparand magnitude. You're still adding two unlike scales; making the match term a *product* makes it quadratic in `Σw` while the gate term stays linear, so the gate residual is negligible when `m*d` is large and dominant when it's small. The match/gate balance in `w` is set by the accidental magnitude of `m*d`, not by design — same disease the `+` had, worse curvature.

**`m*d` also redefines what `w` is.** `m` is compression/keep-evidence; `d` is variation/split-evidence. `m*d` peaks when both are mid and →0 when either →0 — that's an *interaction/refinement-potential* term (worth refining = has signal and unresolved variation), which is defensible. But it's a different quantity than line 164's `vt_(dTT)[0] + sum(gv_)` = plain `m + gate`. Both writers store into `oF.w`, and both feed the *same* consumer: `M = sum([c.w*c.c for c in olp])` (meta 265, comp_caller overlap). So overlap eval now mixes per-instance `m+gv` and per-function `m*d+gv` in one sum. `w` is less of a stable quantity than before, not more. Also note `d` already carries `wTT[1]=wT*avd` out of `vt_`; multiplying by `m` re-weights it a second time with no `ave/avd` frame, so the product's scale is uncontrolled.

If the intent is interaction, normalize it: `(m*d)/(ave*avd*R)` or build the worth inside one `vt_` frame, and make line 164 use the same definition. Right now the two w-definitions are diverging each turn.

**The closure is still unaddressed, and this revision made it sharper to state.** `oF.gv_` is now stored per-site but its only reader in the flow is `sum(oF.gv_)` on the very next line — collapsed to scalar. So:

- For its *stated* purpose ("for gate eval" — deciding which gates to prune), nothing reads the per-site vector. split_oF_/cluster_oF_ (line 882) don't touch `gv_`; they rebuild `oF_` from `body/fc/caller_/typ`.
- For its *actual* use (the w scalar), you don't need the vector — a scalar accumulation suffices.

So the index is orphaned for what it's for and redundant for what it does. And even if a per-site consumer existed, it can't sit after split/cluster: `sum2O` rebuilds `F.gv_map` fresh (meta 381-2), `split_oF_` builds `sub.gmap` fresh (301), and `gv_()` still resolves by lineno (`gv_map[f_back.f_lineno]`, 182) — so the aggregated vector is in the old index space, the rebuilt functions in a new lineno-keyed one with no faithful linenos for synthesized bodies.

Two questions that decide the shape, neither arithmetic:
1. Who reads `oF.gv_` **per-site**, and does that read happen *before* the rebuild or *inside* split_oF_/cluster_oF_? If nobody, drop the vector and keep a scalar.
2. The gate decision (prune which `gv_` site) has to run where the old index is still valid, and `gv_()` has to be re-keyed from lineno to a literal `i` written into the regenerated `FunctionDef`, or the surviving gates re-execute under a broken lookup next regime.

Until 1–2 are answered, this loop's arithmetic is converging onto a value (`oF.gv_`) that the pipeline neither consumes per-site nor carries across the rebuild.

frame_2D_alg/agg_recursion.py
```diff
@@ -880,17 +888,15 @@ def ffeedback(frame, aTT,oTT, aL,oL):  # recompute filters from regime drift; fo
...
                 if not oF.body or any(n.func.id in oF_ for n in cs):
                     for n in cs:
-                        if n.func.id in oF_: n.func.id = oF_[n.func.id]
-                    oF.body=[]; oF.fc=0; oF.g_=[]
-                    for node in ast.iter_child_nodes(oF.fdef):
-                        if (r := build(oF,node)): t,fc=r; oF.body += [t]; oF.fc += fc
-                    compile_oF(oF)
+                        if n.func.id in oF_: n.func.id = oF_[n.func.id]      
+                    oF_ += [oF]
+            inject_oF_(oF_, globals())
```
boris-kz: this is conditional on oF_?

frame_2D_alg/agg_recursion.py
kwcckw: Actually why `w` is the product of m and d?

     oF.w = oF.m*oF.d + sum(oF.gv_)

Previously gv is added to m only?

frame_2D_alg/agg_recursion.py
kwcckw: Yes thanks, so if oF_ os empty, we shouldn't change them at all.

frame_2D_alg/agg_recursion.py
boris-kz: `oF.w is a total function value, m and d are both oF contributions. But each is in 0:1, *9`
I think we need 
`TT = oF.dTT*oF.wTT; m,d = sum(TT[0]), sum(abs(TT[1]))`

boris-kz: def ffeedback(frame, aTT,oTT, aL,oL):  # recompute filters from regime drift; fork: reform oF_ on cross-regime drift

    global ave, avd, oF_
    for oF in oF_:
        if not oF.call: continue
        oF.dTT,oF.c,oF.r = sum_vt(oF.call_)
        tt = oF.dTT * oF.wTT * oF.c  # extensive values
        oF.m, oF.d = sum(tt[0]), sum(abs(tt[1]))
        oF.gv_ = np.sum([c.gv_ for c in oF.call_], axis=0)  # for gate eval
        oF.w = oF.m + oF.d + sum(oF.gv_)  # all extensive: total oF contribution

'''
Arithmetic's done: all three terms extensive (m,d via `×oF.c`, gv_ raw-sum because each instance's gv_ is already extensive), `m+d` not `m*d`. Nothing left to iterate in this loop. So the value of further turns here is zero until the gate pipeline exists — and the source shows that's closer than the memory framing implies.

Crash first: `oF.call` → AttributeError (attr is `call_`). Use `if not oF:` — CF.`__bool__` already returns `bool(f.call_)` (175).

Now the closure, traced, because the migration is **mid-flight in the source**:

`def gv_(v):` takes one arg. Every call site passes two — a literal index (12×`0`, 6×`1`, 3×`2`, none without). So `gv_(expr, i)` against the current signature is a TypeError at all 30 sites. The call sites are already written for the target `gv_(v, i)`; only the function body lags, still doing `gv_map[f_back.f_lineno]` (182). So the "re-key from lineno to literal i" prerequisite is three lines:
```
def gv_(v, i):
    oF = CoF.get()
    if v <= 0: oF.gv_[i] -= v
```
Drop the lineno lookup; `gv_map` is no longer needed at runtime; oF.gv_ sizing comes from i-space (`max i + 1` over the body's gv_ calls), not `len(gv_map)` (155).

That exposes the one contract that genuinely *isn't* traceable from current code, because it's the part split_oF_/cluster_oF_ don't do: **i is function-local** — each function numbers its gv_ from 0. Merge two functions' bodies (cluster_oF_→merge_oF, or split regrouping) and you get two `gv_(_, 0)` in one body, colliding. So body-merge must renumber the i literals *in the regenerated FunctionDef* and remap the gv_ vector slots in lockstep. Right now split/cluster emit `CoF.body` lists and a fresh `gmap` count — not an executable FunctionDef with rewritten `gv_(expr, i)` calls. That regeneration is the actual remaining work, and it's the thing your aggregated `oF.gv_` is waiting on: the vector is in the pre-merge i-space, the rebuilt functions in a renumbered one.

And the consumer you're aggregating *for*: the prune decision is `suppressed value gv_[i]` vs the gate's own cost — keep gates that reject enough to earn their evaluation cost, drop the rest. But per-site cost is collapsed into the scalar `fc` during `F_body_` build; only `gv_[i]` survives. So the prune input is half-missing — you need per-site cost retained alongside the i-indexed gv_, or the vector you're carefully aggregating has nothing to be compared against.

So the untraced contracts, in build order:
1. `gv_(v, i)` — consume the literal already at every call site; delete lineno path. (trivial, unblocks runtime)
2. oF.gv_ sized and indexed by i, not gv_map. (trivial)
3. split/cluster regenerate an `ast.FunctionDef` that **renumbers gv_ i-literals on merge** and remaps the gv_ vector. (the real work; not derivable from current code)
4. retain per-site cost so the prune step has both operands. (gap in F_body_)

Standing one-liners, unchanged: line 164's per-instance `w` is still relative (`vt_(dTT)[0]`) while this is extensive — reconcile, both feed `M=sum(c.w*c.c)` (265); `oF.wTT` is still the default on the static record (set `cent_TT(oF.dTT, oF.r)` first or drop the prefix).

kwcckw: Okay, and this tt: 

    tt = oF.dTT * oF.wTT * oF.c  # extensive values

Should be independent from oF.wTT? 

Because it will be used to compute oF wTT again?

boris-kz: dTT * wTT is what wTT is for. It won't be computed from that dTT alone

frame_2D_alg/agg_recursion.py
boris-kz: unpack compile_oF in inject_oF?

def inject_oF_(oF_, g):

    for oF in oF_:
        oF.fdef = nF_[oF.nF]; oF.g = g
        oF.body=[]; oF.fc=0; oF.g_=[]; oF.gv_=[]
            if (r := build(oF, node)): t,fc = r; oF.body += [t]; oF.fc += fc  
        # inject AST in g, recompile g[name]:
        exec(compile(ast.Module(body=[oF.fdef], type_ignores=[]), '<oF>', 'exec'), oF.g)

frame_2D_alg/agg_recursion.py
kwcckw: Okay, thanks for clarifying.

Also I will be travelling today, will check again much later.

frame_2D_alg/agg_recursion.py
kwcckw: Yes thanks. Looks like we only need compile_oF in inject_oF_.

frame_2D_alg/meta_code.py
```diff
@@ -167,19 +170,22 @@ def inner(*a, **kw):
...
                         F_call_T_[_CoF.nF][j] += oF.dTT  # add,c,r: results per callee to refine the code
+            
+            """
             CoF._cur.reset(_oF)
             return out
         inner.wrapped = True
         return inner
     def __bool__(f): return bool(f.call_)
 
 def Fvt_(N_,TT,c,r):
```
boris-kz: so we don't need this, all data stays in call_?

boris-kz: actually, it seems that sum_vt(call_) won't work, we need sum_vt(oF.vT_) instead:
https://claude.ai/share/7af55fa2-d575-446d-90ea-88fd109b560e

but we don't need oF.N_?
and Fvt_ should be upacked:

oF_[CoF.get().nF].vT_ += [[*LT]]  # +-ve Ls for oF, no oF.N_? 

I pushed partial edits

PR #187, opened 2026-06-26
6/26/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -21,10 +21,13 @@ def vt_(TT, wTT=wTT):  # base eval: multi-variate rel match, rel diff for member
...
-
-    C = sum(n.c for n in N_); R = 0; TT = np.zeros((2,9))
+    fvT = isinstance(list(N_)[0] ,list)  # list to convert set from input E (get_exemplar)
+    C = sum((n[1] if fvT else n.c) for n in N_)
+    R = 0; TT = np.zeros((2,9))
     for n in N_:
-        w = n.c/C; TT += (n.rTT if fr else n.dTT)*w; R += n.r*w
+        if fvT: tt, c, r = n 
+        else:   tt, c, r = (n.rTT if fr else n.dTT), n.c, n.r
+        w = c/C; TT += tt*w; R += r*w
```
kwcckw, line 30: > actually, it seems that sum_vt(call_) won't work, we need sum_vt(oF.vT_) instead: https://claude.ai/share/7af55fa2-d575-446d-90ea-88fd109b560e
> 
> but we don't need oF.N_? and Fvt_ should be upacked:
> 
> oF_[CoF.get().nF].vT_ += [[*LT]] # +-ve Ls for oF, no oF.N_?
> 
> I pushed partial edits

I guess it should be summed the same way as the existing sum_vt? We just need to apply the w by using the c?

frame_2D_alg/agg_recursion.py
```diff
@@ -83,7 +83,7 @@ def cross_comp(root, rr, fC=0):  # core function mediating recursive rng+ and de
 
     N_,G_ = root.N_,[]  # root is Ft, converted below, rc=rdn+olp, comp N_| B_| C_:
     L_,TT,c,r, cLT = comp_C_(N_,rr,fC=1) if fC else comp_N_(combinations(N_,2),rr)
-    oF_[CoF.get().nF].vT_ += [[*cLT]]  # copy default results from comp_s
+    if cLT: oF_[CoF.get().nF].vT_ += [[*cLT]]  # copy default results from comp_s
```
kwcckw, line 86: >  if cLT: oF_[CoF.get().nF].vT_ += [[*cLT]] 

Do we really need `*cLT` here? It could be remained as a tuple rather than a list? Then it will be `+= [cLT]`?

frame_2D_alg/meta_code.py
boris-kz: It's only summed in ffeeddback:

        tt = oF.dTT * oF.wTT * oF.c  # extensive values
        oF.m, oF.d = sum(tt[0]), sum(abs(tt[1]))
        oF.w = oF.m + oF.d + sum(oF.gv_)

frame_2D_alg/agg_recursion.py
```diff
@@ -132,10 +132,12 @@ def proj_V(_N,N, dist, dy_dx, dec):  # _N x N induction
...
-    LT = sum_vt(L_)
-    oF_[CoF.get().nF].vT_ += [[*LT]]  # +-ve Ls for oF, no oF.N_?
-    pL_ = [L for L in L_ if (L.m > ave or L.typ == -1)]
-    return pL_,*sum_vt(pL_), LT  # +ve only, redundant +-ve for oF
+    if L_: 
+        LT = sum_vt(L_)
+        oF_[CoF.get().nF].vT_ += [[*LT]]  # +-ve Ls for oF, no oF.N_?
+        pL_ = [L for L in L_ if (L.m > ave or L.typ == -1)]
+        return pL_,*sum_vt(pL_), LT  # +ve only, redundant +-ve for oF
+    else: return [],0,0,0,[]
```
boris-kz, line 140: I hate these dummy returns

frame_2D_alg/agg_recursion.py
boris-kz: right

frame_2D_alg/meta_code.py
boris-kz: but we need extensive oF.gV_:   oF.gV_ += call gv_ * call.c

    for oF in oF_:
        if not oF.vT_: continue
        oF.dTT, oF.c, oF.r = sum_vt(oF.vT_,oF.wTT)
        tt = oF.dTT * oF.c; oF.m,oF.d = sum(tt[0]), sum(abs(tt[1]))
        oF.w = oF.m + oF.d + sum(oF.gV_)  # all extensive values

frame_2D_alg/agg_recursion.py
boris-kz: instead of oF.vT_, immediately weight by C/R?
```
LTT,C,R = sum_vt(L_); cV = sum(vt_(LTT)) * (C/R)
oF_[CoF.get().nF].V_ += [cV]  # +-ve Ls for oF, no oF.N_?
pL_ = [L for L in L_ if (L.m > ave or L.typ == -1)]
return pL_,*sum_vt(pL_), cV  # +ve only, redundant +-ve for oF
```
...
Confirmed the blocker at line 248. `comp_callers` weights each caller by `c.w * c.c` — **w and c, separately**. cV folds C into the value, so c.c is gone. And comp_callers runs inside ffeedback (via `cluster_oF_`, lines 292/307, reached at 908), so it's the live driver of the oF split/merge layer. Collapse C and that layer reads `c.c = 0` → caller-overlap value → 0. **You can fuse m and d, but not c.** Keep a separate mass accumulator:
```python
oF_[CoF.get().nF].V_ += [cV]; oF_[CoF.get().nF].C_ += [C]
# ffeedback: oF.c = sum(oF.C_); oF.w = sum(oF.V_) + sum(oF.gV_)
```

Then the chain to closure, in order of bite:

**1. m/d fusion is safe — one real win.** Nothing on the oF side reads `oF.m` vs `oF.d` apart; `comp_callers` uses `w` (the sum), and the global filter `ave,avd = vt_(aTT)` (903) comes from `pack_seg` regime segments, not per-oF dTT. So collapsing m+d loses nothing currently consumed. Good.

**2. cV bakes R into value — a semantic change, not just buffering.** `w = m+d+gv` has no redundancy discount; `cV = (m+d)·C/R` does. You're redefining what split/cluster maximize (now rdn-penalized value). Maybe right — but then `gV_`, added un-discounted in ffeedback, is on a different basis: value rdn-scaled, gates not. Same homogeneity wrinkle as the c-basis one two turns back, now in R. And mechanically `C/R` needs eps (R is a mass-weighted *mean* r, can →0), and note it's `C / mean_r`, not `C / Σr` — inflates value at low rdn, which is likely the intent but isn't total-redundancy normalization.

**3. Composition fork — buffering at the leaf silently switches gross→partition.** Inside `comp_N_`, `CoF.get().nF` is comp_N_'s index, so cV lands on comp_N_'s prototype only. ffeedback's loop is flat per function type — it never rolls leaf into parent. The old `Fvt_(*cLt)` in `cross_comp` was exactly what re-credited comp value to `cross_comp`. Drop it (you must — see 5) and `cross_comp.w` no longer contains comp value: each value is now counted once, under its producing function (partition), not gross per caller. If you want `cross_comp`'s aggregate to include what it delegates, leaf-buffering alone won't give it.

**4. dTT is foreclosed.** The summed (2,9) is gone. The commented `F_call_T_[i][oF.nF] += oF.dTT` (155) / `F_call_T_` (219) is unwired per-callsite, per-param feedback — the granular filter update the canonical design points at. A scalar can't be re-expanded into it. If regime-level scalar `w` is all you'll ever feed back, fine; if per-param `wTT`/per-callsite tuning is on the roadmap, keep dTT (buffer the tuple, derive cV in ffeedback instead of at the site).

**5. Caller will crash.** Return is now `pL_,*sum_vt(pL_),cV` → last element is the scalar cV. `cross_comp` does `L_,TT,c,r,cLt = comp_N_(...); Fvt_(*cLt)` → `Fvt_(*float)` → splat-on-float TypeError. The caller-side capture must go too — which is consistent with leaf-buffering, but lands you back on fork 3. Decide 3 first; 5 follows from it.

Net: the scalar collapse is viable for value, but it is not a pure optimization — it (a) requires a parallel `C_` to keep mass alive for `comp_callers`, (b) redefines `w` to be rdn-discounted, (c) commits you to partition accounting, and (d) closes the per-param dTT-feedback path. (a) is mandatory; (b)–(d) are choices the snippet makes implicitly.

frame_2D_alg/meta_code.py
kwcckw: So oF.dTT should be summed from vT_ only? It should be summed from call_'s params too?

frame_2D_alg/agg_recursion.py
kwcckw: Okay, then we should check the output from cross_comp instead. It shouldn't return anything here if L_ is empty.

frame_2D_alg/meta_code.py
boris-kz: call_ vals are not oF own vals

kwcckw: > call_ vals are not oF own vals
> but we need extensive oF.gV_: oF.gV_ += call gv_ * call.c

 Since call_'s vals are not oF own vals, only gV_ is a special case and it's extended by call_'s gV_?

frame_2D_alg/agg_recursion.py
kwcckw: > instead of oF.vT_, immediately weight by C/R?
> cV bakes R into value — a semantic change, not just buffering

So the immediate weighting by C/R produces a different result? It's no longer be normalized by the ratio of c to the total C, and it's actually using the ratio of C/R now. This is actually different with the process in sum_vt?

boris-kz: Yes, because these values are supposed to be extensive

frame_2D_alg/meta_code.py
boris-kz: both V_ (was vT_) and gV_ are oF own vals, V_ is from operations and gV_ from blocking calls

kwcckw: Yes, but from your comment above: 

> but we need extensive oF.gV_: oF.gV_ += call gv_ * call.c

The gV_ is extensive, that's mean it should be sum extensively with their calls? Or it's just referring to the weighting by C/R now?

boris-kz: it's just a gate, the value is in blocking the calls

kwcckw: Okay thanks, then we shouldn't sum the calls' values. The oF params should be summed strictly within the function only (blocking values and operations).

boris-kz: so instead of Fvt_:

def FV_(F, tt,c,r):

    tF = oF_[F.get().nF]; tt *= tF.wTT; V = (sum(tt[0]) + sum(abs(tt[1]))) * c/r  # extensive
    tF.V_ += [V]; tF.C_ += [c]

in comp_N_:

cV = FV_(CoF.get(), *sum_vt(L_))  # +-ve Ls for oF, no oF.N_?
pL_ = [L for L in L_ if (L.m > ave or L.typ == -1)]
return pL_,*sum_vt(pL_), cV  # +ve only, redundant +-ve for oF

boris-kz: Confirmed the decisive fact at line 893: **ffeedback still reads `oF.call_`, not `V_`.** It does `sum_vt(oF.call_)` → instance `.dTT`, then derives m,d. `FV_` writes `V_`/`C_` on the *prototype* and sets no instance `.dTT`. So as proposed, `V_` is write-only and ffeedback computes m=d=0 — the exact dead-end from turn one, relocated. `FV_` changes nothing until ffeedback is rewritten to consume `V_`/`C_`. That rewrite *is* the edit; `FV_` is half of it.

First, the good news that closes the dTT worry from last turn. I traced every reader of per-oF `dTT`:
- **ave/avd** (903) come from `vt_(aTT)`, and `aTT` is the *regime* level dTT from `pack_seg` over `frame.H` (927) — not per-oF.
- **wTT** is inherited (`F.wTT = root.wTT`, 512; pulled as tt-coefs at 55), derived from root/regime dTT by the updater at 62 — not refreshed per-oF in ffeedback.
- The only live reader of `oF.dTT` is ffeedback 894 itself (→ m,d → w).

So nothing outside ffeedback needs per-oF `dTT`. `FV_` precomputing the scalar `m+d` is safe — the foreclosure I flagged doesn't bite, because `F_call_T_` is unwired and no live path reads aggregated per-oF dTT. Collapse is legitimate.

Now the forced ffeedback rewrite and its fallout:

```python
for oF in oF_:
    if not oF.V_: continue
    oF.c = sum(oF.C_)
    oF.w = sum(oF.V_) + sum(oF.gv_)   # V already = (m+d)·c/r·wTT
```

**Delete 893–895.** No per-oF `dTT`, no `tt`, no separate `oF.m`/`oF.d`. Verified nothing reads them: `comp_callers` (248) uses `oF.w` and `oF.c`; `pack_seg` uses level dTT; `sum2O` uses fc/fw/fr. m/d can vanish cleanly.

**Delete 896 — it's a standalone bug your rewrite must drop.** `oF.gv_ = np.sum([c.gv_ for c in oF.call_])` sums *instance* gv_, but `gv_()` (170) writes to the **prototype** `oF_[nF].gv_`. Instances' gv_ are empty → 896 overwrites the real accumulated prototype gv_ with zeros. Use `oF.gv_` as-is. (Latent today, fatal once you rely on gates in `w`.)

Then the three things `FV_` itself needs:

**`c/r` → `c/(r or eps)`.** `r` is the mass-weighted *mean* R from `sum_vt`; at base recursion all `n.r=0` ⟹ R=0 ⟹ `c/r` = inf into `V`. comp_N_ at the bottom hits this. Concrete crash.

**`tt *= tF.wTT` → `tt = tt * tF.wTT`.** Today `tt` is a fresh `sum_vt` accumulator, so in-place is safe — but `*=` mutates, and the moment the same array reaches `FV_` twice it squares wTT. That "twice" is exactly the composition case below. New-array costs nothing and disarms it.

**Composition fork — `FV_` silently picks partition.** `F.get().nF` is the *running* function's index, so `FV_` credits the producer's own prototype (leaf). ffeedback's loop is per-prototype with no roll-up, so `cross_comp.w` will *not* include comp/cluster value — each value counted once, under the function that produced it. If you keep a caller-side `FV_` (old line 89) to also credit the parent, you both double-count *and* hit the wTT-squaring from the shared `cLt` array. So caller-side capture must go → partition accounting. For `split_oF_`/`cluster_oF_`, which compare functions by their own throughput, partition is the right basis — but it's a choice the snippet makes implicitly, not a wash.

One inherited issue, not caused by `FV_`: `V_`/`C_` (like `call_` at 155) are never cleared, and prototypes persist across elevations. So `oF.w` is lifetime-cumulative, not per-elevation. If `split_oF_` should see this-elevation value, clear `V_`/`C_`/`call_` at the elevation boundary; otherwise accept cumulative. WIP-level, but it determines whether the regime-drift comparison sees a rate or a running total.

Net: `FV_` is correct in content (it is precisely the scalar `w` needs) but inert without the ffeedback switch to `V_`/`C_`; that switch deletes 893–896, retires `oF.m`/`oF.d`, and commits you to partition accounting — plus the eps and new-array fixes inside `FV_`.

kwcckw: For tF, originally we have tF.c, why we need tF.C_ here?

boris-kz: from claude:

Confirmed the blocker at line 248. `comp_callers` weights each caller by `c.w * c.c` — **w and c, separately**. cV folds C into the value, so c.c is gone. And comp_callers runs inside ffeedback (via `cluster_oF_`, lines 292/307, reached at 908), so it's the live driver of the oF split/merge layer. Collapse C and that layer reads `c.c = 0` → caller-overlap value → 0. **You can fuse m and d, but not c.** Keep a separate mass accumulator:
```python
oF_[CoF.get().nF].V_ += [cV]; oF_[CoF.get().nF].C_ += [C]
# ffeedback: oF.c = sum(oF.C_); oF.w = sum(oF.V_) + sum(oF.gV_)
```

Then the chain to closure, in order of bite:

**1. m/d fusion is safe — one real win.** Nothing on the oF side reads `oF.m` vs `oF.d` apart; `comp_callers` uses `w` (the sum), and the global filter `ave,avd = vt_(aTT)` (903) comes from `pack_seg` regime segments, not per-oF dTT. So collapsing m+d loses nothing currently consumed. Good.

**2. cV bakes R into value — a semantic change, not just buffering.** `w = m+d+gv` has no redundancy discount; `cV = (m+d)·C/R` does. You're redefining what split/cluster maximize (now rdn-penalized value). Maybe right — but then `gV_`, added un-discounted in ffeedback, is on a different basis: value rdn-scaled, gates not. Same homogeneity wrinkle as the c-basis one two turns back, now in R. And mechanically `C/R` needs eps (R is a mass-weighted *mean* r, can →0), and note it's `C / mean_r`, not `C / Σr` — inflates value at low rdn, which is likely the intent but isn't total-redundancy normalization.

**3. Composition fork — buffering at the leaf silently switches gross→partition.** Inside `comp_N_`, `CoF.get().nF` is comp_N_'s index, so cV lands on comp_N_'s prototype only. ffeedback's loop is flat per function type — it never rolls leaf into parent. The old `Fvt_(*cLt)` in `cross_comp` was exactly what re-credited comp value to `cross_comp`. Drop it (you must — see 5) and `cross_comp.w` no longer contains comp value: each value is now counted once, under its producing function (partition), not gross per caller. If you want `cross_comp`'s aggregate to include what it delegates, leaf-buffering alone won't give it.

**4. dTT is foreclosed.** The summed (2,9) is gone. The commented `F_call_T_[i][oF.nF] += oF.dTT` (155) / `F_call_T_` (219) is unwired per-callsite, per-param feedback — the granular filter update the canonical design points at. A scalar can't be re-expanded into it. If regime-level scalar `w` is all you'll ever feed back, fine; if per-param `wTT`/per-callsite tuning is on the roadmap, keep dTT (buffer the tuple, derive cV in ffeedback instead of at the site).

**5. Caller will crash.** Return is now `pL_,*sum_vt(pL_),cV` → last element is the scalar cV. `cross_comp` does `L_,TT,c,r,cLt = comp_N_(...); Fvt_(*cLt)` → `Fvt_(*float)` → splat-on-float TypeError. The caller-side capture must go too — which is consistent with leaf-buffering, but lands you back on fork 3. Decide 3 first; 5 follows from it.

Net: the scalar collapse is viable for value, but it is not a pure optimization — it (a) requires a parallel `C_` to keep mass alive for `comp_callers`, (b) redefines `w` to be rdn-discounted, (c) commits you to partition accounting, and (d) closes the per-param dTT-feedback path. (a) is mandatory; (b)–(d) are choices the snippet makes implicitly.

kwcckw: I see, so that's because we still need c.w * c.c  in comp_caller. But we could have just sum them into c rather than the array C_?

PR #188, opened 2026-06-27
6/27/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -901,10 +906,10 @@ def ffeedback(frame, aTT,oTT, aL,oL):  # recompute filters from regime drift; fo
         # filters *= ave
         if oL := pack_seg(frame,'oH', wBac, cBac**2, oTT):
             dTT += oL.dTT- oTT; oTT=oL.dTT; dc+=oL.c-_oc; dr+=oL.r-_or
-            sF_, rF_ = split_oF_()  # splits, remaining oF_s
-            smF_,rF_ = cluster_oF_(sF_+ rF_)  # splits+merges + remaining
-            oF_ =[]; _oF_ = smF_+rF_  # dict oF_
-            for F in _oF_: _oF_[F] = _oF_.get(_oF_[F],_oF_[F])  # collapse A->A'->A''?
```
kwcckw, line 907: Actually there's some problem with the mapping here. I changed the code so that oF will be mapped to multiple new split oFs from `split_oF_`, then multiple oFs will be mapped to the merged oF in `cluster_oF_`. 

After the split and merge:

    _oF_ = list(dict.fromkeys(_oF_.values())) 

The code above will be used to removed the mapping the preserve only the final split and merged oFs.

frame_2D_alg/meta_code.py
```diff
@@ -271,42 +274,41 @@ def split_oF_():  # divisive clustering
             grp_ += [grp]
             for grp in grp_:  # single refinement
                 fc = sum([get_fc(prim) for prim in grp])
-                sub = CoF(root=oF,fc=fc,body=grp); sub.caller_ = copy(oF.caller_)
-                sF_ += [sub]
-        else: rF_ += [oF]
-    return sF_, rF_
+                # typ remain unchanged?
+                sub = CoF(root=oF,fc=fc,body=grp,typ=oF.typ); sub.caller_ = copy(oF.caller_)
+                sub.fdef = set([oF.fdef])  # inherit node from oF
```
kwcckw, line 279: I think we need an array based fdef here because after the merging, we may get multiple ast nodes per oF, which we will need them for call_sites later.

frame_2D_alg/agg_recursion.py
```diff
@@ -55,7 +55,7 @@
 ttFrm,tAgg,ttTrc, ttN_,ttC_,ttN,ttF, ttE,ttcN,ttcC,ttcP,ttsG, ttBac,ttPrj, ttVct = [F.wTT for F in oF_]
 
 def FV_(F, tt,c,r):
-    tF = oF_[F.nF]; tt *= tF.wTT; V = (sum(tt[0]) + sum(abs(tt[1]))) * c/r  # extensive
+    tF = oF_[F.nF]; tt =  tt*tF.wTT; V = (sum(tt[0]) + sum(abs(tt[1]))) * c/(r or eps)  # extensive (tt = tt * tF.wTT to prevent overwrites into input.dTT)
```
boris-kz, line 58: default r should be 1

frame_2D_alg/agg_recursion.py
```diff
@@ -87,19 +87,21 @@ def cent_TT(dTT, r):  # EM-like weight attr matches | diffs by their match to th
...
-            E_ = get_exemplars({N for L in L_ for N in L.N_}, r,c)
-            G_,r = cluster_N(root, E_,r,c)  # cluster_C, _P, eval?
-            if G_:
-                if not root.typ: F2N(root)  # promote at 1st sub+ or agg+
-                root.H += [sum2F(L_,root,froot=1)]  # dLev per L_
-                root.Nt = sum2F(G_,root,froot=2); L=len(G_)-1  #| C_?
-                if gv_(vt_(TT,tAgg)[0]* (wAgg*L) - ave* (r+cAgg*L)):
-                    G_ = cross_comp(root.Nt,r)  # agg+
+    LVt = comp_C_(N_,rr,fC=1) if fC else comp_N_(combinations(N_,2),rr)
+    if LVt:
```
boris-kz, line 91: use walrus

frame_2D_alg/agg_recursion.py
boris-kz: what exactly is the problem with mapping?

frame_2D_alg/meta_code.py
boris-kz: I don't get, it's a new oF in new oF_, everything affected is new, old versions are no longer relevant?

frame_2D_alg/agg_recursion.py
```diff
@@ -87,19 +87,21 @@ def cent_TT(dTT, r):  # EM-like weight attr matches | diffs by their match to th
...
+        if L_:  # +ve only (L_ may empty when there's just negative links)
+            root.L_ = L_; L= len(L_)-1  # val=m+d /clust, m/comp
+            if gv_((sum(vt_(TT,ttE))*(wE*L) - (ave+avd)*(r+cE*L)) * c):  # if +ve, store neg gate values
+                E_ = get_exemplars({N for L in L_ for N in L.N_}, r,c)
+                G_,r = cluster_N(root, E_,r,c)  # cluster_C, _P, eval?
+                if G_:
+                    if not root.typ: F2N(root)  # promote at 1st sub+ or agg+
+                    root.H += [sum2F(L_,root,froot=1)]  # dLev per L_
+                    root.Nt = sum2F(G_,root,froot=2); L=len(G_)-1  #| C_?
+                    if gv_(vt_(TT,tAgg)[0]* (wAgg*L) - ave* (r+cAgg*L)):
```
boris-kz, line 103: if gv_((vt_(TT,tAgg)[0]* (wAgg*L) - ave* (r+cAgg*L)) * c):  # *c for extensive?

frame_2D_alg/agg_recursion.py
```diff
@@ -135,11 +137,12 @@ def proj_V(_N,N, dist, dy_dx, dec):  # _N x N induction
                 L_+= [pL]; N.rim+=[pL]; _N.rim += [pL]; N_+=pL.N_  # add neg C?
                 # oF.N_ is +-Ls
         else: break  # beyond initial induction range, re-sort by proj_V?
-    for N in set(N_):
-        if N.rim: N.Rt = sum2F(N.rim)
-    cV = FV_(CoF.get(), *sum_vt(L_))  # +-ve Ls for oF, no oF.N_?
-    pL_ = [L for L in L_ if (L.m > ave or L.typ == -1)]
-    return pL_,*sum_vt(pL_), cV  # +ve only, redundant +-ve for oF
+    if L_:
```
boris-kz, line 140: I don't think L_ should be empty here, we need to re-add len test in the caller?

frame_2D_alg/agg_recursion.py
```diff
@@ -891,7 +895,8 @@ def ffeedback(frame, aTT,oTT, aL,oL):  # recompute filters from regime drift; fo
 
     global ave, avd, oF_
     for oF in oF_:
-        if oF.V_: oF.w = sum(oF.V) + sum(oF.gV_)  # all extensive
+        if oF.V_: oF.w = sum(oF.V_) + sum(oF.gV_)  # all extensive
+        if oF.C_: oF.c = sum(oF.C_)  # not sure, looks redundant if we can just accumulate into c
```
boris-kz, line 899: yes, just oF.c

frame_2D_alg/agg_recursion.py
```diff
@@ -931,7 +936,7 @@ def pack_seg(frame, nF, w, c, _dTT):  # drift-gated regime termination for aH an
 if __name__ == "__main__":  # './images/toucan_small.jpg' './images/raccoon_eye.jpeg', add larger global image
     trace_func(vars()); g = vars()
     parse_funcs(["agg_recursion.py"])   # populate nF_
-    for oF in oF_: oF.fdef = nF_[oF.nF]  # update fdef with ast node
+    for oF in oF_: oF.fdef = set([nF_[oF.nF]])  # update fdef with ast node
```
boris-kz, line 939: that makes no sense, set destroys the order?

frame_2D_alg/agg_recursion.py
kwcckw: The existing code actually doesn't have any mapping at all. Both `split_oF_` and `cluster_oF_` pack the split or merged oF + remaining oFs, which there will be no mapping in the final output because `_oF_` is not a dict:

    _oF_ = smF_+rF_  # dict oF_

Hence this mapping will not work as intended:

    for F in _oF_: _oF_[F] = _oF_.get(_oF_[F],_oF_[F]) 
    
The intended workflow is to map the original oF to the split or merged new oFs?

frame_2D_alg/meta_code.py
kwcckw: This `fdef ` should be point to the ast node, but the merged oF should gets multiple ast nodes now?

Then we need it in the ffeedback to update the func.id and eventually rebuild the bodyu from `build`?

           for F in _oF_:  # update fDefs
                cs = call_sites(F.fdef)
                if not F.body or any(n.func.id in oF_ for n in cs):
                    for n in cs:
                        if n.func.id in oF_: n.func.id = oF_[n.func.id]
                    oF_ += [F]

frame_2D_alg/agg_recursion.py
kwcckw: Thanks for the reminder again.

frame_2D_alg/agg_recursion.py
kwcckw: This `L_` is conditional on `m`, which is computed from the first projected pTT. But we only can get this info after all pairs are formed, unless we set the limit to `if m > 0 or not L_`?

frame_2D_alg/agg_recursion.py
kwcckw: Okay right, it can be changed to a list here.

frame_2D_alg/agg_recursion.py
boris-kz: everything is rebuilt, we don't need old oF_, etc

frame_2D_alg/meta_code.py
```diff
@@ -21,10 +21,13 @@ def vt_(TT, wTT=wTT):  # base eval: multi-variate rel match, rel diff for member
     return m, d
 
 def sum_vt(N_, fr=0, fm=0, wTT=wTT, fdiv=1):  # basic weighted sum of CN|CF list
-
-    C = sum(n.c for n in N_); R = 0; TT = np.zeros((2,9))
+    fvT = isinstance(list(N_)[0] ,list)  # list to convert set from input E (get_exemplar)
```
boris-kz, line 24: this should be external.

frame_2D_alg/agg_recursion.py
boris-kz: ok, right

frame_2D_alg/agg_recursion.py
kwcckw: Okay, with rebuilt, we should remove the mapping:  

     for F in _oF_: _oF_[F] = _oF_.get(_oF_[F],_oF_[F]) 

Not sure why this is included previously.

frame_2D_alg/agg_recursion.py
```diff
@@ -300,7 +304,7 @@ def get_exemplars(N_,_r,_c):  # multi-layer non-maximum suppression -> sparse se
         else:
             break  # the rest of N_ is weaker, trace via rims
     else: E_ = set([N_[0]])  # no inhibition, any N can be seed
```
boris-kz, line 306: this else is a bug, should be 

    if E_: FV_(CoF.get(), *sum_vt(list(E_)))
    else:  E_ = set([N_[0]])  # no gain, no inhibition, any N can be seed
    return E_

frame_2D_alg/meta_code.py
boris-kz: Yes, merged oF will be replacing its components,  but why if not F.body ?

kwcckw: I think that was for the new oF where their body is empty? Since the body should be rebuilt again later.

frame_2D_alg/agg_recursion.py
kwcckw: Yes thanks. I think that was accidently removed.

frame_2D_alg/meta_code.py
boris-kz: they should all be rebuilt first, keeping old oFs as callees.
then we need to init oF_dict with both keys and vals = old oF_.
then splits and merges should replace their parent oFs in vals, then replace mapping keys in call sites?
it's simple with merges, the unresolved question is how to pick one of the splits as a call site replacement?

boris-kz: The "pick one" question is the tell that splits don't belong in the call-site remap at all. Look at what `split_oF_` actually produces (266–275): it scans `body` in source order and breaks on `comp_prim` mismatch, so `grp_` is an **ordered partition of one body into contiguous segments** — sequential stages, not alternative forks. `[a,b,c,d,e]` with a~b~c, d~e → `[a,b,c]` then `[d,e]`. The parent runs seg1 *then* seg2 *then* seg3.

So a caller that invoked `X` expected the whole sequence. Splitting X into X1;X2;X3 doesn't give the call site a *choice* of child — it would have to run all three in order. Merge and split are duals on opposite ends:

- **merge: N→1 identity collapse.** Several callee names become one. The `func.id = oF_[old]` rename is exactly right — callee identity changes, call *shape* is unchanged. Many keys → one val, rewrite each site's name.
- **split: 1→N expansion.** One call would become a threaded pipeline `s1=X1(a); s2=X2(s1); s3=X3(s2)`. That's not a rename — it's replacing one `ast.Call` node with a *block* of calls. The rename machinery structurally cannot express it.

So the rename-based remap handles merges and is *incapable* of splits, regardless of which child you'd pick. That's the real boundary, not the picking.

The resolution that dissolves the dilemma: **splits never touch call sites.** Keep the parent name `X` as a thin dispatcher whose new body is the pipeline over the freshly-registered sub-oFs — callers keep calling `X`, unchanged. Merges act on *identity* (visible to callers → rename at call sites); splits act on *composition* (internal → rewrite the parent's own body, parent name preserved, zero call-site edits). The symmetry you were reaching for doesn't hold because the two operate at different layers.

Which makes your init scheme merge-only: `oF_dict = {name: oF for old}` (identity), merges overwrite vals (several keys → the one merged oF), rewrite keys at call sites. Splits don't get a key — they don't remap an existing callee; they spawn new oFs and rewrite the *parent's body*. So splits are a separate pass, not entries in the call-site dict.

And the part that's genuinely unhandled isn't picking — it's the **segment interface.** `sub = CoF(body=grp)` (274) just holds an op slice; nothing derives the locals crossing each `grp` boundary. seg2 reads what seg1 defined (`L_`, etc.). To make segments independently callable (whether as a pipeline inside `X` or anywhere), `split_oF_` has to compute, per boundary, the live set: names defined-before-and-used-after. Those become the args/returns threading the pipeline. Until that liveness analysis exists, the children aren't well-formed functions and there's no pipeline body to install in `X`. That's the open problem the "which child" framing was standing in for.

kwcckw: > then splits and merges should replace their parent oFs in vals, then replace mapping keys in call sites?

Yes, this is similar with my current commit, where the split and merge is replacing the mapping in the dict.

> it's simple with merges, the unresolved question is how to pick one of the splits as a call site replacement?
  
I think multiple split oFs may still need to point into the same old oF, but the difference is just their body? And the fdef will be different since they will have part of the old oF body only.

boris-kz: splits should point to new forking gates, that replace their parent
also we can update all call sites after each split / call, not in batches?

boris-kz: Yes — gate-in-slot is the right move, and it collapses claim 2 too. Trace it:

**Split = factor the parent into a gate-skeleton + branch oFs.** The "forking gate" is just the existing `gv_` pattern (92, 357, 365…) promoted to the cut boundary. The parent keeps its gates and forks; each guarded branch lifts into a child oF the gate calls. Crucially, "replace their parent" = the gate inherits the parent's **name / nF / prototype slot**. So callers still name the parent, recompile swaps its body to the gate, and resolution lands on the gate with **zero call-site rewrite**. Splits contribute no entries to the remap dict — only merges do. That confirms the asymmetry: merges collapse identity (N names → 1, callers must be rewritten), splits preserve it (gate in-slot, callers untouched, children are new internal callees).

This forces one correction to `split_oF_`: the cut should be at the **gate/fork sites**, not `comp_prim` adjacency (266–269). The current criterion groups adjacent similar ops — a horizontal slice — but a forking gate selects branches. For children to *be* the gated branches, you partition the body at its `gv_` nodes: each branch → child, gates retained in the parent skeleton. The op-similarity grouping is misaligned with the gate semantics you're now specifying.

Still open, unchanged by this: the **branch interface**. `sub = CoF(body=grp)` (274) carries only an op slice. A child the gate *calls* needs the locals defined-before-the-gate-and-read-in-the-branch as params, and its products as returns. The gate computes the fork value `v` from those locals and threads required state in. `split_oF_` does no liveness analysis, so the children aren't yet callable. Plus the registry slots from before — new children need `iF_`/`nF_`/prototype entries or they're invisible as callees and have no buffering target.

**Incremental vs batch — under gate-in-slot the question dissolves.** In general, no: splits feed cluster (908, `cluster_oF_(sF_+rF_)`), so an old name's *terminal* identity isn't fixed until both passes finish — a split child can then be merged. And `cluster_oF_` isn't a sequence of discrete merges; it's one global convergent re-clustering (288–296) whose merge values read `caller_` via `comp_callers` (246, 292) — i.e., the call-site topology. Rewrite call sites mid-pipeline and you (a) feed the clustering an inconsistent, half-rewritten caller graph, and (b) have to re-rewrite anything cluster then dissolves. The collapse line `_oF_[F] = _oF_.get(_oF_[F],_oF_[F])` exists precisely to chase each name split→merge to its terminal form and rewrite *once*.

But with splits in-slot (above), splits emit no rewrites, and all merges come out of the single `cluster_oF_` convergence together. So the only rewrites — merge renames — naturally land in one batch at cluster's end. "After each split/call" and "batched" converge to the same single point. You don't gain anything by incrementalizing; you'd only reintroduce the moving-`caller_` hazard. The real simplification you're after is already bought by gate-in-slot: the remap dict becomes merge-only, the split→merge collapse chain disappears, and there's exactly one rewrite + one `inject_oF_` per ffeedback.

kwcckw: > splits should point to new forking gates, that replace their parent

But the split is not limited to the gv_ gate only right?

So theoretically each split creates new gv_ gate?

PR #189, opened 2026-06-28
6/28/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -259,9 +259,18 @@ def comp_prim(_n,n):
 def get_fc(n):
     return n.fc if isinstance(n,CoF) else costs.get(n[0],0)+sum(get_fc(c) for c in n[1]) if isinstance(n,tuple) else costs.get(type(n),0)
 
-# not updated:
-def split_oF_():  # divisive clustering
-    sF_, rF_ = [], []
+
+def remap(_oF_, old, new):  # remap by updating with call_sites
```
kwcckw: This remap is not fully correct yet. The question is on the prior discussion, do we need to create new gv_ gate in each split? Or split is only done on each gv_ gate?

If we need to split the body and create new ast node in oF.fdef, that will be much more complicated.

boris-kz: Of course it must be a new gate, else the is no point in splitting

kwcckw: Okay, but it will be something like a virtual gate where new gv will be added to parent's gv_, and a sub oF will be created on the split body as the gated section?

Then sub oF 's fdef will be remained the same, only the body is different? The new split gated section in the parent body should be replaced by the sub oF now.

frame_2D_alg/agg_recursion.py
```diff
@@ -853,7 +853,7 @@ def fill_frame(_iy,_ix, elev, T):  # expand level_frame from pixel-level seed ti
         T_ = []
         while T and gv_(sum(vt_(T.dTT, T.wTT*ttFrm*T.c)) * wFrm - (ave+avd)*(T.r+(cFrm*T.c))):
             frame[y,x]=T; T_+=[T]; dy_dx = T.box[2:] -T.box[:2]
-            pTT, pc = proj_N(T, np.hypot(*dy_dx), dy_dx, elev, T.c)
+            pTT, pc = proj_N(T, np.hypot(*dy_dx), dy_dx, elev or 1, T.c)
```
boris-kz: elev+1

frame_2D_alg/meta_code.py
boris-kz: there is nothing virtual about it, the splits are functionally different, so they must be called in different cases. not sure how exactly to set the conditions

frame_2D_alg/agg_recursion.py
```diff
@@ -55,7 +55,7 @@
 ttFrm,tAgg,ttTrc, ttN_,ttC_,ttN,ttF, ttE,ttcN,ttcC,ttcP,ttsG, ttBac,ttPrj, ttVct = [F.wTT for F in oF_]
```
boris-kz: sum(vt_()) is always wrong for relative vals, they add up to 1, we need to use

V = (sum(tt[0]) + sum(abs(tt[1]))) * c/r

frame_2D_alg/meta_code.py
kwcckw: Yes, initially i'm thinking it's virtual because we can't really manually set each condition after the split. So if they must be called under different cases, i think we should formalize or categorize the current condition of existing `gv_` first? Then we can systematically group them and use the similar condition following the grouped conditions on existing `gv_`?

frame_2D_alg/agg_recursion.py
kwcckw: For example in frame_H, fill_frame:

    sum(vt_(T.dTT, T.wTT*ttFrm*T.c)) * wFrm - (ave+avd)*(T.r+(cFrm*T.c))
    
If we use sum, that `@ wTT[0]` is no longer relevant and we no longer need the weight from wTT and ttFrm?

boris-kz: @ wTT is correct, but sum must be extensive, so it's sumV instead of sum(vt_()):

sumV(tt,c,r): return (sum(tt[0]) + sum(abs(tt[1]))) * c/r

frame_2D_alg/meta_code.py
boris-kz: generally, the functions that need splitting already have internally gated forks, we just need to make them external

frame_2D_alg/agg_recursion.py
kwcckw: But if we use @ wTT, we are getting single m or d:

    m = (m_/t_) @ wTT[0]; d = (ad_/t_) @ wTT[1]
    
And if we replace with sumV above, how do we really using @wTT again?

boris-kz: tt should be weighted: * wTT

frame_2D_alg/meta_code.py
kwcckw: > generally, the functions that need splitting already have internally gated forks, we just need to make them external

Then i guess we should split only on the gated area, where it's already guarded by `gv_`? Else we must create new conditions on each new added gv.

frame_2D_alg/agg_recursion.py
kwcckw: Okay, then it should be without @ since that will sum all values. We should be parsing `T.dTT * T.wTT * ttFrm` to sumV instead.

boris-kz: for example in cross_comp:
`if gv_(sumV(TT*ttE,c,r)*(wE*L) - (ave+avd)* (r+cE*L)* c): `
`if gv_((sumV(TT*tAgg,c,r)* (wAgg*L) - ave* (r+cAgg*L))* c):`

def sumV(tt,c,r, fd=0): 

    return sum(tt[0]) + fd or sum(abs(tt[1])) * c/r

frame_2D_alg/agg_recursion.py
```diff
@@ -904,16 +904,9 @@ def ffeedback(frame, aTT,oTT, aL,oL):  # recompute filters from regime drift; fo
         if oL := pack_seg(frame,'oH', wBac, cBac**2, oTT):
             dTT += oL.dTT- oTT; oTT=oL.dTT; dc+=oL.c-_oc; dr+=oL.r-_or
             oF_site_ = {}  # map oFs to their call sites
```
boris-kz: so we need to set initial mapping here: `(key= [call_sites(F,_F) for _F in oF_],  val = F) `?

frame_2D_alg/meta_code.py
boris-kz: The splitting is by callers, not by any similarity in the body?

...
Answer on cheapest replacement: the gate's whole value is that it keeps the parent's name and signature. Callers still call the parent name; it now resolves to the sequencer. So zero callers change → there are no sites to record. oF_site_ stores a mapping nobody reads on the split path.
Cheapest replacement is therefore: hollow the parent's CoF in place (body → the sequence of sub-calls, fdef → gate fdef) and append the subs. Parent keeps its nF slot, so inject_oF_ works on it unchanged; only the subs need the synthesized-fdef branch. That's a list splice plus name binds — no map.
The map only earns its keep on the merge path, where constituent names vanish from callers and you must repoint their Call nodes. Even there, don't precompute it — derive sites on demand with call_sites(caller_fdef) at merge time.
The actual blocker the map distracts from: a sequencer over extracted segments only preserves behavior if you thread live locals across the cuts as params/returns. That's a liveness analysis, and it's the real cost of split. The bind/splice bookkeeping is noise next to it. Design that.

kwcckw: >  only the subs need the synthesized-fdef branch. That's a list splice plus name binds — no map.

This is suggesting creating new fdef in the sub, which is still complicated since that involves creating new ast node?

frame_2D_alg/agg_recursion.py
kwcckw: I see, thanks.

frame_2D_alg/agg_recursion.py
kwcckw: The call_sites returns the callee oFs of oF,  so if we follow your method above, the key will be a group of multiple oFs callee, mapping to the oF, rather than the single oF mapping?

frame_2D_alg/meta_code.py
boris-kz: it's the same function name, with replaced body

frame_2D_alg/agg_recursion.py
boris-kz: `if gv_((sumV(TT*tAgg,c,r, fd=0)* (wAgg*L) - ave* (r+cAgg*L))* c):`

frame_2D_alg/agg_recursion.py
boris-kz: not sure, key=oF.nF?

kwcckw: But nF may not work if we merge oFs? Since it may get multiple nFs if we merge oFs?

frame_2D_alg/meta_code.py
kwcckw: Okay, so the intended workflow is creating new sub with same fdef (same ast node reference and name), but with different body only?

frame_2D_alg/agg_recursion.py
boris-kz: why multiple?

frame_2D_alg/meta_code.py
boris-kz: yes, the body is the gate

kwcckw: But it's not a real gate with ifExp node + conditions? It will be a list of nodes?

frame_2D_alg/agg_recursion.py
kwcckw: The oF.nF is the index of oF in oF_, when we merge oFs, their nF should consists of multiple nFs now? Since there will be no single nF able to represent the merged oF's nFs now?

frame_2D_alg/meta_code.py
boris-kz: same gate just wrapped in oF

frame_2D_alg/agg_recursion.py
boris-kz: single new oF must replace old oFs, that's what the dict is for

boris-kz: Incoming sites, keyed by callee — because the use is repointing a merged component in *other* oFs' calls, so you look the component up and get the calls to it:

```
oF_site_ = {oF: [] for oF in oF_}
for fd in nF_:                                   # every caller body
    for n in call_sites(fd):                     # Calls in fd, func.id in iF_
        oF_site_[oF_[iF_[n.func.id]]] += [(fd, n)]
```

Each old oF → list of `(containing_fd, Call node)` that invoke it. Store the pair, not the bare node: you mutate `n.func.id` to the composite's name to redirect, and you need `fd` to recompile the caller after.

`call_sites(fd)` alone is the wrong direction — it's outgoing (calls *inside* fd). Inverting it across all `nF_`, as above, gives incoming, which is what redirection consumes.

Discriminant on keys: key by oF if the map travels with the reform results; key by name (`{name: []}`, `oF_site_[n.func.id]`) if the only consumer is the repoint pass — then redirection is just `n.func.id = composite_name` over `oF_site_[component_name]`, no oF lookup at all. Name-keyed is the leaner one given the stated use.

frame_2D_alg/meta_code.py
kwcckw: Okay, so it's just packing multiple nodes in the body into an oF. Since it's a gated sub, we need to add new oF.typ into this oF? Something like oF.typ = gate?

frame_2D_alg/agg_recursion.py
kwcckw: Okay, by using this scheme, if we create new oF with new nF, we need to add new oF_, nF_ right after each split or merge? Because they will be used again in the subsequent merge or split.

boris-kz: Everything should be replaced, I told you that about ten times already

frame_2D_alg/meta_code.py
boris-kz: yes, and set conditions for calling one of them, initially the same as they were internally

boris-kz: so I am not sure which of your edits here are relevant, I won't be pushing them

frame_2D_alg/agg_recursion.py
boris-kz: so please use that to replace all cases of sum(vt_())

frame_2D_alg/meta_code.py
kwcckw: > yes, and set conditions for calling one of them, initially the same as they were internally

Okay, then we probably need to pack the conditions of initial gv_ too? So that we can reuse it in a similar way? 

> so I am not sure which of your edits here are relevant, I won't be pushing them

Yes, it will be different now. From your posted LLM:

    key by name ({name: []}, oF_site_[n.func.id]) 

    call_sites(fd) alone is the wrong direction

Those have different structure, and I need to check and try them first.

PR #190, opened 2026-06-29
6/29/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -259,24 +259,31 @@ def comp_prim(_n,n):
...
-    sF_, rF_ = [], []
-    for oF in oF_:
-        if (len(oF.body)-1) * wL > ave:  # * split w,c
-            _n= oF.body[0]; grp=[_n]; grp_=[]
-            for n in oF.body[1:]:
-                if comp_prim(_n,n): grp+=[n]
-                else: grp_+= [grp]; grp =[n]
-                _n=n
+# very initial draft
+def split_oF_(oF_site_):  # divisive clustering
```
kwcckw, line 263: > yes, and set conditions for calling one of them, initially the same as they were internally

Actually if we split based on oF_sites_, we only split on each gate now?

Each sub's body should be the body or code after each gate, and the body or code before any gate should be the leftover, which we still need to remain it in the original oF?

For the conditions, since each split is at the gate, so there's no changes needed on the conditions?

frame_2D_alg/meta_code.py
```diff
@@ -259,24 +259,31 @@ def comp_prim(_n,n):
...
+
+    def get_body(grp):
```
boris-kz, line 265: how is that different from build?

frame_2D_alg/meta_code.py
boris-kz: it doesn't have to be current gv_, just any fork with different callers

frame_2D_alg/meta_code.py
kwcckw: This is actually building the body from grp, which it could be part of the whole body, while build creates all nodes of the function.

If we split a function fork within the gv_, the new body should be the nodes from prior forks to the function node? Or the new body should start from the function node until the end or the next split?

frame_2D_alg/meta_code.py
kwcckw: But why it needs different callers? There's some fork with single callers only, for example, frame_H only has frame_H as callers:

               if gv_(PV__[y,x] - ave):
                    T = frame_H(image, iy,ix, Ly,Lx, Y,X, rV, max_elev=elev)[0]  # expand new tile to current level, no fb
                    
We still can split it?

frame_2D_alg/agg_recursion.py
boris-kz, line 116: C = min(_N.c,N.c); R = (_N.r+N.r)/2; M = (_N.m+N.m)/2
        if M*(C/R) * wPrj > ave * (cPrj+r+C+R):  # not oF, spec / link:

frame_2D_alg/meta_code.py
boris-kz: I think the only reason to split is to avoid calling unused segments;
def comp_prim(_n,n):

    if isinstance(_n,CoF) or isinstance(n,CoF):
        return comp_callers(_n,n) > ave if isinstance(_n,CoF) and isinstance(n,CoF) and n.nF==_n.nF else 0

but this is all tentative

frame_2D_alg/meta_code.py
boris-kz: The gating sequence should be conserved from the parent, we just need to replace execution forks with the new oFs, then run build over resulting sequence

frame_2D_alg/agg_recursion.py
kwcckw: But wouldn't be recompute M from TT is the better way instead of getting average of `_N` and `N`'s m?

boris-kz: it's actually both?
`if (sumV((eTT+iTT)*ttPrj,C,R) + M*(C/R)) * wPrj > ave * (cPrj+r+C+R):`

kwcckw: I think there will be redundancy here since N.m is computed from iTT?

frame_2D_alg/meta_code.py
kwcckw: > we just need to replace execution forks with the new oFs, then run build over resulting sequence

The problem is the `build` creates nodes from the ast function, which it will include all nodes of the function.

If we need to replace only the execution forks, we just need to overwrite the execution fork by replacing the oF in the body and inject the new ast node?

frame_2D_alg/meta_code.py
kwcckw: Could you explain further on how a segment can be unused?

It's only possible only if the gated section is not executed?

boris-kz: This is the case of gate selecting one of several forks, unused is the forks not taken

frame_2D_alg/meta_code.py
boris-kz: no, AST doesn't unpack functions

frame_2D_alg/agg_recursion.py
boris-kz: right, so no  M*(C/R)

frame_2D_alg/meta_code.py
kwcckw: Yes, so that's where we need inject_oF_ to rebuild the replaced oF's node? 

            if (r := build(oF,node)): t,fc = r; oF.body += [t]; oF.fc += fc

frame_2D_alg/meta_code.py
kwcckw: > This is the case of gate selecting one of several forks, unused is the forks not taken

Okay, and to identify the unused fork, we only can identify it by oF.gv_? Which means we should include gv's value in evaluating the splitting process.

boris-kz: unused by specific callers, that means assigning callers to each fork in advace, then using comp_callers(_n,n)

frame_2D_alg/meta_code.py
boris-kz: of course every new oF must run build, why is that even a quesion?
But AST node doesn't change, it's the same old oF with converted body

frame_2D_alg/agg_recursion.py
kwcckw: Also for sumV:

      def sumV(tt,c,r, fd=0):
          return sum(tt[0]) + fd or sum(abs(tt[1])) * c/r
          
Why we need the fd and the summing to m?

Also the return should be  `(sum(tt[0]) + fd or sum(abs(tt[1]))) * c/r`, else the `c/r` applies to the d only.

frame_2D_alg/meta_code.py
kwcckw: > unused by specific callers, that means assigning callers to each fork in advace, then using comp_callers(_n,n)

I see, that's actually in oF_site_ now, where it 's already packing each pair of the caller function def （`caller_fd`) and the site nodes (`oF_cs`) in advance. So instead of `comp_prim` , i guess we just need to have `comp_callers`.

frame_2D_alg/agg_recursion.py
boris-kz: + fd is when we need m+d. yes on the brackets

kwcckw: I see, then the existing fd is wrong because it reads `sum(tt[0]) + fd` first, and then `or sum(abs(tt[1]))`, which will never sum the d values.

We need `and` instead `or` here: 

    return (sum(tt[0]) + (fd and sum(abs(tt[1])))) * c/r
    
So d values will be summed only if fd is 1.

boris-kz: (fd and sum(abs(tt[1]))) is boolean, the sum is trashed

return (sum(tt[0]) + (sum(abs(tt[1])) if fd else 0)) * c/r

frame_2D_alg/meta_code.py
boris-kz: but that's only for oFs, now we need to add callers to raw forks, not sure how

frame_2D_alg/meta_code.py
```diff
@@ -259,24 +259,31 @@ def comp_prim(_n,n):
...
+        pass
+
+    for oF, sites in oF_site_.items():
+        
+        if (len(sites)-1) * wL > ave:  # at least 2 oF functions within the body
```
boris-kz, line 270: we are splitting code segments, they don't have to be oFs

frame_2D_alg/meta_code.py
kwcckw: > but that's only for oFs, now we need to add callers to raw forks, not sure how

The caller should be remained the same after the split or merge? We just need to update the mapping with the same caller, but point to the new forks?

frame_2D_alg/meta_code.py
kwcckw: Actually if we splitting code segments, the `call_sites` is wrong? That's only retrieve the oF's functions, while we need all nodes?

      def call_sites(fd):  # FunctionDef
          return [n for n in ast.walk(fd) if isinstance(n,ast.Call) and isinstance(n.func,ast.Name) and n.func.id in iF_]
          
That ` and n.func.id in iF_` is actually skipping all other nodes except the oFs.

boris-kz: yes, all that needs to change

frame_2D_alg/meta_code.py
boris-kz: the caller doesn't need to know the new oFs, they will get new callers after ffeedback

PR #191, opened 2026-06-30
6/30/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -54,7 +54,7 @@
 wFrm,wAgg,wTrc, wN_,wC_,wN,wF, wE,wcN,wcC,wcP,wsG, wBac,wPrj, wVct ) = [F.fc for F in oF_]  # ave gain/call, init = cost
 ttFrm,ttA,ttTrc,ttN_,ttC_,ttN,ttF,ttE,ttcN,ttcC,ttcP, ttsG,ttBac,ttPrj,ttVct = [F.wTT for F in oF_]
 
-def sumV(tt,c,r, fd=0):
+def sumV(tt,c,r, fd=1):
```
kwcckw: I set fd = 1 as default here,because most of the eval needs to sum both m and d? Only a few evals such as the eval for cross_comp needs the m value.

frame_2D_alg/agg_recursion.py
```diff
@@ -95,7 +95,7 @@ def cross_comp(root, rr, fC=0):  # core function mediating recursive rng+ and de
         oF_[CoF.get().nF].V_ += [cV]  # combined comp_ results
         if L_:  # +ve only (L_ may empty when there's just negative links)
             root.L_ = L_; L= len(L_)-1  # val=m+d /clust, m/comp
-            if gv_(sumV(TT*ttcN,c,r)*(wcN*L) - (ave+avd)* (r+ccN*L)* c):  # if +ve, store neg gate values
+            if gv_(sumV(TT*ttcN,c,r,0)*(wcN*L) - (ave+avd)* (r+ccN*L)* c):  # if +ve, store neg gate values
```
kwcckw: Sorry, this is a mistake, the fd = 0 should be for the eval of cross_comp below:

    if gv_((sumV(TT*ttA,c,r,0)*(wAgg*L) - ave* (r+cAgg*L))* c):

frame_2D_alg/agg_recursion.py
```diff
@@ -427,7 +427,7 @@ def cluster_C(Ft, E_,_r,_c):  # form centroids by clustering exemplar surround v
     if out_:
         for n in [N for C in out_ for N in C.N_]:  # exemplar V + sum n match_dev to Cs, m* ||C rvals:
             n.exe = (n.d if n.typ==1 else n.m) + np.sum(n.m_) - ave
-        if gv_(vt_(DTT,Ft.root.wTT*ttcC)[0]*wcC*(len(out_)-1) - ave*(r+ccC*(len(out_)-1))):
+        if gv_(sumV(DTT*Ft.root.wTT*ttcC,0)*wcC*(len(out_)-1) - ave*(r+ccC*(len(out_)-1))):
```
kwcckw: For the `vt_` here, actually it's just sumV with fd = 0?

boris-kz: right, all these gv_ s are extensive

frame_2D_alg/agg_recursion.py
boris-kz: that's mostly because fd=0 evals are not converted to sumV yet. that will be for all comps

kwcckw: For example, in comp_N_:

    if gv_(abs(m)*wN - ave*(r+cN)):
    
So actually this m should be computed via sumV now using pTT?

frame_2D_alg/meta_code.py
```diff
@@ -270,12 +269,14 @@ def split_oF_():  # divisive clustering
...
-        else: rF_ += [oF]
-    return sF_, rF_
+            if len(grp_)>1:
+                sites = oF_site_.pop(oF)  # remove the split oF?
+                for grp in grp_:  # single refinement
+                    caller_ = set([caller for p in grp if isinstance(p,CoF) for caller in p.caller_])
+                    fc = sum([get_fc(prim) for prim in grp])
+                    sub = CoF(root=oF,fc=fc,body=grp); sub.caller_ = caller_
+                    sub_caller_fd = {nF_[caller.nF] for caller in caller_} 
+                    oF_site_[sub] = [(caller_fd, oF_cs) for (caller_fd, oF_cs) in sites if caller_fd in sub_caller_fd]   # pack only sites with the current sub's caller
```
kwcckw: > the caller doesn't need to know the new oFs, they will get new callers after ffeedback

But they can be assigned right after the sub is formed? Because the split sub should be still have the same caller.

frame_2D_alg/agg_recursion.py

frame_2D_alg/meta_code.py
boris-kz: no, those callers are stale after feedback

kwcckw: Okay right, the caller might be split or merge, but we still can map the existing caller to new caller if there's any changes on those callers after the feedback?

boris-kz: I don't think it matters, recent callers should have a priority, so each feedback cycle can start from the scratch

kwcckw: Yes, everything will be rebuilt anyway, and the split and merge actually will be just changing the mapping for each block or line of code?

Actually the whole split and merge mapping process is still very fuzzy to me, could you clarify further on the mapping process?

Each split should map the split block of code to the new raw fork (new oF), and the caller is unidentified until the rebuild process?

Merging should work the same by mapping multiple oFs to new fork, and each new fork should have the same caller, else they won't be merged?

boris-kz: split simply nests fork oFs in parrent oF, we already discussed that.
merged oF removes fork oFs altogether and replaces them in all caller bodies.

boris-kz: Claude on this PR:

Now thread B — meta_code forks/callers/feedback:Reviewed both threads against source. Responses below, each closed to the open contract change.

## Thread A — `sumV` fd default, `vt_` vs `sumV` in gates

Resolution is correct as stated, but one conversion is still owed and it's the one under discussion.

`sumV(tt,c,r,fd=1)` (line 57) returns `sum(m_) + (sum|d_| if fd)` scaled by `c/r` — extensive. `vt_` (meta_code 17) returns the intensive form: each attr normalized by `m_+|d_|`, then `@wTT`. These are not interchangeable; they differ by the per-attr max-comparand normalization, not just by including `d`.

- fd discriminant: fd=0 = predictive/comp gate (m only), fd=1 = clustering/membership gate (m+d). Default fd=1 is right — clustering gates dominate.
- kwcckw's "all these gv_ s are extensive" is the operative point: every `gv_` compares value against an extensive threshold `ave*(r+c…)`, so the value side must also be extensive. `comp_N_` line 133 `gv_(abs(m)*wN - ave*(r+cN))` feeds it `m` from `vt_` (intensive) → type mismatch with its own threshold. The model to copy is line 116, `gv_`-free but same shape: `sumV((eTT+iTT)*ttPrj,C,R,0)`.

So line 133's gate should be `sumV(pTT*ttN_, c, r, 0)`, not `abs(m)`.

One thing the thread didn't trace: `vt_(pTT,ttN_)` at line 131 is still needed — the negative branch (line 138) stores `m=m,d=d` on the neg `CL`. So this is *add an extensive gate*, not *replace vt_ with sumV*. vt_ stays for the link's stored m/d; sumV supplies the gate value. Same split will recur at every fd=0 comp gate still on vt_ (this is the "not converted yet" set).

## Thread B — split vs merge, caller maintenance

boris's two-line resolution is the right model and it resolves kwcckw's confusion, but the code doesn't implement it yet, and the interface has drifted. Tracing it out:

**The asymmetry answers the "same caller else won't merge" question.** `comp_callers` (247) is the merge *criterion* — shared callers predict mergeability. It is not a post-merge invariant. Post-merge the caller bodies are physically rewritten; that's "replaces them in all caller bodies." Split touches no caller body (nesting is internal, fork selection via the `E` gate oF), so callers keep calling the parent unchanged — hence split needs no remapping.

**"start from scratch" is consistent because there are two caller stores, one transient.** `caller_` (the CoF set, built at trace time, meta_code 158) is dynamic and rebuilt on every re-trace after `inject_oF_`. Mapping old→new `caller_` across a cycle is pointless — it's regenerated. That's the whole content of boris's last three replies, and it's correct given the trace wrapper.

**The open contract change — and a drift I'd flag now:** the mechanism boris means by "replace in all caller bodies" is the *static* site map, not `caller_`. `ffeedback` (894) already builds it: `oF_site_ = {oF: [(fdef, call_node)]}` (908–910), then passes it to `split_oF_(oF_site_)` and `clust_oF_(oF_site_)`. But:

1. meta_code `split_oF_()` takes no args and `cluster_oF_(oF_)` takes the flat list — neither consumes `oF_site_`, and neither rewrites a `Call` node. The merge-side call-site substitution (the actual "replace in caller bodies") is unimplemented.
2. Import/name mismatch: agg_recursion imports `clust_oF_`, meta_code defines `cluster_oF_`.
3. `oF_ = list(dict.fromkeys(oF_site_.values()))` (913) collects the site-lists, not oFs — will not yield the new oF set.

So to close boris's model: `cluster_oF_` (merge path) must take `oF_site_`, and for each merged group rewrite every mapped `Call.func.id` to the merged oF's name so `inject_oF_` recompiles the callers against it; `split_oF_` leaves sites untouched and only nests (root=oF already correct at 274) plus inserts the selection gate. `caller_` needs no cross-cycle mapping — rebuilt on re-trace. That's every contract still open in the thread.

boris-kz: so the criterion for split should be comp_callers only, there are no redundant sequences in parent oF,
but for merge it's the complete comp_prim, so it may still have different callers?

kwcckw: > parent unchanged — hence split needs no remapping.

Okay, so we not really need mapping in split_oF since they still have the same caller if we split within the function.

> split_oF_ leaves sites untouched and only nests (root=oF already correct at 274) plus inserts the selection gate. 

Actually i don't quite get this selection gate. This selection gate should be inherited from parent, and inject into the code so that the code will run this gate to eval the new fork in the next run?

kwcckw: > but for merge it's the complete comp_prim, so it may still have different callers?

Yes, because each oF might be called from different callers. We only need comp_callers and comp_body, as in the existing code?

boris-kz: I am not really sure. the way I explained it split simply wraps the sequence in new oF, and that's to keep the meta-data, but we need that meta-data to trigger the split in the first place?

boris-kz: Split criterion shoeld be tracing granularity, not comp_callers: (len(body)-1)*wL > ave,

accumulate dTT per op-run / body position instead of per-oF. Then candidate segments carry value before wrapping, and the size + comp_prim bootstrap can be replaced by a value gate. As written, split is necessarily the structural bootstrap and value refinement is deferred to the next cycle's merge.

kwcckw: > am not really sure. the way I explained it split simply wraps the sequence in new oF, and that's to keep the meta-data, but we need that meta-data to trigger the split in the first place?

Yes, but collecting dTT per op-run might not be realistic since we may not have dTT in each line of code. So i think the dTT should be cumulative from the first line of the function so that we can have the gate value in any block of code?

frame_2D_alg/meta_code.py
```diff
@@ -259,9 +259,8 @@ def comp_prim(_n,n):
 def get_fc(n):
     return n.fc if isinstance(n,CoF) else costs.get(n[0],0)+sum(get_fc(c) for c in n[1]) if isinstance(n,tuple) else costs.get(type(n),0)
 
-# not updated:
-def split_oF_():  # divisive clustering
-    sF_, rF_ = [], []
+# not fully edited
+def split_oF_(oF_site_):  # divisive clustering
```
boris-kz: oF_site_ should be per oF, we can't pass it here

frame_2D_alg/meta_code.py
boris-kz: it can be parent oF.w * len(gated_segment_AST): approximate gain from encapsulating the the segment

frame_2D_alg/meta_code.py
kwcckw: This oF_site_ is actually per oF_, because it packs the caller and the sites for each oF in oF_. If it's per oF, it's just packing the same thing by unpacking oF_?

frame_2D_alg/meta_code.py
kwcckw: Okay, but it should be the ratio to the total?

     oF.w * len(gated_segment_AST) /  total AST length ?

frame_2D_alg/meta_code.py
boris-kz: we need oF call sites in all other oFs
sorry I didn't have much time today, will check tomorrow

frame_2D_alg/meta_code.py
boris-kz: total is not relevant, we are encapsulating the segment

frame_2D_alg/meta_code.py
kwcckw: Okay, at this point, i think oF_ sites should be relevant for merge only? Since split's caller will be remained the same, and we need oF_site to remap the oF after the merge, as in the suggestion above:

      cluster_oF_ (merge path) must take oF_site_, and for each merged group rewrite every mapped Call.func.id to the merged oF's name so inject_oF_ recompiles the callers against it;

PR #192, opened 2026-07-01
7/1/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -54,8 +54,9 @@
 wFrm,wAgg,wTrc, wN_,wC_,wN,wF, wE,wcN,wcC,wcP,wsG, wBac,wPrj, wVct ) = [F.fc for F in oF_]  # ave gain/call, init = cost
 ttFrm,ttA,ttTrc,ttN_,ttC_,ttN,ttF,ttE,ttcN,ttcC,ttcP, ttsG,ttBac,ttPrj,ttVct = [F.wTT for F in oF_]
 
-def sumV(tt,c,r, fd=1):
-    return (sum(tt[0]) + (sum(abs(tt[1])) if fd else 0)) * c/r
+def sumV(tt,c,r, fd=0, fsep=0):
```
kwcckw, line 57: I added fsep here because for comp, we need both m and d instead of a single summed value.

frame_2D_alg/meta_code.py
```diff
@@ -261,14 +261,17 @@ def get_fc(n):
...
     sF_, rF_ = [], []
     for oF in oF_:
         # or if oF.w * len(gated_segment_AST): approximate gain from encapsulating the segment? same for merge but with component oFs?
         if (len(oF.body)-1) * wL > ave:  # * split w,c;  need to add callers to raw code forks too?
-            _n= oF.body[0]; grp=[_n]; grp_=[]
+            _n= oF.body[0]; grp=[_n]; grp_=[]; len_seg = get_segl(_n);
             for n in oF.body[1:]:
-                if comp_prim(_n,n): grp+=[n]
-                else: grp_+= [grp]; grp =[n]
+                if comp_prim(_n,n) and oF.w*(len_seg + get_segl(n))<ave: grp+=[n]; len_seg+=get_segl(n)  # eval segment length when packing new body?
```
kwcckw, line 273: I think the eval  of `oF.w*len_seg` should be used before we packing new node here?

frame_2D_alg/meta_code.py
```diff
@@ -306,6 +309,9 @@ def clust_oF_(oF_):  # cluster Ts if called together, global only
                 nT = sum2O(t_)
                 nT.memb = gV; fC = sum(t.fc for t in t_); nT.cmpr = fC - fC/len(t_)
                 smF_ += [nT]
+                for fork in t_:
+                    for (caller_fd, site) in fork.sites:
+                        site.func.id = nF_[nT.nF].name  # with this method, we need to set nT's nF_ first
```
kwcckw, line 314: This edit is from the prior suggestion:

    cluster_oF_ (merge path) must take oF_site_, and for each merged group rewrite every mapped Call.func.id to the merged oF's name so inject_oF_ recompiles the callers against it; 

But looks like we need the new nT's nF first before we can update their func.id and we won't have this until the rebuild process later.

frame_2D_alg/meta_code.py
boris-kz: This whole sequence seems to be wrong, consecutive prims should never match? the callers may match, but they are only known for oFs.
So the new scheme we were talking about is segmenting by ifExp, then converting the segment to oF if
oF.w * cost(segment) > ave?

kwcckw: > , consecutive prims should never match? the callers may match, but they are only known for oFs.

Yes, but their caller should be the same too if they are from the same oF, so this comparison method is actually flawed.

> So the new scheme we were talking about is segmenting by ifExp, then converting the segment to oF if
oF.w * cost(segment) > ave?

This is actually segmenting each gv_ gate, and there will be leftover? 

Take cross_comp as example:

          N_,G_ = root.N_,[]  # root is Ft, converted below, rc=rdn+olp, comp N_| B_| C_:
          if Lt := comp_C_(N_,rr,fC=1) if fC else comp_N_(combinations(N_,2),rr):
              L_,TT,c,r, cV = Lt
              if L_:  # +ve only (L_ may empty when there's just negative links)
                  if gv_(sumV(TT*ttcN,c,r,1)*(wcN*L) - (ave+avd)* (r+ccN*L)* c):  # if +ve, store neg gate values
                      G_,r = cluster_N(root, E_,r,c)  # cluster_C, _P, eval?
                      if G_:
                          if gv_((sumV(TT*ttA,c,r)*(wAgg*L) - ave* (r+cAgg*L))* c):  # extensive
                              G_ = cross_comp(root.Nt,r)  # agg+
          return G_
          
          
 We might split it into 3 sub oFs here? 
 
 The first sub:
 
                N_,G_ = root.N_,[]  # root is Ft, converted below, rc=rdn+olp, comp N_| B_| C_:
                if Lt := comp_C_(N_,rr,fC=1) if fC else comp_N_(combinations(N_,2),rr):
                    L_,TT,c,r, cV = Lt
                    if L_:  # +ve only (L_ may empty when there's just negative links)
                        
 The second sub:
 
         if gv_(sumV(TT*ttcN,c,r,1)*(wcN*L) - (ave+avd)* (r+ccN*L)* c):  # if +ve, store neg gate values
                G_,r = cluster_N(root, E_,r,c)  # cluster_C, _P, eval?
                if G_:
                    root.Nt = sum2F(G_,root,froot=2); L=len(G_)-1  #| C_?****
                    
 The last sub:
 
         if gv_((sumV(TT*ttA,c,r)*(wAgg*L) - ave* (r+cAgg*L))* c):  # extensive
            G_ = cross_comp(root.Nt,r)  # agg+

frame_2D_alg/agg_recursion.py
```diff
@@ -128,7 +129,7 @@ def proj_V(_N,N, dist, dy_dx, dec):  # _N x N induction
     for pL in sorted(pairs, key=lambda x: x[0]):  # proximity prior, test compared?
         dist, dy_dx, _N,N, c = pL  # rim angl is not canonic
         pTT = proj_V(_N,N, dist, dy_dx, root.m if root!=2 else decay** (dist/((_N.span+N.span)/2)))  # based on current rim
-        lr = r+ (N.r+_N.r)/2; m,d = vt_(pTT,ttN_)  # +|-match certainty
+        lr = r+ (N.r+_N.r)/2; m,d = sumV(pTT*ttN_, c, lr,fsep=1)
```
boris-kz, line 132: I think m,d should still be from vt_, normalized because they cancel each other in projection
sumV may be just for clustering, because it could be for both forks

frame_2D_alg/agg_recursion.py
boris-kz: if both they should be normalized, so it's still vt_

frame_2D_alg/meta_code.py
boris-kz: yes, probably split by gv_ first, and new gv_ may be formed from plain ifExp, as intermediate encapsulation

frame_2D_alg/meta_code.py
boris-kz: just assign incremental numerical nF?

kwcckw: So do you mean the site should be replaced by incremental numerical nF rather than the ast node?

           for fdef in nF_:  # map oFs to call sites:
                for n in call_sites(fdef):  # calls in fdef, func.id in iF_
                    oF_[iF_[n.func.id]].sites += [(fdef, n)]     
                    
 In that case the `n` in sites above (extracted from my ffeedback) should be replaced by the numerical nF instead.

frame_2D_alg/meta_code.py
kwcckw: > and new gv_ may be formed from plain ifExp, as intermediate encapsulation

Okay, so this is actually split gv_ and any of the ifExp forks. Also this gv_ encapsulation should be run in the next run after ffeedback? Which mean it should be added to code as an additional eval?

frame_2D_alg/agg_recursion.py
kwcckw: Alright, so we actually not really need sumV in comp.

frame_2D_alg/meta_code.py
boris-kz: it's split oF into gv_oFs, and encapsulate ifExp sequences into gv_s

frame_2D_alg/meta_code.py
boris-kz: fdef is a new oF, its nF is assigned in clust_oF_ as global last_nF, which will be incremented after the assignment

frame_2D_alg/agg_recursion.py
boris-kz: yes, 

        if vt_((eTT+iTT)*ttPrj)[0] *(C/R) *wPrj > ave *(cPrj+r+C+R):  # not oF, spec / link:

we only need sumV when m and d should be independent from each other

frame_2D_alg/meta_code.py
kwcckw: > it's split oF into gv_oFs and encapsulate ifExp sequences into gv_s

Okay, and after the split, those new sub oFs should be independent of old oF. So if there's leftover block of code in the old oF, the rebuilt old oF's gv_ should have reduced element now? So in this case, why we still need to pack split ifExp sequence into old oF.gv_?

frame_2D_alg/meta_code.py
kwcckw: Yes, this is the index of new oF in oF_, which is packed in oF.nF. But in oF's sites, we are packing the actual ast node n, which is not exist for the new oF yet?

frame_2D_alg/agg_recursion.py
kwcckw: Okay, so those fd = 0 should be reverted, and sumV always sum both m and d?

boris-kz: It's still good for clustering, it has independent Bt

frame_2D_alg/meta_code.py
boris-kz: it's gv_(), not oF.gV_. These are two different levels of encapsulation for conditional sequences

frame_2D_alg/meta_code.py
boris-kz: this is replacing new callee oF in old caller oF

frame_2D_alg/meta_code.py
kwcckw: Do you referring to this gv_?

      def gv_(v, i=None):
      
          if v > 0: return v
          else: oF_[CoF.get().nF].gV_[i] -= v  # double -ve -> +ve
          
This is still accessing the same oF.gv_, why it's a different level of encapsulation? Or do you referring to something else?

frame_2D_alg/meta_code.py
kwcckw: Yes, i guess we need change `call_sites` so that it uses the index of oF instead of the ast node since we don't have the ast node of new merged oF.

frame_2D_alg/agg_recursion.py
```diff
@@ -54,8 +54,9 @@
 wFrm,wAgg,wTrc, wN_,wC_,wN,wF, wE,wcN,wcC,wcP,wsG, wBac,wPrj, wVct ) = [F.fc for F in oF_]  # ave gain/call, init = cost
```
boris-kz, line 54: >review and respond to the discussion on attached PR, meta_code is not updated

Reconstructed the PR diff from the HTML and traced it against the mounted files. Four threads; my read on each, then the contract gaps.

**agg — `sumV` `fsep`/`fd` (0–7)**

`fsep` is redundant; drop it. Three eval contexts, two primitives already cover them:

- comp (`L.m/L.d`, `comp_N`, `comp_N_`, `comp_F`) → `vt_` (normalized). The diff's `sumV(...,fsep=1)` returns *unnormalized* separated m,d — not what comp wants. Your comment-1/2 line is the correct one: revert. `comp_N`: `m,d = sumV(TT*ttN_,c,r,fsep=1)` → `vt_(TT,ttN_)`; `comp_N_`: `sumV(pTT*ttN_,c,lr,fsep=1)` → `vt_(pTT,ttN_)`; `comp_F` same.
- clustering gates (`cluster_N`/`cluster_C` reform, `sum2G`, agg gate) → `sumV` fd=1 (both forks, magnitude). The diff's `,1` is right — keep.
- projection gate (`proj_V`) → per your comment 4, `vt_(...)[0]` (normalized m), not `sumV(fd=0)` magnitude.

Discriminant for killing `fsep` entirely: is there any site needing *unnormalized separated* m,d? None in the diff — comp wants normalized (`vt_`), clustering wants summed (`sumV` fd=1). If none, remove `fsep` and keep `fd` default = 1 (or make all `sumV` callers explicit).

**meta thread A — split by `gv_` (8–15)**

Your 11/13/15 closed the design. The PR's `split_oF_` does **not** implement it — it still groups `oF.body` by `comp_prim` over consecutive elements, i.e. exactly the "consecutive prims share a caller" case you rejected in 9. `get_segl`/`len_seg` only gates run length; it doesn't change the (wrong) grouping key.

Untraced: `split_oF_` should cut `oF.body` at the `('gv_',l)` markers `build()` emits — one sub-oF per top-level gate block, **nested where gates nest** (in `cross_comp` the agg+ gate sits inside the cluster gate → nested sub, not a flat 3-way as kwcckw sketched). Encapsulate a block iff `oF.w * sum(get_fc(seg)) > ave`. Remove the `comp_prim` loop.

kwcckw 14 (does old `oF.gv_` shrink): yes — extraction moves that gate's `g_`/`gv_` entries into the sub-oF and leaves a call in the old body; the ifExp→`gv_` step *adds* new gates elsewhere. The lists diverge — consistent with your 15.

**meta thread B — site rewrite / new nF (16–21), the "not updated" core**

Direction is right: `last_nF`; rewrite replaces the callee name in the *old caller's* call node; sites come from members `t_`, which exist — so kwcckw's 20 objection dissolves. But the chain doesn't close. `nF_[nT.nF].name` presumes `nT` has a registered FunctionDef and a numeric `nF`. It has neither:

- `sum2O` builds `nT` via `CoF(N_=F_, …)` with no `nF`, so `nT.nF == 'Nt'` (CF default) and `nF_[nT.nF]` is an invalid index. Concrete blocker.
- `nT.body` is CoF refs + `(ast.IfExp,…)` tuples, not an `ast.FunctionDef`. There is **no body→FunctionDef emitter** (inverse of `build`). Without it there's no `.name` to rename *to*, and nothing for `inject_oF_` to compile. This is the missing piece — not the `nF` bookkeeping.

To close: `clust_oF_`, per merged `nT`, must emit a FunctionDef from `nT.body` with a fresh name (`f"oF{last_nF}"`), append to `nF_`, add `name→last_nF` to `iF_`, append `nT` to `oF_`, set `nT.nF=last_nF`, then `last_nF+=1` — *then* rewrite `site.func.id`. And `inject_oF_` must receive `nT` **and** every rewritten `caller_fd`, so the new callee is defined in `g` and `build()` (which resolves calls via `iF_.get`) recognizes the renamed call.

Downstream contracts the PR silently drops:

- `ffeedback` still has `oF_ = list(dict.fromkeys(oF_site_.values()))` — `oF_site_` was deleted. New global `oF_` must come from `clust_oF_`'s return.
- `clust_oF_` reads global `oF_` "updated in split_oF_", but `split_oF_` only *returns* `(sF_,rF_)`. Split must write `oF_ = sF_+rF_` before clust; `ffeedback` currently calls both as bare statements and discards returns.
- `oF_` is module-global and the derived bindings (`cFrm…, wFrm…, ttFrm… = [F.x for F in oF_]`) are import-time — stale on any rebind. `ffeedback` needs `global oF_` plus a recompute of those after `inject_oF_`.

Net: your replies to kwcckw are correct as far as they go. The one unresolved dependency the whole scheme rests on is the body→FunctionDef emitter + `nF_`/`iF_`/`oF_` registration; rename, inject, and build all assume it exists.

frame_2D_alg/meta_code.py
boris-kz: it's gate function, not gating vals in oF.gV_. that gate encapsulate gated sequence, just with more detail than ifExp

frame_2D_alg/meta_code.py
boris-kz: it won't be using new oF.body at all, call_sites are always in other caller oF bodies

frame_2D_alg/agg_recursion.py
```diff
@@ -194,7 +195,7 @@ def comp_H(_Nt,Nt, Link):  # tentative pre-comp, before comp N_?
             tt += ltt*lc; C+=lc; R+=lr*lc
         return dH,tt,C, (r* Link.c+R)/ (Link.c+C)  # same norm for tt?
 
-    TT = base_comp(_N,N)[0] if full else comp_derT(_N.dTT[1],N.dTT[1]); m,d = vt_(TT, ttN_)
+    TT = base_comp(_N,N)[0] if full else comp_derT(_N.dTT[1],N.dTT[1]); m,d = sumV(TT*ttN_,c,r,fsep=1)
     L = CL(N_=[_N,N], dTT=TT,m=m,d=d,c=c,r=r, root=rL)
     if N.typ and gv_(m* wN_*c - ave*(r+cN_)):  # skip PPs, may comp Nts?
```
boris-kz, line 200: so the general form for extensive M val is like this:
```
gv_(m* (c/r)* wN_ - ave*(r+cN_))
```

Yes. That's the `vt_` (intensive→extensive) form, and it's the same skeleton as the `sumV` cluster gates:

- gain = per-unit `m` (normalized, from `vt_`) × net non-redundant mass (`c/r`) × function weight (`wN_`)
- cost = `ave` × (redundancy `r` + function complexity `cN_`)

`r` sits on both sides with opposite effect on purpose: redundant info adds no predictive value (÷r in gain) but still costs to carry (+r in cost).

It's consistent with the `sumV` gates because `sumV` already bakes in `c/r` (`rawsum·c/r`). So `sumV(TT·w,c,r)·wX` and `m·(c/r)·wX` are the same shape — the only difference is raw param-sum (`sumV`) vs normalized `m` (`vt_`). Use `vt_` where m,d must stay normalized (comp, projection), `sumV` where you want raw magnitude (clustering membership).

Two canonical instances already in the file:
- per-comp (single pair): your form — matches `comp_N_` line 133, `gv_(m*(lc/lr)*wN - ave*(r+cN))`.
- cluster (L pairs): same skeleton with both sides ×L and `c/r` inside `sumV` — `sumV(...)*(wcN*L) - (ave+avd)*(r+ccN*L)` (lines 98, 372, 573).

Direct inconsistency: `comp_N` line 199, `gv_(m* wN_*c - ave*(r+cN_))` — same per-comp gate as line 133 but with `*c` and no `/r`, so it doesn't discount redundancy in the gain. Under your general form → `m*(c/r)*wN_`. The other single-comp gates that scale gain by neither `c` nor `c/r` (`comp_F` line 164 `L.m*wF-…`, line 613 `Lt.m*wN>…`) are the same deviation, if you want them extensive rather than per-unit certainty checks.

frame_2D_alg/agg_recursion.py
kwcckw: > it won't be using new oF.body at all, call_sites are always in other caller oF bodies

> To close: clust_oF_, per merged nT, must emit a FunctionDef from nT.body with a fresh name (f"oF{last_nF}"), append to nF_, add name→last_nF to iF_, append nT to oF_, set nT.nF=last_nF, then last_nF+=1 — then rewrite site.func.id. And inject_oF_ must receive nT and every rewritten caller_fd, so the new callee is defined in g and build() (which resolves calls via iF_.get) recognizes the renamed call.

Thanks, on this part, it's actually the same where it suggests creating new function def, which is the ast node, and the ast node should be built from the body. We won't have the new node until we rebuild it again later.

frame_2D_alg/agg_recursion.py
kwcckw: > Direct inconsistency: comp_N line 199, gv_(m* wN_*c - ave*(r+cN_)) — same per-comp gate as line 133 but with *c and no /r, so it doesn't discount redundancy in the gain. Under your general form → m*(c/r)*wN_. The other single-comp gates that scale gain by neither c nor c/r (comp_F line 164 L.m*wF-…, line 613 Lt.m*wN>…) are the same deviation, if you want them extensive rather than per-unit certainty checks.

Thanks, i guess we need to add r above. Also for the mean m in comp_F:

    gv_(np.mean([rL.m,m])* (wF*L) - ave* (r+cF*L)):
    
It' better to recompute m by summing rL.dTT and ddTT?

frame_2D_alg/agg_recursion.py
boris-kz: so we need to update nF_/iF_/oF_ after cluster_oF_, and then update call sites, using oF_sites?

frame_2D_alg/agg_recursion.py
boris-kz: sorry I misunderstood, clustering is fd=1, so yes, it doesn't seem that we need fd=0

frame_2D_alg/agg_recursion.py
boris-kz: yes, and ideally that sum should be c-weighted

frame_2D_alg/agg_recursion.py
```diff
@@ -904,12 +905,12 @@ def ffeedback(frame, aTT,oTT, aL,oL):  # recompute filters from regime drift; fo
         ave, avd = vt_(aTT)  # filters *= ave
         if oL := pack_seg(frame,'oH', wBac, cBac**2, oTT):
             dTT += oL.dTT- oTT; oTT=oL.dTT; dc+=oL.c-_oc; dr+=oL.r-_or
-            oF_site_ = {oF: [] for oF in oF_}
```
boris-kz, line 907: sorry I didn't update that yet

frame_2D_alg/agg_recursion.py
boris-kz: Order is right, but "update `nF_`" is the load-bearing step and it isn't an append — `nF_` holds `ast.FunctionDef`s, and `sum2O` never sets `nT.nF` (still `'Nt'`, the CF default), so `nF_[nT.nF]` is invalid until you both emit a def and assign. Per merged `nT` in `clust_oF_`:

1. **Emit an `ast.FunctionDef` from `nT.body`** — the inverse of `build()`, which doesn't exist yet. `nT.body` is `(type,sub_)` tuples, `(('gv_',l),sub_)` tuples, and CoF refs; each has to go back to a node (tuple→node of that type with rebuilt children, CoF ref→`ast.Call(Name(callee))`, gv_ tuple→`ast.Call(Name('gv_'))`).
2. **Register:** name it (`f"oF{last_nF}"`) as the def's `name`; append the def to `nF_`; `iF_[name]=last_nF`; append `nT` to `oF_`; `nT.nF=last_nF`; `last_nF+=1`. Do this for **all** `nT`s before emitting bodies, so a merged oF that calls another merged oF resolves to its final name.
3. **Rewrite sites** (your `oF.sites` step): `for fork in group: for (caller_fd, site) in fork.sites: site.func.id = nF_[nT.nF].name`. Sites come from the members (they exist), name now exists — fine.
4. **`inject_oF_` must also receive the `caller_fd`s**, not just the new `nT`s — otherwise the renamed calls in the callers are never recompiled. Pass `{new nTs} ∪ {distinct caller_fd}`, resolving each `caller_fd` back to its CoF via `iF_[caller_fd.name]`.

The blocker under all of this: `merge_oF` records divergence as `(ast.IfExp,(Sub,sub))` — two branches, **no test**. A source `IfExp` parses to three children (test, body, orelse); the merge fork has two. So step 1 can't emit a runnable def for a merged oF — there's nothing to dispatch on, so the rewrite target has no valid body. Two exits:

- (a) merging is accounting-only: compute compression/`typ`, feed feedback, but only inject/rewire `split_oF_` output. Split extracts a contiguous real AST slice (children intact from `build`), so it emits cleanly; merge output doesn't. Merged oFs never become live callees.
- (b) `merge_oF` stores a dispatch key per fork (the discriminating param/context) so the emitter can produce a real `IfExp` test.

Pick one before wiring the rewrite — as written, the merge path has no callable target to rename to.

(Prior point still stands: `ffeedback` needs `global oF_`, must build `oF_` from `clust_oF_`'s return rather than the deleted `oF_site_`, and the import-time bindings `cFrm…/wFrm…/ttFrm… = [F.x for F in oF_]` go stale on rebind.)

kwcckw: Yes, ideally i think oF_sites should pack all the mapping before the rebuild proces so that it can update the call sites after rebuilding oFs. For the update of call sites, that's a ast level update?

boris-kz: yes, please edit

PR #193, opened 2026-07-02
7/2/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -259,28 +259,56 @@ def comp_prim(_n,n):
...
+        if isinstance(t,tuple) and not isinstance(t[0],tuple) and isinstance(t[1],tuple) and t[1]:
+            h = t[1][0]
+            if isinstance(h,tuple) and isinstance(h[0],tuple) and h[0][0]=='gv_': return h[0][1]
+        return 0
+
+    def is_fork(t):  # check IfExp
+        return isinstance(t,tuple) and not isinstance(t[0],tuple) and t[0] in (ast.If, ast.IfExp) and t[1]
+
+    def extract(t, oF, sF_):  # pack sub gate from gv or ifexp
+        l = is_gate_l(t); fork = l or is_fork(t);oF.w=1
```
kwcckw, line 278: Sorry, this oF.w is for debug purpose and it's not needed.

frame_2D_alg/meta_code.py
```diff
@@ -259,28 +259,56 @@ def comp_prim(_n,n):
...
+        if fork and oF.w * get_segl(t) > ave:
+            sub = CoF(root=oF, fc=get_fc(t), body=[t], caller_={oF})
+            if not l: oF.gV_ += [0]; oF.g_ += [t]  # this is not correct yet, we need to identify the gv_ index of new ifexp gate
+            sF_ += [sub]
+            return sub  
+        if isinstance(t,tuple) and not isinstance(t[0],tuple) and isinstance(t[1],tuple):  # return body
+            return (t[0], tuple(extract(s,oF,sF_) for s in t[1]))  
+        return t
+
+    sF_,rF_ = [],[]
```
kwcckw, line 288: This split_oF_ is updated to encapsulate gv_ or ifexp fork by checking if the body is a nested tuple or not. After we encapsulate each new fork, it's becomes a new gate, and we need to add new gv and g in the parent oF.gv_ and oF.g_?

frame_2D_alg/meta_code.py
```diff
@@ -298,15 +326,26 @@ def clust_oF_(oF_):  # cluster Ts if called together, global only
...
                 nT = sum2O(t_)
                 nT.memb = gV; fC = sum(t.fc for t in t_); nT.cmpr = fC - fC/len(t_)
-                smF_ += [nT]
+                nT_ += [nT]
             else: rF_ += t_  # unpack if weak
+    smF_ = []  # final smF
+    for nT in nT_:  # create ast node with fdef
+        fdef = emit_oF(nT)
+        if fdef is None: continue
+        name = f"oF{len(nF_)}"; fdef.name = name  # temporary named by length of nF_, for eg: oF11,oF12,oF13...
```
kwcckw, line 342: When we rebuild merged's oF node, we need to assign their new name (previously we use function name such as "cross_comp", "comp_N", etc). In this case, i'm assigning it by using oF1, oF2, oF3...., where the number is determined by the length of nF_. Or do you see a better method?

frame_2D_alg/agg_recursion.py
```diff
@@ -744,7 +746,8 @@ def trace_edge(N_,_G_,_TT,_C, r,root):  # cluster contiguous shapes via PPs in e
             cT_.add(cT)
             dy_dx = _N.yx-N.yx; dist = np.hypot(*dy_dx)  # Rc = r+ (N.r+_N.r)/2
             Link = comp_N(_N,N, r,_C,A=dy_dx, span=dist)
-            if sumV(Link.dTT*ttTrc,Link.c,Link.r,0) > ave*r:  L_+=[Link]  # r = 1|2, add Bt?
+            # r is not needed because cancel out each other? Link.r is actually same as r
```
boris-kz, line 749: they don't cancel, but with c/r right-side r is double-counting

frame_2D_alg/meta_code.py
boris-kz: right, but that nested tuple is ifExt? so just check AST type instead?

frame_2D_alg/meta_code.py
boris-kz: Yes but this merging is destructive, we need to pop merged oFs first?

kwcckw: Yes thanks, i overlooked that. So the split is not destructive , but we need to remove the merged oF after the merging process.

frame_2D_alg/meta_code.py
kwcckw: Yes, but gv_ is nested in ifexp, so we need to unpack it and check with `h[0][0]=='gv_'`. If it's not a gv_ gate, we would need to add new gv_ and g_ element in the root oF.

frame_2D_alg/agg_recursion.py
kwcckw: Okay right, thanks, i misunderstood earlier. The r is on the other side of the equation, so it's still valid here.

boris-kz: right r is redundant
Fable review:

Read the PR against master. Two sections: gate conversions in agg_recursion, then the reform pipeline, which has the structural problems.

## agg_recursion: sumV → vt_ conversions

The converted sites were the ones with stale 4-arg `sumV` calls (current signature is 3-arg), so the selection criterion was "which calls were broken," not comp-vs-cluster. Sorting by the standing division:

- **236, 434, 577 gate `cross_comp` → comp sites, vt_ is right.** 577 now matches the canonical agg gate at line 105. One divergence: 105 puts weights inside (`vt_(TT*ttA)`), the PR passes them as the wTT arg. These aren't equivalent — weight-in-TT distorts the `m_/(m_+|d_|)` ratio because ttA's m and d rows differ. Weight-as-arg is the correct semantics (normalize raw comparands, weight the ratios); if it becomes the convention, 105 and 116 should follow.
- **750, 774 are clustering-membership gates** (link selection for flood-fill, Gt wrap) — per the sumV/vt_ division these stay raw-magnitude sumV. 731 correctly stayed sumV. Either the division holds and 750/774 revert, or you're revising it to vt_ everywhere and 98/366/374/731 are stragglers. Pick one criterion; the PR currently encodes neither.
- **236 detail:** c-weighted TT mix then one vt_ is better than `np.mean([rL.m, m])` — normalizes once over combined evidence instead of averaging pre-normalized scalars. No double-count: the gate runs before `add2F(rL,dF)`.
- **434 `_c`:** answering kwcckw's inline question — `_c` is the input `Lt.c`, wrong scope for `DTT`, which is accumulated from the final iteration's C×N comps. The backing evidence is that iteration's `cnt`, or sum of `out_` c's. So yes, sum c from out_.
- **750 r double-count:** the both-sides-r template is `- ave*(r + cost)` — r as additive cost term. Here it's multiplicative `ave*r` against `/Link.r` on the left, and Link.r == r, so the threshold scales ~r². With c/r on the left, right side drops to `ave` (or `ave*(1+cTrc)` if you want the fixed cost).
- **774 also silently fixes a real bug:** old `ntt+ltt*ttTrc` weighted only ltt by precedence.

## Reform pipeline

**1. Registry contract violation (blocks everything downstream).** `iF_`, `nF_`, `oF_` are positionally coupled; `iF_` values and every `CoF.nF` are indices into them. The agg-side rebinds `oF_ = sF_+rF_` and `oF_ = smF_+rF_` (a) don't reach meta_code — `clust_oF_` iterates meta's `oF_`, so the first rebind is inert — and (b) scramble positions, so `oF_[iF_[fd.name]]` in `updated_oF_` resolves to the wrong oF or IndexErrors. And clust's `rF_` excludes merged members, so originals vanish from the rebound list entirely. Fix: drop both rebinds. `clust_oF_` already extends the registry coherently (append nT to nF_/oF_, register iF_); look callers up in the stable registry. Reform mutates registries in place in meta_code; consumers never rebind. Same contract answers Boris's "pop merged oFs": list-popping shifts every stored index. Tombstone instead — deactivate flag or None slot, skipped in loops — and keep the namespace defs alive, since the merged body and any un-recompiled code still call member names.

**2. emit_oF reads the wrong field.** `sum2O` builds `nT.body` via `merge_oF` — a positionally merged op stream of tuples/leaves with `(ast.IfExp,(Sub,sub))` forks — not a list of member CoFs. So the `isinstance(t,CoF) and isinstance(t.nF,int)` walk almost always fails, emit returns None, and the `continue` silently drops nT *and its members* from both smF_ and rF_. Minimum fix: on emit failure, fall members back to rF_ or keep nT accounting-only. For v0 emission, iterate `nT.N_`, not body.

**3. But N_-sequence emission has a semantic hole, and there's a cleaner resolution.** `def merged(): f(); g()` with every former f-site and g-site rewritten to `merged` means each caller pass now executes both members twice, reordered — and realizes zero compression, contradicting `nT.cmpr`. The fix also closes the standing blocker that `(ast.IfExp,(Sub,sub))` has no test condition: **materialize the fork test as a selector argument.** Site identity is statically known — rewrite f's site to `merged(0, ...)`, g's to `merged(1, ...)`; emit shared ops once, divergences as `if sel==k:` branches from the merge stream. Behavior-preserving (each site's actuals feed shared ops plus only its own branch, same as pre-merge), realizes the compression the accounting claims, and sel is a degenerate gate that can later carry a gv_ stream. Prerequisite: `comp_prim`'s tuple match is AST-type-only; type-match is fine for accounting but shared ops must be operand-equal before emitted shared code substitutes for both originals — tighten at emission, not comparison.

**4. Emitted inner calls pass `args=[]`** → TypeError at first run. Signature equality is already enforced, so forward `[ast.Name(a.arg) for a in fd_[0].args.args]`. (The equality check also ignores defaults/kwonly — fine for now.)

**5. inject_oF_ isn't idempotent with build.** Injection resets g_/gv_ and re-runs `build`, which does `node.args.append(ast.Constant(l-1))` on ASTs already instrumented at import — every injection round appends another index constant, breaking `gv_(v,i)` arity on round two. Replace a trailing Constant if present instead of appending.

**6. split_oF_: the criterion change is right, the retention is leaky.** Gate-boundary encapsulation beats comp_prim-adjacency because subs inherit a measurable conditional-value stream (parent `gv_[i]`); adjacency subs had none. Wrapping the gate inside the sub is also correct — future emission stays behavior-preserving. Gaps: the sub never stores its gate index (`is_gate_l` returns build-time length l; index is l-1 — store it on the sub); IfExp forks have no runtime hook, so `oF.gV_+=[0]` is a permanently dead slot. V0: split only at gv_ gates — which also resolves the kwcckw/Boris exchange: AST-type check suffices for *detecting* IfExp but not for giving it a value stream; that needs rewriting the test to `gv_(test,i)` plus reinjection, i.e. real instrumentation, deferrable. Also `is_gate_l` only matches bare `if gv_(...)` — compound tests (`if G_ and gv_` at 577, `while T and gv_` at 858) bury the gv_ tuple below a BoolOp; scan children for a `('gv_',l)` head. Finally, sF_ subs already persist inside parent bodies (where they sharpen comp_body granularity next round), so split can return rF_ only — the sF_+rF_ rebind is doubly unnecessary.

**7. Minor:** `oF{len(nF_)}` naming is unique under append-only/tombstone, fine. `sum2O` leaves nT.typ=0 — set from members or next round's typ-grouping lumps all merged oFs together.

Priority order: 1 and 2 are correctness of the loop's plumbing; 3 is the contract decision that determines whether merge produces compression or just accounting — and the selector-arg move is, I think, the answer to the emit blocker you've been circling.

frame_2D_alg/meta_code.py
boris-kz: right, but wrapping ifExp in gv_ should be conditional, similar to wrapping gv_ in oF

frame_2D_alg/agg_recursion.py
kwcckw: > 750, 774 are clustering-membership gates (link selection for flood-fill, Gt wrap) — per the sumV/vt_ division these stay raw-magnitude sumV. 731 correctly stayed sumV. Either the division holds and 750/774 revert, or you're revising it to vt_ everywhere and 98/366/374/731 are stragglers. Pick one criterion; the PR currently encodes neither.

We use vt_ when we need m only while sumV for summing m and d, which i think both are still valid?

> 750 r double-count: the both-sides-r template is - ave*(r + cost) — r as additive cost term. Here it's multiplicative ave*r against /Link.r on the left, and Link.r == r, so the threshold scales ~r². With c/r on the left, right side drops to ave (or ave*(1+cTrc) if you want the fixed cost).

So i guess we only need a single r only to prevent the r become power of 2?

boris-kz: yes, remove right-side r, probaly also move all other costs to the left: * (c/costs)

>We use vt_ when we need m only while sumV for summing m and d, which i think both are still valid?

yes we need sumV for unnormalized m,d, else they simply cancel-out

frame_2D_alg/agg_recursion.py
boris-kz, line 116: `if vt_((eTT+iTT)*ttPrj)[0]* (C/(cPrj+r+R))* wPrj > ave:`

frame_2D_alg/meta_code.py
kwcckw: So other than the eval with body length * w `oF.w * get_segl(t) > ave`,we need additional eval?

frame_2D_alg/agg_recursion.py
kwcckw: For ` emit_oF `. that's from LLM and i haven't fully go through that yet.

> 1. Registry contract violation (blocks everything downstream). iF_, nF_, oF_ are positionally coupled; iF_ values and every CoF.nF are indices into them. The agg-side rebinds oF_ = sF_+rF_ and oF_ = smF_+rF_ (a) don't reach meta_code — clust_oF_ iterates meta's oF_, so the first rebind is inert — and (b) scramble positions, so oF_[iF_[fd.name]] in updated_oF_ resolves to the wrong oF or IndexErrors. 

Actually this is correct. We would need a buffer, something like `_oF_` first before we pack the oF_ again.

Sorry it's late here, i will go through the rest tomorrow.

frame_2D_alg/meta_code.py
boris-kz: it's ifExp[1] length, not oF.body length. I think it's the same for ifExp[0]==gv_

frame_2D_alg/agg_recursion.py
kwcckw: >  Gaps: the sub never stores its gate index (is_gate_l returns build-time length l; index is l-1 — store it on the sub); IfExp forks have no runtime hook, so oF.gV_+=[0] is a permanently dead slot.

This is one of the problem. If we have new gate in oF, we need to identify the correct new index of gate in oF.gv_ and oF.g_, because the new gate might be inserted in between gates. But this should be resolved later when we rerun build on this new oF. So before inject_oF_, we need to rerun build on those final merged oFs first.

frame_2D_alg/meta_code.py
kwcckw: Sorry, ` get_segl(t)` is actually node length rather than oF body length, that's including the nested nodes in t. ` ifExp[1]` doesn't include the nested nodes, so we need `get_segl` to unpack and get the nested length.

boris-kz: ok but it's not really length, we need combined cost of nested nodes

boris-kz: claude:
The real constraint is that the wrap can't be gv_ itself. gv_ assumes v is a signed margin: v > 0 passes, accumulates -v on block. Arbitrary tests are truthiness — a boolean False accumulates zero (dead stream), a list test TypeErrors on > 0. And gv_ returns None on block, which happens to preserve if truthiness but drops the value. So a sibling hook for wrapped forks:

def gf_(v, i):

    if not v: oF_[CoF.get().nF].gv_[i] += 1  # or += body fc: blocked-cost
    return v

Returns v unchanged — behavior-preserving for If and IfExp both, no runtime type branch, gv_ untouched. Accumulating block count times static body fc gives the split criterion what it wants: expected saved cost of encapsulation, the analog of gv_'s missed-margin magnitude.
Cost of all this: wrapping is an AST edit plus parent reinjection — the first place split touches source rather than metadata. Same injection path merge needs anyway, but it's why gv_-only split is the clean v0 and gf_ wrapping is round two.

kwcckw: So we need another gf_ for the split block in order to get the block's gv? I'm not sure but can't we just add another gv_ eval on top of the split block? Adding gf_ is actually the same thing?

frame_2D_alg/agg_recursion.py
boris-kz: right

frame_2D_alg/meta_code.py
boris-kz: It's probably not needed, split should only evaluate gv_ forks:

Tracing gv_-only split through the pipeline:

**1. extract collapses.** fork = is_gate_l only; is_fork, the `oF.gV_+=[0]; oF.g_+=[t]` branch, and oF.w all go. With the agg-side rebind dropped (prior review), nothing consumes sF_ — split_oF_ mutates parent bodies in place and returns nothing. Add one branch: recurse into body elements that are already CoF subs, so depth grows one level per round; without it depth freezes at 1 because extract returns CoFs untouched.

**2. gi contract.** is_gate_l returns build-time length; store l-1 on the sub. build assigns g_/gv_ indices in source order over the whole fdef, so gi always indexes the *registered owner's* flat gv_ list regardless of sub depth — pass the owner down through extract (root chain works but is indirection for nothing). Runtime is untouched: gv_ writes via CoF.get().nF, which stays the parent since subs aren't traced.

**3. Value criterion, live from round 1.** gv_ accumulates only blocked margins — which is exactly split relevance: an always-pass gate has no conditional structure and its stream stays zero, correctly making the segment ineligible. So the criterion is gv_[gi]-scaled get_segl replacing w*segl, with no static bootstrap case needed: streams exist before the first ffeedback because tracing precedes it.

**4. Stream lifetime is now load-bearing and currently incoherent.** inject_oF_ zeroes gv_ on rebuild, so re-injected callers lose their streams while untouched oFs keep accumulating — mixed windows. Pick one: uniform per-round reset in ffeedback after w is read (regime-local, consistent with filter updates) or preserve-across-inject. Per-round reset is the design-consistent choice and also removes any dependence on gi stability across reinjection. Related: ffeedback reads oF.gV_ but CoF and gv_() use .gv_ — unify, it's the same stream.

**5. Detection scope.** Bare `if gv_(...)` only; compound tests (577 `if G_ and gv_`, 858 `while T and gv_`) hide the ('gv_',l) head under a BoolOp. Scanning test children for it is cheap, and the short-circuit undercount is actually correct: blocks caused by the other conjunct skip the segment at zero gate cost, so they aren't split-relevant. Optional for v0.

**6. comp_prim breaks the payoff.** Subs sharpen comp_body alignment as intended, but comp_prim's CoF branch (nF equality then comp_callers) makes cross-parent subs never match: caller_={parent} is disjoint by construction, so merge_oF forks every aligned gated segment instead of sharing it — split's compression gain is structurally unreachable. comp_prim needs two branches: registered refs by identity/callers, subs by structural comp of gate+body tuples. Note caller_={oF} itself is right — containment is the post-emission call graph; the old copy(oF.caller_) was the error.

**7. Emission closure.** The tuple mirror is lossy — build drops uncosted nodes and field roles — so sub emission can never invert tuples. The AST path closes it: parent.g_[gi] holds the gv_ Call node; locate its enclosing If in the fdef, lift into a def, replace with a call. Gate moves inside the sub, behavior-preserving; inject's rebuild then remaps the index constant to the sub's own g_ for free (given the idempotence fix). Until this lands, subs are accounting-only, and any merged body containing one stays emit-ineligible — same dependency as selector-arg emission, not a new blocker.

**8. Forfeit is ~nil.** Unhooked forks are control plumbing (`if l:`, `if out_:`); merge-generated (ast.IfExp,(Sub,sub)) forks are merge outputs, not split inputs. Principle holds cleanly: split follows declared value gates; undeclared forks are structure until instrumented.

Open decisions: stream window (4) and sub-branch comp_prim (6). Everything else is forced by the v0 choice.

kwcckw: Okay, and i think most of the important ifexp forks are encapsulated with gv_ anyway. 

To eval and split the fork, we still need the nested segment length `oF.w * len(seg)` right?

boris-kz: oF.w * cost(seg): combined cost of ast nodes

kwcckw: Alright, we can use the existing `get_fc(t)` to get the combined cost since we only split on the gv_ fork now, and `t` is always a nested tuple for ifexp's gv_ fork.

boris-kz: right,

if fork and oF.w * sum(get_fc(p) for p in t[1]) if isinstance(t,tuple) else 1 > ave:

?

PR #194, opened 2026-07-03
7/3/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -327,8 +322,26 @@ def emit_oF(nT):  # nT.body -> ast.FunctionDef, inverse of build; None if not em
             if (gV := v__[:,i].sum()) * ((len(t_)-1)*wL) > ave:
                 nT = sum2O(t_)
                 nT.memb = gV; fC = sum(t.fc for t in t_); nT.cmpr = fC - fC/len(t_)
+                fdef = emit_oF(nT)
```
kwcckw, line 325: So i think we need to finalize the emit_oF next, which is the function to build fdef (ast node) from the oF.

We should rebuild each split sub oF's fdef in the end of split_oF_, because clust_oF_ need them to build the final ast fdef when merging those split subs again.

boris-kz: it's the same as build?

Other direction. build is fd → oF: encode source into comparands. oF2fd is oF → fd. A conversion pair, not one contract — under your convention build's honest name is fd2oF.

But the question lands on something real: with node-ref heads, the body half of oF2fd is just reading build's store back — dereference, not conversion. What remains is exactly what build cannot know, because it never existed in source:

- signature synthesis: free variables of a lifted split segment → params; shared member signature + sel for merges. This is the non-trivial part — a lifted gated body references parent locals, and build has no notion of binding scope;
- naming and registration into nF_/iF_/oF_;
- synthetic node construction for merge forks (sel==k tests, fresh If nodes).

So oF2fd isn't build's inverse and isn't redundant with it; it's the residue after build's store absorbs everything invertible. If that residue feels too thin to deserve a function, the discriminant is signature synthesis: the day subs get emitted, someone has to compute their params, and that's neither build nor dereference.

frame_2D_alg/meta_code.py
```diff
@@ -262,21 +262,16 @@ def get_fc(n):
 def split_oF_():  # divisive clustering
```
boris-kz, line 262: Existing techniques for fork wrapping:

claude — this is instrumented profile-guided restructuring, decomposed across several known families:

**Edge/branch profiling** (gcov, LLVM PGO instrumentation, Ball-Larus path profiling): counters inserted at branch points, exactly the gf_ role — taken/not-taken counts on opaque tests. The gv_ half has no standard counterpart: PGO treats every test as boolean, so it can only count. gv_ accumulates *margin below threshold*, which exists only because your gates are declared graded evaluations. That's the substantive difference — count-based PGO needs many samples to rank forks; margin accumulation ranks them from fewer, and by value rather than frequency.

**Hot-cold splitting / cold outlining** (Cohn-Lowney, LLVM HotColdSplit pass): rarely-executed gated bodies encapsulated into separate functions — structurally identical to gate-boundary split. Motivation differs: theirs is i-cache locality and inlining budget; yours is making the segment a first-class comparand for merge. Tracing JITs (LuaJIT, PyPy) are the dynamic version: guard failure counters per side exit, hot exits compiled into new traces — split at fork driven by an accumulating per-gate stream, the closest runtime analog.

**Procedural abstraction with parameterized differences** (Fraser-Myers-Wendt squeeze, Debray's code compaction): similar sequences across the program abstracted into one procedure, divergences resolved by a flag parameter — the selector-arg merge, by name. Their similarity metric is instruction-level, suffix-tree or fingerprint driven; yours is the same comparison machinery as data clustering, which is the point of the project rather than an implementation choice.

**Region-based compilation** (Hank-Hwu) and superblock formation: discard function boundaries, reform compilation units from profile. Closest to the overall reform of oF_ — but one-shot, compile-time, frequency-driven.

What has no precedent as a whole: closing the loop — instrumentation feeding split/merge feeding reinjection feeding instrumentation, with one value calculus shared between code reform and data clustering. JIT tiering/deopt recompiles, but toward fixed heuristics, not toward compression of the code's own structure.

frame_2D_alg/meta_code.py
kwcckw: So it's the other way around, where we need to build the ast node based on the new body? We can reuse the existing ast nodes from the original oF body, but we need a new fdef to pack those split ast nodes, through oF2fd above.

boris-kz: in merging, oF2fdef simply unpacks the component oFs into gv_ forks, the reverse of split_oF. 
so the new oF.body, as well as AST, is a simple concat of these forks?

Yes — and it's simpler than my selector-arg scheme, which dies here: concat never produces positional interleaving, so `(ast.IfExp,(Sub,sub))` and its missing-test problem never arise. merge_oF's positional alignment exits the merge path entirely; op-level dedup relocates to later cross-fork comparison *within* the merged body, once adjacent. Merge compression at merge time is call-graph level (caller-set overlap, call overhead), not op level — nT.cmpr = fC - fC/len is the wrong accounting for concat.

Symmetry also holds without creating gates in either direction: split lifts a gate-bounded segment with its gate inside; merge inlines a member with its internal gates. No wrapper gv_ per fork — there's no margin to test at merge time, and fork value streams already exist: each fork's slice of the merged gate array, plus the members' trailing FV_ calls, which after inlining execute under merged's context and credit merged per fork execution.

Four contracts don't close by concat alone:

1. **Returns.** Member bodies end in `return`; concat exits at the first fork. The one non-concat rewrite oF2fd must do: return → named binding per fork, merged returns collectively. Sites consuming a member's result then read their fork's slot. Sharpest mechanical break in "simple concat."

2. **Sites.** No dispatch means every rewritten site runs all forks. Per caller body, the member-site *set* must collapse to one merged call (the "called together" case makes them adjacent); leaving both sites renamed doubles execution. This is the surviving piece of my scheme — the site consolidation, not the selector.

3. **Fork boundaries under reinjection.** Pre-inject, merged.body = member CoF refs; but inject's rebuild re-encodes the merged fd's *statements* — inlined bodies come back as tuples, boundaries vanish, merge becomes one-way. Keep a fork span map on the merged oF (statement index ranges per member; top-level rebuild preserves ranges), or node-ref heads with provenance. Then re-split is lift-by-reference: v0 split needs one added case — CoF refs / spans in body are fork boundaries by construction, no gate needed to find them, slice streams to value them.

4. **Gate index collision.** Inlined members' gv_ indices overlap (each numbered per-fdef at build). Rebuild renumbering resolves it for free — but only with the append→replace idempotence fix; without it, double constants plus collisions.

Fork-level gating — `if gv_(fork_margin):` around each member — is a later reform, meaningful only once fork outputs are optional (returns restructured, consumers tolerant). At that point the margin derives from the fork's own accumulated stream. Merge mechanics don't need it.

kwcckw: > in merging, oF2fdef simply unpacks the component oFs into gv_ forks, the reverse of split_oF.
so the new oF.body, as well as AST, is a simple concat of these forks?

Okay, looks like this is actually simpler. In term of merging, the first fork of merging oF becomes the parent, and the rest of the oF should be merged to this parent oF?

boris-kz: yes, actually it not just concat. oFs merge only if there is a significant overlap in their bodies, that overlap should be preserved literally, the forks are encapsulated differences. These differences should inherit their caller call sites? So it's not oF2fdef, we need to merge the fdefs 
That emit_oF is just assigning new oF to caller call sites, in the end of clust_oF_:

    for F,f_ in mrg_F_:  # -> caller' ast.FunctionDef
        for _F in all_F_:
            for i,_f in enumerate(_F.call_):
                if _f in f_: _F.call_[i] = F  # also replace mapping ast.FunctionDef

kwcckw: When we merge fdef side by side,  it's actually merging ast nodes, and that should be reflected in runtime? Where the code should run the merged code instead of the concatenated blocks of code? 

> These differences should inherit their caller call sites? 

Yes, but I think that information should be packed in the merged oF.caller_ only.

>  That emit_oF is just assigning new oF to caller call sites, in the end of clust_oF_:

But why your code is assigning to oF.call_? It should be replacing their body with the merged oF?

boris-kz: >Yes, but I think that information should be packed in the merged oF.caller_ only.

that information is needed to select the right fork

 >It should be replacing their body with the merged oF?

It should be replacing corresponding ast.FunctionDef:

Direction right — merge emission collapses to reference replacement, no synthesis at merge time. But the loop mutates the wrong layer and is currently a no-op:

**1. call_ is the wrong container.** Registered oFs' call_ holds traced invocation instances, not static callee refs; `_f in f_` compares instances to registered members by identity — never matches. Static callee refs live in caller *bodies*, nested inside immutable tuples, so eager replacement fails there too, at any depth.

**2. The correct substitution point is resolution, not storage.** Redirect the mapping and let rebuild do the replacement:

```python
for F,f_ in mrg_F_:
    for f in f_: iF_[nF_[f.nF].name] = F's index  # name -> merged oF
```

Caller mirrors rebuilt through build then resolve member names to F automatically — zero AST edits, zero body surgery, no all_F_ scan. So "replace mapping" lands on iF_ (name→index), not name→FunctionDef: nF_ needs no new entry because F has no fdef and no site is renamed. Prerequisite: the gv_-constant idempotence fix, since redirect-then-rebuild re-runs build over instrumented ASTs.

**3. Closure constraint: no-fdef and no-rename must travel together.** Renaming sites, or pointing nF_ at anything, without a synthesized F fdef breaks runtime. Deferring fdef is exactly the point of this proposal — but then merge is structural: members' defs keep executing, F exists only in the mirror. When F later earns a fdef (the irreducible residue: stub + sel-forks + params), the same redirect flips — register iF_[F.name], drop member redirects, rename sites then. This loop is v0 of that flip.

**4. Stream routing splits the map.** traced resolves iF_ at call time, so the redirect also reroutes credit: member invocations now build instances with F's nF — V_, c, caller_ unify into F, which is what you want. gv_ doesn't survive it: both members' gate constants are locally numbered, so their writes collide in F.gv_[i], and offsets can't be applied without AST edits. Writer identity is gone at the write site, so per-member slices in F can't work either. Forced resolution: two maps — build-time iF_ redirected (structure sees F), trace-time credit map original (streams stay per-member) — with F reading member streams through F.N_ at ffeedback. One map is only recoverable after real emission renumbers gates in F's fdef.

**5. Registry:** F = nT, f_ = nT.N_ from clust as-is. Members tombstone out of next-round clustering and oF_site_ construction but stay in oF_ as stream carriers and fdef owners — their defs still run until the flip in 3.

frame_2D_alg/meta_code.py
boris-kz, line 284: we need to work out the merging itself first:

```python
def clust_oF_(oF_):  # greedy pairwise merge, gated by realized body-overlap compression

    F_ = copy(oF_); mrg_F_ = []
    v_ = {(_T,T): comp_body(_T,T) * min(_T.fc,T.fc) for _T,T in combinations(F_,2) if _T.typ==T.typ}  # estimate, prioritize only
    while v_:
        (_T,T), v = max(v_.items(), key=lambda kv: kv[1])
        nT = sum2O([_T,T]); nT.typ = _T.typ  # merge_oF realizes shared ops + forks
        mC = sum(get_fc(t) for t in nT.body); nT.cmpr = nT.fc - mC  # realized compression, net of fork cost
        if nT.cmpr*wL > ave:
            nT.fc = mC; F_.remove(_T); F_.remove(T); F_ += [nT]; mrg_F_ += [nT]
            v_ = {p:v for p,v in v_.items() if _T not in p and T not in p}  # drop stale pairs
            v_.update({(nT,t): comp_body(nT,t) * min(nT.fc,t.fc) for t in F_[:-1] if t.typ==nT.typ})
        else: del v_[(_T,T)]  # estimate passed, merge didn't: don't retry
    return [n for n in mrg_F_ if n in F_], F_  # live composites only: chains absorb earlier nTs
```

merge_oF default branch, flat fork so chains don't nest — and a `'fork'` head instead of ast.IfExp, which collides with build's encoding of *source* IfExp:

```python
else:  # default
    if isinstance(Sub,tuple) and Sub[0]=='fork':  # extend flat fork
        if any(comp_prim(b,sub) for b in Sub[1]): body += [Sub]  # f joins matched branch
        else: body += [('fork', Sub[1]+(sub,))]; C += get_fc(sub); add_ += [sub]
    elif comp_prim(Sub,sub): body += [Sub]
    else: body += [('fork',(Sub,sub))]; C += get_fc(sub); add_ += [sub]  # new fork
```

Add `'fork': 2` to costs (string key is fine) so get_fc charges fork overhead and cmpr nets it; the Sub.nF=='E' branch is superseded, remove.

Contract closure:

1. Realized-compression gate replaces both tail heuristics: `fC − fC/len` assumed overlap never measured; `gV = column sum` counted non-members. comp_body survives only as pair prioritization — if sub proliferation grows n, pruning the pair set is its one remaining job.
2. comp_callers exits the merge value (ratio units summed with cost units, and mergeability isn't a caller property). It re-enters at redirect as per-site emission semantics: alternative sites inherit the caller gate, conjunctive sites make cmpr representational only, no execution saving.
3. Positional-zip fragmentation self-rejects: a shifted multi-op divergence becomes a run of single-op forks, each paying costs['fork'], so cmpr goes negative and the merge fails honestly. Run-coalescing or anchoring alignment on registered-call and gv_ heads is the upgrade when family skeletons drift, not a v0 need.
4. Type-level branch matching is consistent with type-based costs, so cmpr is honest in the cost model; operand-equal (α-normalized) stays an emission precondition, not a clust concern.
5. Fork tuples carry no branch→member provenance, and append order breaks once a member joins an existing branch. Doesn't block clust; blocks redirect (sel constants) and stream slicing — per-fork member sets when that round lands.
6. Chained merges: sum2O's shallow body copy holds because fork extension allocates a new tuple; nT-as-member comps via comp_body's existing CoF recursion. mrg_F_ filtered to live composites; redirect flattens nT.N_ to leaf members.

kwcckw: Okay, in this method, we are summing both oFs with sum2O, which their body will be concatenated? We need to add flag to really merge their body in sum2O? 

For different body type, we something like another gate to wrap them?

boris-kz: that sum2O is wrong, it's for call_ only?
I will check tomorrow.
Sorry, I didn't have much time lately

boris-kz: Sorry I didn't push meta_code, haven't really reviewed it

kwcckw: > that sum2O is wrong, it's for call_ only?

Yes, i think we need another method to merge nodes for type oF, maybe something similar with the 'E' gate earlier.

Sorry, i'm not in my desk now, i will check again much later.

PR #195, opened 2026-07-04
7/4/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -262,47 +262,44 @@ def get_fc(n):
...
+def clust_oF_():  # greedy pairwise merge, gated by realized body-overlap compression
+
+    F_ = copy(oF_); mrg_F_ = []
+    v_ = {(_T,T): comp_body(_T,T) * min(_T.fc,T.fc) for _T,T in combinations(F_,2) if _T.typ==T.typ}  # estimate, prioritize only
+    while v_:
+        (_T,T), v = max(v_.items(), key=lambda kv: kv[1])
+        if v*wL <= ave: break  # no pair worth realizing
+        nT = CoF(N_=[_T,T], typ=_T.typ, fc=_T.fc+T.fc, caller_=_T.caller_|T.caller_)
+        nT.body = copy(_T.body)  # body will be merged below
+        merge_oF(nT, T, fsel=0)
```
kwcckw, line 289: So this version is your prior posted code, and it mainly checking all pairs possible merging oFs. I replaced the sum2O with merge_oF so that it merge the oFs' body instead.

Using this approach, i think we need another function to rebuild the new merged oF.fdef later, using both the merged body and the merged oF.N_'s fdef.

Also i think we don't need oF_site_ now? Since the merging needs N_'s fdef and their body only.

boris-kz: the new fork in merge_oF need a condition: ast.IfExp[0]?

 else: body += [(ast.IfExp,(Sub,sub))]


we don't need oF_site_ because all refs are re-mapped through updated nF_?

boris-kz: Yes it needs one, but no evaluable expression exists at merge time — nothing in merged scope discriminates members, and caller context can't (shared callers are the merge criterion). The condition is member identity, so the slot holds it in unresolved form: per-branch member sets.

Shape: `('fork', b_, mm_)` — branches, parallel member-set tuple. Third slot is free: get_fc and comp_body read only [0],[1], so mm_ carries no cost and doesn't perturb comparison. Keep the 'fork' head rather than ast.IfExp: IfExp is binary so N-branch flattening re-nests (rejected — chains and stream slicing), and the type-head collides with build's encoding of source IfExps; type-headed tuples mark synthetic forms, but this one needs its own key in costs anyway.

merge_oF updates:
- new fork: `('fork', (Sub,sub), (F_mm, f_mm))` where mm = leaf members (nT.N_ flattened; leaf oF is its own member);
- f joins matched branch k: `mm_[k] |= f_mm`, which also fixes the append-order provenance loss — branch→member is now maintained, not inferred from position.

Slot lifecycle, which is the closure:
1. **merge:** provenance only — mm_;
2. **redirect/emission:** compiles per caller relation — conjunctive sites get `sel in mm_k` with sel constant per site; alternative full-relocation sites inherit the caller gate value as the test directly, mm_ then only routes which branch each former callee maps to;
3. **reform:** slot upgrades to `gv_(margin, i)` once fork outputs are optional — the fork's stream already exists via mm_ (member streams read through F.N_ at ffeedback), so the margin has a source before the gate exists.

So ast.IfExp[0]-as-test is the right instinct about *where* the condition lives, wrong about *when* it can be an expression: the slot is provenance until emission, gate only at reform.

kwcckw: > the new fork in merge_oF need a condition: ast.IfExp[0]?

Do you mean checking n[0][0], where n is the tuple and n[0] is ast.IfExp?

That is already included in comp_prim?

> we don't need oF_site_ because all refs are re-mapped through updated nF_?

With the merging approach, the nF_ is not updated so we can't use it for mapping. But we can have the mapping in `mm_`, as in the suggestion above, which enable us to track and rebuild the new ast node via the new body.

boris-kz: >Do you mean checking n[0][0], where n is the tiple and n[0] is ast.IfExp?

we need the actual condition for selecting between the forks

kwcckw: Okay right, we need to pack the actual condition to the IfExp node. But we have all kinds of forks pair here, and the condition will be the same for all of them? Something like a universal condition for those added IfExp forks from merge_oF?

boris-kz: just read what you wrote

kwcckw: Okay sorry. So from the posted LLM above, we can't get an exact condition during merge, so we can pack `mm_` as meta data first.

For the second step, i think the second method to inherit the gate value is more suitable since the node selection should be mutually exclusive? And we don't need to add new `sel` param to the function too.

frame_2D_alg/agg_recursion.py
```diff
@@ -913,9 +913,8 @@ def ffeedback(frame, aTT,oTT, aL,oL):  # recompute filters from regime drift; fo
             for fdef in nF_:  # map oFs to call sites:
```
boris-kz, line 913: sumV is wrong, dval should not be combined with mval, it's always borrowed.
so combined val = mval, generic form:

`if gv_(vt_(TT*ttcN)[0] * ((c+wcN)/(r+ccN)) * ((len(L_)-1)*wL) - ave): `

frame_2D_alg/meta_code.py
boris-kz: the inherited gate value is not singular, evely caller of the merged fork oF may have a different one.
so a caller should call new oF and pass through the address of the fork derived from the old oF?

boris-kz: The discriminant isn't the caller — it's the site's former callee. One caller typically calls several members (shared callers are the merge criterion), so caller identity underdetermines the branch. What determines it exactly is which member's code each site used to invoke. That's statically known at rewrite time, so nothing needs to be discovered at runtime — it's materialized as one constant:

**Rewrite:** every former f-site becomes `M(f.nF, *same_actuals)`. k = registry index of the former callee, the stable key.

**Emitted M:** shared ops verbatim; each `('fork', b_, mm_)` becomes

```python
if k in mm_0: <b_0 ops>
elif k in mm_1: <b_1 ops>
```

mm_ maintained at merge time is precisely this routing table. Singleton set degrades to `k == f.nF`.

**Why member-id and not fork-branch position:** one constant resolves *all* forks in M consistently — each fork's mm_ partitions the same member set, k routes through every fork with no per-fork plumbing. And it's restructure-stable: when M merges again or a member joins an existing branch, only M's emitted tests change; every site keeps passing the same k. Branch positions would cascade site rewrites on each restructure.

**Gates stay caller-side:** `if gv_(v,i): M(f.nF, ...)` — the caller's gate decides whether the call fires; k decides which body runs inside. Per-caller gate diversity is thereby handled with zero transmission. The complementary-branch case computes k inline: `M(f.nF if gv_(v,i) else g.nF, ...)`.

**Rejected alternatives:** runtime caller lookup via CoF.get() — fails exactly the conjunctive case above, and couples dispatch to tracing. Passing the gate value into M — relocates the gate, making every call unconditional; destroys the gate's saving.

**Zero-rewrite variant** if you want emission before touching sites: redefine each member as a stub `def f(...): return M(f.nF, ...)`. No site edits, no iF_ changes, old names stay valid for un-recompiled code; cost is one extra frame per call. Transitional only — compression realizes when sites rewrite and stubs drop.

kwcckw: > the inherited gate value is not singular, evely caller of the merged fork oF may have a different one. so a caller should call new oF and pass through the address of the fork derived from the old oF?

I see, this is closer to the first method, where we need the site information. From LLM above, we only need k (site info) to resolve all the forks' routing, caller triggers the gate, and forks information should be still within the forks?

> k = registry index of the former callee, the stable key.

So we still need oF_site_ to pack this info? Or actually this could be mapped with nF_?

frame_2D_alg/agg_recursion.py
kwcckw: We could have borrow d val in sumV when we need the combined val?

boris-kz: Combined val is m

frame_2D_alg/meta_code.py
boris-kz: The forks should be assigned their old oF, which should also be passed by the caller, so the fork is called when they match

kwcckw: I see, but the matching process should be done within the fork parsed by the caller right? The caller didn't decide on which fork right? It's only used to parse the site info.

frame_2D_alg/agg_recursion.py
kwcckw: Okay, then we actually need to revert all of this. From the comment in vt_, the borrow should be valid in the boundary Bt only.

PR #196, opened 2026-07-05
7/5/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -368,17 +370,47 @@ def merge_oF(F,f, fsel=1):  # combine aligned ops, if-fork per miss, no inline r
                 C+=get_fc(sub); add_+=[sub]
             body += [Sub]
         else:  # default
-            if comp_prim(Sub,sub): body += [Sub]  # add count?
-            else: body += [(ast.IfExp,(Sub,sub))]; C += get_fc(sub); add_+=[sub]
+            if is_fork(Sub):  # Sub is prior added IfExp ast fork
+                if   comp_prim(Sub[1][0],sub): body += [(ast.IfExp, Sub[1], Sub[-1]|{f})]  # sub same as if branch, update the o_ with f
+                elif comp_prim(Sub[1][1],sub): body += [Sub]  # sub is the same as else branch, nothing to assign here
+                else: body += [(ast.IfExp,(Sub,sub),o_)]; C += get_fc(sub); add_+=[sub]  # nest in new fork
```
kwcckw, line 376: > new fork: ('fork', (Sub,sub), (F_mm, f_mm)) where mm = leaf members (nT.N_ flattened; leaf oF is its own member);

So i think we only need `F_mm`? Because we only need `F_mm` to route the if branch, and the rest should be in else branch?

frame_2D_alg/meta_code.py
```diff
@@ -368,17 +370,47 @@ def merge_oF(F,f, fsel=1):  # combine aligned ops, if-fork per miss, no inline r
...
+            elif comp_prim(Sub,sub): body += [Sub]
+            else: body += [(ast.IfExp,(Sub,sub),o_)]; C += get_fc(sub); add_+=[sub]  # new fork
             # new fork
-    if len(f.body) > len(F.body): offs = f.body[i+1:]; body+=offs; add_+=offs
+    if len(f.body) > len(F.body): offs = f.body[i+1:]; body+=offs; add_+=offs  # offsets should be added to a new fork?
```
kwcckw, line 380: When both Fs body has different length, and if `f` has longer body, instead of concatenating the longer f's body into F, we should add new IfExp fork on the offset here?

boris-kz: It's as with any diffs, if the diff is expensive we add a conditional fork, else a simple concat

frame_2D_alg/meta_code.py
```diff
@@ -353,6 +354,7 @@ def merge_oF(F,f, fsel=1):  # combine aligned ops, if-fork per miss, no inline r
 
     body,add_ = [],[]  # replace F.body: op sequence in func
     C = 0  # compression
+    o_ = set(F.N_) or {F}   # old oFs (or the whitelist of the if branch)
```
boris-kz, line 357: this is sub_, but why is it a set? and these sub_Fs should really be assigned per fork, else it's pointless?

frame_2D_alg/meta_code.py
boris-kz: branching is symmetrical, doesn't matter which component oF came first

frame_2D_alg/agg_recursion.py
```diff
@@ -868,7 +868,7 @@ def fill_frame(_iy,_ix, elev, T):  # expand level_frame from pixel-level seed ti
             else: break
         if T_:
             TT,C,R = sum_vt(T_, wTT=ttFrm); R += elev
-            if vt_(TT*ttFrm)[0] * ((C+wFrm)/(R+cFrm) > ave):
+            if vt_(TT*ttFrm)[0] * (C+wFrm)/(R+cFrm) > ave:
```
boris-kz, line 871: ((C+wFrm)/(R+cFrm))

also, dval / total in v_ is wrong, that makes it stronger when the match is weak, while it should be borrowing from match

claude — ad_/t_ is a miss share, not a value: it peaks exactly where m vanishes, so vt_ currently rewards noise. The /t_ normalization itself is fine as rdn accounting; the error is using the miss share directly as d value. Value needs a lender term:

```python
m = (m_/t_) @ wTT[0]
d = (ad_/t_) @ wTT[1] * max(m, 0)   # d borrows from co-projected m
```

Scalar borrow, not per-param product: proj_TT projects forks wholesale (TT[0]*dec, TT[1]*cos_d*dec), so the lender is the pattern's overall projectability, and per-param m_/t_ * ad_/t_ leaks sign from deviation-based params. Clip at 0: negative match lends nothing. Behavioral consequence worth noting: since t = m+|d|, borrowed d peaks at half-match (rel m = rel d = 1/2) and vanishes at both extremes — no value in pure noise, none in pure repetition. That shape is correct for "informative change in a projectable context."

The unresolved contract is the lender's identity, and it splits by channel:

1. der+: d of matched links borrows from those links' own m — local lender, vt_'s default above is right.
2. Contrast (B_ selection, Bt): boundary d is high-d low-m by construction; borrowing from its own m near-zeroes it. Its lender is the core it delimits — which is exactly what Bt.brrw already does (rroot.m * decay * span ratio). So vt_'s old comment "borrow in G.Bt only?" is inverted: borrow everywhere except Bt's own d, which borrows externally. Concretely: B_ selection in cluster_N becomes d * rM(core N_) > avd, and Bt gate in sum2G uses brrw, not locally-borrowed bd. Simplest plumbing: vt_(TT, wTT, lm=None) — default lender own m, contrast sites pass core rel m.
3. ffeedback: ave,avd = vt_(aTT) must use unborrowed d. Filters are cost/expectation stats; borrowed avd would shrink when global match is weak, lowering the d threshold in noise — inverted incentive. Pass lm=1 there (raw miss share), same for the inline vt_ clone at the filter-update site.
4. sumV from last turn collapses into this: M + |D|*rM is the extensive twin of the same rule; one borrow definition, two scales.
5. proj_TT cert: magnitude-like, raw d defensible; either is consistent, low stakes.

No further contract changes: F.m/F.d fields, sum_vt fm branch, and all gv_ gates inherit the corrected d through vt_ without semantic edits.

kwcckw: I see, so `d = (ad_/t_) @ wTT[1] * max(m, 0) ` scales the value in the same way, but it doesn't change the sign.

> Contrast (B_ selection, Bt):

I think this is correct? It's already has high d in Bt, so borrowing using m will scale the values down?

frame_2D_alg/meta_code.py
kwcckw: > branching is symmetrical, doesn't matter which component oF came first
> this is sub_, but why is it a set? and these sub_Fs should really be assigned per fork, else it's pointless?

Yes thanks. I misunderstood earlier. `F.N_`'s leaf can't contain duplicated leaf, so using set is redundant here.

Also I was thinking we only have a single F and f here, so F.N_ and f.N_ will be the same across all subs, so `F_mm` and `f_mm` only need to be retrieved once per F and f pair with `get_mm`? Even it's per fork, their leaf will be the same?

    def get_mm(F):  # leaf members via N_ flattening
        return {m for n in F.N_ for m in get_mm(n)} if F.N_ else {F}

frame_2D_alg/meta_code.py
kwcckw: So we can use something similar with ` oF.fw * sum(get_fc(p) for p in t[1]) > ave` in the split function to eval for the new fork on those offsets body?

frame_2D_alg/agg_recursion.py
boris-kz:
```
def val_(TT, wTT=wTT, fd=0):  # multi-variate rel match for membership, rel diff for boundary

    m_, d_ = TT; ad_ = np.abs(d_); t_ = eps_(m_+ad_)  # ~ max comparand
    vm = (m_/t_) @ wTT[0]  # /= total = rdn
    if fd: 
        vd = (ad_ @ wTT[1] - avd) * ((vm+ave)/2)  # borrow from mean(known cis_vm, unknown trans_vm)?
        return vm,vd
    else: return vm

```
Structure is right (m-only default, borrow inside the d branch, ave as the unknown-trans stand-in — commensurable since ffeedback sets filters via this same function's units). Four problems:

1. **Fork normalization asymmetry.** vm is relative (m_/t_), but vd uses raw ad_ — extensive. Downstream sites add or co-gate them (sum2G: lm+ld; CF.m/CF.d), so mixed units corrupt those gates. Restore ad_/t_; extensive scaling stays at call sites via c, as before.

2. **avd inside val_ double-subtracts.** Every consumer already subtracts a redundancy-scaled threshold: B_ selection `d > avd*(_r-1)`, Bt gate `bd*wAgg > avd*(br+cAgg*L)`. Baking -avd in at unit r both double-counts and freezes the r-scaling. Also ordering: (D-avd)*rate scales the deficit by the lender, so sub-threshold contrast in high-m context reads as less negative — threshold is cost, cost doesn't borrow. Lead: val_ returns borrowed value only; gates keep avd*r. Alternative — keep -avd inside and strip it from all gates — touches more sites for no gain.

3. **Negative lender.** vm < -ave flips vd sign: strong anti-match context making contrast negatively valued. Anti-match doesn't project, so it lends nothing: clip, (max(vm,0)+ave)/2.

4. **Hard-coded boundary lender breaks der+.** (vm+ave)/2 assumes one unknown side. In der+ both comparands are known; its d should borrow from own link vm, and averaging with ave under-borrows at high vm. One formula covers both via a lender arg:

```python
def val_(TT, wTT=wTT, fd=0, lm=None):  # rel match for membership; borrowed rel diff per lender

    vm = (m_/t_) @ wTT[0]
    if fd:
        vd = (ad_/t_ @ wTT[1]) * (max(vm,0) + (ave if lm is None else lm)) /2  # lm: trans default ave; der+ passes vm -> rate=vm
        return vm, vd
    return vm
```

lm=vm collapses the mean to vm itself — der+ and boundary unify without a mode flag.

Closure: (a) ffeedback can't set avd from fd=1 output — borrowed vd defining avd is circular; it needs the raw rel d (inline ad_/t_ @ w, or lm=1 convention). (b) Refinability stands: once trans_cluster forms the adjacent core, pass its vm as lm, replacing ave. (c) Extensive twin for cluster gates: M + |D|*rate, same rate — sumV's replacement, unchanged from before. No other contracts move.

frame_2D_alg/meta_code.py
boris-kz: F_mm and f_mm are from LLM, it's a horrible notation, I don't even know what that means

It's a single F and f per pair, but merge is across many pairs

frame_2D_alg/meta_code.py
boris-kz: right

frame_2D_alg/agg_recursion.py
boris-kz: 1, 2 — retracted, and your avd definition is what dissolves both. Consistency is per-fork, not cross-fork: vm pairs with ave in normalized units, vd pairs with avd in raw ad_@wTT[1] units, and since ffeedback derives each filter from its own fork's stream, units track automatically. Combined gates decompose as (vm-ave)+(d-avd) — two consistent pairs, which is why (ave+avd) thresholds were never the mixed-unit problem I claimed. And normalizing d by t_ would be worse than inconsistent: ad_/t_ is ~complementary to m_/t_ per param, so normalized vd mostly duplicates vm and loses the extensive magnitude that borrowing is supposed to discount. Raw ad_ with internal avd subtraction is right.

3 — retracted. Clipping assumed anti-match doesn't project. It does: it predicts continued mismatch, so a negative cis lend legitimately debits the ave prior. Sign flips only when anti-match exceeds expected match, and the resulting negative vd is functional — gv_ accumulates it as reform signal. Clipping censors that.

4 — retracted; I misidentified the lender. It isn't the other comparand (known in der+), it's the projected match of the d's future realization — der+ yield or trans-side recurrence — unknown in both channels. (vm+ave)/2 is shrinkage of that projection toward the prior; one estimator, no mode flag.

Closure, four contracts:

- ffeedback: avd = average of ad_@wTT[1] computed inline, not from the fd branch (its output is post-subtraction, post-borrow — circular). ave from the fd=0 stream as now.
- Redundancy on the d threshold: avd inside val_ is at unit r, gates currently scale it (avd*(_r-1), avd*(br+...)). A residual avd*(r-1) subtracted outside would be un-borrowed while the unit part is borrowed — inconsistent. Pass r in: (ad_@wTT[1] - avd*r) * (vm+ave)/2.
- Gates consuming fd=1 vd drop avd from their threshold side; ave and complexity costs stay.
- Bt: ave in the rate is the prior for the unknown adjacent match. Where the adjacent core is actually formed — rroot in sum2G, trans core after trans_cluster — substitute its m for ave: (vm + adj_m)/2. Bt.brrw then becomes this substitution rather than a second lend stacked on val_'s internal rate.

Extensive cluster gates keep the M + |D|*rate form with the same rate; nothing else moves.

frame_2D_alg/meta_code.py
kwcckw: Yes, merge_oF is per pair, and all of pair's subs should have the same `o_` (the whitelist of IfExp) since it's originated from the same F and f?

boris-kz: merge is just one component of clust_oF_, over multiple components

frame_2D_alg/agg_recursion.py
kwcckw: >  (vm+ave)/2 is shrinkage of that projection toward the prior; one estimator,

So if `*(vm+ave)/2` is the shrinkage of the projection towards the prior, why we need +ave/2 (averaging with ave) when we could have just use `vm`? 

      def val_(TT, wTT=wTT, fd=0):  # multi-variate rel match for membership, rel diff for boundary
      
          m_, d_ = TT; ad_ = np.abs(d_); t_ = eps_(m_+ad_)  # ~ max comparand
          vm = (m_/t_) @ wTT[0]  # /= total = rdn
          if fd: 
              vd = (ad_ @ wTT[1] - avd) * ((vm+ave)/2)  # borrow from mean(known cis_vm, unknown trans_vm)?
              return vm,vd
          else: return vm

boris-kz: (vm+ave)/2
is mean (internal + external) m that vd is borrowing from 

ave = decay = .3; avd = 20  # mean sum( abs(dTT[1]) * wTT[1]), the borrower

kwcckw: We usually have `ave` as the cost part, but here it's summed with vm as external val, which is a special case? This is actually not quite consistent with rest of the code?

boris-kz: works the same, it's an opportunity cost
but it may be just vm/2, not sure

boris-kz: Yes — vm/2 is better, and the /2 isn't arbitrary. The rate is the mean of two lenders, (cis_vm + trans_vm)/2, and the question is what fills the trans slot before measurement. ave presumes the unknown side lends at prior odds unconditionally. But a lend needs a projected relation to travel through, and in CogAlg only match projects: at vm=0 nothing projects this d's context across the boundary (or forward to der+ realization), so generic expected match elsewhere is unreachable — the trans lend is mediated by cis match, ~vm*ave, not free-standing ave. Folding that in gives vm(1+ave)/2, and any constant factor is absorbable into wTT[1] by feedback. The only non-absorbable difference between the two forms is the asymptote at vm→0: ave/2 keeps lending in zero-match context — the same weak-match-preserves-d inversion this thread has been stripping out — while vm/2 goes to zero. That's the discriminant, and it favors vm/2.

So: trans slot initialized 0, substituted with measured m when the trans core forms — rate = (vm + trans_m)/2, same functional form before and after refinement, which (vm+ave)/2 doesn't give.

Closure:

- val_: rate = (vm + lm)/2, lm default 0; keep lm as the substitution hook. der+ needs no special case — its future-yield prior is likewise cis-mediated, so vm/2 uniformly.
- Bt: cis lender remains core m, so rate = core_m/2 pre-trans, (core_m + trans_m)/2 post. Bt.brrw's decay*span factor survives as projection decay applied to the substituted lm.
- No clipping, as settled: negative vm/2 makes vd negative for supra-threshold d in anti-match context — functional as gv_ reform signal.
- Bootstrap check: B_ selection stays raw d > avd*r, so halved pre-trans rates delay Bt evaluation, not boundary collection; recoverable via avd feedback.
- ffeedback unaffected: ave from fd=0 stream, avd from raw ad_ stream, both pre-rate.
- Extensive twin M + |D|*rate inherits the same rate. Nothing else moves.

frame_2D_alg/meta_code.py
kwcckw: Yes, and when we merge 2 oFs' body, IfExp fork is created when the body node is different. If we need to assign per fork's sub_Fs, it will be something like `(ast.IfExp,(Sub,sub),(SubF, subF)`

But this SubF will be the F or F.N_ (Or the leaf)? Because Sub or sub may not be an oF, and they only can be referenced via the parent F or f. This reference will be used to route the If or else fork later.

frame_2D_alg/agg_recursion.py
kwcckw: I see, thanks for clarifying.

frame_2D_alg/meta_code.py
```diff
@@ -284,12 +286,11 @@ def clust_oF_():  # pairwise merge for body compression
     while v_:
         (_T,T), v = max(v_.items(), key=lambda kv: kv[1])
         if v*wL <= ave: break  # no pair worth realizing
-        nT = CoF(N_=[_T,T], typ=_T.typ, fc=_T.fc+T.fc, caller_=_T.caller_|T.caller_)
+        nT = CoF(N_=[_T], typ=_T.typ, fc=_T.fc, caller_=copy(_T.caller_))
         nT.body = copy(_T.body)  # body will be merged below
-        merge_oF(nT, T, fsel=0)
+        merge_oF(nT, T, fsel=0); _T.N_ += [T]; _T.fc += T.fc; _T.caller_ |= T.caller_  # or move this to merge_oF too?
```
boris-kz, line 291: Just F.N_ += [f], fc and caller_ should be recomputed?

frame_2D_alg/agg_recursion.py
kwcckw: > Yes — vm/2 is better, and the /2 isn't arbitrary. The rate is the mean of two lenders, (cis_vm + trans_vm)/2, and the question is what fills the trans slot before measurement. ave presumes the unknown side lends at prior odds unconditionally. But a lend needs a projected relation to travel through, and in CogAlg only match projects: at vm=0 nothing projects this d's context across the boundary (or forward to der+ realization), so generic expected match elsewhere is unreachable — the trans lend is mediated by cis match, ~vm*ave, not free-standing ave. Folding that in gives vm(1+ave)/2, and any constant factor is absorbable into wTT[1] by feedback. The only non-absorbable difference between the two forms is the asymptote at vm→0: ave/2 keeps lending in zero-match context — the same weak-match-preserves-d inversion this thread has been stripping out — while vm/2 goes to zero. That's the discriminant, and it favors vm/2.
> 
> So: trans slot initialized 0, substituted with measured m when the trans core forms — rate = (vm + trans_m)/2, same functional form before and after refinement, which (vm+ave)/2 doesn't give.
> 
> Closure:
> 
>     * val_: rate = (vm + lm)/2, lm default 0; keep lm as the substitution hook. der+ needs no special case — its future-yield prior is likewise cis-mediated, so vm/2 uniformly.
> 
>     * Bt: cis lender remains core m, so rate = core_m/2 pre-trans, (core_m + trans_m)/2 post. Bt.brrw's decay*span factor survives as projection decay applied to the substituted lm.
> 
>     * No clipping, as settled: negative vm/2 makes vd negative for supra-threshold d in anti-match context — functional as gv_ reform signal.
> 
>     * Bootstrap check: B_ selection stays raw d > avd*r, so halved pre-trans rates delay Bt evaluation, not boundary collection; recoverable via avd feedback.
> 
>     * ffeedback unaffected: ave from fd=0 stream, avd from raw ad_ stream, both pre-rate.
> 
>     * Extensive twin M + |D|*rate inherits the same rate. Nothing else moves.

From above, the trans m should be optional , and parsed from the dfork's or L's m if it's available?

frame_2D_alg/meta_code.py
kwcckw: Thanks, there'a a typo above. 

It should be nT.N_ += [T] (nT instead of `_T`). For fc, yes, i think it should be recomputed from the new nT.body since the merge doesn't concatenating the body now. For caller_, i think it should be union of both? Both oF's callers are still valid after we merge them?

Also i think this should be done right after the merge_oF and we shouldn't do it in batch, because we might merge this nT again in the next while loop.

boris-kz: it's like call_, old callers are not relevant?

frame_2D_alg/agg_recursion.py
boris-kz: I don't think we need it

frame_2D_alg/meta_code.py
kwcckw: Okay right, the old callers are in nT.N_'s caller_ anyway, and nT's caller_ should be filled in the next run.

frame_2D_alg/meta_code.py
boris-kz: just the original sub_F per fork, for a handshake

PR #197, opened 2026-07-06
7/6/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -338,13 +339,19 @@ def merge_oF(F,f, fsel=1):  # combine aligned ops, if-fork per miss, no inline r
         # not revised, need to assigned _F per fork:
         else:  # default
             if is_fork(Sub):  # Sub is prior added IfExp ast fork
-                if   comp_prim(Sub[1][0],sub): body += [(ast.IfExp, Sub[1], Sub[-1]|{f})]  # sub same as if branch, update the o_ with f
+                if   comp_prim(Sub[1][0],sub): body += [(ast.IfExp, Sub[1], (*Sub[2], f))]  # sub same as if branc, add new f into the F and f pair
```
kwcckw, line 342: > It's a single F and f per pair, but merge is across many pairs

I think it could be more than F and f when we merge IfExp with another body again?

It will be more than a pair of F and f?

boris-kz: that's exactly what I said

frame_2D_alg/agg_recursion.py
```diff
@@ -555,7 +555,7 @@ def sum2G(ft_, fTT, root=None, init=1):  # core clustering function
         if len(ft_)>1: L_,_,ltt,lc,lr=ft_[1]; L_+=root.L_; ltt+=root.Lt.dTT; lc+=root.Lt.c; lr+=root.Lt.r; ft_[1]=L_,_,ltt,lc,lr
     Ft_ = []
     for ft, nF in zip_longest(ft_,('Nt','Lt','Bt')):
-        if ft: n_,_,tt,c,r = ft; Ft_+= [CF(N_=n_,nF=nF,dTT=tt,m=(vt:=val_(tt,wTT,1)),d=vt[1],c=c,r=r)]
+        if ft: n_,_,tt,c,r = ft; Ft_+= [CF(N_=n_,nF=nF,dTT=tt,m=(vt:=val_(tt,wTT,1))[0],d=vt[1],c=c,r=r)]
```
boris-kz, line 558: thanks

frame_2D_alg/meta_code.py
kwcckw: Thanks for the clarification.

frame_2D_alg/meta_code.py
boris-kz, line 321: so this is destructive split or merge of whole oFs, just to make existing process more efficient.
but eventually we need to generate incrementally complex hierarchical oFs.
that means non-destructive and partial split or merge: oFs may remove their code blocs, or encapsulate them for selective recursion, and these code blocs may be borrowed from other oFs.

frame_2D_alg/meta_code.py
boris-kz: f should be assigned to the new fork, not the whole F.body
something like this:

            if is_fork(Sub):  # Sub is previously added IfExp
                if comp_prim(Sub[1][0],sub):  # this should new sub-fork, if incomplete match:
                    Sub[1][1] += [sub]  # f should be assigned to the fork, not the whole F.body

boris-kz: Lead: don't label items in slot 1 — repurpose slot 2 as a parallel source list. Slot 2 `(F,f)` is currently write-only (nothing reads `Sub[2]` except your commented line), so it's free to restructure, and every existing consumer of slot 1 — `comp_prim` head check, `comp_body` elementwise zip over `n[1]`, `get_fc` sum over `n[1]` — stays untouched. Pairing f into the items themselves (`(f,alt)` wraps) breaks all three with unwrapping.

Fork record:

```python
(ast.IfExp, alt_, src_)  # src_[j] = set of fs that contributed alt_[j]
```

Creation (347): `body += [(ast.IfExp, (Sub,sub), ({*F.N_}, {f}))]` — alt0's label is the members merged so far, which is correct in clust_oF_ because N_ grows per merge; in sum2O N_ is preset, so set it incrementally there or pass accumulated sources.

Match to alternative j (replaces 341-344):

```python
for j,alt in enumerate(Sub[1]):
    if comp_prim(alt,sub):
        if isinstance(alt,CoF): add2O(alt,sub)
        body += [(ast.IfExp, Sub[1], (*Sub[2][:j], Sub[2][j]|{f}, *Sub[2][j+1:]))]; fin=1; break
if not fin:
    body += [(ast.IfExp, (*Sub[1],sub), (*Sub[2],{f}))]; C += get_fc(sub); add_ += [sub]
```

Rebuild the tuple, never mutate in place: forks are shared with `_T.body` via the shallow copy at 293, and the merge can still be rejected at 297. Immutable inner tuples make this hazard structural rather than disciplinary.

Two boundaries to keep clean:

- Don't overload `CoF.caller_` for this. caller_ is call-graph provenance (who invokes the oF); src_ is merge provenance (which f diverged at this position). Different relations — a CoF alternative gets f in the fork's src_[j], not in its own caller_.
- Unforked matches (346) need no label: no deviation, provenance is implicitly all sources. Labels exist only where branches do. Tail If at 352 takes one label for the whole segment: `(ast.If, tuple(offs), {f})`.

frame_2D_alg/meta_code.py
kwcckw: > means non-destructive and partial split or merge: oFs may remove their code blocs, or encapsulate them for selective recursion, and these code blocs may be borrowed from other oFs.

But under what condition we need the complex hierarchical oFs? That should be conditional?

Actually the existing split only encapsulating the block of code under the gate, for removal, it will totally remove the code block operation from the oF?

For the code borrowing, it should be something like a pointer?

frame_2D_alg/meta_code.py
kwcckw: > f should be assigned to the new fork, not the whole F.body something like this:
> 
> ```
>         if is_fork(Sub):  # Sub is previously added IfExp
>             if comp_prim(Sub[1][0],sub):  # this should new sub-fork, if incomplete match:
>                 Sub[1][1] += [sub]  # f should be assigned to the fork, not the whole F.body
> ```

Okay, i was thinking if both Sub[1][0] and sub are the same (positive comp_prim), we should only need 1 or them.

> (ast.IfExp, alt_, src_)  # src_[j] = set of fs that contributed alt_[j]

Instead of separating alt and src, we can use a dict in `alt_` slot so that each key is source, while value is alt_?

frame_2D_alg/meta_code.py
boris-kz: everything is conditional, based on oF.w
borrowing would incorporate the block, refs or not doesn't really matter

Lead read: 
this converges meta_code onto the agg architecture itself, and the destructive/non-destructive question resolves into your standard clustering order. Blocks (gate-led segments, the unit `split` already wraps in CoF) become the elementary nodes; oFs become compositions of block refs. Then: split = connectivity clustering within an oF (gates and dataflow are the native links), borrowing = a block belonging to multiple oFs = overlapping membership, which by your own definitions is only legal in centroid clustering. So cross-oF block sharing is centroid formation over the block pool, connectivity-first within each oF — same order as agg. Non-destructive stops being a design choice: centroids retain exemplars, and hierarchical levels require lower nodes to persist.

Implication chain:

1. **Unit shift.** Blocks get their own fc, fw, dTT. The container exists (split's CoF); the contract change is per-block value accumulation — gv_/gV_ already index gate sites, so dTT attaches there, not only at callee call sites.

2. **Borrowing forces namespace lifting.** A shared block must be a single module-level traced def with free vars promoted to params. The comparand therefore acquires an interface: def/use signature. Block comp keys on signature + structure, not nF identity (comp_prim currently requires same nF for CoF pairs). This is the largest new contract — without signature alignment, borrowed code isn't runnable, and merge validity becomes checkable rather than assumed-by-typ.

3. **Comparison pool changes.** Cross-comp moves from oF pairs (comp_body over whole bodies) to block pairs over the global block pool; borrowing candidates come from there. oF pairs are then compared coarsely by block-ref overlap — comp_callers-style set match — which is far cheaper than zip-aligned op comparison.

4. **Site map generalizes.** Sites are now of two kinds: Call nodes (borrowed lifted blocks) and inline gate sites (encapsulated blocks left in place). Invertibility: block tuples keep the original inline AST ref, so encapsulation is reversible — re-inline when caller_ drops to one and gate val stays high, removing call overhead.

5. **Non-destructive → two filters.** Dispatch ave selects composite vs member by projected val; retention ave purges exemplars whose residual val no longer covers storage. The functional payoff is re-split: when a member's measured deviation from its composite exceeds avd after feedback, it's re-promoted — only possible because merge retained it. nT.N_ already retains members; the change is they stay addressable in the hierarchy instead of leaving oF_.

6. **Hierarchy.** oF_ → oF_H, level per reform cycle — frame_H analog. Dispatch walks down: caller context selects oF, gates select blocks, blocks recurse selectively. Encapsulation-for-selective-recursion is your existing gv_ mechanism; the only extension is the gated unit may be shared.

7. **Partial merge comes nearly free.** With blocks factored, merge_oF's per-op zip alignment reduces to ref-set intersection plus residual op merge; IfExp forking moves from op level to block-ref level, surviving at op level only inside blocks. Coarser, cheaper, less brittle.

8. **Block removal.** Gate val persistently below ave → drop the ref. The block persists while caller_ is nonempty; garbage-collect when empty and retention fails. Refcounting is caller_, already present. Feedback needs indexing changes only: per-block dTT into ffeedback, inline avd derivation unchanged.

Discriminants:

- **Lifting vs aliasing** (referencing into a host oF's body): aliasing entangles borrowers with host split/merge and breaks single-def invertibility. Lift, unless single borrower.
- **Block granularity**: gate-led segments vs arbitrary op subsequences. Gates are the value-tested, data-native boundaries — connectivity-first. Arbitrary subsequences require alignment search with no native link. Keep gate-led.
- **One filter vs two**: a single ave forces the current purge-on-merge. Re-split under filter drift is the whole reason for non-destructive, so the retention filter follows necessarily.

Closure: remaining items (exec/compile of lifted defs, tracing shared blocks, F_call_T_ reindexing) are bookkeeping downstream of 2–4. No independent contract change left.

frame_2D_alg/meta_code.py
boris-kz: yes this is really elif fork val_
I guess we can use dict, that's in Sub[1][1] ?

boris-kz: Yes, with one discriminant that decides viability: key direction must be `{alt: src_set}`, not `{f: alt}`. Every downstream consumer iterates slot 1 expecting alternatives — `get_fc` sums `n[1]`, `comp_body` zips `_n[1], n[1]`, both iterate dict keys. Alt-keyed: they keep working unchanged (insertion order preserves the positional zip). Source-keyed: get_fc sums source fc, comp_body compares sources — silently wrong.

Record: `(ast.IfExp, {alt: src_set})`. Creation at 347: `(ast.IfExp, {Sub: {*F.N_}, sub: {f}})`. Match: `{**Sub[1], alt: Sub[1][alt]|{f}}`; miss: `{**Sub[1], sub: {f}}`. The gain over parallel lists is real but modest: no positional coupling to maintain. Note the dict never does keyed lookup — matching stays a linear comp_prim scan, since match is semantic, not identity. It's paired storage, not an index.

Implication chain:

- **Hashability**: alt keys are CoF (identity-hash), AST-record tuples, or leaves — all hashable now. The invariant this imposes: an alternative can never internally contain a fork record (dict inside tuple → unhashable). Holds structurally — merge_oF forks only at body positions, sub-level forks live inside CoF bodies (identity-hashed, dict buried) — but it's an invariant the tuple form didn't need. Corollary worth having anyway: key equality is object identity for AST nodes, so structurally-similar alts from different fs never silently collapse; comp_prim remains the sole match criterion.
- **Rebuild discipline persists**: dicts invite in-place update, but the shared-structure hazard from the shallow copy at 293 is unchanged — rejected merge at 297 must leave `_T.body` intact. Always `{**old, ...}`.
- **Same hazard, deeper**: add2O into a matched CoF alternative mutates an object shared with `_T.body` — a rejected merge still leaves absorbed stats. Either defer stat absorption until cmpr passes (collect (alt,sub) pairs, apply on acceptance) or accept the drift under the lossy contract. Independent of dict-vs-lists, but the dict change is the moment to decide it.
- **split_oF_ ordering becomes a contract**: `split` rebuilds `(t[0], tuple(split(s) for s in t[1]))` and `is_gate_l` indexes `t[1][0]` — over a dict that mangles or throws. No call sites exist yet, so pin it: split runs only on pre-merge bodies, or gets an `is_fork` guard.
- **Tail fork at 352**: same shape for uniformity — `(ast.If, {tuple(offs): {f}})` — so get_fc/comp_body need no second fork format.

Unresolved either way: comp_body's elementwise zip across two forks compares alternatives by insertion order, which is noise — alternatives are unordered. Best-match pairing is the correct estimate; dict neither helps nor hurts there.

frame_2D_alg/meta_code.py
kwcckw: I see, and could you clarify on the oF.w and oF.fw?

oF.fw is process weight, and this oF.w is the gain from the internal val and gate val, added only from ffeedback:

    for oF in oF_:
        if oF.V_: oF.w = sum(oF.V_) + sum(oF.gV_)  # all extensive

So all the conditional eval of split and merge should be based on oF.w only?

boris-kz: I think it's just w, we don't really need separate fw at this point?

frame_2D_alg/meta_code.py
kwcckw: > I guess we can use dict, that's in Sub[1][1] ?

I think it's in Sub[0][0] and Sub[1][1]? 

     (ast.IfExp, [ {Sub: [F]}, {sub: [f]} ]  )

Because we still need to differentiate both forks ops and their source.

frame_2D_alg/meta_code.py
kwcckw: Right now `fw` is added as gain:

      def add2F(F, n, merge=0, fr=0, fo=0):  # unpack for batching in sum2F
      
          if F.c:
          else:
              
I'm not sure if this is overlapping with the w, and it is added from V_ (added from FV_):

    def FV_(F, tt,c,r):  # combine m,d per oF
        tF = oF_[F.nF]; tF.c += c
        V = val_(tt*tF.wTT) * (c/r); tF.V_ += [V]; return V

    for oF in oF_:
        if oF.V_: oF.w = sum(oF.V_) + sum(oF.gV_)  # all extensive

frame_2D_alg/meta_code.py
boris-kz: that's redundant, I think the forks should be sequence of ifExps, each with f in [0] and ops in [1]?

frame_2D_alg/meta_code.py
boris-kz: separate fw would probably be from code compression which is just oF.V_ now, may add more later

frame_2D_alg/meta_code.py
boris-kz: Agreed — per-source entries are the native form once forks are pure deviation records: each f either matches base (no record) or contributes one (f, ops) entry. The dict's alt-keying and src_sets were runtime-dispatch thinking; as provenance for feedback, source-keyed singletons are right, default unlabeled.

Two placement constraints before it closes:

**Entries must nest inside one record per body position, not inline as body items.** A sequence of sibling IfExps in body breaks `zip(F.body, f.body)` — positional alignment is the basis of the whole merge. So:

```python
(ast.IfExp, (Sub, (f1,ops1), (f2,ops2)...))  # ent_[0] = base, unlabeled default
```

**f in t[1][0] of each entry pollutes the two consumers that iterate t[1].** `get_fc` would add f.fc — the source function's entire cost — into fork cost; `comp_body`'s elementwise zip would compare f-labels as content via the CoF branch. Either move f to per-entry slot 2, `(ops, f)` reversed or 3-slot, where consumers don't look (why old slot-2 sources were free), or special-case is_fork in both. One of these is forced; f-as-content is the only real defect in your layout.

Merge step then simplifies past the dict version: `comp_prim(ent_[0], sub)` match → keep record; miss → rebuild `(ast.IfExp, (*Sub[1], (f,sub)))`. No scanning of deviant entries at all — a later f whose ops resemble f1's just appends its own entry. That redundancy is fine and even useful: cross-entry matching is centroid formation over ent_, a later clust pass over fork entries as comparanda, not micro-merging inside merge_oF. Lines 341-344's alternative-matching logic and 345's nesting both disappear.

Unchanged contracts: rebuild-not-mutate (shallow copy at 293, rejection at 297); creation-time base needs no F.N_ attribution — that problem dissolves with the unlabeled default; tail at 352 takes the same shape, `(ast.If, ((f, offs),))`.

Enabled but not required: comp_body across two merged composites can align fork entries by source identity — same f deviating in both is a strong match — replacing positional zip noise.

boris-kz: remove leading ifExp:

The IfExp head is only a type discriminant so consumers dispatching on t[0] can tell fork records from op records. Nothing lowers these to AST anymore, so IfExp is vestigial — any sentinel works. Sub is inside because the record *replaces* the body item, and the position must still expose the base op for zip-alignment. Drop Sub only if base stays as the body item — but then entries become sibling items or a parallel per-position store, and every body walker (zip in merge, comp_body, get_fc) needs filtering plus filtered→raw index mapping to insert new entries adjacent to their base. Containment buys locality.

Lead: extend the base record instead of wrapping it — `(head, subs)` → `(head, subs, ent_)`, ent_ = ((f,sub),...). No new head, no wrapper, and both objections dissolve:

- comp_prim untouched: a deviated position matches incoming subs natively via head equality — the is_fork exclusion disappears entirely, which also kills the dead-matching problem at root
- comp_body untouched: zips n[1], entries invisible — estimates compare bases, correct since deviations are per-source noise
- get_fc: one line if entries count in merged fc — they should, cmpr = Σ individual fc − merged fc and carried deviations are merged cost
- merge: match → keep record as is, nothing to write; miss → `(head, subs, (*ent_,(f,sub)))`

Two edges. Leaf bases can't take a slot — but under the lossy contract a lone leaf's deviation rarely clears ave (your 274/351 gating pattern), so gate entry creation on fw*get_fc(sub) > ave and leaf deviations mostly discard rather than record; no wrapper needed. CoF bases go through the nF=='E' recursive path, deviations land inside sub-oF bodies — slot 2 only ever appears on record tuples.

The tail (349-354) is the one place a tagged record survives: no base item exists to extend, so it keeps its own `(ast.If, ((f,offs),))` shape — single special case instead of a universal wrapper.

frame_2D_alg/meta_code.py
boris-kz, line 323: try to unpack merge_oF, it's too opaque?

frame_2D_alg/meta_code.py
kwcckw: > that's redundant, I think the forks should be sequence of ifExps, each with f in [0] and ops in [1]?

Yes, this looks better. In this case, to determine the if or else branch, it will be based on the index? 


    (ast.IfExp, [ (f1,ops1) , (f2,ops2) ])? 

The f1 is if branch and f2 is else branch.

For the merged branch, if should be nested in ops?

frame_2D_alg/meta_code.py
kwcckw: Okay, and I thinkt that E or goF is no longer relevant? So it's actually could be simpler.

boris-kz: it's gv_ now, or wrapped in gf_ as claude suggested

frame_2D_alg/meta_code.py
boris-kz: I think it's just f:
body += [ast.IfExp(f,sub)]  # test by source only?

kwcckw: But then the original Sub is removed? When we merge both, we still need the Sub?

frame_2D_alg/meta_code.py
kwcckw: Okay, then the existing if section should be removed. That should be replaced with the if is_fork instead.

PR #198, opened 2026-07-07
7/7/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -318,71 +318,56 @@ def inject_oF_(oF_, g):  # inject AST in g, recompile g[name]
...
+                if comp_prim(_Sub,sub):  # check all Subs in if fork to match with sub
+                    # add eval for partial match and significant sub-fork? 
+                    Sub[1] += [(f,sub)]; felse= 1   # test by source only?
+                    break
+            if felse: Sub[2] += [(f,sub)]  # pack to else fork
+            body += [Sub]   
+        elif comp_prim(Sub, sub):
+             body += [Sub]  # match: position stays plain
+        # structure = (type, if_subs, else_subs). Each subs is [(source, nested sub),...]
+        else: body += [(ast.IfExp, [(F,Sub)], [(f,sub)])]  # miss: new fork: consensus branch + f branch
```
kwcckw, line 338: > I think it's just f:
body += [ast.IfExp(f,sub)] # test by source only?

I think we still need the Sub fork, so i added (F, sub) and the structure becomes:

     [(ast.IfExp, [(F,Sub)], [(f,sub)])]

When we merge new body it appends new (f,sub), something like this:

     [(ast.IfExp, [(F,Sub), (F2,Sub2)], [(f,sub),(f2,sub2)])]

Where Sub[1] is If branch and Sub[2] is else branch.

Also since we only merge body in merge_oF, i guess we can rename it to merge_Fbody?

boris-kz: It's just a concat of ifs, including initial fork. else is no op for all folks

kwcckw: Yes, but the concat should be done int the Sub instead of body?

This:

    body += [ast.IfExp(f,sub)]
    
Should be something like:

    Sub += [ast.IfExp(f,sub)]

boris-kz: No, in the body. Your initial Sub is just the first sub

Else is no op

kwcckw: Okay, this is appending the new IfExp fork into the body, but why it carries only the f and sub? The original Sub is still remained in the body without packing into the added IfExp fork?

The added IfExp fork should be a branch between Sub and sub? If we add IfExp fork with a single sub, it's actually a if fork, rather than a if-else fork?

boris-kz: The original Sub is just another sub, it should have no special status

kwcckw: I see, then it should be simpler and "sub" is actually the "if" for , and the rest is else?

boris-kz: There's no else. Actually there is no initial sub ether, it's just adding two forks at once with initial miss. So this initialization is a special case

frame_2D_alg/meta_code.py
boris-kz, line 288: this whole pairs is wrong, we need to sort individual oFs by overall match, then sort their matches and merge them staring from the best

boris-kz: Agreed — this is cluster_C form applied to oFs: exemplar-seeded centroid formation instead of chain agglomeration. Pairwise global-max builds order-dependent chains where composites re-align against composites, compounding alignment drift, and seeds can be peripheral. Seed-centric roots every fork branch in one exemplar body: composite = exemplar + deviations, which is also the only form that later supports fuzzy membership and re-split — chains aren't cleanly invertible to members.

Implication chain:

1. **Same comp cost, different topology.** Overall match per oF = sum of comp_body * min fc over same-typ pairs (value mass, not match count — plain match undervalues large-body compression). The O(n²) comparison stays; what changes is merge scheduling.

2. **Loop restructure.** v_ pair-dict → per-oF sorted match list + total. Outer loop: seeds by descending total; inner loop: absorb from seed's sorted matches into nT, validate each step by realized cmpr. The stale-pair pruning and the mrg_F_ live-composite cleanup both disappear — no chains, each composite born from one seed.

3. **Staleness.** After each absorption nT.body drifts, so remaining match estimates are stale. Recomputing against updated nT per step is exact but costly; keeping static order and gating by realized cmpr matches your existing estimate-prioritizes/realization-validates split. Lead: static order. Discriminant on failure handling: break-on-first-fail trusts an order that drift has degraded; since cmpr check is per-member anyway, skipping a failed member and continuing down the list costs one wasted merge attempt but recovers ordering error. With small pools, break is fine and matches your sorted-scan convention.

4. **Consumption.** Members marked fin, excluded from remaining pools. Next-seed totals should subtract matches to consumed members — an oF whose match mass went to an already-formed cluster either belonged there or failed its cmpr; its residual centrality is what's left. Cheap, since pairwise M is already computed. This is rdn discount to stronger clusters.

5. **Forward compatibility.** fin → membership weight w is the only change needed for fuzzy overlap later; the sorted-absorption structure carries over unchanged into the non-destructive hierarchy.

No other contract touched: merge_oF, cmpr gating, caller_ union, nF registration all unchanged.

kwcckw: > this whole pairs is wrong, we need to sort individual oFs by overall match, then sort their matches and merge them staring from the best

I see, so there will be a single strongest seed, merge with all other possible oFs, repeat until all pairs are checked?

> Same comp cost, different topology. Overall match per oF = sum of comp_body * min fc over same-typ pairs (value mass, not match count — plain match undervalues large-body compression). The O(n²) comparison stays; what changes is merge scheduling.

In term of their match, that will be from all to all comps between the oFs first?

frame_2D_alg/meta_code.py
kwcckw: Please correct me if i'm wrong. So this is just back to add 2 simple if forks, each with Sub and sub?

frame_2D_alg/meta_code.py
boris-kz: right. so all merges will add forks to that oF body

frame_2D_alg/meta_code.py
boris-kz: the two subs per ast site is fork initialization, additional forks in that site will be added one by one. 
so initialization adds nesting to ast site, and all subsequent forks to that site use the same nesting 
there is no special Sub

kwcckw: > the two subs per ast site is fork initialization, additional fork in the same site will be added one by one. so initialization adds nesting to ast site, and all subsequent forks to that site use the same nesting there is no special Sub

I see, so to init 2 subs per ast site in body, it will be:

    body += [ast.IfExp, (f1, sub1), (f2, sub2)]
    
Then subsequent forks to the site will be adding to the same body?

    [ast.IfExp, (f1, sub1), (f2, sub2), (f3, sub3), (f4, sub4)..]

boris-kz: yes, but I don't know if we need them in ifExp

frame_2D_alg/meta_code.py
boris-kz: def clust_oF_():  # pairwise merge for body compression

    F_ = copy(oF_); mrg_F_ = []
    for _F,F in combinations(F_,2): comp_body(_F,F)  # add F.rim, init all typs? 
    for _F in sorted(F_.pop(), key=lambda: _F.w, reversed=True):
        if _F.w > ave:  # sum F.w in comp_body
            if _F.fin: continue
            T = CoF(N_=[_F], typ=_F.typ, body = copy(_F.body), fc=_F.fc, caller_=copy(_F.caller_))
            for F,w in sorted(_F.rim, key=lambda: w, reversed=True):
                if F.fin: continue  # for unique mrg
                if w > ave: F.fin=1; merge_oF(T,F)
                else: break
        else: break
        if len(T.N_) > 1: mrg_F_ += [T]

boris-kz: Chain on the sketch:

1. **comp_body stays pure.** It's recursive — rim/w bookkeeping inside it needs a depth guard to fire only at entry. Do it at the call site: w = C * min fc per pair, appended to both rims. Also make the tail-length penalty symmetric (currently only charges when n is longer than _n): pair w must be order-free since merge direction is decided later, at seeding.

2. **Seed rank = projected compression, not raw sum.** Summing all rim w conflates centrality with typ-pool size and mixes sub-ave mass into the rank. Sum only above-ave rim w — that is the expected realized compression, and the outer gate becomes scale-consistent with ave.

3. **Typ scope.** Keep comp within typ: comp_prim's CoF branch is nF-keyed and conventions differ across typs, so cross-typ w is spurious mass with no mergeable pairs. Rims are per-typ by construction then; "init all typs" reduces to running the same loop per typ partition.

4. **Seed consumption.** Members fin on absorption is right, but the sketch never fins the seed — a later seed's rim can absorb _F into a second composite. Fin the seed only on composite success; a failed seed must stay absorbable.

5. **Validation.** The sketch drops the realized-cmpr gate entirely, trusting the estimate. Per-member revalidation needs body snapshots to revert; a single final gate needs only fin-flag resets, because member bodies are intact — non-destructive makes late validation nearly free. Lead: one final T.cmpr * wL > ave gate, release members on failure. Residual risk is one drifted member sinking the composite — bounded by sorted rim + break, acceptable.

6. **merge_oF call: fsel=0, explicit T.N_ += [F].** The fsel=1 branch gates membership by individual-cost/fork-cost — a different filter — and returns a tuple. Selection here is the w gate; merge_oF should only align.

7. **Copy depth.** copy(_F.body) shares the spine; merge_oF's fork accretion then mutates the exemplar, silently breaking non-destructive. deepcopy clones CoF refs, breaking identity/invertibility. Contract: structural copy — private spine (lists at fork points, since forks accrete), shared leaves (ast and CoF refs).

8. **Bookkeeping.** caller_ union per absorbed member; T.fc recomputed once from body after the loop (merge_oF doesn't maintain it at fsel=0); registration into oF_/nF_ unchanged; the destructive F_.remove disappears — members stay, dispatch redirect decides. Mechanically: sorted(F_.pop()) iterates one popped CoF, not the pool.

```python
def clust_oF_():  # exemplar-seeded merge for body compression

    F_ = copy(oF_); mrg_F_ = []
        if _F.typ == F.typ:
            w = comp_body(_F,F) * min(_F.fc,F.fc)  # comp_body pure, symmetric len penalty
            _F.rim += [(F,w)]; F.rim += [(_F,w)]
            if w > ave: _F.w += w; F.w += w  # projected compression: above-ave w only
        if _F.w*wL <= ave: break
        if _F.fin: continue
        T = CoF(N_=[_F], typ=_F.typ, body=copy_spine(_F.body), fc=_F.fc, caller_=copy(_F.caller_))
        for F,w in sorted(_F.rim, key=lambda t: t[1], reverse=True):
            if w <= ave: break
            if F.fin: continue
            merge_oF(T,F, fsel=0); T.N_ += [F]; T.caller_ |= F.caller_; F.fin = 1
        if len(T.N_) > 1:
            T.fc = sum(get_fc(t) for t in T.body); T.cmpr = sum(f.fc for f in T.N_) - T.fc
            if T.cmpr*wL > ave: _F.fin = 1; mrg_F_ += [T]
            else:
                for F in T.N_[1:]: F.fin = 0  # free revert: member bodies intact
    for T in mrg_F_: oF_.append(T); nF_.append(None); T.nF = len(oF_)-1
```

Closure: comp_body internals, merge_oF alignment, ffeedback, dispatch redirect untouched. The fuzzy stage later swaps fin for membership w without structural change to this loop.

boris-kz: Chain, then code:

1. **Selection out entirely.** fsel, C, add_, add2O/sum2O calls, tuple return — all gone. Membership is decided by clust's w gate and validated by the final cmpr gate; merge_oF only aligns. sum2O's call drops the fsel arg too. Since alignment is where membership is realized, N_ and caller_ bookkeeping move into merge_oF; clust drops its T.N_ += [F], caller union, and fc recompute lines.

2. **Fork = (ast.IfExp, alt_, tag__).** Flat alternative list with parallel member-tag lists, created at first divergence with [copy(F.N_), [m]] — prior members vs new — accreted in place thereafter (private by construction). Untagged element = universal, matched by every member so far; that invariant is what makes copy(F.N_) the correct tag at creation. Flat alts replace fork-in-fork nesting.

3. **is_fork = 3-slot IfExp.** Source ternaries are build's 2-tuples, so they flow through comp_prim as ordinary ops — the conflation fix.

4. **Tails must be tagged.** Inline tail ops are member-specific but read as universal: dispatch executes them for wrong members, and the next member's zip forks against them spuriously. Per-op single-alt forks preserve positional alignment; a grouped If would collapse k positions into one and corrupt alignment for subsequent merges — that's the discriminant, so the fw cost gate goes (worth is the final cmpr gate's job). Mirrored for F's tail when f is shorter: wrap untagged ops with prior members.

5. **Gate pairs: COW + recursion.** Clone when Sub.root is not F (seed-spine or registry refs are shared), with N_ = copy of outer F.N_ so nested fork tags stay in top-member terms; m passes down for the same reason. Mixed gate/op falls through to fork — a gating difference is a fork, not absorption, which deletes the old fin-scan/add2O logic.

6. **fc maintained at exit,** so get_fc on nested clones isn't stale and clust's recompute is redundant. get_fc already handles the new fork via n[1]; tag lists at n[2] are cost-free.

```python
def merge_oF(F, f, m=None):  # align f into F: match | tagged fork per miss; F composite-private, f read-only
    m = m or f  # top member for fork tags, passed down in nested gate merge

    body = []
    for Sub,sub in zip(F.body, f.body):
        if isinstance(Sub,CoF) and Sub.nF=='E' and isinstance(sub,CoF) and sub.nF=='E':  # both gates
            if Sub.root is not F:  # copy-on-write: shared gate from seed spine | split registry
                Sub = CoF(N_=copy(F.N_), body=copy(Sub.body), fc=Sub.fc, caller_=copy(Sub.caller_), root=F, nF='E')
            merge_oF(Sub, sub, m)
            body += [Sub]
        elif is_fork(Sub):  # composite-born, accrete in place
            for alt, tag_ in zip(Sub[1], Sub[2]):
                if comp_prim(alt, sub): tag_ += [m]; break
            else: Sub[1] += [sub]; Sub[2] += [[m]]
            body += [Sub]
        elif comp_prim(Sub, sub): body += [Sub]  # untagged = universal op
        else: body += [(ast.IfExp, [Sub,sub], [copy(F.N_), [m]])]  # fork: prior members | m
    if   len(f.body) > len(F.body): body += [(ast.IfExp,[t],[[m]]) for t in f.body[len(F.body):]]
    elif len(F.body) > len(f.body): body += [t if is_fork(t) else (ast.IfExp,[t],[copy(F.N_)]) for t in F.body[len(f.body):]]

    F.body = body; F.N_ += [m]; F.caller_ |= f.caller_
    F.fc = sum(get_fc(t) for t in body)

def is_fork(t): return isinstance(t,tuple) and len(t)==3 and t[0] is ast.IfExp  # source ternaries are 2-tuples: ordinary ops
```

Closure: comp_body untouched — composites aren't re-compared in seed-centric clust, so it never sees merged forks. Remaining open contract is downstream and was already open: inject/dispatch consumes tag__ to resolve forks per member, replacing gv_-style gating at fork sites.

kwcckw: For clust_oF, i think your version is fine and we not really need to add rim?

After this line:

    if w > ave: F.fin=1; merge_oF(T,F)
    
 I think we need to accumulate w, and recompute fc?

On the merge_oF, that 'E' section is no longer relevant?

boris-kz: not E, but we do need to identify gv_ somehow. But it's probably not relevant for source forks
fc will be recomputed after all merges

rim is needed for merge eval:
for F,w in sorted(_F.rim, key=lambda t: t[1], reverse=True):

frame_2D_alg/meta_code.py
```diff
@@ -291,7 +291,7 @@ def clust_oF_():  # pairwise merge for body compression
         if v*wL <= ave: break  # no pair worth realizing
         nT = CoF(N_=[_T], typ=_T.typ, fc=_T.fc, caller_=copy(_T.caller_))
         nT.body = copy(_T.body)  # merged below
-        merge_oF(nT, T, fsel=0)
+        merge_oF(nT, T)
```
boris-kz, line 294: this whole merge should be done in comp_body
and is_fork is pure fluff

frame_2D_alg/meta_code.py
kwcckw: Okay, the `gv_` should be an existing IfExp fork, and i think we can identify it via t[1][0] or something similar, but it should working the same as the other IfExp fork, so there's no need to identify it?

frame_2D_alg/meta_code.py
kwcckw: The comp_body is for body node comparison, why we need the merge within comp? Those 2 are different operations?

That is_fork actually can be unpacked, so we can remove it.

frame_2D_alg/meta_code.py
boris-kz: we may need it to eval gv_ -> oF conversion

frame_2D_alg/meta_code.py
boris-kz: because comp is where we define forks, exact same process

def clust_oF_():  # exemplar-seeded merge for body compression

    F_ = copy(oF_); T_ = []
    while F_:
        _F = F_.pop()
        if not _F.fin:
            T = CoF(N_=[_F], typ=_F.typ, body=copy(_F.body), fc=_F.fc, caller_=copy(_F.caller_))
            for F in F_:
                if _F.typ == F.typ and not _F.fin:
                    C, body = comp_body(T.body, F.body)
                    if C > ave:
                        T.body = body; T.w += C; _F.fin=1
            if len(T.N_) > 1:
                T.fc = sum(get_fc(t) for t in T.body); T.cmpr = sum(f.fc for f in T.N_) - T.fc
                if T.cmpr > ave: _F.fin=1; T_+= [T]
                else:
                    for F in T.N_[1:]: F.fin = 0
    for T in T_: oF_.append(T); nF_.append(None); T.nF = len(oF_)-1

frame_2D_alg/meta_code.py
kwcckw: Do you mean the process similar to split_oF_, where we pack gates into oF?

But all gv_ should be already eval in split_oF_, and we don't create new gv_ in clust_oF_？

boris-kz: right

frame_2D_alg/meta_code.py
boris-kz: Three restorations, then one real flag:

1. **fin inversion again, both sites**: inner guard checks `_F.fin`, commit sets `_F.fin=1` — first commit kills the seed's own loop and members are never consumed. Guard `not F.fin`, commit `F.fin=1`.
2. **T.N_ never grows** — so `len(T.N_)>1` is always false: no composite registers, the cmpr block and revert are dead. `T.N_ += [F]` plus caller union per commit.
3. **Tag args dropped from comp_body** — fork alternatives need membership at creation (prior members vs new); the source is `T.N_, F`. Without them alt.N_ can't be built.

The flag — your new cmpr gate exposes an fc convention clash: member `.fc` from build counts callee refs at Call cost 3, while `get_fc(CoF)` returns the callee's full body fc. So `sum(f.fc for f in T.N_) - T.fc` is biased negative by the full body cost of every referenced oF, per reference — root-typ bodies are mostly refs, so the gate kills valid composites. Convention fix: int-nF CoF = registered ref, charge Call cost (body counted at its def); str-nF ('E','|') = inline block, charge its fc. Same rule fixes comp_body's identity branch (matching two refs saves a call site, not a body): `if n is _n: return get_fc(n), _n`. Once consistent, T.w ≈ T.cmpr by construction — the gate becomes a cheap cross-check of the cost table; one of the two can go later.

```python
def clust_oF_():  # exemplar-seeded merge for body compression

    F_ = copy(oF_); T_ = []
    while F_:
        _F = F_.pop()
        if not _F.fin:
            T = CoF(N_=[_F], typ=_F.typ, body=_F.body, fc=_F.fc, caller_=copy(_F.caller_))
            for F in F_:
                if _F.typ == F.typ and not F.fin:
                    C, body = comp_body(T.body, F.body, T.N_, F)
                    if C > ave:
                        T.body = body; T.w += C; T.N_ += [F]; T.caller_ |= F.caller_; F.fin = 1
            if len(T.N_) > 1:
                T.fc = sum(get_fc(t) for t in T.body); T.cmpr = sum(f.fc for f in T.N_) - T.fc
                if T.cmpr > ave: _F.fin = 1; T_ += [T]
                else:
                    for F in T.N_[1:]: F.fin = 0
    for T in T_: oF_.append(T); nF_.append(None); T.nF = len(oF_)-1
```
```python
def get_fc(n):
    if isinstance(n,CoF): return n.fc if isinstance(n.nF,str) else costs.get(ast.Call,0)  # inline 'E'|'|' block | registered ref: body counted at def
    if isinstance(n,tuple): return costs.get(n[0],0) + sum(get_fc(c) for c in n[1])
    return costs.get(type(n),0)
```

`copy(_F.body)` dropped: comp_body is pure, commit reassigns, exemplar body is never touched.

boris-kz: Assembled with everything standing: forks as '|' CoFs with alternative blocks in N_, membership in N_ only, ref-vs-inline by int/str nF, identity match = saved call site, inline-only body merge. Assembly resolves one thing left implicit: split-registered gates carry int nF, so on the member side gates are *refs*, not 'E' blocks — different members' gates fork per site and merge at def level, which happens automatically since split appends them to oF_ before clust runs; sites redirect after. The inline-'E' branch therefore fires only on merge-born blocks, i.e. composite-side or in later cycles.

```python
def comp_body(_n, n, _m_, m):  # compare = merge: realized compression C + merged form, pure
    # _n: composite side, may hold 'E' blocks | '|' forks; n: member side, raw ops + registered refs

    def alt(m_,t_): return CoF(nF='A', N_=m_, body=t_, fc=sum(get_fc(t) for t in t_))  # member-subset block
    def fork(a_):   return CoF(nF='|', N_=a_, fc=2 + sum(a.fc for a in a_))  # alts in N_; 2: ast.IfExp

    def seq(_q, q):  # aligned op sequences
        C = 0; Q = []
        for c,t in [comp_body(_t,t,_m_,m) for _t,t in zip(_q,q)]: C += c; Q += [t]
        for _t in _q[len(q):]: Q += [_t if isinstance(_t,CoF) and _t.nF=='|' else fork([alt(copy(_m_),[_t])])]; C -= 2  # m skips
        for t in q[len(_q):]:  Q += [fork([alt([m],[t])])]; C -= 2  # priors skip
        return C, Q

    if isinstance(n,list): return seq(_n, n)  # top bodies
    if isinstance(_n,CoF) and _n.nF=='|':  # fork: merge n into best alt | append alt
        r_ = [comp_body(a.body, [n], a.N_, m) for a in _n.N_]  # nested forks tag in alt-member terms
        i = int(np.argmax([c for c,_ in r_])); c, body = r_[i]
        a_ = copy(_n.N_)
        if c > 0: a_[i] = alt(_n.N_[i].N_+[m], body); return c, fork(a_)
        else:     return -2, fork(a_ + [alt([m],[n])])
    if isinstance(n,CoF) or isinstance(_n,CoF):
        if n is _n: return get_fc(n), _n  # same registered ref | shared block: match = saved call site
        if isinstance(n,CoF) and isinstance(_n,CoF) and isinstance(n.nF,str) and isinstance(_n.nF,str) and n.fc > ave_C:
            C, body = seq(_n.body, n.body)  # inline blocks: recursive body merge
            if C > 0: return C-2, CoF(nF='E', N_=copy(_m_)+[m], body=body, fc=sum(get_fc(t) for t in body),
                                      caller_=_n.caller_|n.caller_)
        return -2, fork([alt(copy(_m_),[_n]), alt([m],[n])])  # diff refs | mixed: fork here, merge at def level
    if isinstance(_n,tuple) and isinstance(n,tuple) and n[0] is _n[0]:
        C, sub_ = seq(_n[1], n[1]); return C + costs.get(n[0],0), (n[0], tuple(sub_))  # deep partial merge
    if type(_n) is type(n) and not isinstance(n,tuple): return costs.get(type(n),0), _n  # leaf
    return -2, fork([alt(copy(_m_),[_n]), alt([m],[n])])  # fork per miss
```

alt gets nF='A' so str-nF = inline holds uniformly for get_fc. Scope stays as agreed: member side never carries '|' — that arises only when composites re-enter the pool as members, so fork-into-fork merge is the open contract for hierarchical cycles, nothing else.

kwcckw: I see, in this section:

    if _F.typ == F.typ and not F.fin:
        C, body = comp_body(T.body, F.body, T.N_, F)
        if C > ave:
            T.body = body; T.w += C; T.N_ += [F]; T.caller_ |= F.caller_; F.fin = 1

So comp_body compare each node in body, and merge them within the func? The merge_oF should be folded into comp_body now?

frame_2D_alg/meta_code.py
```diff
@@ -318,71 +318,56 @@ def inject_oF_(oF_, g):  # inject AST in g, recompile g[name]
...
+            for _F,_Sub in Sub[1]:
+                if comp_prim(_Sub,sub):  # check all Subs in if fork to match with sub
+                    # add eval for partial match and significant sub-fork? 
+                    Sub[1] += [(f,sub)]; felse= 1   # test by source only?
+                    break
+            if felse: Sub[2] += [(f,sub)]  # pack to else fork
+            body += [Sub]   
+        elif comp_prim(Sub, sub):
+             body += [Sub]  # match: position stays plain
+        # structure = (type, if_subs, else_subs). Each subs is [(source, nested sub),...]
```
kwcckw, line 337: > yes, but I don't know if we need them in ifExp

I'm not sure why my comment get deleted, but i think we need them for the if fork routing when rebuild ast node?

frame_2D_alg/meta_code.py
kwcckw: So clust_oF_ should create new gv_ fork? If yes, then we should include this similar section:

        if is_gate_l(t) and oF.w * sum(get_fc(p) for p in t[1]) > ave:  # w is membership weight, it should be fw?
            sub = CoF(root=oF, fc=get_fc(t), body=[t], caller_={oF})
            oF_.append(sub); nF_.append(None); sub.nF = len(oF_)-1

Else there's i don't think we needto identify if the fork is gv_.

boris-kz: comp_body needs to be updated, it replaces merge_oF now, i posted claude version

frame_2D_alg/meta_code.py
boris-kz: right, check above

frame_2D_alg/meta_code.py
boris-kz: probably. I might have deleted it by mistake

frame_2D_alg/meta_code.py
kwcckw: Actually I couldn't understand the code above. The alt above packs member, while fork packs the prior ast IfExp fork?

boris-kz: Not sure, I will check tomorrow

kwcckw: There are several new nFs above, including the prior E, and the new A and |, but I think we only need IfExp fork instead of a nF= E's CoF?

kwcckw: > Not sure, I will check tomorrow

Okay, I will try to create another version first since this is actually not quite readable.

PR #199, opened 2026-07-08
7/8/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -276,6 +276,40 @@ def seq(_q, q):  # aligned op sequences
     if type(_n) is type(n) and not isinstance(n,tuple): return costs.get(type(n),0), _n  # leaf
     return -2, fork([alt(copy(_m_),[_n]), alt([m],[n])])  # fork per miss
 
+
+def comp_body_claude2(_n, n, _m_=None, m=None):  # compare == merge: compression C + merged form (pure, replaces merge_oF)
```
kwcckw, line 280: I edited and refactored the prior code using claude. 

The prior code `comp_body_claude` uses CoF in initializing forks, which should be in the format of `(ast.IfExp, (f,_sub),...)` instead?

boris-kz: right

frame_2D_alg/meta_code.py
```diff
@@ -315,19 +349,28 @@ def split(t, oF):  # pack sub gate from gates
 def clust_oF_():  # exemplar-seeded merge for body compression
 
     F_ = copy(oF_); T_ = []
-    while F_:
```
boris-kz, line 318: what's wrong here? the problem with combinations is that there is no clear seed to initialize new T

frame_2D_alg/meta_code.py
```diff
@@ -315,19 +349,28 @@ def split(t, oF):  # pack sub gate from gates
...
+    for F in F_: F.fin = 0; F.rim = []
+    for _F,F in combinations(F_,2):
+        if _F.typ == F.typ:
+            w,_ = comp_body(_F,F)
+            # w *= min(_F.fc,F.fc)  # this is not needed?
+            if w > ave:
+                _F.w += w; F.w += w  # projected compression: above-ave w only
+                _F.rim += [(F,w)]; F.rim += [(_F,w)]
+                
+    for _F in sorted(F_, key=lambda F: F.w, reverse=True):
```
boris-kz, line 361: do we need this sorting, or the forks can be added in any order?

kwcckw: > what's wrong here? the problem with combinations is that there is no clear seed to initialize new T

This is similar with centroid clustering, which we should start initialize with the highest w's F? Because if F.w is weak, we should break and skip the rest on the clustering?

The `while F_` method is random and the seed should starts with highest w's F?

boris-kz: It makes no difference, we compare all to all, and there is no refinement around the centroid at this point

kwcckw: If we use the combination method:

        if _F.typ == F.typ:
            w,_ = comp_body(_F,F)
            # w *= min(_F.fc,F.fc)  # this is not needed?
            if w > ave:
                _F.w += w; F.w += w  # projected compression: above-ave w only
                
The F with highest match should get highest w from `_F.w += w; F.w += w `, and those Fs should starts as seed?

boris-kz: ok that's if we do centroid-like clustering here, that makes sense.
but then we we are sorting _F.rim, wich is not directly related to T-to-F similarity.
it should probably be the average linkage version, where we need to go through rims of all _Fs in T.N_, not just the seed _F

kwcckw: > wich is not directly related to T-to-F similarity.

Okay right, so it's something like eval with the overlapping rim between `_F.rim` and `T.N_`'s rim?

frame_2D_alg/meta_code.py
boris-kz: Three breaks:

1. **get_fc crashes on forks.** Generic path reads a branch pair (m_, op) as (head, sub_): costs.get(m_) with a list key → TypeError; and branches at [2:] are invisible to [1]-only recursion, so fc would undercount even if it didn't crash. Fork case inline:

```python
def get_fc(n):
    if isinstance(n,CoF): return costs.get(ast.Call,0)  # ref: body counted at def
    if isinstance(n,tuple):
        if n[0] is ast.IfExp and isinstance(n[1][0],list):  # fork: (m_,op) branches in n[1:]
            return costs.get(ast.IfExp,0) + sum(get_fc(op) for m_,op in n[1:])
        return costs.get(n[0],0) + sum(get_fc(c) for c in n[1])
    return costs.get(type(n),0)
```

2. **Live-tag bug at top entry.** _m_=[_n]=[T]: a fork born at member k stores that as the prior-branch tag; when k+1 diverges into a sibling branch, resolving [T] through live N_ puts k+1 in both branches — siblings must partition members. Tags snapshot at creation: `merge_seq(_n.body, n.body, copy(_n.N_), n)`. Purity elsewhere holds (branch updates build new lists), so this copy is the only one needed.

3. **Tail discriminator under-checks.** `_t[0] is ast.IfExp` alone catches build's source ternaries (sub_ at [1] is a tuple); they'd ride the tail untagged and dispatch would execute them for m. Same check as your fork branch, plus the -2 belongs only where a gate is actually created — an existing fork skips m for free, m is absent from all tags:

```python
        for _t in _q[len(q):]:
            if isinstance(_t,tuple) and _t[0] is ast.IfExp and isinstance(_t[1][0],list): Q += [_t]
            else: Q += [gate(_m_, _t)]; C -= 2
```

Bookkeeping: clust's call becomes `comp_body(T, F)`. The open contract shifts rather than grows: with inline-block merge gone, encapsulating expensive divergence is next-cycle split on registered composites — split's detector keys on gv_ heads only, so it must also recognize membership forks to extract branches into registered, borrowable sub-oFs.

frame_2D_alg/meta_code.py
boris-kz: `_F.rim is already in T.N_'s rim, this is extending the scope, and element-wise pair summing replaces comp_body(T,F)`

frame_2D_alg/meta_code.py
```diff
@@ -276,6 +276,40 @@ def seq(_q, q):  # aligned op sequences
     if type(_n) is type(n) and not isinstance(n,tuple): return costs.get(type(n),0), _n  # leaf
     return -2, fork([alt(copy(_m_),[_n]), alt([m],[n])])  # fork per miss
 
+
+def comp_body_claude2(_n, n, _m_=None, m=None):  # compare == merge: compression C + merged form (pure, replaces merge_oF)
+    
+    def gate(f_, sub):             return (ast.IfExp, (f_, sub))              # 1-branch: these members do op, others skip
```
boris-kz, line 282: we need to get rid of this crap

frame_2D_alg/meta_code.py
kwcckw: Yes sorry, it should be looping each `__F` in  `_F.rim`,  and compare `__F.rim` with T.N_'s rim?

> element-wise pair summing replaces comp_body(T,F)

Actually i don't get this, the element pair summing is between `_F` and `__F` in _F.rim when extending the scope?

boris-kz: yes, when extending the oF cluster
all F are already compared, we just need to sum the results from rims

frame_2D_alg/meta_code.py
kwcckw: So in existing comp_body, we have 3 main sections for CoF (oFs), tuple (nested forks) and normal ast node. To fold merge_oF into the comp_body, we actually need to create new ast.IfExp fork for both 3 sections above regardless of their types? Because we need new forks for both `_n` and `n` in the merged body?

boris-kz: right, but these new forks are defined paiwise per F, we need to define them in T.body 
so comp_body shouldn't add forks, T.body will have to be rebuilt anyway

boris-kz: Yes — and it goes further than dropping fork emission: with average linkage, membership is fixed *before* construction, so the incremental fold itself is residue of the old interleaved decide-while-merging. Construction becomes one n-way pass over member bodies, and comp_body returns to its original contract — compare-only, C only. fmrg dies, gate/new_fork/add_fork die, consolidate dies (adjacent grouping happens at build, not as rewrite).

Chain:

1. **Columns replace sequential zips.** Position j across all N_ bodies is one column; members short of j are simply absent — tails need no special case, absence = skip.
2. **Within-column: seeded op clustering by C.** Route each op into best group if C > 0, else seed — add_fork's routing relocated, using compare-only comp_body. Single group covering all members → bare op; partial coverage → gate (1-branch fork); multiple groups → fork with member partition. Same-head tuples in one group merge by recursing on children columns with group members.
3. **Forks are built once, final-form**, branch payload = op list, so adjacent same-partition columns concatenate into sequence branches in the same pass.

```python
def comp_body(_n, n):  # compare only: compression estimate C; construction in form_body

    if isinstance(_n,list):  # op sequences: bodies | tuple children
        C = -2*abs(len(_n)-len(n))
        for _t,t in zip(_n,n): C += comp_body(_t,t)
        return C
    if _n is n: return get_fc(n)  # same ref: saved call site
    if isinstance(_n,tuple) and isinstance(n,tuple) and _n[0] is n[0]:
        return costs.get(_n[0],0) + comp_body(list(_n[1]), list(n[1]))
    if type(_n) is type(n) and not isinstance(_n,(tuple,CoF)): return costs.get(type(n),0)  # leaf
    return -2  # miss: fork in form_body

def form_body(N_):  # render composite body: n-way positional columns, cluster ops per column

    def f(t): return isinstance(t,tuple) and t[0] is ast.IfExp and isinstance(t[1][0],list)  # fork, not source ternary

    def form(col, nT):  # col: [(op,F)] -> bare op | merged op | fork
        grp_ = []
        for t,F in col:
            c_ = [comp_body(g[0][0], t) for g in grp_]
            i = int(np.argmax(c_)) if c_ else 0
            if c_ and c_[i] > 0: grp_[i] += [(t,F)]
            else: grp_ += [[(t,F)]]
        br_ = []
        for g in grp_:
            t0 = g[0][0]; m_ = [F for _,F in g]
            if len(g)==1 or all(t is t0 for t,_ in g) or not isinstance(t0,tuple): op = t0  # single | ref | leaf
            else: op = (t0[0], tuple(form([(t[1][j],F) for t,F in g if len(t[1])>j], len(g))
                                     for j in range(max(len(t[1]) for t,_ in g))))  # same-head: children columns
            br_ += [(m_,[op])]
        if len(br_)==1 and len(br_[0][0])==nT: return br_[0][1][0]  # universal: bare
        return (ast.IfExp, *br_)  # fork | gate if partial column

    bod_ = [F.body for F in N_]; B = []
    for j in range(max(len(b) for b in bod_)):
        t = form([(b[j],F) for b,F in zip(bod_,N_) if len(b)>j], len(N_))
        if f(t) and B and f(B[-1]) and [b[0] for b in B[-1][1:]]==[b[0] for b in t[1:]]:  # same partition: concat
            B[-1] = (ast.IfExp, *[(m_, op_+_op_) for (m_,op_),(__,_op_) in zip(B[-1][1:], t[1:])])
        else: B += [t]
    return B
```
clust construction block:
```python
        if len(N_) > 1:
            T = CoF(N_=N_, typ=_F.typ, body=form_body(N_), caller_=set().union(*[F.caller_ for F in N_]))
            T.fc = sum(get_fc(t) for t in T.body); T.cmpr = sum(f.fc for f in N_) - T.fc
            if T.cmpr > ave: _F.fin = 1; T_ += [T]
            else:
                for F in N_[1:]: F.fin = 0
```

Within-column routing stays greedy first-seeded — same order exposure as the fold had, unchanged. sum2O's body branch delegates to form_body. get_fc keeps the fork case from before (payloads now uniformly lists). Open contracts shrink by one: composite-as-member (fork-bearing columns in comp_body C and in grouping), split extracting membership forks, inject consuming (m_, op_) branches — consolidate gone.

kwcckw: The T.body should be init from _F.body, and merge with the rest of the Fs when we cluster them? So during the cluster process, we still need to compare and create fork when we have both T.body and F.body, which I think we still need to fold the merging process in the comp since me must compare them first before we can merge them.

boris-kz: this is supposed to combine the body from all elements:


def form_body(N_):  # render composite body: n-way positional columns, cluster ops per column

    def fk(t): return isinstance(t,tuple) and t[0] is ast.IfExp and isinstance(t[1][0],list)  # fork

    def form(col, Fn):  # col: [(op,F)] -> bare op | merged op | fork
        grp = {}
        for t,F in col:
            k = t if isinstance(t,CoF) else id(t[0]) if isinstance(t,tuple) else type(t)  # ref | head | leaf type
            grp.setdefault(k,[]).append((t,F))
        branch_ = []
        for g in grp.values():
            t0 = g[0][0]; f_ = [F for _,F in g]
            if len(g)==1 or not isinstance(t0,tuple) or all(t is t0 for t,_ in g):
                op = t0  # single | ref,leaf | shared
            else:  # same-head: children columns
                op = (t0[0], tuple(form([(t[1][j],F) for t,F in g if len(t[1])>j], len(g)) for j in range(max(len(t[1]) for t,_ in g))))
            branch_ += [(f_,[op])]
        if len(branch_)==1 and len(branch_[0][0])==Fn: return branch_[0][1][0]  # universal: bare op
        return (ast.IfExp, *branch_)
    Bod = []
    bod_= [F.body for F in N_]
    for j in range(max(len(b) for b in bod_)):
        t = form([(b[j],F) for b,F in zip(bod_,N_) if len(b)>j], len(N_))
            Bod[-1] = (ast.IfExp, *[(f_, op_+_op_) for (f_,op_),(__,_op_) in zip(Bod[-1][1:], t[1:])])
        else:
            Bod += [t]
    return sum(F.fc for F in N_) - sum(get_fc(t) for t in Bod), Bod

kwcckw: > this is supposed to combine the body from all elements:
 
 Yes, the original code should be doing the same thing, but the form_body above is per F_, and we only need to reform body from a pair of Fs in clust_oF_?

boris-kz: no, it's a body of a new cluster. the clustering element is pairs, but body rebuiling is batched

kwcckw: I see, thanks. So this is rebuilding T.N_'s body. In this line:

    t = form([(b[j],F) for b,F in zip(bod_,N_) if len(b)>j], len(N_))

This is actually merging all Fs' `j` position ops at once, a batch based merging process. 

I will need to step through it first, and in clust_oF_, we only need C to eval the packing of F into T.N_:

          C = comp_body(T.body,F.body, T.N_,F)
          if C>ave: T.w += C; T.N_ += [F]; F.fin = 1

And the form_body should be in the section below ` if T.cmpr > ave`?

        if len(T.N_) > 1:
            T.fc = sum(get_fc(t) for t in T.body); T.cmpr = sum(f.fc for f in T.N_) - T.fc
            if T.cmpr > ave: _F.fin = 1; T.fc, T.body = form_body(T.N_); T_ += [T]

boris-kz: probably, please edit

PR #200, opened 2026-07-09
7/9/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -280,17 +282,19 @@ def form(col, Fn):  # col: [(op,F)] -> bare op | merged op | fork
             else:  # same-head: children columns
                 op = (t0[0], tuple(form([(t[1][j],F) for t,F in g if len(t[1])>j], len(g)) for j in range(max(len(t[1]) for t,_ in g))))
             branch_ += [(f_,[op])]
+        # when single node in branch, eval to add If fork too?
         if len(branch_)==1 and len(branch_[0][0])==Fn: return branch_[0][1][0]  # universal: bare op
```
kwcckw, line 286: When we have single fork, that's when there's offset, i think we can eval the offset body length to add fork on those offsets too, as in the prior code?

frame_2D_alg/meta_code.py
```diff
@@ -280,17 +282,19 @@ def form(col, Fn):  # col: [(op,F)] -> bare op | merged op | fork
...
-        return (ast.IfExp, *branch_)
+        return (ast.IfExp, *branch_)  # always form branch when there's at least 2 same length Fs
     Bod = []
     bod_= [F.body for F in N_]
     for j in range(max(len(b) for b in bod_)):
         t = form([(b[j],F) for b,F in zip(bod_,N_) if len(b)>j], len(N_))
+        # i don't get below? Why we need to merge Bod[-1]? it's in entirely different block?
         if fk(t) and Bod and fk(Bod[-1]) and [b[0] for b in Bod[-1][1:]]==[b[0] for b in t[1:]]:  # same func partition: concat
```
kwcckw, line 293: Actually i don't understand this section. When we merge Fs' body by index j, why we need to concat with the prior index's body here?

frame_2D_alg/meta_code.py
```diff
@@ -238,11 +238,13 @@ def clust_oF_():  # exemplar-seeded merge for body compression
         T = CoF(N_=[_F], typ=_F.typ, body=_F.body, fc=_F.fc, caller_=copy(_F.caller_))
```
boris-kz, line 238: T = CoF(N_=[_F], typ=_F.typ, body=_F.body, fc=_F.fc, caller_=copy(_F.caller_))
        _F.rooT = T  # temporary?
        for F,w in sorted(_F.rim, key=lambda t: t[1], reverse=True):
            T.L_ = {L for N in T.N_ for L in N.rim if not L[0].fin}  # T.N_ is extended per loop
            _L_ = F.rooT.L_ if F.rooT else F.rim  # try merging F.rooT if any
            rim_olp = T.L_.intersection(_L_)  # rooT overlap
            if olp_w := sum(L[1] for L in rim_olp) > ave:
                T.w += olp_w; T.N_ += [L[0] for L in rim_olp]; F.fin = 1  # merge Ts

boris-kz: actually we need recursive T frontier expansion here, as in cluster_N:

        while __L_:
            _L_ = []
                _N = L.N_[0] if L.N_[1].fin else L.N_[1]; in_.add(L)
                if not _N.fin and _N in Ft.N_:
                    m,d = nt_vt(*L.N_)
                    if m > ave * (_r-1):  # cluster nt, L,C_ by combined rim density, add gv_?
                            iM = sum([L.m for L in _N.rim if (L.N_[0] if L.N_[1] is _N else L.N_[1]) in N_])
                            if iM / (span*decay) < ave * (_r-1): continue  # normalized N-to-N_ match, sum for sum2G?
                        N_ += [_N]; L_ += [L]; _N.fin = 1
                        _L_+= [l for l in _N.rim if l not in in_ and (l.N_[0].fin ^ l.N_[1].fin)]   # new frontier links, +|-?
                    elif d > avd * (_r-1): B_ += [L]  # contrast value, exclusive?
            __L_ = list(set(_L_))

frame_2D_alg/meta_code.py
```diff
@@ -238,11 +238,13 @@ def clust_oF_():  # exemplar-seeded merge for body compression
...
-            C, body = form_body(T.N_+[F])
-            if C - T.w > ave:  # realized incremental compression
-                T.body = body; T.w = C; T.N_ += [F]; T.caller_ |= F.caller_; F.fin = 1
+            orim = {nt for N in T.N_ for nt in N.rim if not nt[0].fin}.intersection(F.rim)  # overlapping rim between T.N_'s rim and F.rim
+            ow = sum(Nt[1] for Nt in orim)  # sum w from overlapping rim
+            if ow > ave:  # realized incremental compression
+                T.w += w; T.N_ += [F]; F.fin = 1  # caller should be filled in the next runtime
         if len(T.N_) > 1:
-            T.fc = sum(f.fc for f in T.N_) - T.w; _F.fin = 1; T_ += [T]
+            T.body = form_body(T.N_); T.fc = sum(get_fc(t) for t in T.body)  # this fc computation can be unpacked from comp_body?
```
boris-kz, line 246: yes, it should be in form_body

frame_2D_alg/meta_code.py
kwcckw: I see, the existing code using only the rim from F.rim's rim, but not recursive. The recursive version needs to be eval with this similar m and d `m,d = nt_vt(*L.N_)`?

frame_2D_alg/meta_code.py
kwcckw: It's just using get_fc on each body , right before return from form_body? So i think that can be moved to here?

frame_2D_alg/meta_code.py
boris-kz: yes, this is average linkage clustering

frame_2D_alg/meta_code.py
boris-kz: form_body shouldn't retunr anything, it just adds clutter

frame_2D_alg/meta_code.py
boris-kz: I think it's the case of multiple forks per root ifExt. 
Have you asked Claude?

frame_2D_alg/meta_code.py
boris-kz: yes, that's when the op is conditional on the caller handshake.
else there is no fork, the execution is default

kwcckw: Okay, then we shouldn't add anything here.

frame_2D_alg/meta_code.py
kwcckw: From claude:

form([(b[j],F) …]) is vertical — all members at position j. It produces exactly one entry (a bare op, or a fork if members diverge at that position).
The concat block is horizontal — it looks at position j and position j-1. If both are forks and split the members along the identical partition, it fuses them so the two op-lists sit in the same branches instead of producing two back-to-back forks.

Yes, so it actually "look" into the prior position to check for forks and fuse them if they have the same ops.

frame_2D_alg/meta_code.py
kwcckw: The rim of oF is another oF here, and they do not have m or d. So the recursive expansion should be eval solely by their w?

boris-kz: Rims is links, they have w

kwcckw: Yes, we pack them as tuple of F and w?

    _F.rim += [(F,w)]; F.rim += [(_F,w)]
    
So we can only eval with their w here.

boris-kz: yes.
Four contract breaks in the draft, then the port:

1. **No recursion yet** — candidates stay fixed at the seed's rim; absorbed members' rims extend T.L_ but never the candidate list. Flood needs the cluster_N form: frontier list rebuilt per wave from absorbed members' rims, fin links kept in it — they're the cluster-contact channel.
2. **Absorption target inverted**: the olp gate evaluates F but absorbs its co-neighbors — F itself never joins, and the neighbors enter ungated per node. Absorb F; neighbors surface through the extended frontier and pass their own gates.
3. **Overlap-on-non-fin is order-unstable**: T.L_ filters fin, so shared-external overlap monotonically *shrinks* as clustering proceeds — the same candidate passes early, fails late. cluster_N's own refinement is member-directed iM, which *grows* with T. Use that; direct link w is pre-gated by ave at rim construction.
4. Mechanical: `(F,w)` tuples share w only within one pair, so `T.L_.intersection(_L_)` is generically empty — node-keyed overlap or drop it (dropped: iM replaces it). Walrus binds the bool: `(olp_w := sum(...)) > ave`. rooT must init on all F and set on every absorption, else contact merging only sees seeds. form_body's C is discarded — it's the realized gate; and rendering must defer past merges or rT's render is wasted.

```python

    F_ = copy(oF_); T_ = []
    for F in F_: F.fin = 0; F.w = 0; F.rim = []; F.rooT = None
        if _F.typ == F.typ:
            w = comp_body(_F.body, F.body)  # compression estimate
            if w > ave: _F.w += w; F.w += w; _F.rim += [(F,w)]; F.rim += [(_F,w)]

        if _F.fin: continue
        if _F.w <= ave: break
        T = CoF(N_=[_F], typ=_F.typ, caller_=copy(_F.caller_)); _F.fin = 1; _F.rooT = T; mrg_ = set()
        L_ = _F.rim
        while L_:  # recursive frontier expansion
            _L_ = []
            for F,w in L_:
                if F.fin:
                    rT = F.rooT
                    if rT and rT is not T and rT not in mrg_:  # contact: merge clusters by cross-link density
                        mrg_.add(rT)
                        X = sum(v for N in T.N_ for G,v in N.rim if G.rooT is rT)
                        if X / (np.sqrt(len(T.N_)*len(rT.N_)) * decay) > ave:
                            rN_ = rT.N_; rT.N_ = []  # emptied rT dropped at render
                            for G in rN_: G.rooT = T
                            T.N_ += rN_; T.w += rT.w + X
                            _L_ += [l for G in rN_ for l in G.rim]
                else:
                    if len(T.N_) > 3:  # density refinement, cluster_N iM form: F-to-members, grows with T
                        iM = sum(v for G,v in F.rim if G.rooT is T)
                        if iM / (np.sqrt(len(T.N_)) * decay) < ave: continue
                    T.N_ += [F]; T.w += w; F.fin = 1; F.rooT = T
                    _L_ += F.rim
            L_ = list(set(_L_))
        if len(T.N_) > 1: T_ += [T]

    for T in T_:
            C, T.body = form_body(T.N_)  # render once, post-merges; realized gate
            if C > ave:
            else:
                for F in T.N_: F.fin = 0; F.rooT = None  # release; nothing downstream to undo
```

Contact-merge is the one thing here beyond cluster_N — exclusive fill there skips fin nodes; justified here because w-sorted seeding means a later T reaching an earlier T' is evidence they're one cluster, gated by cross-density X over geometric-mean span. Accepted simplifications: mrg_ memo blocks retry of a rejected rT as T grows; members released by a failed render stay unclustered this pass (seed loop already passed them). Duplicate frontier entries resolve through fin: second hit routes to contact branch, `rT is not T` skips.

kwcckw: I see, looks like it's more complicated because the rim might be already clustered and we need to get their root for T merging purpose.

I think this line ` mrg_.add(rT)` should be removed. We not even need this when all rT.N_ reference should be updated to new T after the merging.

boris-kz: right

kwcckw: Also i'm not sure if we need the decay here? It's using the global decay and we do not have any span or distance between Fs, so all pairs will be sharing the same decay anyway.

boris-kz: we don't. half of that was wrong, here's another draft:

    F_ = copy(oF_); T_ = []
        if _F.typ == F.typ:
            w = comp_body(_F.body, F.body)  # compression estimate
            if w > ave:
                _F.rim += [(F,w)]; F.rim += [(_F,w)]; F.w += w; _F.w += w
        if _F.rooT: continue
        if _F.w <= ave: break
        T = CoF(N_=[_F], typ=_F.typ, caller_=copy(_F.caller_)); _F.fin = 1; _F.rooT = T
        L_ = _F.rim
        while L_:  # recursive frontier expansion
            _L_ = []
            for F,w in L_:
                if rT := F.rooT and rT is not T and rT not in T.N_:  # merge clusters by cross-link density
                    if W := sum(w for N in T.N_ for f,w in N.rim if f.rooT is rT) > ave:
                        for f in _N_ := rT.N_: f.rooT = T
                        T.N_ += _N_; T.w += rT.w + W
                        _L_ += [l[0] for f in _N_ for l in f.rim]
                elif W := sum(w for f,w in F.rim if f.rooT is T) > ave:
                    T.N_ += [F]; T.w += W; F.rooT = T
                    _L_ += F.rim
            L_ = list(set(_L_))
        if len(T.N_) > 1: T_ += [T]
    for T in T_:
            form_body(T)
            if C := T.cmpr > ave:
            else:
                for F in T.N_: F.fin = 0; F.rooT = None  # release; nothing downstream to undo

kwcckw: Okay, i will go through them again later.

PR #201, opened 2026-07-10
7/10/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -304,6 +306,7 @@ def form(col, Fn):  # col: [(op,F)] -> bare op | merged op | fork
         else:
             Bod += [t]
     F.fc = sum(get_fc(n) for n in Bod)
+    F.cmpr = sum(f.fc for f in F.N_) - fc  # compression = original member - merged?
```
kwcckw, line 309: This F.cmpr is compression? So it should be the original F.N_'s fc subtract the new F.fc in order to check the final compression value?

frame_2D_alg/meta_code.py
```diff
@@ -234,29 +234,31 @@ def clust_oF_():  # flood-fill by rim links, cluster_N form: frontier expansion
...
+                        for f in (_N_:= rT.N_): f.rooT = T
                         T.N_ += _N_; T.w += rT.w + W
                         _L_ += [l[0] for f in _N_ for l in f.rim]
-                elif W := sum(w for f,w in F.rim if f.rooT is T) > ave:
+                        rT.N_ = []  # empty the merged T
+                elif F.rooT is None and (W := sum(w for f,w in F.rim if f.rooT is T)) > ave:  # we need bracket, else W is boolean from > ave
                     T.N_ += [F]; T.w += W; F.rooT = T
-                    _L_ += F.rim
+                    n_ = [nt[0] for nt in L_]
+                    _L_ += [nt for nt in F.rim if nt[0] is not _F and nt[0] not in n_]  # rim assignment is bidirectional, prevent cyclic here
```
kwcckw, line 251: Since rim assignment is bidirectional, we to need to exclude the same rim, such as `_F` or the same `F`s in `L_`.

frame_2D_alg/meta_code.py
```diff
@@ -234,29 +234,31 @@ def clust_oF_():  # flood-fill by rim links, cluster_N form: frontier expansion
     for _F in sorted(F_, key=lambda F: F.w, reverse=True):
         if _F.rooT: continue
         if _F.w <= ave: break
-        T = CoF(N_=[_F], typ=_F.typ, caller_=copy(_F.caller_)); _F.fin = 1; _F.rooT = T
```
kwcckw, line 237: I think fin is no longer needed now so it should be removed.

frame_2D_alg/meta_code.py
```diff
@@ -234,29 +234,31 @@ def clust_oF_():  # flood-fill by rim links, cluster_N form: frontier expansion
...
             for F,w in L_:
-                if rT := F.rooT and rT is not T and rT not in T.N_:  # merge clusters by cross-link density
+                if (rT := F.rooT) and rT is not T and rT not in T.N_:  # merge clusters by cross-link density
                     if W := sum(w for N in T.N_ for f,w in N.rim if f.rooT is rT) > ave:
-                        for f in _N_ := rT.N_: f.rooT = T
+                        for f in (_N_:= rT.N_): f.rooT = T
                         T.N_ += _N_; T.w += rT.w + W
                         _L_ += [l[0] for f in _N_ for l in f.rim]
-                elif W := sum(w for f,w in F.rim if f.rooT is T) > ave:
+                        rT.N_ = []  # empty the merged T
```
boris-kz, line 247: rT is removed anyway?

frame_2D_alg/meta_code.py
boris-kz: the way it's done here, the links should contain different F, not both, so they are not the same for both Fs?

frame_2D_alg/meta_code.py
boris-kz: right

frame_2D_alg/meta_code.py
kwcckw: If we don't empty their N_, this may true and we may pack them into oF_:

    for T in T_:
            form_body(T)

frame_2D_alg/meta_code.py
kwcckw: If we need the links to contain different F, we need only single side rim assignment:

    _F.rim += [(F,w)]; F.w += w; _F.w += w

Otherwise we always getting back the same F since `_F`'s rim contains `F` and `F`'s rim contains `_F`, which is cyclic.

frame_2D_alg/meta_code.py
boris-kz: ok right

frame_2D_alg/meta_code.py
boris-kz: yes, will that work?

frame_2D_alg/meta_code.py
boris-kz: have you verified the rest of form_body?

kwcckw: Mostly yes, except this section:

            Bod[-1] = (ast.IfExp, *[(f_, op_+_op_) for (f_,op_),(__,_op_) in zip(Bod[-1][1:], t[1:])])
            
I don't see why we need to merge across different index's body even they are the same.

frame_2D_alg/meta_code.py
kwcckw: Actually no, when rim is single direction, we may miss out some Ns if we check from a single direction only. I think we still need bidirectional rim and check it here.

boris-kz: ok, that's the reason we have bidirectional links, so they are node-neutral

frame_2D_alg/meta_code.py
boris-kz: there is so much crap in it, probably easier to write from the scratch

frame_2D_alg/meta_code.py
boris-kz: def clust_oF_():  # flood-fill by rim links, cluster_N form: frontier expansion + cluster contact-merge

    F_ = copy(oF_); T_ = []
        if _F.typ == F.typ:
                _F.rim += [(_F,F,w)]; F.rim += [(_F.F,w)]; F.w += w; _F.w += w
        if _F.rooT: continue
        if _F.w <= ave: break
        T = CoF(N_=[_F], typ=_F.typ, caller_=copy(_F.caller_)); _F.rooT = T
        L__ = L_ = list(_F.rim)
        while L_:  # extend frontier
            _L_ = []  # new frontier
            for _f,f, w in L_:  # current frontier
                F = f if _f is _F else _f
                l_ = [l for l in (F.rooT.L_ if F.rooT else F.rim) if l not in L__]
                if (W := sum(l[2] for l in l_)) > ave:  # merge clusters by cross-link density
                    T.w += W; _L_ += l_; L__ += l_
                    for n in [n for l in l_ for n in (l[0],l[1])]:
                        n.rooT.N_ = []; n.rooT = T; T.N_ += [n]
            L_ = list(set(_L_))
        if len(T.N_) > 1: T_ += [T]
    for T in T_:
            form_body(T)
            if (C := T.cmpr) > ave:
            else:
                for F in T.N_: F.rooT = None

frame_2D_alg/meta_code.py
kwcckw: Okay, so I think we still can preserve the batch based merging structure, since it's looks correct to create a fork on all ops instead of creating new fork on each new merging.

frame_2D_alg/meta_code.py
boris-kz: Understood — review only.

On your last draft (the `L__` version), against the cluster_N template:

1. `_F.rim += [(_F,F,w)]; F.rim += [(_F.F,w)]` — second is a typo (`_F.F`), and the two entries diverge: one 3-tuple, one 2-tuple. Both unpack sites assume 3.

2. `F = f if _f is _F else _f` — resolves the outside endpoint only for seed links. Once the frontier extends, `_f` is some interior member, the test fails, and you get F = interior node: the gate then evaluates and re-absorbs T's own members. Endpoint selection must key on rooT-membership, not seed identity.

3. The W gate counts `l_` = candidate's rim minus already-seen links — that's F's *outward* mass to third parties, while the F↔T links that are the actual admission evidence sit in L__ and are excluded. Inverted evidence: a hub passes on any weak contact, a tightly-bound F with few outside links fails.

4. Absorption loop `for n in [n for l in l_ for n in (l[0],l[1])]` pulls both endpoints of queued links — third-party nodes enter T ungated, each skipping its own frontier evaluation.

5. `n.rooT.N_ = []; n.rooT = T` — crashes when n.rooT is None (unclustered F), and on the merge path it empties rT without transferring members: rT's non-frontier-adjacent nodes and its accumulated w vanish.

6. `L__ += l_` grows unbounded and duplicates in_'s role; membership tests against a list are linear — your cluster_N in_ is a set for this reason.

7. Failed-render release clears rooT but not w, harmless only because the seed loop is single-pass — flag if you intend re-seeding within the pass.

On the consume-and-rescan port (your cluster_N excerpt semantics): `T.w += w` at absorption undercounts — F's other links into T are consumed as internal without their mass ever landing; iM is computed for the gate anyway, so accumulate that.

kwcckw: With this method:

    l_ = [l for l in (F.rooT.L_ if F.rooT else F.rim) if l not in L__]

We need to pack L_ into T?

Also i think this may reset the same T.N_ when n.rooT is T?

    n.rooT.N_ = []; n.rooT = T; T.N_ += [n]

boris-kz: reset is not a problem?

    F_ = copy(oF_); T_ = []
        if _F.typ == F.typ:
        if _F.rooT is not None: continue
        if _F.w <= ave: break
        T = CoF(N_=[_F], typ=_F.typ, caller_=copy(_F.caller_)); _F.rooT = T
        L_ = list(_F.rim)
        while L_:  # extend frontier
            _L_ = []  # new frontier
            for _f,f, w in L_:  # current frontier
                F = f if _f in T.N_ else _f; inL_,exL_ = [],[]; W=0
                for l in F.rooT.L_ if F.rooT is not None else F.rim:
                    _r, r, w = l
                    if _r in T.N_ or r in T.N_: inL_ += [l]; W += w
                    else: exL_ += [l]
                if W > ave:  # merge clusters by cross-link density
                    T.w += W; _L_ += exL_; T.L_ += inL_
                    for n in [n for l in inL_ for n in (l[0],l[1])]:
                        n.rooT.N_ = []; n.rooT = T; T.N_ += [n]
            L_ = list(set(_L_))
        if len(T.N_) > 1: T_ += [T]
    for T in T_:
            form_body(T)
            if (C := T.cmpr) > ave:
            else:
                for F in T.N_: F.rooT = None

kwcckw: Actually I think it's clearer to separate the section for rooT is None or rooT is available? Since the merging process from rooT is actually different with the packing of F.rim.

boris-kz: merging actually packs ns, I think it should work anyway:

                if W > ave * len(inL_+exL_):  # merge clusters by cross-link density
                    T.w += W; _L_ += exL_; T.L_ += inL_
                    for n in [n for l in inL_ for n in (l[0],l[1]) if n not in T.N_]:
                        n.rooT.N_ = []; n.rooT = T; T.N_ += [n]

kwcckw: Okay, so this `if n not in T.N_` should resolve the issue. I will check again later.

frame_2D_alg/meta_code.py
boris-kz: so please try that

PR #202, opened 2026-07-11
7/11/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -306,6 +309,19 @@ def form(col, Fn):  # col: [(op,F)] -> bare op | merged op | fork
...
     F.body = Bod
 
+def form_body(F):
+    
+    _body_ = [F.body for F in F.N_]
+    max_bodyl = max(len(body) for body in _body_)
+    Body=[]
+    for i in range (max_bodyl):
+        ibody_ = [(F.N_[j], body) for j,body in enumerate(_body_) if len (body)>i]  # i's index bodies and their F from all Fs
+        if len(ibody_)>1: Body += [(ast.IfExp,*ibody_)]  # in the format of (ast.IfExp, (f,t)...)
```
kwcckw, line 319: I added a non recursive version of form_body. 

In the claude version, it has recursive part to merge same index's IfExp forks recursively, so that it reduces the number of IfExp node. To merge those IfExp forks, we will be concatenating their (f, sub)? Something like this?

    (IfExp, (f1, sub1)) + (IfExp, (f2, sub2)) = (IfExp, (f1, sub1) , (f2, sub2))

Other than IfExp, should we merge same ops nodes too?

frame_2D_alg/meta_code.py
```diff
@@ -238,14 +238,17 @@ def clust_oF_():  # flood-fill by rim links, cluster_N form: frontier expansion
             _L_ = []  # new frontier
             for _f,f, w in L_:  # current frontier
                 F = f if _f in T.N_ else _f; inL_,exL_ = [],[]; W=0
+                if F.rooT is T: continue  # skip same root
```
boris-kz, line 241: this is the frontier, F can't have the same rooT

frame_2D_alg/meta_code.py
```diff
@@ -238,14 +238,17 @@ def clust_oF_():  # flood-fill by rim links, cluster_N form: frontier expansion
...
                 for l in [l for n in (F.rooT.N_ if F.rooT is not None else [F]) for l in n.rim]:
                     _r, r, w = l
-                    if _r in T.N_ or r in T.N_: inL_ += [l]; W += w
+                    if _r in T.N_ and r in T.N_: continue  # if both Fs are included, skip?
```
boris-kz, line 244: I don't think that's possible

frame_2D_alg/meta_code.py
boris-kz: we need an eval to wrap fork in ifExp if fc>ave, just concat in the body
all those ts should be ns

kwcckw: The IfExp fork should have multiple (f,sub)s, do you mean eval them to add nested IfExp fork?

frame_2D_alg/meta_code.py
kwcckw: In the first iteration, `f` or `_f` might be the added rim below?

    _r, r, w = l
    
And they might be clustered?

frame_2D_alg/meta_code.py
kwcckw: When we have A,B, C, D, we may add `A,B` from `AB` pair,` C,D` from `CD` pair, and eval the same B,C pair again?

frame_2D_alg/meta_code.py
boris-kz: I meant the opposite, the ops that are not conditional, so they concut directly in the body

frame_2D_alg/meta_code.py
```diff
@@ -296,7 +299,7 @@ def form(col, Fn):  # col: [(op,F)] -> bare op | merged op | fork
         return (ast.IfExp, *branch_)  # always form branch when there's at least 2 same length Fs
     Bod = []
     N_ = F.N_; bod_= [F.body for F in N_]
-    for j in range(max(len(b) for bod in bod_)):
+    for j in range(max(len(bod) for bod in bod_)):
```
boris-kz, line 302: I don't get this whole expression?

frame_2D_alg/meta_code.py
boris-kz: L_ is the frontier, _f and f must be on the opposite sides of that frontier

frame_2D_alg/meta_code.py
kwcckw: This is looping body based on the number of the max body across Fs,so that it can retrieve j's index body across all Fs for batch-based merging purpose.

frame_2D_alg/meta_code.py
kwcckw: I see, that's mean we only need to concat the nested IfExp across the body, i will include that in the next update.

frame_2D_alg/meta_code.py
kwcckw: But the problem is any f could be pair with any other fs in other rim?

So in the subsequent main loop `F_`, some of them are actually clustered.

boris-kz: how about this:

```python
T = CoF(N_=[_F], L_=set(_F.rim), typ=_F.typ, caller_=copy(_F.caller_)); _F.rooT = T
L_ = set(_F.rim)
while L_:  # extend frontier
    _L_ = []  # new frontier
    for _f,f,_ in L_:  # current frontier
        F = f if _f.rooT is T else _f
        if F.rooT is T: continue  # link internalized since queued
        r = F.rooT; n_,inL_,exL_ = [],[],[]; W = 0
        for l in (r.L_ if r is not None else F.rim):
            _n, n, w = l
            if _n.rooT is T or n.rooT is T:  # cross link
                inL_ += [l]; W += w; n_ += [n if _n.rooT is T else _n]
            else: exL_ += [l]  # outward | r-internal: denominator + new frontier
        if W > ave * len(inL_+exL_):  # cross-link density
            T.w += W; _L_ += exL_; T.L_.update(inL_+exL_)
            for n in n_:
                if n.rooT is T: continue  # dup endpoint across cross links
                if (r := n.rooT) is not None: r.N_ = []  # lossy: non-adjacent members orphan
                n.rooT = T; T.N_ += [n]
    L_ = set(_L_)
```

Notes: endpoint and internalized tests via rooT identity — O(1), equivalent to `in T.N_` since rooT is set at every admission. One behavior consequence kept as-is: the wipe clears r.N_ but not r.L_, so an emptied r reached again through a later frontier link scans its stale rim and can readmit its orphans piecemeal — if the loss is meant to be final, wipe r.L_ in the same line; if orphan recovery is acceptable, current form already does it.

kwcckw: So we not really need T.L_ here? 

In this section, i still think both `n` and `_n` may have the same root = T because both of them might be added from prior link, with different pair of ns:

      if _n.rooT is T or n.rooT is T:  # cross link
          inL_ += [l]; W += w; n_ += [n if _n.rooT is T else _n]

boris-kz: yes, because we may be merging a whole cluster at once. But I am not actually sure that's a good idea, seems too coarse

boris-kz: Coarse in a specific way: evidence/commit mismatch. The gate aggregates over r's entire boundary, then commits members the evidence never individually covered — every other admission here is per-element gated (iM). The adjacent-only+wipe variant is worse, not better: evidence aggregated over all of r, commit excludes the members that contributed denominator mass, and the orphaning isn't filter-loss — orphans aren't below any ave, they're collateral.

And I should retract my original justification for contact-merge ("later T reaching earlier T' is evidence they're one cluster"): with a selective density gate that inverts. Contact happens precisely where a candidate failed T1's gate and seeded T2 — the T1↔T2 boundary is sub-ave *by selection*. Contact evidence is weak exactly when contact occurs.

The level argument settles it: merging clusters as units is a next-composition decision — clusters become nodes, evidence becomes comp over their merged bodies, and they cluster there if they actually match. That's the hierarchical cycle you're building anyway (composite-as-member, form over fork-bearing bodies — already the open contract). Within-pass contact-merge duplicates that future function with node-level links as proxy evidence. So: cluster_N's exclusive form was right — `if F.rooT is not None: continue`, one skip covering internalized and foreign-cluster both; the whole r branch, the wipe, n_, and the T.L_/reconstruction question all delete with it. The loop becomes the straight cluster_N port. Under-merge this pass self-corrects at the next level.

Middle option if within-pass correction ever matters: per-node poaching — contacted F transfers alone if its mass into T beats its mass into r (relative iM, node-level gate and commit, granularity-consistent). Discriminant is whether seeding-order misassignment accumulates enough to matter before the next cycle corrects it; at per-typ pools of 3–5 funcs, it can't.

Me: 
this is different from cluster_N because spatial proximity prior does not apply.
that means we should be doing fuzzy assignment here?

Yes — and this diagnosis retires the whole flood apparatus, not just the merge branch. cluster_N's exclusivity and frontier are both licensed by the spatial prior: links exist only between proximate nodes, so candidates must be *reached* (frontier walk), and clusters partition space, so membership is exclusive (a node is in one place). Here rims are a complete same-typ similarity graph — every candidate is one hop away, so "frontier" spans the pool in one wave, and exclusive first-commit is pure seeding-order artifact: the ranking-dependence you already rejected, resurfacing through the borrowed form. Similarity without locality is cluster_C's domain.

What the port takes from cluster_C:

1. **Plural graded membership**: F.root_ as (T, C) list — the field CN already reserves ("reciprocal roots, Cs not Bs"). rooT, fin, wipe, orphaning all go.
2. **Reform loop replaces expansion**: seed composites from top-w exemplars, evaluate every same-typ F against every composite per pass, re-render, iterate until membership deltas subside (Up analog). Simpler than cluster_C, which still needs rim-walked frontiers for candidate generation — here the candidate set is just the typ pool. Order-free by construction: refinement erases seeding, which finally makes "ranking doesn't matter" structural rather than assumed.
3. **Exemplar suppression** has a similarity-space analog: don't seed an F that's already a strong match of a stronger seed — NMS with match replacing distance.

Membership evidence target: the true centroid is the merged body, so refinement wants comp_body against fork-bearing composites — the open composite-as-member contract becomes this form's first consumer, not a future cycle's. Bridge until it exists: mean pairwise C to T.N_ from rims — zero new comparisons, and it is exactly your earlier average-linkage proposal. That was the right borrow; the flood was the wrong one.

Overlap needs no exclusivity rule — rdn resolves it at the realized gate: memberships ordered by C, F's fc savings credited once to the top membership; in lower composites F contributes branch cost minus matched mass, so it survives only where nearly free (mostly coincides with existing branches). Exclusive-ish assignment emerges, genuine overlap persists where cheap — same accounting as cluster_C's olp-as-rdn.

Contact-merge's one real function — correcting boundary misassignment from seeding order — is what reassignment does natively; cluster-level merge proper stays a next-composition decision.

kwcckw: The fuzzy method is the same membership approach for each body in F? Each body is assigned with soft membership scores of different clusters, and we will merge the actual body only in the final step after the membership finalization?

boris-kz: right, with possibly multiple assignments if the weaker ones still justify the redundancy

kwcckw: > right, with possibly multiple assignments if the weaker ones still justify the redundancy

So the weaker node could be in multiple oFs? But during the runtime, the line of code from that node will be run more than a single time from multiple functions?

boris-kz: yes, in principle we may have multiple functions preforming the same operation

kwcckw: i see, previously we use max weight to select the final membership of each member, so with multiple assignments here, it's will be eval purely based on redundancy? As long each member's rdn < ave, they will be included as the final member?

frame_2D_alg/meta_code.py
```diff
@@ -238,14 +238,17 @@ def clust_oF_():  # flood-fill by rim links, cluster_N form: frontier expansion
             _L_ = []  # new frontier
```
boris-kz, line 238: Didn't check it yet:

```python
def clust_oF_():  # centroid form: exemplar seeds, fuzzy membership by mean link C, reform to convergence

    F_ = copy(oF_); T_ = []
    for F in F_: F.rim = []; F.w = 0; F.root_ = []
        if _F.typ == F.typ:
                L = (_F,F,w); _F.rim += [L]; F.rim += [L]; _F.w += w; F.w += w
    E_ = []
    for F in sorted(F_, key=lambda F: F.w, reverse=True):  # exemplars: NMS in similarity space
        if F.w <= ave: break
        if not any((n if _n is F else _n) in E_ for _n,n,w in F.rim): E_ += [F]
    for E in E_:
        T = CoF(N_=[E], typ=E.typ); E.root_ = [(T,E.w)]; T_ += [T]  # seed C = E.w: pass-1 stand-in
    upd = 1
    while upd:  # reform memberships to convergence
        upd = 0
        for F in F_:
            r_ = []
            for T in T_:
                if F.typ == T.typ:
                    n_ = [G for G in T.N_ if G is not F]
                    C = sum(w for _n,n,w in F.rim if (n if _n is F else _n) in n_) / len(n_) if n_ else F.w
                    if C > ave: r_ += [(C,T)]
            r_.sort(key=lambda t: t[0], reverse=True)
            root_ = [(T,C) for i,(C,T) in enumerate(r_) if C > ave*(i+1)]  # rdn-scaled gate per extra root
            if [t for t,_ in root_] != [t for t,_ in F.root_]: upd = 1
            F.root_ = root_
        if upd:
            for T in T_: T.N_ = [F for F in F_ if any(t is T for t,_ in F.root_)]
            T_ = [T for T in T_ if T.N_]
    for T in sorted(T_, key=lambda T: sum(C for F in T.N_ for t,C in F.root_ if t is T), reverse=True):
        if len(T.N_) > 1:
            C, body = form_body(T.N_)  # realized: sum member fc - merged fc
            fc = sum(F.fc for F in T.N_) - C
            C -= sum(F.fc for F in T.N_ if F.root_[0][0] is not T)  # rdn: member fc credited once, at primary root
            if C > ave:
                T.body = body; T.fc = fc; T.w = C
                T.caller_ = set().union(*[F.caller_ for F in T.N_])
            else:
                for F in T.N_: F.root_ = [r for r in F.root_ if r[0] is not T]  # release: next roots promote
```

Decisions in it:

1. Membership C = mean link w over *all* of T.N_, zero-filling unlinked members — coverage penalty is the point of average linkage; dividing by linked count only would reward partial fit.
2. rdn is two-stage: at assignment, the gate scales per extra root (second membership needs > 2·ave); at render, member fc is credited only to the primary root, so a duplicate or parasitic composite strips to C ≤ 0 and self-releases. Render order strongest-first + release keeps the credit chain consistent: a failed T's members promote their next root before weaker Ts render.
3. Seed persistence: singleton T has no mean, so pass-1 self-membership uses E.w — wrong scale (sum vs mean), harmless for one pass, dies at first reform.
4. rooT, fin, frontier, contact-merge, wipe — all gone; originals persist, release is pure bookkeeping.

Open: oscillation cap on the reform loop if a 2-cycle ever shows (flat structure makes it possible, rdn gating makes it unlikely); membership target upgrades from rim-mean to comp_body against T.body once fork-aware comp exists — one line swaps in the eval.

kwcckw: The C above is computed usng the mean link w, and I think that's correct since it measures the possible overlapping roots using w.The other params such as c and dTT may not be directly relevant here?

boris-kz: yes, all the data params are stale

kwcckw: Okay, I will draft another simpler version later.

PR #203, opened 2026-07-12
7/12/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -286,12 +287,33 @@ def comp_body(_n, n):  # compare only: compression estimate C; construction in f
...
+            
+    _body_ = [f.body for f in F.N_]
     Bod = []
     for i in range(max(len(b) for b in _body_)):
-        ibod_ = [(F.N_[j], body) for j, body in enumerate(_body_) if len(body) > i]  # i's index bodies and their F from all Fs
+        ibod_ = [([F.N_[j]], body[i]) for j, body in enumerate(_body_) if len(body) > i]  # i's index bodies and their F from all Fs
         if len(ibod_) > 1:
-            Bod += [(ast.IfExp, *ibod_)]  # in the format of (ast.IfExp, (f,t)...)
+            form_forks(ibod_)  # ibod_ is [([f],t),([f],t)...]
+            Bod += [(ast.IfExp, *ibod_)]  # in the format of (ast.IfExp, ([f],t)...)
```
kwcckw, line 316: If we need to merge IfExp forks, looks like we need `(f,t)` in the format of `([f],t)` instead since we may get different F on each different IfExp fork.

frame_2D_alg/meta_code.py
```diff
@@ -225,39 +225,40 @@ def call_sites(fd):  # FunctionDef
...
                 L = (_F,F,w); _F.rim += [L]; F.rim += [L]; _F.w += w; F.w += w
-    E_ = []
+    T_ = []
     for F in sorted(F_, key=lambda F: F.w, reverse=True):  # exemplars: NMS in similarity space
-        if F.w <= ave: break
-        if not any((n if _n is F else _n) in E_ for _n,n,w in F.rim): E_ += [F]
-    for E in E_:
-        T = CoF(N_=[E], typ=E.typ); E.root_ = [(T,E.w)]
-        T_ += [T]  # seed C = E.w: pass-1 stand-in
+        # if F.w <= ave: break
```
kwcckw, line 235: Sorry, this is still needed.

frame_2D_alg/meta_code.py
```diff
@@ -225,39 +225,40 @@ def call_sites(fd):  # FunctionDef
...
+        if not any((n if _n is F else _n).root_ for _n,n,w in F.rim):    
+            T = CoF(N_=[F], typ=F.typ); F._root_ = [(T,F.w)]
+            T._N_ = list({(_f if f is F else f) for _f,f,w in F.rim})  # frontier
```
kwcckw, line 238: So i edited the code to use frontier via rim instead of checking all Fs. The C computation is remained the same and the iterations stop when we have constant roots in Fs after the refinement.

boris-kz: There's no real frontier here, proximity only matters in space-time?

kwcckw: Yes, so for oFs, frontier should be those unchecked Fs, and they should have the same typ and high compression value, added from this section:

        if _F.typ == F.typ:
                L = (_F,F,w); _F.rim += [L]; F.rim += [L]; _F.w += w; F.w += w
                
By looping all Fs, it's actually redundant to check for typ and w again.

boris-kz: they are all checked in comp_body. The only way to define frontier is to sort F.rim by cmpr, so you don't need `T._N_`

boris-kz: What it recovers: proximity's role as *search prior*. In space, nearest-first ordering is cheap and predicts match (decay). Here cmpr-rank is that ordering — strongest-first traversal, deterministic in link weights rather than in seeding or pop order. Global form is cleaner than per-T frontiers: one edge list sorted by cmpr descending, union-find over components, gate per edge — Kruskal-style, fully order-determined by the data. A per-T sorted frontier reintroduces seed choice as an order source; the global sort doesn't, so it fits "ranking doesn't matter" better than any flood we drafted.

What it can't recover: the exclusivity license. Components still partition, and a boundary F linking two components comparably is still assigned by accident of rank ties — that's the part fuzzy membership owns, unchanged.

The interesting point is that connectivity actually fits the *objective* better than the centroid form does for formation: realized compression is transitive — comp_N~comp_C ops overlap, comp_C~comp_F ops overlap, and the merged body of all three compresses by the sum of shared substructure even if comp_N~comp_F directly match little. Mean-linkage membership under-admits exactly those chains; the realized cmpr gate would have accepted them. Chaining, the classic single-linkage pathology, isn't pathological when the objective is additive over shared ops and every admission is validated by realized ΔC.

So the two forms split by function, which is just agg's order recovered with rank replacing distance: connectivity by sorted cmpr for cluster formation (transitive objective, single pass, gate = link cmpr or realized ΔC per union), then centroid/fuzzy as refinement — exemplars per component, membership by mean C or body-C, overlap with rdn — owning boundary assignment and dispatch fit, which are not transitive. The reform loop then starts near-converged instead of from raw seeds, and the contact-merge question dissolves: both-clustered edges are exactly what the refinement stage adjudicates.

One scaling note: sorting presupposes cmpr, so it can't reduce the O(n²) comp_body cost the way spatial locality reduces comparison count. The analog there is incremental resolution, not incremental distance: cheap coarse rank first (comp_callers overlap, fc bands), fine comp_body only on the top slice. Irrelevant at 3–5 per typ, but it's the structural answer if pools grow.

frame_2D_alg/meta_code.py
```diff
@@ -225,39 +225,40 @@ def call_sites(fd):  # FunctionDef
...
-
+                T.N_ = [F for F in F_ if any(t is T for _,t in F.root_)]  # rebuild T.N_ from F.root_
+                T._N_ = list({(n if _n is F else _n) for F in T.N_ for _n,n,w in F.rim} - set(T.N_))  # extend frontier via member rims
+            T_ = [T for T in T_ if T.N_]  # remove if no new root
+        for F in F_: F._root_ = F.root_; F.root_ = []  # reset
+    # not updated
     for T in sorted(T_, key=lambda T: sum(C for F in T.N_ for t,C in F.root_ if t is T), reverse=True):
         if len(T.N_) > 1:
-            fc, cmpr, bod = form_body(T)
+            fc, cmpr, bod = form_body(T,0)
```
boris-kz, line 261: so cmpr here is L.w, not per F
it's computed in comp_body, we don't need it here?

frame_2D_alg/meta_code.py
kwcckw: > they are all checked in comp_body. The only way to define frontier is to sort F.rim by cmpr, so you don't need `T._N_`

Okay, but a single F.rim doesn't cover all possible rim, so it should recursively retrieve all Fs' rim first, and then only sort it?

frame_2D_alg/meta_code.py
kwcckw: I think both are computing compression, it's just here we are using the sum oF 's fc - T's new body's fc, while in comp_body it's computing the compression side by side using their nodes, which is more precise. 

But when we merge oFs into T, we can't retrieve per T's compression since it's pack per F pair in rim only?

frame_2D_alg/meta_code.py
boris-kz: it's all-to-all if we remove typ checking

frame_2D_alg/meta_code.py
boris-kz: it will be in T.L_

frame_2D_alg/meta_code.py
kwcckw: We don't need that same type clustering? But if we merge across types, we need to define again the new type?

boris-kz: the types are the clusters, it's the same thing

kwcckw: Okay, some it will be something like this?

        for T in T_:
            for _F in T.N_:
                for l in sorted(_F.rim, key=lambda rim: rim[2], reverse=True):
                    if l[2]<ave: break
                    F = l[1] if [0] is _F else l[2]
                    n_ = [f for f in T.N_ if f is not F]
                    C = sum(w for _n,n,w in F.rim if (n if _n is F else _n) in n_) / len(n_) if n_ else F.w
                    if C > ave: F.root_ += [(C,T)]

boris-kz: I just pushed:

    F_ = copy(oF_); T_ = []
        if (w := comp_body(_F.body, F.body)) > ave:  # compression
        if _F.w <= ave: break
        w_,N_,L_ = [_F.fc], [_F], set(_F.rim)
        for _f,f, w in sorted(_F.rim, key=lambda l: l[2], reverse=True):
            F = f if _f.root_[-1] is T else _f
            if w > ave * len([rw for rw in F.rw_ if rw>w]):  # rdn = n stronger memberships
                w_+=[w]; N_+=[F]; L_.update(F.rim)
        if len(N_) > 1 and (W :=sum(w_)) > ave:
            T = CoF(w=W, N_=N_, L_=L_)
            for f,w in zip(N_,w_): f.root_+=[T]; f.rw_+=[w]
            T_ += [T]
    for T in T_:
            form_body(T)
            if (C := T.cmpr) > ave:
            else:
                for F in T.N_: F.root_.pop()

frame_2D_alg/meta_code.py
boris-kz: sorry I haven't checked that yet

frame_2D_alg/meta_code.py
kwcckw: In your version:

      for _f,f, w in sorted(_F.rim, key=lambda l: l[2], reverse=True):
            F = f if _f.root_[-1] is T else _f
            if w > ave * len([rw for rw in F.rw_ if rw>w]):  # rdn = n stronger memberships
                w_+=[w]; N_+=[F]; L_.update(F.rim)
                
We don't have T yet, F is the f that's not in the local N_ var?

Also we don't have F.rw_ yet in the first iteration?

So the following section will never be true:

        if len(N_) > 1 and (W :=sum(w_)) > ave:
            T = CoF(w=W, N_=N_, L_=L_)
            for f,w in zip(N_,w_): f.root_+=[T]; f.rw_+=[w]
            T_ += [T]

boris-kz: it's in the init. I will check tomorrow

claude:
Deferred-commit fixes 2/3/6-adjacent problems at once — root_/rw_ written post-loop, so rdn counts can't see the forming T and release order is per-T append... except release still pops blind (see 4). Against this draft:

1. **`_f.root_[-1] is T` — T is unbound** (this-T doesn't exist until after the loop; first seed: NameError, later seeds: stale T from a previous iteration). With deferred commit the resolution can't reference T at all — membership in the *forming* cluster is N_: `F = f if _f in N_ else _f`. That also generalizes correctly if the loop ever extends past the seed's rim, where `_f is _F` wouldn't.

2. **L_ is collected but never traversed** — the loop still runs over `sorted(_F.rim)` only, so L_.update(F.rim) accumulates a frontier that nothing consumes: one-hop stars again, chains still invisible (my #4 last round, unaddressed). Either iterate the growing set — replace the for with a sorted-pop over L_, re-sorting or heapifying as it grows, skipping links with both endpoints in N_ — or drop L_ from formation and keep it only as T's rim record. As drafted it's the latter with wasted set-builds inside the admission branch.

3. **w_[0] scale clash persists** (my #5): _F.fc seeds w_, so W mixes a cost into a sum of matches and the W > ave gate is inflated by exactly fc — a seed with one weak link passes on fc alone. Seed's stand-in should be F.w-mean-scale or just 0 with the gate on links only: `W := sum(w_[1:])`.

4. **Blind pop now misfires more, not less**: with multi-membership, F absorbed into T1 and T2 releases from T2 → pop removes T2 iff it was last-appended — true only if no later T admitted F. T_ renders in formation order, releases happen during render — an F in T3 (failed) and T5 (pending) pops T5's entry if T5 formed later. Identity strip both lists: `i = F.root_.index(T); F.root_.pop(i); F.rw_.pop(i)`.

5. **rw_/root_ never reset** in the init loop — stale memberships from prior passes inflate rdn counts and misdirect release. `F.root_ = []; F.rw_ = []` next to rim reset; rooT init is now the dead one.

6. Same open choices carried: every above-ave F seeds (no suppression — each strong pair forms twice as two stars unless rdn gate kills the second; it does only if w ties break right), and typ guard still absent from the pair loop.

kwcckw: Thanks, this should be helpful.

kwcckw: But F.rw is init empty, so this will never get true:

    if w > ave * len([rw for rw in F.rw_ if rw>w]):
    
And we only append rw_ when the eval above is true.

So i think ` len([rw for rw in F.rw_ if rw>w])` should be use only if F.rw_ is not empty.

boris-kz: It's len -1

PR #204, opened 2026-07-13
7/13/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -225,29 +225,28 @@ def call_sites(fd):  # FunctionDef
...
         if (w := comp_body(_F.body, F.body)) > ave:  # compression
             _F.rim += [(_F,F,w)]; F.rim += [(_F,F,w)]; F.w += w; _F.w += w
     for _F in sorted(F_, key=lambda F: F.w, reverse=True):
         if _F.w <= ave: break
         w_,N_,L_ = [_F.fc], [_F], set(_F.rim)
         for _f,f, w in sorted(_F.rim, key=lambda l: l[2], reverse=True):
-            F = f if _f.root_[-1] is T else _f
-            if w > ave * len([rw for rw in F.rw_ if rw>w]):  # rdn = n stronger memberships
+            F = f if _f in N_ else _f
+            if w > ave * (len([rw for rw in F.rw_ if rw>w])-1):  # rdn = n stronger memberships
```
kwcckw, line 237: > It's len -1

Thanks, i was confused earlier.

Also the current clust_oF_ is actually no longer the centroid method with refinement, and it's just a one pass clustering with w?

boris-kz: Yes, centroid clustering will be the next step, once the clustering hierarchical. It doesn't matter make sense for destructive merging

kwcckw: > clustering hierarchica

Hierarchical clustering here would be reclustering those Ts? But that will be in the next feedback?

boris-kz: Yes, and it will have an option to pack element oFs instead of merging their bodies.
but that only makes sense when we have a lot of oFs, I am still figuring how new ops can be added by projection, basically a recursion of existing ops

frame_2D_alg/meta_code.py
```diff
@@ -225,29 +225,28 @@ def call_sites(fd):  # FunctionDef
...
-    for F in F_: F.rim = []; F.rooT = None
+    for F in F_: F.rim = []; F.root_ = []; F.rw_ = []
     for _F,F in combinations(F_,2):
         if (w := comp_body(_F.body, F.body)) > ave:  # compression
             _F.rim += [(_F,F,w)]; F.rim += [(_F,F,w)]; F.w += w; _F.w += w
     for _F in sorted(F_, key=lambda F: F.w, reverse=True):
         if _F.w <= ave: break
         w_,N_,L_ = [_F.fc], [_F], set(_F.rim)
         for _f,f, w in sorted(_F.rim, key=lambda l: l[2], reverse=True):
-            F = f if _f.root_[-1] is T else _f
```
boris-kz, line 236: what's wrong here?

frame_2D_alg/meta_code.py
```diff
@@ -225,29 +225,28 @@ def call_sites(fd):  # FunctionDef
...
-            if w > ave * len([rw for rw in F.rw_ if rw>w]):  # rdn = n stronger memberships
+            F = f if _f in N_ else _f
+            if w > ave * (len([rw for rw in F.rw_ if rw>w])-1):  # rdn = n stronger memberships
                 w_+=[w]; N_+=[F]; L_.update(F.rim)
         if len(N_) > 1 and (W :=sum(w_)) > ave:
-            T = CoF(w=W, N_=N_, L_=L_)
+            T = CoF(w=W, N_=N_); T.L_ = L_; T.cmpr = sum(L[2] for L in L_ if (L[0] in N_ and L[1] in N_))  # cmpr computed with only L with both Ns in N_?
```
boris-kz, line 240: I guess we can use H instead of L_ here, it's not used in CoF anyway

frame_2D_alg/meta_code.py
boris-kz: we have gv_(), that returns a value but is much lighter that oF? Currently it's function is negative: to gate an op sequence, but it can be extended to positive: to iterate over that sequence?

boris-kz: **1. The positive extension is an If→While flip at the gv_ site, not a runtime change to gv_.** gv_ receives an already-evaluated scalar, so it can't re-run the test itself. But a While re-evaluates its test per pass by construction — so "iterate the sequence" = build/inject rewrites `if gv_(expr): body` to `while gv_(expr): body` at sites whose accounting justifies it. Single-node AST substitution, cost ~1 in your costs table, no new CoF, no new call. fill_frame already uses `while ... and gv_(...)`, so the flip generates an existing idiom rather than a new structure. This is the lightest generator output — below sub-oF and pack in the hierarchy: gv_ flip < split-to-sub-oF < pack/merge.

**2. Validity condition: the body must rebind operands of the gate expression.** Otherwise While runs zero or forever. Checkable statically at build time: assignment targets in the gated body intersect names in the test. For comp/cluster sequences this holds by construction — der+/rng+ always rebind (next iteration consumes previous output). Where it fails, the flip needs an inserted depth-cost term in the test, which is projection again — but the clean cases need nothing, because re-testing actual value per iteration is your compare-projected-to-actual principle applied directly: no predicted iteration count, just marginal gate per pass.

**3. Trigger signal requires symmetric accounting, which is a contract change.** gV_ currently records only misses (double-negative). The flip criterion is per-site realized surplus consistently exceeding body fc — invisible today, since the positive branch returns without recording and FV_ only has oF resolution. Storing surplus in the same slot collides with the miss convention (both would accumulate positive). So: second per-site accumulator, or signed sum with count. Downstream: build() inits gV_ slots, inject_oF_ resets them, and ffeedback does `oF.w = sum(oF.V_) + sum(oF.gV_)` — that line already mixes blocked value with realized value as worth; adding surpluses changes its meaning again. Needs a decision, not a patch.

**4. The flip creates body divergence unless comp_prim treats If/While as one prim.** After flipping one oF's site, its unflipped twin elsewhere mismatches (type inequality, C -= 2), suppressing exactly the merges that should still happen. Consistent fix within your framework: gate-headed If and While match structurally, iteration flag carried as a d — structure is match, loop-ness is a diff.

**5. Consistency check this passes:** the hand-written recursion in cross_comp (rr, elev loops) is what this generator would produce from flat code — the mechanism's output space contains the code you already wrote. The negative fork of the same accounting is already consumed by split_oF_ (is_gate_l keys on gv_-headed forks). So gv_ sites become the common generative currency: high miss → split out; high surplus → iterate; and only cross-context transfer forces promotion to oF.

kwcckw: > it will have an option to pack element oFs instead of merging their bodies

Could you clarify if the packing element oFs is packing their body only?

> new ops can be added by projection

Something like instead of comparing, split and cluster those oFs, we add new oFs by simply using the projection?

frame_2D_alg/meta_code.py
kwcckw: Their root_ is init empty, so we won't get any root here?

frame_2D_alg/meta_code.py
kwcckw: Okay, and for cmpr, it should be those rim with both Fs in T.N_? Since those are the internal w?

frame_2D_alg/meta_code.py
boris-kz: it could assigned in prior _F loop.
so, `if _f.root_ and _f.root_[-1] is T` ?

frame_2D_alg/meta_code.py
boris-kz: it will be the whole oF, body-only is merging
on projection, I posted from claude above

frame_2D_alg/meta_code.py
boris-kz: currently it's the same as W: sum L ws

frame_2D_alg/meta_code.py
kwcckw: > So gv_ sites become the common generative currency: high miss → split out; high surplus → iterate; and only cross-context transfer forces promotion to oF.

So something like gv_ will be used to test across all nodes or code, high miss = split the line or function, high surplus = continue to run the next line?

frame_2D_alg/meta_code.py
kwcckw: The prior F loop's T is for the terminated T, but here it's for current T, which T's N_ is N_, so we need to check with N_?

frame_2D_alg/meta_code.py
kwcckw: I see, then we could have just use `T.cmpr = W` here.

frame_2D_alg/meta_code.py
boris-kz: This is about recursion over that sequence. 
Existing methods:

Several established methods each match one role of gv_; none combine all three.

**1. Dijkstra's guarded commands — the syntactic form, including your proposed extension.** A guard gating a command sequence, with two composition forms: if-fi (execute a guarded alternative once) and do-od (iterate the same guarded set until all guards fail). The If→While flip is exactly the if-fi→do-od move, so the conditional↔iteration duality you're proposing has a formal precedent: iteration is the fixpoint of the guarded conditional, not a separate construct. No value accounting though — guards are pure booleans.

**2. Rational metareasoning (Russell & Wefald) — the semantic content.** Expected value of computation: execute a step iff estimated benefit minus cost is positive; anytime algorithms with myopic stopping rules iterate while marginal value of the next pass exceeds its cost. That is the meaning of `while gv_(V - cost)`. Their framing is decision-theoretic estimation before the step; yours re-measures actual value per pass, which is simpler and matches your projected-vs-actual principle.

**3. Profile-guided optimization and branch predictors — the per-site accounting.** gV_[i] indexed by static gate position is structurally a per-branch profile counter (saturating counters in hardware, edge profiles in compilers). PGO uses these to restructure code offline: hot/cold splitting (= split_oF_ on gate forks), inlining by call-site profile, loop transforms. Closest match to the accounting-drives-reform loop — but PGO optimizes cost only; the benefit side of the ledger doesn't exist there because ordinary code has no per-op predictive value.

**4. ACT-R production utilities — per-unit value-minus-cost gating without backprop.** Each production rule carries utility = expected gain minus cost, updated incrementally from outcomes, gating which rule fires. Closest cognitive-architecture analog, and notably gradient-free. Neural conditional computation (ACT/PonderNet halting units, MoE routing, early exit) does the iterate-while-gate-holds pattern, but gates are learned parameters trained end-to-end — the fitting you reject.

**5. The unclaimed part is the combination.** Metareasoning gates computation but never rewrites code; PGO rewrites code but only for speed; guarded commands have the structure but no ledger; ACT-R has the ledger but fixed rule syntax. A gate whose miss/surplus statistics feed structural reform (split, flip, merge) in the same value currency as the data clustering — that specific closure I can't point to an established method for. Worth checking program-synthesis literature on cost-driven rewriting if you want certainty, but the known systems there (superoptimization) are stochastic search, outside your constraints.

Me:
gv_ doesn't rewrite the code, it only provides the data for it. so it's the same as ACT-R?

Yes at that level — I drew the line in the wrong place last message. As a data-provider, gv_ is functionally ACT-R utility tracking: per-unit, gradient-free, value-minus-cost statistics from actual execution. And I understated ACT-R's reform side: production compilation does rewrite — rules that fire in sequence get merged into one, with utility competition deciding adoption. So ledger plus one reform operator exists there. Remaining differences that are genuine:

**1. The statistic is not your decision variable.** ACT-R fires the production with highest accumulated utility — the learned average *is* the gate. gv_ gates on fresh measured value of the current input; gV_/V_ accumulate purely for downstream reform and filter updates. ACT-R's ledger is the controller; yours is the observer. Consequence: ACT-R can misfire on stale statistics, your gate can't — it re-measures per call, and the ledger errs only in reform decisions, which are lower-frequency and reversible.

**2. No credit assignment.** ACT-R rewards are exogenous and propagate backward over recently fired rules with temporal decay — a credit-assignment scheme, with its usual smearing problems. gv_ needs none: match is locally measurable at each site, so the ledger is defined per-op without a task or reward channel. This is the doctrinal difference, but it's also mechanical — it's what makes per-site accounting exact rather than estimated.

**3. Reform operator origin.** Compilation is one hand-written rule (merge sequential pair, substitute retrieval), triggered by firing adjacency — it never inspects rule content. Your reform operators are comp/cluster applied to bodies: comparison-driven, same algorithm as the data side, with split and the If→While flip having no ACT-R counterpart. ACT-R's rule syntax is fixed; only rule count changes.

**4. Granularity.** Per whole production vs per gate site within a function — sub-rule resolution is what makes split_oF_ possible at all.

So: same interface as ACT-R (per-unit value stats attached to executable units), different in that the accumulated statistic neither drives the gate nor requires reward, and the rewriter that consumes it compares code content. If you want a one-line placement: gv_ = ACT-R utility tracking with intrinsic currency, demoted from controller to sensor.

frame_2D_alg/meta_code.py
boris-kz: all Ts can be in root_ of all Fs. 
But current T is actually not assigned yet, so N_ is right, thanks

frame_2D_alg/meta_code.py
kwcckw: I see, but ACT-R won't be exactly similar since it requires back-prop, which prone to error, while gv_ is more accurate with the local value decision.

boris-kz: Reform operator is not required, and backprop here is same as our feedback, not local

frame_2D_alg/meta_code.py
```diff
@@ -225,29 +225,28 @@ def call_sites(fd):  # FunctionDef
...
             for f,w in zip(N_,w_): f.root_+=[T]; f.rw_+=[w]
             T_ += [T]
     for T in T_:
         if len(T.N_) > 1:  # skip Ts emptied by contact-merge
```
boris-kz, line 244: we don't have this contact-merge anymore, so this eval is redundant?

boris-kz: Your two annotations are the real items; mechanics first:

1. **`-1` in the rdn count is unconditional but the seed-entry it corrects is conditional**: rw_ holds an fc-scale entry only for Fs that seeded their own T earlier in the loop. For never-seeded F the count is deflated by 1 — first membership gated at ave·(−1), i.e. any positive w admits, even sub-ave ones already filtered... no, rim links are all > ave, so for the *first* membership it's harmless; for the second it wrongly gates at ave·0 → free second membership for every F. Cleaner: don't put fc in rw_ at all — seed commits `w_=[_F.fc]` into rw_ via the zip. Replace the seed stand-in with the seed's own admission-scale value (0 or _F.w normalized) and drop the −1; fc then stays out of both the rdn count and W.

2. **W still mixes fc into the sum** (third round): `sum(w_)` with w_[0]=_F.fc — the W>ave gate passes on cost mass. Same fix as 1 resolves both.

3. **`H=L_`**: CF.H is the hierarchy slot — packed lower levels, and form_body/render may treat it as such downstream. T's link record belongs in L_ (CF.L_ takes a set fine). If this is deliberate — links-as-lower-level — then it's a contract change to H's meaning for oFs, worth a comment, not a silent kwarg.

4. Loop still one-hop (L_ collected, never traversed) — standing from last round, your call whether stars suffice this pass.

Your annotations:

5. **"rdn not final in this loop"** — correct, and it's ordering bias, not just staleness: rdn sees only memberships from *stronger-seeded* Ts (loop is sorted by seed w), so early Ts admit rdn-free what later Ts pay for. That's half-right by design — priority to stronger clusters is your standard rdn semantics — but the count is over *committed* memberships only, while the final overlap includes later Ts. Final rdn exists only after all Ts form.

6. **"only if rdn can be reassigned"** — the render loop is where it can: memberships and rw_ are complete there, so re-gate each F per T with final counts — `w > ave * len(stronger rw_)` — drop failures from T.N_ before form_body, then render and credit. That single post-pass is the degenerate reform iteration: one Jacobi step of the fuzzy loop we cut, applied to final counts. If it changes membership materially you've reinvented the convergence loop; at these pool sizes one step is plausibly enough, and the structure is already in your render loop's shape.

frame_2D_alg/meta_code.py
kwcckw: Okay, this maybe indirect since the neg val will be packed in oF.gv_, which will be used in the ffeedback, and affect the next iteration results.

To iterate the positive sequence, we can't have gv_ on each line, so it should be in added from ast level?

frame_2D_alg/meta_code.py
kwcckw: Yes, thanks.

All T is form with N_ > 1 anyway `if len(N_) > 1 and (W :=sum(w_)) > ave:`, and we don't have any pruning, so their N_ size always >1 here.

boris-kz: That rdn should be after final comp_body(T,F) for F in T.N_, as in centroid clustering but not recursive
rim sorting is not relevant

boris-kz: Agreed, and this cleans up three standing sores at once: the fc-scale contamination of w_/rw_ dies structurally (all membership weights are comp_body outputs on one scale), rdn-during-formation ordering bias dies (counts computed once, complete), and pairwise w stops doing a job it was never evidence for — pairwise C to one member isn't membership in the composite. Division of labor becomes clean: rims = candidate grouping only (connectivity, unweighted links suffice), centroid = form_body output, membership = F against centroid, rdn across final memberships, one pass.

Consequences:

1. **comp_body(T,F) forces the composite-side fork branch now** — T.body carries forks, and comparing F's op against a fork tuple currently scores head-mismatch −2 even when F matches a branch. Membership needs: fork → max over branch scores (no fork cost on match — fitting an existing alternative is free). This was the open composite-as-member contract; it just became the membership evaluator rather than a next-cycle item. Detection must match whatever your edited form_body emits.

2. **Formation choice determines whether rdn has any work**: exclusive components per typ (flood over unweighted rims) → likely one T per typ at current pool sizes → each F has one membership, rdn degenerate, the gate reduces to prune-poor-fits. Overlapping formation (star per strong seed, no fin) → rdn adjudicates genuinely. The machinery is identical either way; components are the conservative default.

3. **Membership can extend, not just prune**: F gated into a T it wasn't formed in is the centroid-clustering payoff; with one-component-per-typ it's unreachable, so it's a comment until formation overlaps or typs share conventions.

4. **One re-render after membership** — pruning stales bodies. Then credit and register. "Not recursive" = stop there: no second membership pass against re-rendered centroids.

Skeleton, mechanics only:

```python
        if _F.typ==F.typ and comp_body(_F.body, F.body) > ave: _F.rim += [F]; F.rim += [_F]
    for _F in F_:  # exclusive connectivity components
        if _F.fin: continue
        N_ = [_F]; _F.fin = 1; q = list(_F.rim)
        while q:
            F = q.pop()
            if not F.fin: F.fin = 1; N_ += [F]; q += F.rim
        if len(N_) > 1: T_ += [CoF(N_=N_, typ=_F.typ)]
    for T in T_: form_body(T)  # centroids
    for F in F_:  # membership + rdn, single pass
        r_ = sorted([(comp_body(T.body, F.body), T) for T in T_ if T.typ==F.typ], key=lambda t: t[0], reverse=True)
        F.root_,F.rw_ = [],[]
        for i,(w,T) in enumerate(r_):
            if w > ave*(i+1): F.root_ += [T]; F.rw_ += [w]  # rdn = i stronger memberships
            elif F in T.N_: T.N_.remove(F)  # prune; extension: T.N_ += [F] on gate-in, once formation overlaps
    for T in T_:
        if len(T.N_) > 1:
            form_body(T)  # re-render pruned N_
            C = T.cmpr - sum(F.fc for F in T.N_ if F.root_[0] is not T)
            if C > ave: T.w = C; oF_.append(T); nF_.append(None); T.nF = len(oF_)-1
            else:
                for F in T.N_: i = F.root_.index(T); F.root_.pop(i); F.rw_.pop(i)
```

Known bias, accepted per cluster_C: F is inside the centroid it's scored against — self-match inflates rw uniformly enough at these sizes; exclusion would need per-F re-renders, not worth it.

boris-kz: Ts should be initialized per exemplar F, so we need a this initial comp-cluster in nested loop, not in combinations?

Yes — the pairwise phase earned its place only while membership evidence was link weights; with membership now comp_body against the centroid, rims and F.w are dead structure, and formation reduces to cluster_C init: seed per exemplar, inner loop admits by comp to the seed. Consequences:

1. **The one-hop objection dissolves properly**: stars miss chains at formation, but the membership pass runs against *composite* bodies — C matching B's branch inside T is admitted there — so the extension branch goes live instead of commented. Transitive capture moves to where the evidence actually is (merged body), rather than being simulated by frontier traversal.
2. **Comparison count drops**: intra-star non-seed pairs were never used downstream; membership adds |F_|×|T_| composite comps, which the plan required anyway.
3. **Seed-order dependence shrinks to "which composites exist"** — membership re-adjudicates every F against all final centroids, so consumption order can't misassign, only under-seed. Consistent with no sorting.
4. Lazy T creation: a seed that admits nobody stays unfinned and absorbable by a later star — no singleton Ts to drop.

```python
def clust_oF_():  # cluster_C form: exemplar-seeded stars, centroid membership + rdn, single pass

    F_ = copy(oF_); T_ = []
    for F in F_: F.fin = 0; F.root_ = []; F.rw_ = []
    for _F in F_:  # dual loop: comp-cluster per exemplar
        if _F.fin: continue
        T = None
        for F in F_:
            if F is not _F and F.typ == _F.typ and not F.fin:
                if comp_body(_F.body, F.body) > ave:
                    if T is None: T = CoF(N_=[_F], typ=_F.typ); _F.fin = 1; T_ += [T]
                    T.N_ += [F]; F.fin = 1
    for T in T_: form_body(T)  # init centroids
    for F in F_:  # membership vs centroids, rdn over final counts
        r_ = sorted([(comp_body(T.body, F.body), T) for T in T_ if T.typ==F.typ], key=lambda t: t[0], reverse=True)
        F.root_, F.rw_ = [],[]
        for i,(w,T) in enumerate(r_):
            if w > ave*(i+1):
                F.root_ += [T]; F.rw_ += [w]
                if F not in T.N_: T.N_ += [F]  # extension: chain capture via composite match
            elif F in T.N_: T.N_.remove(F)  # prune
    for T in T_:
        if len(T.N_) > 1:
            form_body(T)  # re-render adjusted N_
            C = T.cmpr - sum(F.fc for F in T.N_ if F.root_[0] is not T)  # fc credited at primary root
            if C > ave: T.w = C; oF_.append(T); nF_.append(None); T.nF = len(oF_)-1
            else:
                for F in T.N_: i = F.root_.index(T); F.root_.pop(i); F.rw_.pop(i)
```

Unchanged requirement, now on the critical path twice: comp_body must score composite-side forks as max over branches, matching whatever your edited form_body emits — both the membership comps and nothing else touch it.

kwcckw: >  F.typ == _F.typ 

This should be removed?

> for T in T_: form_body(T)  # init centroids

This is redundant, since we may copy over the body instead.

> That rdn should be after final comp_body(T,F) for F in T.N_, as in centroid clustering but not recursive
rim sorting is not relevant
> r_ = sorted([(comp_body(T.body, F.body), T) for T in T_ if T.typ==F.typ], key=lambda t: t[0], reverse=True)

Without multiple iteration's refinement, in the first F,  r is computed to the inital T.body that with a single N only ?

boris-kz: yes, most of that is crap.
we need a nested loop, the outer one forming Ts, the inner refining membership by rdn.
but then we to recompute the bodies, so it's at least two-pass.

kwcckw: The first pass is copy the first N body since we only have a single N? But in the inner loop, we do need recompute body when we have new members or there's changes in the member.

boris-kz: no, the body should be formed from N_, then recomputed after comp(T, N_) and redundancy adjustment

kwcckw: For initial T, their N_ is single F and we don't have multiple Fs in N_ yet? We should add new N only after the comp in the refinement?

boris-kz: For initial T, N_ is added from exemplar F.rim. The draft I just pushed is probably all wrong

kwcckw: I see, I will add that later, thanks.

PR #205, opened 2026-07-14
7/14/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -221,31 +221,36 @@ def call_sites(fd):  # FunctionDef
...
-    for T in T_: T.N_ = []  # rebuilt by membership
-    for F in F_:
-        for i, (w, T) in enumerate(sorted([(comp_body(T.body, F.body), T) for T in T_], key=lambda t: t[0], reverse=True)):
-            if w > ave * (i + 1):
-                F.root_ += [T]; F.rw_ += [w]; T.N_ += [F]
-            else: break  # w falls, threshold grows
+            _N_ = F.N_; F.root_ = []; F.N_ = []; F.rw_ = []
+            for i, (w, T) in enumerate(sorted([(comp_body(T.body, F.body), T) for T in _T_], key=lambda t: t[0], reverse=True)):
+                if w > ave * (i + 1):  # rdn beased on their w?
+                    F.root_ += [T]; F.rw_ += [w]; T.N_ += [F]; T.w += w
```
kwcckw, line 241: I added a draft of T refinement using while loop here. For T.w, it should be sum from all of their member's rw?

boris-kz: right

frame_2D_alg/meta_code.py
```diff
@@ -221,31 +221,36 @@ def call_sites(fd):  # FunctionDef
...
+def clust_oF_():
+
+    F_ = copy(oF_); _T_ = []
+    for F in F_: F.rim = []; F.root_ = []; F.rw_ = []
+    for _F, F in combinations(F_, 2): 
+        if comp_body(_F.body, F.body) > ave: 
+            _F.rim += [F]; F.rim += [_F]
+    for _F in F_:  # init T with F and rim
+        if not _F.rim: continue
+        T = CoF(N_=[_F] + _F.rim);form_body(T); _T_ += [T]  # exemplars are Fs with non empty rim
```
boris-kz, line 233: we can probably subtract w summed over rim overlap, for sparse exemplars

kwcckw: So we still need to pack w into rim when we compare between Ns here:

    for _F, F in combinations(F_, 2): 
        if comp_body(_F.body, F.body) > ave: 
            _F.rim += [F]; F.rim += [_F]
            
But how exactly we can subtract w summed over rim overlap per F to select exemplar?

boris-kz: it will be rdn += 1 per link where _F.w > F.w, where each w is summed over F.rim

boris-kz: for _F, F in combinations(F_, 2):
        if (w := comp_body(_F.body, F.body)) > ave: _F.rim += [F,w]; F.rim += [_F,w]; _F.w += w; F.w += w
    for _F in F_:
        _F.w = sum([w * (_F.w/(_F.w+F.w)) for F,w in _F.rim])  # /= rdn 
        if _F.w > ave: 
            T = CoF(N_=[_F] + _F.rim); form_body(T); _T_ += [T]  # exemplars are Fs with non empty rim

kwcckw: I see, but this F.w is between Fs. When we compare F with T during the refinement process, T.w should be summed from the w between F and T, rather than F.w?

boris-kz: yes, T.w is not computed above

boris-kz: anyway, here we need a dense array mapping T.N_ to F_ and F.root_ to T_, same as in cluster_P

boris-kz: def clust_oF_():  # simplified oF rim-mediated centroid clustering

    F_ = copy(oF_)
    for F in F_: F.rim = []
    for _F, F in combinations(F_,2):
        if (w := comp_body(_F.body, F.body)) > ave: _F.rim += [F,w]; F.rim += [_F,w]; _F.w += w; F.w += w
    _T_ = []
    for _F in F_:
        _F.w = sum([w * (_F.w/(_F.w+F.w)) for F,w in _F.rim])  # /= rdn to stronger F in rim
        if _F.w > ave:
            N_ = [_F]+ [L[0] for L in _F.rim]; T = CoF(N_=N_, L_= [0 for _ in N_])  # no T.L_,T.w yet
            form_body(T); _T_ += [T]
    while _T_:
        T_,nT_ = [],[]; W = DW = 0
        root__ = [[] for f in F_]  # dense, replace in parallel
        for i,T in enumerate(_T_):
            Tw = Dw = 0; N_,L_ = [],[]
            for j, w in enumerate(T.L_):  # aligned with F_ in all Ts
                w = comp_body(T.body, F.body)
                N_+= [F]; L_+= [w]; Tw += w; Dw += F_[j].root_[i]
            DW += Dw; W += Tw
            if Dw > ave:
                T.N_= N_; T.L_= L_; T.w = W  # add adjust each w for rdn to stronger roots?  
                T_+= [T]; nT_+=[T]
            else: nT_ += [[]]
        if any(nT_) and DW > ave:   # continue refinement as long there's changes in the members
            for T,nT in zip(T_,nT_):
                if nT: form_body(T)  # rebuild from new N_
        else: break
    for T in T_:
        if T.w > ave: oF_.append(T); nF_.append(None); T.nF = len(oF_) - 1
        else:
            for F in T.N_: i = F.root_.index(T); F.root_.pop(i); F.rw_.pop(i)

boris-kz: the while loop is cluster_P's role (parallel dense refine), the rim seeding is cluster_C's role compressed to one shot. Folding both into one function is fine at 15 oFs. The correspondence map, and where it breaks:

**Transfers directly:**

1. **Dense matrix + delta convergence**: `_md__`/`md__` (N×C), `conv = abs(md__-_md__).sum()` scaled — exactly what Dw/root__ should be. One w matrix rows F_, cols _T_, delta per pass. Kills root_/rw_ plumbing (items 6, 11): cluster_P resets root_/m_ each cycle and lets sum2F repopulate them; the oF analog is form_body (or a sum2O wrapper) doing the reciprocal push, not manual pops.

2. **Merge redundant Cs**: pairwise comp between centroids, add2F on match — the missing duplicate-T merge (item 13). Needed here more than in cluster_P, since your seeding is per-F, not per-exemplar: near-duplicates are guaranteed.

3. **Prune by centroid strength, finalize on convergence**: `c.m*wcP > ave*c.r*ccP` survives, conv + no churn breaks — the Tw/Dw separation (item 8). cluster_P's exit condition (conv AND no removed/new AND stable count) is the right contract; yours exits on the wrong variable.

4. **Ambiguity-gated seeding** (`new_` from Ns whose second-best match is high): optional at your scale, but it's the principled replacement for "every strong F seeds a T".

**Breaks, two places:**

5. **Mean shift needs a weighted mean; form_body is discrete.** cluster_P's update is `sum2F(N_, md__[:,i,0])` — every C re-formed from *all* Ns, membership soft, weights continuous. AST ops can't be weight-averaged: body merge is set-union with forks. So the update step degrades to hard select-then-merge: threshold w per member each pass, form_body from survivors. That makes clust_oF_ k-means-shaped rather than mean-shift-shaped — reassignment is discrete, convergence can oscillate between two member sets. Matrix delta still works as the test, but expect a cycle guard (cap passes or detect repeat membership sets) instead of guaranteed monotone convergence.

6. **Comparand scale.** cluster_P compares in val_ units (relative match, ave-commensurate). comp_body returns raw shared cost. For the analogy to hold, normalize: shared cost over min(fc) of the comparands — relative compression, which is also your min/max match definition, so this is the derivation-consistent fix for item 3, not just a patch.

And item 12 restated in cluster_P terms: base_comp works because the centroid is an *average*, comparable to any member. Your centroid body is a fork *union* — comp_body must score a member against its best branch (max over fork branches, fork cost once), otherwise the centroid is systematically farther from every member than members are from each other, and refinement anti-converges. That's the one comp_body change required before the cluster_P shape gives sane deltas.

boris-kz:
```python
def clust_oF_():  # oF centroid clustering, cluster_P analog: parallel refine, _T_ varies via merge|seed

    F_ = copy(oF_); Ln = len(F_)
    for F in F_: F.rim = []; F.w = 0
    for _F,F in combinations(F_,2):  # w = relative compression: shared / min cost, ave-commensurate
        if (w := comp_body(_F.body,F.body) / min(_F.fc,F.fc)) > ave:
    w_ = [sum([w * F.w/(F.w+_F.w) for _F,w in F.rim]) for F in F_]  # /= rdn to stronger F in rim, raw-w snapshot
    _T_ = []
    for F,w in zip(F_,w_):
        F.w = w
        if w > ave: T = CoF(N_=[F]+[rF for rF,_ in F.rim]); form_body(T); _T_ += [T]
    _w__ = np.zeros((Ln,len(_T_)))  # FxT, cols aligned to _T_, no priors
    while _T_:
        Lt = len(_T_); L = Lt*Ln
        w__ = np.zeros((Ln,Lt)); O = 0
        for j,F in enumerate(F_):
            for i,T in enumerate(_T_): w__[j,i] = comp_body(T.body,F.body) / min(T.fc,F.fc)
            r_ = w__[j]; O += r_.sum() - r_.max()  # cross-T ambiguity, gates merge|seed
        for i,T in enumerate(_T_):  # discrete mean shift: hard reselect members, rebuild body
            T.N_ = [F for F,w in zip(F_,w__[:,i]) if w > ave]; T.L_ = [w for w in w__[:,i] if w > ave]
            T.w = w__[:,i].sum() / Ln  # intensive, ~C.m
            if T.N_: form_body(T)
        conv = w__.shape==_w__.shape and w__.sum()* np.abs(w__-_w__).sum()* wcO <= ave* ccO*L  # total membership delta
        removed = []
        if O*wcO > ave* ccO*L:  # merge redundant Ts
            for i,_T in enumerate(_T_):
                if _T in removed: continue
                for T in _T_[i+1:]:
                    if T in removed or comp_body(_T.body,T.body) / min(_T.fc,T.fc) <= ave: continue
                    _T.N_ = list(set(_T.N_+T.N_)); _T.w += T.w; form_body(_T); removed += [T]  # L_,w refresh next cycle
        new_ = [CoF(N_=[F], body=copy(F.body), fc=F.fc) for j,F in enumerate(F_)
                if np.sort(w__[j])[-2:].min() > ave]  # seed 2nd-overlap Fs
        _T_ = [T for T in _T_ if T not in removed and T.w > ave] + new_  # prune, +seeds
        if conv and not (removed or new_) and len(_T_)==Lt: break
        _w__ = w__
    for T in _T_:
        oF_.append(T); nF_.append(None); T.nF = len(oF_)-1  # already pruned by T.w
```

Companion edit in comp_body, inserted after the CoF branch, before the list branch — without it refinement anti-converges (item 12):

```python
    f_,f = [isinstance(t,tuple) and len(t)>1 and t[0] is ast.IfExp for t in (_n,n)]
    if f_ or f:  # centroid fork: score best branch pair, single fork cost; form_body branch = ([f],op)
        return max(comp_body(_b,b) for _b in ([x[1] for x in _n[1:]] if f_ else [_n]) for b in ([x[1] for x in n[1:]] if f else [n])) - 2
```

Deviations from cluster_P, all forced:

1. **Hard membership replaces sum2F mean shift**: bodies can't be weight-averaged, so per cycle each T reselects members by w > ave and form_body rebuilds from them. Consequence: reassignment is discrete, oscillation between member sets is possible; conv delta + merge usually absorb it, else cap cycles.

2. **No r, c**: oFs carry only fc; r-scaled thresholds become count-scaled (L), and T.w is the column mean (intensive, C.m analog) rather than c-weighted.

3. **gv_ dropped**: clust_oF_ isn't traced, no CoF context for the gate; plain comparisons with wcO/ccO.

4. **root_/m_ bookkeeping dropped**: the matrix is the membership state; nothing downstream reads F.root_ for oFs, and final registration needs no per-member pops.

5. **new_ seeds bare F** (Copy_ analog: N_=[F], body copy), not F+rim — rim is seed-phase only, matching cluster_P where seeds are single-N centroids.

6. Merge leaves merged _T's L_/w stale for one cycle — same as add2F running after md__ in cluster_P.

Untraced residue: form_body sets fc but branch format differs between form_body and _form_body (([f],op) vs (f_,op_list)); the comp_body fork clause assumes the former, which is what this function produces. If you switch T construction to _form_body, x[1] becomes a list and the clause needs one more line.

kwcckw: In the latest code above, actually all Ts will be compared with all Fs, and rim is use to init seed Ts only.

> L_+= [w]

T.L_ is actually ws, that's replacing the original rw_?

Also this section below is redundant? We don't need to merge them when they fail to cluster?

      if O*wcO > ave* ccO*L:  # merge redundant Ts
            for i,_T in enumerate(_T_):
                if _T in removed: continue
                for T in _T_[i+1:]:
                    if T in removed or comp_body(_T.body,T.body) / min(_T.fc,T.fc) <= ave: continue
                    _T.N_ = list(set(_T.N_+T.N_)); _T.w += T.w; form_body(_T); removed += [T]  # L_,w refresh next cycle

boris-kz: right

kwcckw: For the convergence:

    conv = w__.shape==_w__.shape and w__.sum()* np.abs(w__-_w__).sum()* wcO <= ave* ccO*L  # total membership delta

We just need to check if there's changes in T's member right? I think that could be checked with `(_w__>ave) == (w__ > ave)`?

boris-kz: dual criterion is better, as in my version:

    if any(nT_) and DW > ave:   # continue refinement as long there's changes in the members
        for T,nT in zip(T_,nT_):
            if nT: form_body(T)  # rebuild from new N_

boris-kz: so I don't think we need to force it into cluster_P template, some edits by claude:

```python

    F_ = copy(oF_)
    for F in F_: F.rim = []; F.w = 0
    for _F, F in combinations(F_,2):  # w = relative compression: shared / min cost, ave-commensurate
    w_ = [sum([w * F.w/(F.w+_F.w) for _F,w in F.rim]) for F in F_]  # /= rdn to stronger F in rim, from raw-w snapshot
    _T_ = []
    for F,w in zip(F_,w_):
        F.w = w
        if w > ave:
            T = CoF(N_= [F]+[rF for rF,_ in F.rim], L_= [0 for _ in F_])  # L_: dense prior w_, aligned with F_ in all Ts
            form_body(T); _T_ += [T]
    out_ = []
    while _T_:
        T_ = []
        for T in _T_:
            Tw = Dw = 0; N_,L_ = [],[]
            for j,F in enumerate(F_):
                w = comp_body(T.body, F.body) / min(T.fc, F.fc)
                L_ += [w]; Tw += w; Dw += abs(w - T.L_[j])
                if w > ave: N_ += [F]  # hard member cutoff, L_ stays dense
            T.L_ = L_; T.w = Tw
            if Tw > ave:  # else weak: drop
                if Dw > ave: T.N_ = N_; form_body(T); T_ += [T]  # rebuild from new members, refine
                else: out_ += [T]  # converged
        _T_ = T_
    for T in out_:
```

Notes:

1. The dense prior lives in T.L_ — your "aligned with F_ in all Ts" comment, now actually true. Kills root_/rw_/root__ and all pops; no reciprocal bookkeeping anywhere.

2. Per-T exit is your shape: weak → dropped mid-loop, converged → out_, else refine. So the final loop only registers; no second Tw test needed.

3. Seed L_ = zeros makes first-pass Dw large by construction — every seed refines at least once, no special-casing the first cycle.

4. Overlap unmanaged, per your call: duplicates converge and both register. Two knobs when you want fewer: the seed filter (w_ threshold) and the member cutoff — they're the same `ave` now; splitting them is the first refinement.

5. Unguarded: discrete membership can oscillate between two member sets, Dw never settling — while then doesn't terminate. If it bites, cheapest fix in this shape is finalize on non-decreasing Dw (one prior scalar on T), not a pass cap.

6. The comp_body fork clause from my previous message is still required — without it converged bodies with forks score depressed w against their own members and Tw-pruning eats valid Ts.

kwcckw: But i don't get your Dw?

 Dw += F_[j].root_[i]
 
`F_[j].root_[i]` is their root, but not w?

Also that root_[i] indexing may not reference to any root, since F.root_ is not mapped to all Ts.

kwcckw: > so I don't think we need to force it into cluster_P template, some edits by claude:
> 
> ```python
> def clust_oF_():  # simplified oF rim-mediated centroid clustering
> 
>     F_ = copy(oF_)
>     for F in F_: F.rim = []; F.w = 0
>     for _F, F in combinations(F_,2):  # w = relative compression: shared / min cost, ave-commensurate
>         if (w := comp_body(_F.body, F.body) / min(_F.fc, F.fc)) > ave:
>             _F.rim += [(F,w)]; F.rim += [(_F,w)]; _F.w += w; F.w += w
>     w_ = [sum([w * F.w/(F.w+_F.w) for _F,w in F.rim]) for F in F_]  # /= rdn to stronger F in rim, from raw-w snapshot
>     _T_ = []
>     for F,w in zip(F_,w_):
>         F.w = w
>         if w > ave:
>             T = CoF(N_= [F]+[rF for rF,_ in F.rim], L_= [0 for _ in F_])  # L_: dense prior w_, aligned with F_ in all Ts
>             form_body(T); _T_ += [T]
>     out_ = []
>     while _T_:
>         T_ = []
>         for T in _T_:
>             Tw = Dw = 0; N_,L_ = [],[]
>             for j,F in enumerate(F_):
>                 w = comp_body(T.body, F.body) / min(T.fc, F.fc)
>                 L_ += [w]; Tw += w; Dw += abs(w - T.L_[j])
>                 if w > ave: N_ += [F]  # hard member cutoff, L_ stays dense
>             T.L_ = L_; T.w = Tw
>             if Tw > ave:  # else weak: drop
>                 if Dw > ave: T.N_ = N_; form_body(T); T_ += [T]  # rebuild from new members, refine
>                 else: out_ += [T]  # converged
>         _T_ = T_
>     for T in out_:
>         oF_.append(T); nF_.append(None); T.nF = len(oF_)-1
> ```
> 
> Notes:
> 
>     1. The dense prior lives in T.L_ — your "aligned with F_ in all Ts" comment, now actually true. Kills root_/rw_/root__ and all pops; no reciprocal bookkeeping anywhere.
> 
>     2. Per-T exit is your shape: weak → dropped mid-loop, converged → out_, else refine. So the final loop only registers; no second Tw test needed.
> 
>     3. Seed L_ = zeros makes first-pass Dw large by construction — every seed refines at least once, no special-casing the first cycle.
> 
>     4. Overlap unmanaged, per your call: duplicates converge and both register. Two knobs when you want fewer: the seed filter (w_ threshold) and the member cutoff — they're the same `ave` now; splitting them is the first refinement.
> 
>     5. Unguarded: discrete membership can oscillate between two member sets, Dw never settling — while then doesn't terminate. If it bites, cheapest fix in this shape is finalize on non-decreasing Dw (one prior scalar on T), not a pass cap.
> 
>     6. The comp_body fork clause from my previous message is still required — without it converged bodies with forks score depressed w against their own members and Tw-pruning eats valid Ts.

I see, this looks better.

boris-kz: >F_[j].root_[i] is their root, but not w?
Also that root_[i] indexing may not reference to any root, since F.root_ is not mapped to all Ts.

root_ should be mapped to all Ts, so it's just rw_

kwcckw: Okay, so it's actually `abs(w - F_[j].root_[i].rw_[j])` in your version.

In the latest code above:

    else: out_ += [T]  # converged
    
Once T is converged, we no longer need to refine them right? Since each T is actually independent of each other?

boris-kz: there is some overlap, but it's probably not a problem. so:

    for i,T in enumerate(out_): T.nF = i
    global oF_; oF_ = out_

boris-kz: not sure this is correct:

Right intent — generational replacement instead of accumulation. Four contracts left open, one syntax error:

1. `global oF_` must be the first line of the function — `F_ = copy(oF_)` reads the name earlier in scope, which is a SyntaxError, not a warning.

2. **Coverage**: out_ holds only strong Ts. Originals in no surviving T vanish from oF_ while their call sites in agg_recursion still execute — traced/gv_ then index a list that no longer contains them. Uncovered Fs must carry over.

3. **nF_ realignment**: inject_oF_ reads `nF_[oF.nF]`. Carried originals keep their fdef; Ts have none until a T.body → FunctionDef renderer exists (currently missing — build strips to skeletons, nothing goes back). nF_ must be rebuilt in the same order.

4. **Statistics transfer**: form_body sets only body/fc. As sole survivors, Ts need caller_ (comp_callers input) and c from members, or the next generation compares empty sets.

Closed version of your block:

```python
    covered = {F for T in out_ for F in T.N_}
    out_ += [F for F in F_ if F not in covered]           # uncovered originals, still called
    for T in out_:
        if not isinstance(T.nF, int):                     # new T, original nF is int
            T.caller_ = set().union(*[F.caller_ for F in T.N_]); T.c = sum(F.c for F in T.N_)
    nF_new = [nF_[F.nF] if isinstance(F.nF, int) else None for F in out_]
    for i,F in enumerate(out_): F.nF = i
    oF_ = out_; nF_ = nF_new                              # global oF_, nF_ declared at top
```

Residue, not closable inside this function:

5. iF_, F_call_T_, F_call_i_ stay aligned to the old generation. traced and gv_ resolve names through iF_ at runtime, so between this replacement and re-injection, no traced function may run: merged members' names are gone from the new numbering. The rebuild of all three belongs to the render+inject stage, which is also where call sites of merged members get redirected to their T (with overlap: strongest-root T). Until that stage exists, oF_ = out_ is a commitment the rest of the file can't yet honor — appending was the version that didn't require it.

6. Minor: T.typ defaults to 0; members of a T are near-uniform in typ, inherit it, since typ blocks sub_comp downstream.

kwcckw: > there is some overlap, but it's probably not a problem. so:

Yes, there should be overlapping in their N_, but that's expected since their w are now separated in T.L_.

PR #207, opened 2026-07-16
7/16/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -195,24 +195,26 @@ def clust_oF_():  # simplified oF rim-mediated centroid clustering
...
             form_body(T); T_ += [T]
-    nT = len(T_)
-    while nT:
-        nT = 0
-        for T in [t for t in T_ if not t.fin]:
-            Tw = Dw = 0; N_,L_ = [],[]
+    floop = 1; _w__ = [[0 for F in F_] for T in T_]  # we just need a flag isntead of nT?
+    while floop:
+        floop = 0
+        w__ = [[0 for F in T.N_] for T in T_]
```
kwcckw, line 201: So i restructured the code and the prior empty list in N_ adds a lot clutter, especially when we need to get T.N_ in subsequent loops again. As a workaround, a `w__` is used and w__[i] is map to each T and w__[i][j] is map to each F in T.N_.

boris-kz: thanks

frame_2D_alg/meta_code.py
```diff
@@ -195,24 +195,26 @@ def clust_oF_():  # simplified oF rim-mediated centroid clustering
...
-                if w > ave: Tw += w; N_ += [F]
-                else: N_ += [[]]  # hard member cutoff, L_,N_ stay aligned
-            T.N_ = N_; T.L_ = L_; T.w = Tw / (len(L_))  # members, weights, mean w
-            if Tw > ave and Dw > ave: form_body(T); nT += 1  # rebuild from remaining members, refine
+                w = (comp_body(T.body, F.body) / min(T.fc, F.fc) if F else 0) 
+                if F.root_: w/=sum(_T.w for _T in F.root_)
+                Dw += abs(w - _w__[i][j]); w__[i][j] = w
+                if w>ave: N_ += [F]; w_ += [w]; F.root_.add(T)
+            T.N_ = N_; T.L_ = w_
+            if w_: T.w = sum(w_) / (len(w_))  # members, weights, mean w
```
boris-kz, line 211: if w_: T.w = np.mean(w_)

kwcckw: Yes, this is actually their mean, thanks.

frame_2D_alg/meta_code.py
```diff
@@ -185,7 +185,7 @@ def parse_funcs(paths):
 def clust_oF_():  # simplified oF rim-mediated centroid clustering
```
boris-kz, line 185: I think incremental composition should be produced by alternating cluster_N and cluster_C, neither of which should be independenty recursive. That will make cluster_N parallelizable by splitting it across many small windows. that solves both the chaining problem and the sequential bottleneck inherent in floodfill.

boris-kz: the discussion that led to this, the details in the end: https://claude.ai/share/cdf3121e-b691-424c-9cc7-3b6ecd1bf30f

kwcckw: > segments can be produced by cluster_N in small windows, that we have cluster_C in larger windows, then splicing

But how do we select element for the larger windows in the first place?  We still need another process to cluster them into different windows first? Or that will be completely random since it will be spliced and recluster later?

boris-kz: the elements are still connectivity clusters: Gs, but now next step is sum2G-mediated same-frame CN or CC:


            if gv_(Vn* (wcC-wcN)* (mdecay(L_)-decay) - Av* (lr+1+(ccC-ccN)*L)):
                r +=1; G_,r = cluster_C(G.Nt,E_,r,c)  # higher V, low decay, eval cluster_P

in the new scheme the next step will always be CC in a larger frame

boris-kz: Claude:
the strongest argument for it is error-mode interleaving. Recursive cluster_N compounds its own systematic bias — transitive drift deepens level after level with nothing to break the chains; recursive cluster_C compounds spatial ungroundedness — types of types get progressively less anchored. Alternation inserts each operator's corrective between every step of the other: C breaks N's chains by regrouping on parameters, N splits C's types into instances by adjacency. One refinement to the phrasing: you're not removing recursion, you're relocating it to the scheduler, where it becomes external and preemptible. Strict 1:1 alternation is the special case; your filter logic can repeat an operator while its marginal yield still dominates. The improvement is that recursion is now interruptible by the other operator rather than committed inside one.

Parallelization: yes, and the fragment revision is what makes it free. Parallel connected-component labeling always needed a global merge phase because it wanted complete components; you don't — window cuts just mint extra seeds, and the stitch is the splice you were doing anyway. Windowing also caps transitive depth, which is the chaining pathology exactly: link errors stay bounded at window scale, and global grouping is delegated to parameters instead of transitivity.

On flood-fill: its serial depth is the object's extent — frontier step k+1 waits on k. Your depth becomes window diameter plus a few C iterations plus a near-logarithmic merge with parallel union-find. Genuine span reduction. One honesty clause: parallel CCL in the Shiloach-Vishkin lineage already solves the raw bottleneck without any of this. Your actual delta is that their merge tree is representationally sterile while yours is the representation — fragments, types, tokens — so one restructuring buys both the algorithmic win and the compositional structure.

The transferred cost: long thin structures now exist only through splicing, so their recovery makes intrinsic, orientation-relative fragment parameters load-bearing for correctness, not just efficiency. The chaining problem isn't deleted; it's converted into the splicing problem — deferred, evidence-gated, revisable. Which is where you wanted it.

kwcckw: So we should have cluster_C here instead? But that would be cluster_C for the smaller windows, and cluster_N for larger windows? Since the initial G_ is clustered via a larger window in cluster_N first?

boris-kz: No, we start with the smallest windows and connectivity clustering

kwcckw: > No, we start with the smallest windows and connectivity clustering
> in the new scheme the next step will always be CC in a larger frame

The smallest window here is referring to each tile?

Something like after we get CC in sum2G, we splice them from all tiles to get larger frame before proceed to the next step?

boris-kz: yes, tile. 
Then frame will first compare and splice tile box terminated Gs, and then run CC over all >ave Gs. 
So initial CC is still G-internal sub-clustering, but then it spreads across Gs

boris-kz: CN consumes linkable objects: real coordinates, endpoints, link slots. 
A centroid has none — its position is an average lying on no actual structure, its parameters possibly off-manifold; the mean of a left-curving and a right-curving fragment is straight. 
So the only form in which CC's accumulated evidence can enter extended CN is an instance: 
the exemplar — the member maximizing match to the centroid, the medoid, computable for free during membership refinement by tracking the argmax. The centroid stays internal to CC as an estimator; the exemplar is its externally usable form. Prototype for measuring, exemplar for touching.

One refinement: a single exemplar per type biases long-range match toward its particular pose. Better a small coreset — the medoid plus a few maximal-spread or high multi-membership members, the latter doubling as bridges for cross-type splices. Exemplar-based clustering (k-medoids, affinity propagation) formalizes exactly this output, and exemplar-SVMs made the same argument against category means for structured matching: averaging destroys the geometry that matching needs. Max- versus average-pooling is the distributed echo — forwarding the argmax instance beats forwarding the mean wherever structure must survive the summary.

What CC supplies besides exemplars is scheduling rather than objects: counts, variance, summed match per type — deciding where extended CN is worth its range cost. Even block pruning at range is exemplar-mediated: compare exemplars first, open member-level comparison only where they match — branch-and-bound, Barnes-Hut with a real node standing in for the center of mass. So CC's product splits by consumer: summaries feed filters, exemplars feed CN. "Mostly exemplars" is right for the structural path; the statistics ride along for control.

kwcckw: > but then it spreads across Gs

It spreads acros Gs from multiple tiles right?

> a single exemplar per type biases long-range match toward its particular pose. Better a small coreset — the medoid plus a few maximal-spread or high multi-membership members

So this is suggesting multiple member based exemplars, as in clust_oF_, where we init exemplar with F and their rim?

boris-kz: yes, within new frame
actually we may still have sub-clustering CNs, and then global CC across them. 
max N.w in C.N_ will be exemplars for extended CN, within the same frame as CC?

boris-kz: Accepting your operator assignment over my last-turn framing: the intra-G pass is CN mechanically — partition through links by the sign and kind of minted derivatives — while functioning as the proposal stage for CC. Cleaner than my "degenerate CC": every structural operation is N, every statistical one is C, no special cases. Note what the level template has become: N-sub → C-global → N-extended. That's literally the sandwich I couldn't find published — recognition between two proximity stages — now explicit in your own schedule.

Exemplar as max N.w in C.N_: correct, the medoid by membership, free from refinement bookkeeping. Two implementation notes. Read it off at CC termination, not mid-iteration — w shifts while centroids move, and a drifting exemplar corrupts any comparison already opened against it. And top-1 is the right start; the upgrade path when single-pose bias shows is top-k plus members with high secondary weights — those double as bridges for cross-type splices.

Same window: yes, and it's forced three ways. One scope multiplication per level, owned by CC — a second one at CN breaks the constant-work pyramid. The gate can't outrun the gater — exemplars only type what CC's window saw, so CN beyond it is ungated comparison, re-importing both walls. And longer-range links aren't lost, they're deferred to the level whose window spans them — same principle as rare-type unification. Invariant: one window per level, exhausted by the N/C/N triple, expanded only between levels.

Last, the reconciliation with one-to-one splicing: exemplar comparison is the screening step — the allocator deciding which G-pairs are worth opening. Actual links still form member-to-member, one-to-one, inside opened blocks. Exemplars measure; members touch.

kwcckw: > max N.w in C.N_ will be exemplars for extended CN, within the same frame as CC?

That N with max w in C.N_ mostly is the seed for that C too? So a same N could be used as exemplar to form C and the subsequent N clustering?

Also this is from C to N, how about from N to C? When we have Ns, the exemplar selection is the same as in existing `get_exemplars`, which is based on rim's m?

boris-kz: For next CN, N.w should be computed by comp to C, not laterally.
For N to C, current exemplar selection should be fine

kwcckw: Okay right and for this process:

>> Then frame will first compare and splice tile box terminated Gs, and then run CC over all >ave Gs.

We splice Gs from multiple tiles before have the centroid clustering. After when we have Cs, the next cross_comp will be using Ct's C_ that packing all tiles' Cs? But then in the next N clustering, we won't get different tiles now after we splice Gs from different tiles?

boris-kz: It will be Ct.N_s, and we don't need new tiles, CN is a new composition level from the same tiles

PR #208, opened 2026-07-17
7/17/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -564,12 +564,18 @@ def sum2G(ft_, fTT, root=None, init=1):  # core clustering function
...
             if gv_(Vn* (wcC-wcN)* (mdecay(L_)-decay) - Av* (lr+1+(ccC-ccN)*L)):
                 # just nested CN, then CC only between sub-Gs in frame?
                 r +=1; G_,r = cluster_C(G.Nt,E_,r,c)  # higher V, low decay, eval cluster_P
             else:      G_,r = cluster_N(G.Nt,E_,r,c)  # updates G
-            if G_ and gv_(val_(G.Nt.dTT,G.wTT*ttA) * ((G.c+wAgg) /(G.r+r+cAgg)) * ((len(G_)-1)**2 *wL) - ave):  # if full cross_comp?
-                cross_comp(G.Nt,r)
+            '''
+            # nested CN
+            if gv_(Vn* (wcN)* (mdecay(L_)-decay) - Av* (lr+1+ccN*L)):
+                G_,r = cluster_N(G.Nt,E_,r,c)  # updates G
```
kwcckw: For nested CN, do we still need the cross_comp here? Or we just need cluster_N to pack new sub into G.N_?

frame_2D_alg/agg_recursion.py
```diff
@@ -879,8 +885,17 @@ def fill_frame(_iy,_ix, elev, T):  # expand level_frame from pixel-level seed ti
     if not T or not max_elev: return T, ([], np.zeros((2,9)),0,0,0)  # frame_H(0) = pixel tile
     while elev < max_elev:
         tile_,C,R = fill_frame(iY,iX, elev, T)  # project from seed tile
-        if tile_: # sparse,2D
-            Fr = sum2F(tile_)  # higher-scope tile( oH( aH
+        G__ = [G for T in tile_ for G in T.N_ for sub in G.N_ if sub.sub == G.sub]  # sub+'s N.sub should be == G
```
kwcckw: I added draft here to splice tiles' N_'s sub before selecting exemplar for the subsequent centroid clustering.

After we get the CC here, we need another process to select exemplar based on their member w before feed into the cross_comp again? 

Also my concern is still on the tiles because after CN->CC->CN, when we merged all tiles during CN->CC, the subsequent CN->CC doesn't have tiles to merge anymore since it's already merged?

frame_2D_alg/agg_recursion.py
boris-kz: Yes, replace G.N_

boris-kz: actually this is still fine:

            if gv_(Vn* (wcC-wcN)* (mdecay(L_)-decay) - Av* (lr+1+(ccC-ccN)*L)):
                r +=1; G_,r = cluster_C(G.Nt,E_,r,c)  # low decay: CC nodes
            else:      G_,r = cluster_N(G.Nt,E_,r,c)  # CC sub-Gs, if any

frame_2D_alg/agg_recursion.py
```diff
@@ -564,12 +564,18 @@ def sum2G(ft_, fTT, root=None, init=1):  # core clustering function
...
                 r +=1; G_,r = cluster_C(G.Nt,E_,r,c)  # higher V, low decay, eval cluster_P
             else:      G_,r = cluster_N(G.Nt,E_,r,c)  # updates G
-            if G_ and gv_(val_(G.Nt.dTT,G.wTT*ttA) * ((G.c+wAgg) /(G.r+r+cAgg)) * ((len(G_)-1)**2 *wL) - ave):  # if full cross_comp?
-                cross_comp(G.Nt,r)
+            '''
+            # nested CN
+            if gv_(Vn* (wcN)* (mdecay(L_)-decay) - Av* (lr+1+ccN*L)):
+                G_,r = cluster_N(G.Nt,E_,r,c)  # updates G
+                # below no longer needed?
+                # if G_ and gv_(val_(G.Nt.dTT,G.wTT*ttA) * ((G.c+wAgg) /(G.r+r+cAgg)) * ((len(G_)-1)**2 *wL) - ave):  # if full cross_comp?
```
boris-kz: I think it's something like this:

    if G.Lt:  # sub+
        Lt = G.Lt; L_,lm,ld,lr = Lt.N_,Lt.m,Lt.d,Lt.r; L=len(L_)-1; Av = ave+avd
        if gv_(Vn := (lm+ld)*wcN - Av* (lr+1+ccN*L)):  # for cluster_N
            c = G.Lt.c; E_ = get_exemplars({N for link in L_ for N in link.N_}, r,c)
            if gv_(Vn * (wcC-wcN) * (mdecay(L_)-decay) - Av * (lr+1+(ccC-ccN)*L)):
                r+=1; G_,r = cluster_C(G.Nt,E_,r,c)  # low decay: CC nodes
            else:     G_,r = cluster_N(G.Nt,E_,r,c)  # CC sub-Gs, if any
    if G.Ct and gv_(val_(G.Nt.dTT,G.wTT*ttA) * ((G.c+wAgg) /(G.r+r+cAgg)) * ((len(G.N_)-1)**2 *wL) - ave):
        cross_comp(G.Ct,r)  # cross_comp G.N_ via Ct.N_ exemplars, still sub+

kwcckw: > It will be Ct.N_s, and we don't need new tiles, CN is a new composition level from the same tiles

This is actually almost the same? We are not actually splicing Gs before performing any sub+?

boris-kz: this is still sub+, in the same tile, there's nothing to splice
splicing is a separate process, in frame_H.
and then we need to start with cluster_C(frame.N_), not cross_comp

frame_2D_alg/agg_recursion.py
boris-kz: CN->CC is sub+ only, 
agg+ will be something like cross_comp(E_ = select in frame.Ct.N_),
secondary to to cluster_C(frame.N_)

frame_2D_alg/agg_recursion.py
kwcckw: > CN->CC is sub+ only, agg+ will be something like
cross_comp(E_ = select in frame.Ct.N_)

I see, so the spliced Ns should be in Fr, added from tile_?

Fr.Ct.N_ should be those Cs added from each tile, and we need to run cluster_C on those C's N_, and exemplars are selected based on N's weight on the C?

boris-kz: yes, actually not just exemplars, I think the whole Fr.N_ for cross-comp should be pruned Fr.Ct.N_, formed across tiles.
Splicing is only for Gs terminated at tile box, if they share termination coords

kwcckw: > pruned Fr.Ct.N_,

We need to prune Fr.Ct.N_? But i think there's shouldn't be any overlapping between tiles, so why we need to prune them?

> Splicing is only for Gs terminated at tile box, if they share termination coords

This should be complicated because we need to identify if the G is located in the corner or side of the tile, where it may splice in different ways if it's located in corner compare to side.

boris-kz: the whole point of centroid clustering now is select Ns with high w to C, otherwise long-range cross-comp is prohibitive

kwcckw: > the whole point of centroid clustering now is select Ns with high w to C

We already have the pruning section in cluster_P, and the weak N to their C are discarded. Pruning Ct.N_ is actually pruning those Cs, which is a different case and why we need to do that?

boris-kz: cluster_P is not a default, pruning should be. Yes, it's actually pruning concat C.N_s

kwcckw: Okay, since we prune Cs' N_ here, and Cs are spliced from all tiles, i guess we can splice those individual C with tile's connected Cs first, and then only prune them?

boris-kz: Right, so that w will be sum N.root_

kwcckw: > Right, so that w will be sum N.root_

When we merge Cs, their N.w should be recompute? Since the w is only relative to the original C, but not the merged C.

boris-kz: I don't think it matters, it still shows how representative the N is.
so it's something like this:

    if (Ct := G.Ct) and (len(E_ := [n for C in Ct.N_ for n in C.N_ if n.w>ave])-1)**2 *wL > ave:
        if sum(e.w for e in E_) * ((Ct.c+wAgg) / (Ct.r+r+cAgg)):
            cross_comp(E_,r)  # cross_comp Ct.N_'N_, still sub+

kwcckw: Okay, but we can't cross_comp E_, that E_ needs to be summed into Nt first, and we didn't pack the resulting G_ or Nt from cross_comp?

boris-kz: right,
            G.Nt = sum2F(e_,G)
            cross_comp(G,r)  # cross_comp Ct.N_'N_, still sub+

kwcckw: > I don't think it matters, it still shows how representative the N is. so it's something like this:
> 
> ```
> if (Ct := G.Ct) and (len(E_ := [n for C in Ct.N_ for n in C.N_ if n.w>ave])-1)**2 *wL > ave:
>     if sum(e.w for e in E_) * ((Ct.c+wAgg) / (Ct.r+r+cAgg)):
>         cross_comp(E_,r)  # cross_comp Ct.N_'N_, still sub+
> ```

But could you clarify if this is in sum2G? G.Ct should be empty before cluster_C, so this whole section is valid only when there's C fork?

boris-kz: yes, for base tiles it's sub+ in sum2G only, default cluster_C and agg+ is higher tiles only

frame_2D_alg/agg_recursion.py
boris-kz: so it's similar to sub+:

        if tile_:  # sparse,2D
            # add eval:
            E_ = get_exemplars({n for N in Fr.N_ for n in N.N_}, R,C)
            cluster_C(Fr,E_, R,C)  # to select cross_comp, as in sub+, no direct agg+
            if (Ct := Fr.Ct) and (len(e_ := [n for C in Ct.N_ for n in C.N_ if n.w > ave]) - 1) ** 2 * wL > ave:  # or just eval Ct?
                if sum(e.w for e in e_) * ((Ct.c+wAgg) / (Ct.r+ R+cAgg)):
                    Fr.Nt = sum2F(e_, Fr)
                    if cross_comp(Fr,R):  # agg+, spec-> tN_,tC_,tL_, proj comb N_'L_?
                            Fr,aTT,oTT,aH,oH = ffeedback(Fr, aTT,oTT,aH,oH)  
                            # term,form oH(aH?

frame_2D_alg/agg_recursion.py
```diff
@@ -879,8 +885,17 @@ def fill_frame(_iy,_ix, elev, T):  # expand level_frame from pixel-level seed ti
     if not T or not max_elev: return T, ([], np.zeros((2,9)),0,0,0)  # frame_H(0) = pixel tile
     while elev < max_elev:
         tile_,C,R = fill_frame(iY,iX, elev, T)  # project from seed tile
-        if tile_: # sparse,2D
-            Fr = sum2F(tile_)  # higher-scope tile( oH( aH
```
boris-kz, line 883: >>Splicing is only for Gs terminated at tile box, if they share termination coords

>This should be complicated because we need to identify if the G is located in the corner or side of the >tile, where it may splice in different ways if it's located in corner compare to side.

I don't know how to make it simpler.
so it will be here, before cluster_C

frame_2D_alg/agg_recursion.py
kwcckw: > default cluster_C and agg+ is higher tiles only

I see, so we have cluster_C in both elev+ an sub+ now.

frame_2D_alg/agg_recursion.py
kwcckw: Yes, the splicing should be here, but that should be before getting the exemplar and cluster_C?

        # add eval:
        merge_G_(Fr)
        E_ = get_exemplars({n for N in Fr.N_ for n in N.N_}, R,C)

frame_2D_alg/agg_recursion.py
boris-kz: But this is for spliced G_s, 
we still init cross_comp(tile_), that's a base representation, as pixels in base tiles?

kwcckw: Yes, i guess we can include that first before splicing those Gs in the second step.

PR #209, opened 2026-07-18
7/18/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -570,6 +570,9 @@ def sum2G(ft_, fTT, root=None, init=1):  # core clustering function
             else:      G_,r = cluster_N(G.Nt,E_,r,c)  # updates G
             if G_ and gv_(val_(G.Nt.dTT,G.wTT*ttA) * ((G.c+wAgg) /(G.r+r+cAgg)) * ((len(G_)-1)**2 *wL) - ave):  # if full cross_comp?
                 cross_comp(G.Nt,r)
+            # this is actually done in cluster_C already
+            # if G.Ct and gv_(val_(G.Nt.dTT,G.wTT*ttA) * ((G.c+wAgg) /(G.r+r+cAgg)) * ((len(G.N_)-1)**2 *wL) - ave):
+            #     cross_comp(G.Ct,r)  # cross_comp G.N_ via Ct.N_ exemplars, still sub+
```
kwcckw, line 575: This section is added from your prior comment, but i think it's already done in the cluster_C? We already have cross_comp of Ct within cluster_C.

frame_2D_alg/agg_recursion.py
```diff
@@ -881,11 +884,37 @@ def fill_frame(_iy,_ix, elev, T):  # expand level_frame from pixel-level seed ti
...
-                    Fr,aTT,oTT,aH,oH = ffeedback(Fr, aTT,oTT,aH,oH)  # term,form oH(aH
+            cross_comp(Fr.Nt, rr=0)  # add base representation
+            N__ = [N for tile in tile_ for N in tile.N_]
+            # splice Ns
+            for _N in N__:
+                for i, N in enumerate(N__):
+                    if _N is N: continue
+                    if _N.box[0] <= N.box[2]+1 and N.box[0] <= _N.box[2]+1 and _N.box[1] <= N.box[3]+1 and N.box[1] <= _N.box[3]+1:  # check adjacency          
+                        # not sure, i guess we need to comp again to eval for merge?
+                        L = comp_N(_N,N, (_N.r+N.r)/2, min(_N.c,N.c), A=(dy_dx:=_N.yx-N.yx), span=np.hypot(*dy_dx))
```
kwcckw, line 895: To eval for splicing between Gs, other than eval using their box, i guess we at least need to compare them and eval using the resulting m?

frame_2D_alg/agg_recursion.py
```diff
@@ -881,11 +884,37 @@ def fill_frame(_iy,_ix, elev, T):  # expand level_frame from pixel-level seed ti
...
+                            add2F(_N,N,1); add_Nt(_N)          # merge N into _N
+                            _N.m,_N.d = val_(_N.dTT, fd=1)     # recompute m and d
+                            for l in N.rim: l.N_[l.N_.index(N)] = _N  # update N.rim's nt from N to _N
+                            _N.rim += N.rim        # merge rim
+                            _N.Rt = sum2F(_N.rim)  # update Rt
+                            N__[i] = _N            # update N to _N                
+            N__ = list(set(N__))  # remove duplicated merged Ns
+            # add eval:  (i think this can be default? we have Ct eval below anyway)
+            E_ = get_exemplars(N__, R,C)
+            Fr.H += [Fr.Nt]; Fr.Nt = sum2F(N__)  # pack exsiting base tile into H first?
```
kwcckw, line 907: For base representation (tiles), i guess we can pack them into H first, before we pack the splice Gs into Nt.N_?

frame_2D_alg/agg_recursion.py
boris-kz: yes but that's cross_comp Ct.N_, we need to change that to
cross_comp( G.N_:= [n for C in Ct.N_ for n in C.N_ if n.w>ave])

frame_2D_alg/agg_recursion.py
boris-kz: the tiles are temporary representations, unpacked by cluster_C, so we probably don't need them in H.
this tile-level fill_frame is similar to base frame_blobs, both form "blobs" latter unpacked in sub+.
the difference is additional tile.N_ splicing on higher levels, base pixels don't have N_

frame_2D_alg/agg_recursion.py
boris-kz: yes, just base_comp?

frame_2D_alg/agg_recursion.py
kwcckw: Ct is summed from out_:

      cross_comp(Ct,r)  # all distant Cs, seq C_ in eigenvector = argmax(root.wTT)?

So the cross_comp is actually comparing C's N_, which is the same input N_ to the cluster_C?

frame_2D_alg/agg_recursion.py
kwcckw: Yes, that's better, and we can remove the redundant rim now, thanks.

frame_2D_alg/agg_recursion.py
kwcckw: I see, then we not really need sum Fr with tiles:

    Fr = sum2F(tile_)
    
The Fr could be summed with those unpacked tiles' N_ instead.

frame_2D_alg/agg_recursion.py
boris-kz: yes, but it's selective by combined w

frame_2D_alg/agg_recursion.py
boris-kz: ok we probably do need the tiles in H[0], it's the lowest processing level

frame_2D_alg/agg_recursion.py
```diff
@@ -881,11 +884,37 @@ def fill_frame(_iy,_ix, elev, T):  # expand level_frame from pixel-level seed ti
...
             Fr = sum2F(tile_)  # higher-scope tile( oH( aH
-            if cross_comp(Fr.Nt, rr=0):  # spec-> tN_,tC_,tL_, proj comb N_'L_?
-                if elev and ffb:  # ffb =1 in main, no ffeedback in added tiles
-                    Fr,aTT,oTT,aH,oH = ffeedback(Fr, aTT,oTT,aH,oH)  # term,form oH(aH
+            cross_comp(Fr.Nt, rr=0)  # add base representation
+            N__ = [N for tile in tile_ for N in tile.N_]
+            # splice Ns
+            for _N in N__:
+                for i, N in enumerate(N__):
+                    if _N is N: continue
```
boris-kz, line 892: how is that possible?

frame_2D_alg/agg_recursion.py
boris-kz: if (Ct := G.Ct):
        e_ = {n for C in Ct.N_ for n in C.N_}
        for e in e_: e.w = sum(e.root_)  # combine memberships: sub+ cross_comp(representative Ns)
        e_ = [e for e in e_ if e.w]
        if sum(e.w for e in e_) * ((Ct.c+wAgg) / (Ct.r+r+cAgg)) * ((len(e_)-1)**2 * wL):  # or eval Ct?
            cross_comp(sum2F(e_,G), r)  # Ct.N_'N_

frame_2D_alg/agg_recursion.py
kwcckw: Under this scheme, i have another `N__[i] = _N ` below, so that the subsequent `_N` could be the merged N. The reason is the splicing of `N` into `_N` should have new boundary and we need to recheck `_N` boundary again for new splicing Ns.

frame_2D_alg/agg_recursion.py
```diff
@@ -881,11 +884,37 @@ def fill_frame(_iy,_ix, elev, T):  # expand level_frame from pixel-level seed ti
...
+                    if _N.box[0] <= N.box[2]+1 and N.box[0] <= _N.box[2]+1 and _N.box[1] <= N.box[3]+1 and N.box[1] <= _N.box[3]+1:  # check adjacency          
+                        # not sure, i guess we need to comp again to eval for merge?
+                        L = comp_N(_N,N, (_N.r+N.r)/2, min(_N.c,N.c), A=(dy_dx:=_N.yx-N.yx), span=np.hypot(*dy_dx))
+                        if gv_(L.m - ave*(L.r)):  # merge eval
+                            _N.rim.remove(L); N.rim.remove(L)  # remove the added L from comp_N above
```
boris-kz, line 897: your comments here are pure fluff.
I didn't check the code yet

kwcckw: Yes, some of them are no longer relevant. This rim removal should be removed now with base_comp:

    _N.rim.remove(L); N.rim.remove(L) 
    
But i think add2F, add_Nt and rim related operations are still relevant to compute the merged G attrs.

boris-kz: the comments are dumb regardless

kwcckw: Okay sorry, i get what you mean now. Those comments are really pointless and redundant.

PR #211, opened 2026-07-20
7/20/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -795,7 +821,9 @@ def trace_edge(N_,_G_,_TT,_C, r,root):  # cluster contiguous shapes via PPs in e
             cT_.add(cT)
             dy_dx = _N.yx-N.yx; dist = np.hypot(*dy_dx)  # Rc = r+ (N.r+_N.r)/2
             Link = comp_N(_N,N, r,_C,A=dy_dx, span=dist)
-            if val_(Link.dTT,ttTrc,1)[1] * ((Link.c+wTrc)/(r+cTrc)) > ave:  L_+=[Link]  # r = 1|2, add Bt?
+            if val_(Link.dTT,ttTrc,1)[1] * ((Link.c+wTrc)/(r+cTrc)) > ave:  
+                L_+=[Link]  # r = 1|2, add Bt?
+                N.rim += [Link]; _N.rim += [Link]  # or change PP's typ to 3 in F2N? The existing comp_N doesn't pack rim in PPs, and we need the rim in flood fill below
```
kwcckw, line 826: We need rim in flood fill below, but comp_N adds link only if N.typ == 3, and PP/typ == 0, so we need to add it externally here.

frame_2D_alg/agg_recursion.py
```diff
@@ -121,13 +121,30 @@ def proj_V(_N,N, dist, dy_dx, dec, r):  # _N x N induction
 
 def comp_N_(_pairs, r, tnF=None, root=2):  # incremental-distance cross_comp, max dist depends on prior match
 
+    # not fully review, from fable:
+    def proj_T(N, tile):  # to splice tiles: if N projection to tile box y|Y|x|X: tile.y_|Y_|x_|X_ += [N]
```
kwcckw, line 125: This is from fable, not fully review yet. Basically the projection to tile box is based on the distance of N to each of the tile's edges?

boris-kz: yes, these sides is where the splicing may occur
this function is redundant, we have proj_N

frame_2D_alg/agg_recursion.py
```diff
@@ -121,13 +121,30 @@ def proj_V(_N,N, dist, dy_dx, dec, r):  # _N x N induction
...
+                s_ += [N]
+
     pairs, olp_ = [],[]  # no olp_?
     for _N,N in _pairs:  # get all-to-all pre-links
         if _N.sub != N.sub: continue  # or comp x composition?
         if N is _N: olp_ += [N]  # overlap = unit match, no miss
         else: dy_dx = _N.yx-N.yx; dist = np.hypot(*dy_dx); c = min(_N.c,N.c); pairs += [[dist, dy_dx, _N,N, c]]
     # to splice tiles:
     # if _N projection to tile box y|Y|x|X: tile. y_|Y_|x_|X_ += [N]
+    if pairs:
```
boris-kz, line 144: why?

kwcckw: Since the projection is per N, i think we can move the N projection to sum2G (in each G.N_) instead of this section?

frame_2D_alg/agg_recursion.py
kwcckw: I see, so we just need to compute the distance of N to each side, where A is [dy,dx], r and c will be based on N, and decay is default value?

     proj_N(N, dist, A,_r,_c, dec=1)

boris-kz: sorry, this whole scheme is way too coarse.
splicing should really be called from comp_N spec sequence:

        if N.typ < 3:  # L | C | Nt?
        else:  # CN

if we need to merge / unpack Ns instead of forming a link

frame_2D_alg/agg_recursion.py
boris-kz: so this is no longer relevant

frame_2D_alg/agg_recursion.py
kwcckw: > we need to merge / unpack Ns instead of forming a link

If it's in comp_N spec sequence, the merging is still limited to single tile only?

boris-kz: no, this would be after that default comp tiles, where the tiles are unpacked if they match
so that's the case of two-level spec:

        if N.typ < 3:  # L | C | Nt?

between tiles, then between Ns:

        else:  # CN

kwcckw: >  where the tiles are unpacked if they match

So we still need the cross_comp between tiles (the prior verison), and the splicing should be done after that?

boris-kz: Yes, just nt_ = comp_N_(tile_), then cluster_C( get_exemplars(nt_))

kwcckw: This is comparing tile_:

    nt_ = comp_N_(tile_)

But cluster_C should be clustering tile_'s N_?

boris-kz: ok right, there is no spec, just get exemplars(_N_+N_)

kwcckw: So if the merging is in the spec, it will be in the comp_F or right after comp_F? 

Then after the merging, multiple tiles might have the same merged N.

boris-kz: they can't be the same, tiles don't overlap.
so we need a light xcomp for that:

def cross_comp(Ct, R,root, fC=0):  # calls cluster_N, sub+, over exemplars spliced from C_

    e_ = {n for C in Ct.N_ for n in C.N_}; G_ = []
    for e in e_: e.w = sum(e.m_)  # combine memberships
    e_ = [e for e in e_ if e.w]
    if sum(e.w for e in e_) * ((Ct.c+wAgg) / (Ct.r+R+cAgg)) * ((len(e_)-1) **2 * wL):
        # root = Ft, converted below, rc=rdn+olp, sub+ cross_comp select Ct.N_'N_s:
        if Lt := comp_C_(e_,R,fC=1) if fC else comp_N_(combinations(e_,2), R):
            L_,TT,c,r, cV = Lt
            root.L_ = L_  # val=m+d /clust, m/comp
            if gv_(val_(TT*ttcN) * ((c+wcN)/(r+ccN)) * ((len(L_)-1)*wL) - ave):  # return +ve, store -ve gate vals
                G_,r = cluster_N(root,E_,r,c)  #> sum2G
                if G_:
                    root.Nt = sum2F(G_,root,froot=2)  #| C_?
                    if Ct := root.Ct:
                        cross_comp(Ct,r,root) # sub+'agg+
    return G_

boris-kz: def xcomp(N_, merge=1):  # light version for contiguous tiles, also Bs?

    out_ = []
    for N in N_:
        if N.fin: continue  # merged
        for _N in N.rim:  # contiguous only, maybe in 2D and deep
            m = val_(base_comp(_N,N)[0])
            if m > ave:
                add2F(_N.N, merge); N.fin = 1
        out_ += [N]
    return out_

kwcckw: > they can't be the same, tiles don't overlap.

If we splice Gs across tiles, multiple tiles maybe referenced to a same spliced G?

Also the xcomp above is for contiguous tiles, where each N is tile?

Or it is for the merged N_ from different tiles?

boris-kz: yes, contiguous tiles.
splicing forms tile blobs, it doesn't affect Ns until final cross_comp, with spec in comp_N:

        if tile_:  # sparse,2Dwes
            Fr.H += [sum2F(tile_)]  # minimaly processed
            Fr.N_ = xcomp(tile_)  # merge if match
            for t in Fr.N_:  # tile "blobs"
                if t.m * t.c * wcC > ave * (t.r+ccC):
                    cluster_C(Fr, get_exemplars(t.N_,R,C), R,C)
            if (Ct := Fr.Ct):  # no direct agg+
                cross_comp(Ct,R,Fr)

kwcckw: Okay, this is actually much more simpler now when we merge contiguous tiles, and their Ns will be in the same merged tile now.

Also this `cluster_C(Fr, get_exemplars(t.N_,R,C), R,C)` is per tile_ or tile? It's parsing Fr for all tiles in `tile_`, but exemplars are computed from T.N_ only?

boris-kz: cluster_C is per merged Tile.
not sure if Ct should be per Tile or Fr now

boris-kz: Read the current agg_recursion. The design ordering itself is sound: centroids first is the cheap pass (centroid-to-member `base_comp`, frontier via formation rims, no new pairwise comps), membership then gates the expensive `comp_N_`/`cluster_N`, and cross-tile binding correctly falls to the distance-sorted global comp since rims are intra-tile. But the revision changed the caller contract of `cluster_C`/`cluster_N`, and several call sites are still on the old contract — tracing to closure:

**1. Fork-vs-node contract split — the structural break.** Both clusterers assume `Ft` is a fork: pool = `Ft.N_`, owner = `Ft.root`. `sum2G` honors this (`G.Nt`); `frame_H` passes `Fr` and `cross_comp` passes `root` (CN). Consequences:
- `frame_H`'s `cluster_C(Fr,...)`: pool becomes `Fr.N_` = tiles, not member nodes, so the `n not in N_` init wipes the exemplar's self-membership set two lines earlier; final attach `Ft.root.Ct = Ct` hits `Fr.root` = None → crash if the gate passes, else `Fr.Ct` stays empty and step 3 never runs. Intended target is `Fr.Ct`.
- `cross_comp`'s `cluster_N(root, E_...)`: membership test `_N in Ft.N_` fails (`Fr.N_` = tiles, e_ are member Gs) → flood-fill can't expand past seeds → `(len(N_)-1)*wL` = 0 → `G_` empty, silent no-op. Also `Ft.root.root` crashes on the frame path.
- The reform block `rG = Ft.root; rG.Nt...` targets one level too high from `cross_comp` — and `cross_comp` then assigns `root.Nt` again: two conflicting Nt writes. Same mislevel on the boundary path: `cross_comp(Bt, br, Bt)` → F2N → `cluster_N` → `rG = Bt.root = G`, so boundary clusters overwrite the core `G.Nt`.

**2. frame_H control flow.** `return Fr` is inside the while → one iteration ever; `if elev and ffb` is False at elev=0 → break path always taken → ffeedback and frame expansion are dead code. And per-tile `cluster_C` attach is an overwrite (`Ft.root.Ct = Ct`), not accumulation — only the last tile's centroids would survive even with correct targeting.

**3. Step 1 merge never happens.** Nothing populates tile rims — `fill_frame` has the 2D grid but adjacency is never written to `rim`, so `xcomp` iterates empty rims. Inside `xcomp`: `add2F(_N.N, merge)` should be `add2F(_N, N, merge)`; absorbed N (fin=1) is still appended to `out_`; no break after absorption, so one N can merge into several `_N`s.

**4. "Most representative" is not implemented.** `cross_comp`'s entry gate is a truthiness test (`if sum(...) * ... * ...:`), no `> ave`. `e.w = sum(e.m_)` is computed then unused — all nonzero-membership members enter `combinations(e_,2)`. Note `cluster_C` already writes representativeness into `n.exe` and `cluster_N` gates on exe at depth, but `cross_comp`/`get_exemplars` ignore it and re-derive from Rt — two parallel selection channels. Also `comp_C_`'s fC merge path is dead: no caller passes `fC=1`, so similar per-tile centroids are never merged before node-level comp.

**5. sum2G placeholder root.** `cluster_N` calls `sum2G(ft_, ttcN, CN())` → `root.root is None` misidentifies every sub-G as a tile (gets y_/x_ attrs), `rroot=0` kills `Bt.brrw`, and G's up-chain is a detached node. Pass `Ft.root`. With that fixed, step 4's recursion inside `sum2G` (cluster_C|N → `G.Ct` → `cross_comp(G.Ct, r, G)`) is the correctly-wired loop — the tail call in `cross_comp` (`if Ct := root.Ct`) duplicates it one level up and re-enters on the *same* Ct (its refresh, the `sum2F(C_, root=Ft.root.Ct)` splice, targets the wrong level), terminating only by r inflation.

Smaller, but corrupting `e.w`: `cluster_P` resets `root_` but not `m_/d_` before the final `out_` sum2F, so members keep last-cycle memberships plus new ones — `e.w` double-counts. `cluster_N` init `sum2F(N.rim, N.Rt)` accumulates onto the Rt already assigned in `comp_N_` — rim counted twice for downstream projections. And the sub+ branch choice `gv_(Vn*(wcC-wcN)*(mdecay(L_)-decay)...)` flips direction with the sign of `wcC-wcN`, which feedback adjusts — encode "low slope → cluster_C" explicitly.

Unifying fix: one convention — always pass the fork whose `N_` is the pool being clustered (`t.Nt` / `Fr.Nt` with members spliced in, `G.Nt`), attach results to `Ft.root.<fork>`, reform `Ft` itself. That closes 1, most of 2, and the mislevel in 5–6 in one contract change.xcomp. The current `cross_comp(Bt, br, Bt)` is wrong on three counts, and even a corrected version would be redundant:

**1. As wired, it isn't der+ at all.** `cross_comp` unpacks `e_ = {n for C in Ct.N_ for n in C.N_}` — with Ct=Bt that's the *endpoint nodes* of boundary links, not the links. So it re-compares core/adjacent nodes under a wrong root (and its `cluster_N` reform overwrites `G.Nt` via `Ft.root`, per the mislevel from last time). To make it real der+ you'd need `e_ = Bt.N_` directly.

**2. Inter-G contour comparison is already covered one level up.** `comp_N` typ==3 fork-wise spec does `comp_F(_N.Bt, N.Bt)` — contours of different Gs get compared when the Gs are. So the only within-G job left for Bt is organizing contrast links into a contour summary. That's correlation clustering by contiguity + alignment, which is exactly xcomp's operation: after `F2N`, `base_comp` on link-nodes compares their ders, so m-merge = alignment-merge — merging aligned contrast links into contour segments is the right semantics.

**3. Selection is the criterion.** `comp_N_`'s machinery (proj_V, distance sort, prelinks, exemplars) exists to solve *which pairs to compare* under uncertainty. Bt has no selection problem: boundary links are adjacent by construction — rim = shared endpoint (derivable from `L.N_` intersection, no spatial grid). Same logic that justified xcomp for tiles. And d-value is mostly non-predictive on its own — it borrows to the core (`brrw`), which only needs the merged `Bt.m`, so a full per-G pipeline on the dd fork fails cost amortization; your gates would kill it almost always anyway.

Escalation path for the rare strong contour: keep the existing `bd*(wAgg*L) > avd*(...)` gate for F2N, xcomp-merge into segments, and promote surviving segments into the core's next-level `N_` — let the main pipeline carry them, rather than running a parallel hierarchy inside Bt.

Prerequisites: xcomp needs the fixes from before (rim population — here from endpoint-sharing, the `add2F(_N.N, merge)` typo, merged-N still appended to `out_`), plus a d-biased merge gate: for contours the merge test should weigh aligned d, not just m.

kwcckw: Actually why cluster_C is per merged Tile?

It should be per tile_, then Ct is per Fr?

If cluster_C is per merged Tile, we may get multiple Cts. Then those Cts, feedback will be per Tile too?

boris-kz: the whole point of comparing tiles is to merge them, so their Ns can then be merged in comp_N spec sequence. we may need a separate process to compare tile blobs, not sure.

kwcckw: > the whole point of comparing tiles is to merge them, so their Ns can then be merged in comp_N spec sequence.

But their Ns are not merged in the spec sequence? From comp_N, tLs is packed to L.H, and in trans_cluster, we are merging tL'nt's root again, which is those tiles?

frame_2D_alg/agg_recursion.py
```diff
@@ -121,13 +121,30 @@ def proj_V(_N,N, dist, dy_dx, dec, r):  # _N x N induction
 
 def comp_N_(_pairs, r, tnF=None, root=2):  # incremental-distance cross_comp, max dist depends on prior match
 
+    # not fully review, from fable:
+    def proj_T(N, tile):  # to splice tiles: if N projection to tile box y|Y|x|X: tile.y_|Y_|x_|X_ += [N]
+        y0,x0,yn,xn = tile.box
```
boris-kz, line 126: spec can form merged sub_Ns, but that currently doesn't replace their root Ns, this would be in 

def sort_H(H, fi):  # lev.rc = complementary to root.rc and priority index in H, if selective and aligned

    i_ = []  # priority indices
    for i, lay in enumerate(sorted(H.node_, key=lambda lay: [lay.m,lay.d][fi], reverse=True)):
        di = lay.i - i  # lay index in H
        lay.r += di  # derR - valR
        i_ += [lay.i]
    H.i_ = i_  # H priority indices: node/m | link/d
    if fi > 1:
        H.root.node_ = H.node_
    # more advanced ordering: dH | H as medoid cluster of layers, nested cent_TT across layers?

kwcckw: This sort_H is sorting layers based on priority indices. For spec, how is this can be used to merged NS?

boris-kz: The Ns will be merged based on trans links in their rim anyway

PR #212, opened 2026-07-21
7/21/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -108,15 +108,21 @@ def cross_comp(Ft, R,root, fC=0, fB=0):  # calls cluster_N, sub+, over exemplars
     return G_
 
 def xcomp(N_, merge=1):  # light version for contiguous tiles, also Bs?
-    out_ = []
+    for _N,N in combinations(N_,2):
+        if (abs(_N.box[1]-N.box[3]) == 1 and _N.box[0] == N.box[0]) or (abs(_N.box[0]-N.box[2]) == 1 and _N.box[1] == N.box[1]):  # only adjacent, no overlap
+            _N.rim += [N]; N.rim += [_N]
```
kwcckw: So i think we need to add adjacent tiles as rim first, and then merge them below.

In this scheme, `_N.fin` is the merged root N is there's any prior merging, and we need to replace `_N` to `N` in `N_`: `N_[N_.index(_N)] = N` since we may get new rim after the merging.

frame_2D_alg/agg_recursion.py
```diff
@@ -897,7 +903,8 @@ def fill_frame(_iy,_ix, elev, T):  # expand level_frame from pixel-level seed ti
             Fr.N_ = xcomp(tile_)  # merge if match
             for t in Fr.N_:  # tile "blobs"
                 if t.m * t.c * wcC > ave * (t.r+ccC):
-                    cluster_C(Fr, get_exemplars(t.N_,R,C), R,C)
+                    cluster_C(t.Nt, get_exemplars(t.N_,R,C), R,C)
+            Fr.Ct = sum2F([t for t in Fr.N_ for C in t.Ct.N_])  # sum from individual tile's Cs instead?
```
kwcckw: Under the current structure, i guess we can sum tiles' Cs to Fr.Ct?

frame_2D_alg/agg_recursion.py
```diff
@@ -108,15 +108,21 @@ def cross_comp(Ft, R,root, fC=0, fB=0):  # calls cluster_N, sub+, over exemplars
     return G_
 
 def xcomp(N_, merge=1):  # light version for contiguous tiles, also Bs?
-    out_ = []
```
kwcckw: > The Ns will be merged based on trans links in their rim anyway

From trans_cluster:

    def trans_cluster(G):
        for L in G.L_:
            for lev in L.H[1:]:  # L.H[0] is direct ders
                for tFt in lev.N_:  # Lt doesn't form trans-links, Ct is not root-constrained?
                    for tL in tFt.N_:
                        if tL.m*wcN > ave*ccN:  # merge trans_link.N_.roots
                            rt0 = getattr(tL.N_[0].root,'root',None); rt1 = getattr(tL.N_[1].root,'root',None)  # CNs
                            if rt0 and rt1 and rt0 != rt1:
                                if rt1.H: add_H(rt0.H, rt1.H, rt0, fN=1)
                                rt0.N_ += rt1.N_; add_Nt(rt0)  # recompute Nt attrs / G
                L.Nt,L.Bt,L.Ct = CF(),CF(),CF()

Those Ns are tL.'s nt, which they will be used to merge rt only? We didn't merge them using trans link?

boris-kz: rt0.N_ += rt1.N_; add_Nt(rt0)  # recompute Nt attrs / G
                L.Nt,L.Bt,L.Ct = CF(),CF(),CF()
            # merge roots

the idea is to merge nt roots

frame_2D_alg/agg_recursion.py
boris-kz: the rim should be added in 

    def fill_frame(_iy,_ix, elev, T, adj_T_):  # expand level_frame from pixel-level seed tile, similar to frame_b;obs

        PV__ = np.zeros([Ly,Lx])
        T_ = []
        while T and gv_(val_(T.dTT*T.wTT*ttFrm) * ((T.c+wFrm)/(T.r+cFrm)) - ave):
            pTT, pc = proj_N(T, np.hypot(*dy_dx), dy_dx, elev,T.c)  # no proj r?
            if gv_(ave - val_(pTT*T.wTT*ttFrm) * ((pc+wFrm)/(T.r+elev+cFrm))):  # inverted val
                proj_focus(PV__,y,x,T)
                y,x = np.unravel_index(pv__.argmax(), PV__.shape)
                if gv_(PV__[y,x] - ave):
                    _T.rim += [T]
                    for t in [T] + adj_T_: t.rim += [_T]  # add to 2D-adjacents' rim 
                else: break
            else: break

frame_2D_alg/agg_recursion.py
```diff
@@ -108,15 +108,21 @@ def cross_comp(Ft, R,root, fC=0, fB=0):  # calls cluster_N, sub+, over exemplars
...
+    for _N,N in combinations(N_,2):
+        if (abs(_N.box[1]-N.box[3]) == 1 and _N.box[0] == N.box[0]) or (abs(_N.box[0]-N.box[2]) == 1 and _N.box[1] == N.box[1]):  # only adjacent, no overlap
+            _N.rim += [N]; N.rim += [_N]
     for N in N_:
-        if N.fin: continue  # merged
```
boris-kz: N can be merged too

frame_2D_alg/agg_recursion.py
kwcckw: But nt's roots are tile, which we already merge them in xcomp?

frame_2D_alg/agg_recursion.py
kwcckw: Yes, but i have this `N_[N_.index(_N)] = N` below, so if N merge any `_N` in prior iteration, it should be recheck for their new rim again.

frame_2D_alg/agg_recursion.py
boris-kz: no, this is for sub_Ns, the roots are the Ns

frame_2D_alg/agg_recursion.py
boris-kz: I think that's more awkward

frame_2D_alg/agg_recursion.py
boris-kz: I guess, thanks

frame_2D_alg/agg_recursion.py
```diff
@@ -108,15 +108,21 @@ def cross_comp(Ft, R,root, fC=0, fB=0):  # calls cluster_N, sub+, over exemplars
...
+                while _N.fin:
+                    _N = _N.fin
+                    if _N is N: break   
+            if N is _N: continue     
             m = val_(base_comp(_N,N)[0])
             if m > ave:
-                add2F(_N.N, merge); N.fin = 1
-        out_ += [N]
-    return out_
+                add2F(N,_N, merge); _N.fin = N
```
boris-kz: _N.fin = N?

frame_2D_alg/agg_recursion.py
kwcckw: The sub comp is retrieved from tile.Ft.N_ `_N_,N_=_F.N_,F.N_`:

    if _F.nF == F.nF:  # sub-comp
        _N_,N_=_F.N_,F.N_; nF=F.nF; l = nF=='Lt'
        if  _N_ and N_:
            if l: Np_ = [[_n,n] for _n,n in zip(_N_,N_) if _n and n]  # same forks
            L = len(Np_)-1
            tt = (rL.dTT*rL.c + ddTT*dF.c) / (rL.c+dF.c)
            if gv_(val_(tt,ttF)* (c/r)* (wF*L) - ave* (r+cF*L)):
                if l: L_= [L for Np in Np_ for L in comp_F(*Np, r,rL=dF).N_]; TT,C,R = sum_vt(L_, wTT=ttF)

Where F is tile.Ft, and tL.nt is actually tile.N_, so tL.nt is referencing tile now?

frame_2D_alg/agg_recursion.py
kwcckw: Okay, without replacement, we still need to check if N.fin, and if it's merged, we need to get the prior merged N and check N.rim again? Because we didn't check N.rim after the merging.

frame_2D_alg/agg_recursion.py
kwcckw: This is the reference to the merged N after the merging, so that we can retrieve it to check for new rim again.

frame_2D_alg/agg_recursion.py
boris-kz: that's for Ns, I am talking about sub_Ns in comp_N spec:

        if N.typ < 3:  # L | C | Nt, merge?

frame_2D_alg/agg_recursion.py
boris-kz: I guess we need to merge rims in 
add2F(_N.N, merge)?

frame_2D_alg/agg_recursion.py
boris-kz: just merge the rims

frame_2D_alg/agg_recursion.py
kwcckw: > just merge the rims

Yes, i think we can have `N.rim += _N.rim` externally, and this will grow the N.rim while looping it.

frame_2D_alg/agg_recursion.py
boris-kz: actually, we can just merge the tiles base on their m, similar to blobs based on G
comparison is kind of pointless because they are not clusters

kwcckw: We can use G to identify if the blobs are similar. For tiles, group of high m or low m tiles may not be similar?

boris-kz: the tiles don't need to be similar, they are just arbitraty aggregates of Gs

kwcckw: Okay, in term of merge eval, it will be based on their m difference? Something like `abs(_tile.m - tile.m) < ave`?

boris-kz: actually it doesn't matter, the tiles should be small enough to just splice their N_s anyway:

        tile_,C,R = fill_frame(iY,iX, elev, T)  # also prior 2D adjacents, project from seed tile
        if tile_: # sparse,2D
            Fr.Nt = sum2F(tile_)
            Fr.H += [Copy_(Fr.Nt)]  # min processed level in higher-scope tile( oH( aH
            Fr.N_ = [g for t in tile_ for g in t.N_]
            cluster_C(Fr, get_exemplars(Fr.N_,R,C), R,C)
            if Ct := Fr.Ct:  # agg+ over edge exemplars
                cross_comp(Ct,R,Fr)

kwcckw: Okay, but there's a problem here when we apply cluster_C on the spliced N_ without comp because their rim is still empty?

We would need the rim to get the frontier during cluster_C. I guess we still need to compare them first?

boris-kz: In the new scheme, CC is be done when CN terminates by some discontinuity, which is what happens in vect_edge. So the rim in these Ns is actually Bt?

kwcckw: But Bt is also limited to single tile, and it won't be crossing to another tile?

boris-kz: yes, we probably need to constrain CC by proximity instead of existing rim. I will check tomorrow

kwcckw: Okay, so i guess we will need another process to check all existing Ns to get their distance first.

boris-kz: Yes, actually this is the same as forming pLs in comp_N_?

kwcckw: Yes, so do we need a same process in Cs' comp_C_ too? Right now comp_C is actually still using the existing comp_N.

PR #213, opened 2026-07-22
7/22/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -883,9 +887,23 @@ def fill_frame(_iy,_ix, elev, T, adj_T_=[]):  # expand level_frame from pixel-le
...
+            for pL in sorted(pairs, key=lambda x: x[0]):  
+                dist, dy_dx, _N,N, lc = pL 
+                pTT = proj_V(_N,N, dist, dy_dx, decay** (dist/((_N.span+N.span)/2)), Fr.r)
+                m,d = val_(pTT,ttN_,1)
+                if m > 0: 
+                    pL = CL(typ=-1, N_=[_N,N],dTT=pTT,m=m,d=d,c=lc,r=(_N.r+N.r)/2, angl=[dy_dx,1],span=dist)
+                    _N.rim += [pL]; N.rim += [pL]
+            '''
+
+            comp_N_(list(combinations(Fr.N_,2)), Fr.r,fpL=1)
```
kwcckw, line 905: To compute the similar pL in N.rim here, we actually just need to add fpL to comp_N_? I think adding fpL is better than the similar section of code above.

frame_2D_alg/agg_recursion.py
```diff
@@ -867,8 +867,12 @@ def fill_frame(_iy,_ix, elev, T, adj_T_=[]):  # expand level_frame from pixel-le
                 if gv_(PV__[y,x] - ave):
                     iy = _iy+ (y-cy)*(T.box[2]-T.box[0]); ix = _ix+ (x-cx)*(T.box[3]-T.box[1])
                     _T = frame_H(image, iy,ix, Ly,Lx, Y,X, rV, max_elev=elev)  # expand new tile to current level, no fb
-                    _T.rim += [T]
-                    for t in [T] + adj_T_: t.rim += [_T]  # add to 2D-adjacents' rim
+                    if _T:   
+                        _T.rim += [T]
+                        for t in [T] + adj_T_: t.rim += [_T]  # add to 2D-adjacents' rim
+                        adj_T_ += [_T]
```
kwcckw, line 873: This `adj_T_` should be local since we shouldn't have adjacent Ts with different elevations?

Also adj_T_ should be init with T, and pack the subsequent `_T`s?

boris-kz: Not sure this is still relevant, we probably need global tile_ N_ splicing now

kwcckw: This global tile_'s N_ should be still limited to specific elevation right?

boris-kz: Yes, it will only extend for higher-agg Gs on higher levels

frame_2D_alg/agg_recursion.py
boris-kz: just make it a separate function:

def pL_(_pairs, root, r):
    
    def proj_V(_N, N, dist, dy_dx, dec, r):  # _N x N induction
        Dec = dec or decay ** ((dist / ((_N.span + N.span) / 2)))
        iTT = (_N.dTT + N.dTT) * Dec
        C = min(_N.c, N.c); R = (_N.r + N.r) / 2
        if val_((eTT + iTT) * ttPrj) * (C / (cPrj + r + R)) * wPrj > ave:  # not oF, spec / link:
            eTT += proj_N(N, dist, dy_dx, r, N.c, dec)[0]  # pTT/ L_,B_,rim, if pV >0
            eTT += proj_N(_N, dist, -dy_dx, r, _N.c, dec)[0]  # reverse direction
        return iTT + eTT
    
    N_, L_ = [],[]
        dist, dy_dx, _N,N, lc = pL  # rim angl is not canonic
        pTT = proj_V(_N,N, dist, dy_dx, root.m if root!=2 else decay** (dist/((_N.span+N.span)/2)), r)  # based on current rim
        m,d = val_(pTT,ttN_,1); lr = r+ (N.r+_N.r)/2  # +|-match certainty

    return pairs

kwcckw: I think the following section is redundant in pL_? Since they should be recomputed using the pairs again.

          dist, dy_dx, _N,N, lc = pL  # rim angl is not canonic
          pTT = proj_V(_N,N, dist, dy_dx, root.m if root!=2 else decay** (dist/((_N.span+N.span)/2)), r)  # based on current rim
          m,d = val_(pTT,ttN_,1); lr = r+ (N.r+_N.r)/2  # +|-match certainty

boris-kz: actual comps are selective, we need projected vals to select pLs

kwcckw: Yes, so this pL_ is actually exclusive for frame_H only? Because comp_N_ needs both actual comps and pLs, and if we include and N_ and L_ packing, it's actually almost the same as comp_N_ now.

boris-kz: comp_N_ now is getting exemplars, we may need pL_ there for more selectivity.
but primarily it's for cluster_C, in both sub+ and agg+

def proj_L_(pairs, root, r, max=20, fall=1):

    def proj_V(_N, N, dist, dy_dx, dec, r):  # _N x N induction
        Dec = dec or decay ** ((dist / ((_N.span + N.span) / 2)))
        iTT = (_N.dTT + N.dTT) * Dec
        C = min(_N.c, N.c); R = (_N.r + N.r) / 2
        if val_((eTT + iTT) * ttPrj) * (C / (cPrj + r + R)) * wPrj > ave:  # not oF, spec / link:
            eTT += proj_N(N, dist, dy_dx, r, N.c, dec)[0]  # pTT/ L_,B_,rim, if pV >0
            eTT += proj_N(_N, dist, -dy_dx, r, _N.c, dec)[0]  # reverse direction
        return iTT + eTT

    N_, pL_, olp_ = [],[],[]  # no olp_?
    for _N,N in pairs:  # get all-to-all pre-links
        else: 
            dy_dx = _N.yx-N.yx; dist = np.hypot(*dy_dx)  # rim angl is not canonic
            if dist < max:
                pTT = proj_V(_N,N, dist, dy_dx, root.m if root!=2 else decay** (dist/((_N.span+N.span)/2)), r)  # based on current rim
                m,d = val_(pTT,ttN_,1) 
                if fall or m > ave:
                    lc = min(_N.c, N.c); lr = r + (N.r+_N.r) / 2  # +|-match certainty
                    pL_ += [[dist, dy_dx, _N, N, lc, lr]]
    return pL_

boris-kz: Actually, for cluster_C we need to pack pLs in N rim, pL_ is not needed?

kwcckw: > comp_N_ now is getting exemplars, we may need pL_ there for more selectivity.

So pL_ is the prerequisite for the comp_N to enable a higher selectivity? Right now pL is actually skipping comp_N.

> Actually, for cluster_C we need to pack pLs in N rim, pL_ is not needed?

Yes, so we can init pL instance instead of packing the tuple.

boris-kz: > Right now pL is actually skipping comp_N.

comp_N will be for selected pL_
so we can call proj_L_ in cross_comp too:

        if pL_ := proj_L_(combinations(e_,2), root, R):
            if Lt := comp_C_(pL_,R,fC=1) if fC else comp_N_(pL_, R):

kwcckw: Okay, and pL_ should be CL instead of tuple now?

boris-kz: why CL?

kwcckw: For frame_H, we need pLs to be CL and packed into N.rim instead? Otherwise ,we need to do that externally.

boris-kz: I think it's easier to keep separate n.prim:

    N_, pL_, olp_ = [],[],[]  # no olp_?
    for _N,N in pairs:  #-> all-to-all pre-links
        else:
            dy_dx = _N.yx-N.yx; dist = np.hypot(*dy_dx)  # rim angl is not canonic
            if dist < max:
                pTT = proj_V(_N,N, dist, dy_dx, root.m if root!=2 else decay** (dist/((_N.span+N.span)/2)), r)  # based on current rim
                m,d = val_(pTT,ttN_,1)
                if fall or m > ave:
                    lc = min(_N.c, N.c); lr = r + (N.r+_N.r) / 2  # +|-match certainty
                    pL = [dist, dy_dx, _N, N, lc, lr, pTT, m,d]
                    for l_ in pL_,_N.prim,N.prim: l_+=[pL]
    return pL_

def comp_N_(pL_, r, tnF=None, root=2):  # incremental-distance cross_comp, max dist depends on prior match

    L_,N_ = [],[]
    for dist, dy_dx, _N,N, lc,lr, pTT,m,d in pL_:
        if m > 0:
            if gv_(m*(lc/lr)*wN - ave*(r+cN)):  # comp if marginally predictable, update N.Rt pair eval, ave / proj surprise value?
                L_+= [Link]; N_+= [_N,N]
    if L_:
        for N in set(N_):
            if N.rim: N.Rt = sum2F(N.rim)
        cV = FV_(CoF.get(), *sum_vt(L_))  # +-ve Ls for oF, no oF.N_?
        return L_, cV  # +ve only, redundant +-ve for oF

frame_2D_alg/agg_recursion.py
boris-kz, line 860: this expands frame via last tile, please edit to expand via viable frontier as in cluster_C instead

boris-kz: Edited `fill_frame`, frontier-based:

```python
def fill_frame(_iy,_ix, elev, T):  # expand level_frame from pixel-level seed tile, similar to frame_blobs

    PV__ = np.zeros([Ly,Lx])
    T_ = []
    while True:
        frame[y,x] = T or 0  # mark scanned, failed T is excluded from T_ and PV__
        if T and gv_(val_(T.dTT*T.wTT*ttFrm) * ((T.c+wFrm)/(T.r+cFrm)) - ave):  # viable tile
            T_ += [T]; dy_dx = T.box[2:] -T.box[:2]
            pTT, pc = proj_N(T, np.hypot(*dy_dx), dy_dx, elev,T.c)  # no proj r?
            if gv_(val_(pTT*T.wTT*ttFrm) * ((pc+wFrm)/(T.r+elev+cFrm)) - ave):  # T projects beyond own box
                proj_focus(PV__,y,x,T)  # radial accum of projected vals in PV__ = frontier, ~ C._N_ in cluster_C
        pv__ = PV__.copy(); pv__[frame != None] = 0  # viable frontier = projected - scanned cells
        y,x = np.unravel_index(pv__.argmax(), pv__.shape)  # strongest frontier cell, projected from all tiles vs. last only
        if gv_(pv__[y,x] - ave):
            iy = _iy+ (y-cy)*dy_dx[0]; ix = _ix+ (x-cx)*dy_dx[1]  # pitch = last viable tile box, per-lev constant instead?
            T = frame_H(image, iy,ix, Ly,Lx, Y,X, rV, max_elev=elev)  # fill selected cell, no fb
        else: break  # no viable frontier left
    if T_:
            return T_,C,R
    return [],0,0
```

Changes:

- PV__ is now the frontier proper, analog of `C._N_`: every viable, outward-predictive tile radiates into it via proj_focus, next fill = argmax over unscanned cells, termination = no unscanned cell > ave. The main fix: a weak new tile no longer terminates expansion — it's marked scanned and skipped, the loop continues from the rest of the frontier, which the old code abandoned.

- Un-inverted the proj_N gate: its only remaining role is whether T seeds the frontier, so it should pass on strong projection. The old `ave - val` + `else: break` made strong direct projection a dead end, which only made sense as a placeholder in the last-tile flow. Flip the sign back if you actually intended uncertainty-driven scanning there.

- `frame[y,x] = T or 0`: failed recursive frame_H (None or c=0, per CN.__bool__) marks the cell scanned instead of crashing the old while-eval on `T.dTT`, and the `frame != None` mask keeps argmax off it.

- iy,ix pitch: kept your box-derived scaling but from the last *viable* tile (dy_dx persists; always set before first use, since a >ave PV cell implies a prior proj_focus). Open issue: with frontier selection the argmax cell may have been projected by a different tile than the one defining pitch, so pitch should probably be a per-level constant (window size × elev scale) rather than any tile's content box.

One structural difference from cluster_C remains deliberate: frontier is consumed best-first, one cell per iteration, not batch-per-pass — each inclusion is a frame_H call that itself updates PV__, so sequential argmax is the right equivalent. Say if you want the project's agg_recursion.py updated in place.

kwcckw: > y,x = np.unravel_index(pv__.argmax(), pv__.shape)

This is expanding via max PV, but that doesn't guarante the new location is in adjacent to the current tile? Or the max selection should be selected among the 4 or 8 adjacent tiles only?

frame_2D_alg/agg_recursion.py
kwcckw: Okay, I guess prim should be under Rt too?

frame_2D_alg/agg_recursion.py
boris-kz: they are all suppsed to be adjacent, that's how the frontier expands 
the problem with the current scheme is that it add only one tile per prior tile, it should be up to 4 new tiles

frame_2D_alg/agg_recursion.py
boris-kz: probably just +ve pLs, above the uncertainty margin

frame_2D_alg/agg_recursion.py
boris-kz: Fair — in my version the "frontier" was only the nonzero support of PV__ minus scanned cells, i.e. implicit in the map, and selection was still one global argmax per pass. Structurally all I did was stop breaking on a weak tile. Here it is as an actual structure — an explicit per-wave candidate set, batch-evaluated like `_C.N_+_C._N_` in cluster_C, recomputed per wave:

```python
def fill_frame(_iy,_ix, elev, T):  # expand level_frame from pixel-level seed tile, similar to frame_blobs

    cy,cx = (Ly-1)//2,(Lx-1)//2
    PV__ = np.zeros([Ly,Lx])
    T_ = []; Tyx_ = [(T,cy,cx)]  # eval wave, init = seed
    while Tyx_:
        for T,y,x in Tyx_:  # eval all wave members, ~ _C.N_+_C._N_ loop in cluster_C
            frame[y,x] = T or 0  # mark scanned, failed T excluded from T_ and PV__
            if T and gv_(val_(T.dTT*T.wTT*ttFrm) * ((T.c+wFrm)/(T.r+cFrm)) - ave):  # viable tile
                T_ += [T]; dy_dx = T.box[2:] -T.box[:2]
                pTT, pc = proj_N(T, np.hypot(*dy_dx), dy_dx, elev,T.c)  # no proj r?
                if gv_(val_(pTT*T.wTT*ttFrm) * ((pc+wFrm)/(T.r+elev+cFrm)) - ave):  # T projects beyond own box
                    proj_focus(PV__,y,x,T)  # radiate projected val
        front_ = np.argwhere((PV__ > ave) & (frame==None))  # frontier: unscanned cells that crossed ave this wave, ~ new C._N_
        Tyx_ = [(frame_H(image, _iy+(y-cy)*dy_dx[0], _ix+(x-cx)*dy_dx[1], Ly,Lx, Y,X, rV, max_elev=elev), y,x)
                for y,x in front_]  # fill all viable-frontier cells
    if T_:
            return T_,C,R
    return [],0,0
```

The frontier is `front_`: all unscanned cells whose accumulated projection cleared ave after this wave's radiation. Every one of them is filled in the next wave, all its viable tiles radiate, and the frontier is recomputed — candidacy by accumulated projection from all members, inclusion by post-fill viability, exactly the two-stage structure of cluster_C (rim-adjacency → comp to C). It converges to the same closure as argmax-sequential, since radiation is strictly additive (negative-V tiles radiate nothing — proj_focus breaks before accum) and scanning is permanent, but the wave form is the real difference: frame_H calls within a wave are independent, so this is parallelizable per wave, which argmax-one-at-a-time never was.

One residual gap to cluster_C, your call: inclusion here is by the new tile's standalone val, not by comp of the tile to the growing frame (cluster_C includes by base_comp to the centroid). The fuller analog would gate `T_ += [T]` on match of T to sum_vt(T_) or to its projecting neighbors — one comp_N per new tile.

Pitch caveat stands: `dy_dx` in the comprehension is the last viable tile's box, arbitrary for cells radiated by other tiles; a per-level constant would be cleaner.

kwcckw: > they are all suppsed to be adjacent, that's how the frontier expands the problem with the current scheme is that it add only one tile per prior tile, it should be up to 4 new tiles

If there's at most 4 frontier tiles, it's just the top, bottom, left and right direction tiles of existing tile in the first iteration? Then PV__ will be used to select the tiles with PV > ave. 

In subsequent iterations, we may get multiple tiles now, so new `front_` should be retrieved from all of the prior frontiers. I think we would need `_front_` and `front_`, where  `front_` should be new frontier of prior `_front_`.

boris-kz: frontier: _T_ += 4 _Ts for each scanned T, unless already scanned or outside the frame.
then we need to eval each _T:
if PV__[T.Y,T.X] > ave:

     T = frame_H(image, iy,ix, Ly,Lx, Y,X, rV, max_elev=elev)  # expand new tile to current level, no fb

kwcckw: Yes, so this extension is similar with the floodfill process where each tile is extending in 4 new directions.

boris-kz: Per-direction projection generating 4 actual prospective _Ts, in your notation:

```python
def fill_frame(_iy,_ix, elev, T, Ly,Lx, image, rV):  # expand level_frame from pixel-level seed tile, similar to frame_blobs

    cy,cx = (Ly-1)//2,(Lx-1)//2  # seed cell = center
    PV__ = np.zeros([Ly,Lx])
    T_ = []; __T_ = [(T,cy,cx)]  # eval wave, init = seed cell
    while __T_:
        _T_ = []  # next wave
        for _T,y,x in __T_:  # eval all wave members
            if frame[y,x] is not None: continue  # already scanned
            if not _T.N_:  # prospective _T, no content: form actual tile if accumulated projected value
                if gv_(PV__[y,x] - ave):
                    iy = _iy+ (y-cy)*(_T.box[2]-_T.box[0]); ix = _ix+ (x-cx)*(_T.box[3]-_T.box[1])
                    T = frame_H(image, iy,ix, Ly,Lx, Y,X, rV, max_elev=elev)  # expand new tile to current level, no fb
                else: continue  # not marked: may re-enter next waves with accumulated PV__
            else: T = _T  # pre-formed seed
            frame[y,x] = T or 0  # mark scanned, failed T excluded from T_ and PV__
            if T and gv_(val_(T.dTT*T.wTT*ttFrm) * ((T.c+wFrm)/(T.r+cFrm)) - ave):  # viable tile
                T_ += [T]; dy,dx = T.box[2:] -T.box[:2]
                for _y,_x,A in ((y-1,x,(-dy,0)), (y+1,x,(dy,0)), (y,x-1,(0,-dx)), (y,x+1,(0,dx))):  # project in 4 directions
                    if 0<=_y<Ly and 0<=_x<Lx and frame[_y,_x] is None:  # unless scanned or outside frame
                        A = np.array(A)
                        pTT,pc = proj_N(T, np.hypot(*A), A, elev,T.c)  # directional projection, dist = span along axis
                        PV__[_y,_x] += val_(pTT*T.wTT*ttFrm) * ((pc+wFrm)/(T.r+elev+cFrm))  # accum per cell, from all projecting tiles
                        _T_ += [(CN(dTT=pTT,c=pc, box=T.box+np.tile(A,2)), _y,_x)]  # 4 new prospective _Ts
        __T_ = _T_
    if T_:
            return T_,C,R
    return [],0,0
```

What changed:

- Each viable T calls proj_N once per direction, with A = axial span vector, so cos_d inside proj_N gives per-direction alignment of T's angl — this replaces both the single diagonal proj_N and proj_focus. The old projection gate's val expression became the graded accumulant: PV__[_y,_x] += that val, summing over all tiles projecting into the cell across waves; the form gate stays PV__ > ave at pop.

- Each new _T is a real prospective tile: directional pTT, pc, and parent box shifted by A — so it carries its own pitch, and later you can comp projected to actual tile after frame_H for surprise (your comp_prj_dH note).

- Seed with your (T,cy,cx) init needs the `not _T.N_` test: prospective _Ts have no content nodes, the pre-formed seed does. Without the bypass the seed hits PV__=0 at the gate and dies, or re-forms through frame_H.

Dropped the extension through failed tiles: without radial proj_focus, no value lands beyond adjacent cells, so such candidates could never clear ave. If you want scanning to survive weak gaps, either let failed-but-formed T project too (move the direction loop out of the viable branch — weak tiles project weakly, self-gating), or re-project a blocked _T one step further with decay.

kwcckw: Thanks, i will go through them later.

PR #214, opened 2026-07-23
7/23/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -891,26 +891,27 @@ def frame_H(image, iY,iX, Ly,Lx, Y,X, rV, max_elev=4, ffb=0):
...
-                y,x = np.unravel_index(pv__.argmax(), PV__.shape)
-                if gv_(PV__[y,x] - ave):
-                    iy = _iy+ (y-cy)*(T.box[2]-T.box[0]); ix = _ix+ (x-cx)*(T.box[3]-T.box[1])
-                    T = frame_H(image, iy,ix, Ly,Lx, Y,X, rV, max_elev=elev)  # expand new tile to current level, no fb
-                else: break
-            else: break
-        if T_:
-            TT,C,R = sum_vt(T_, wTT=ttFrm); R += elev
+        _T_ = [[T,cy,cx]]; T_ = []; fT_ = []
+        frame[cy,cx]=T  # only T from first main call is not Fr
```
kwcckw: In the existing code structure, the input T for the first fill_frame from `__main__` frame_H will be T from vect_edge, rather than the Fr from frame_H since we can't rerun frame_H with the same T again. This is one of the inconsistency in the current structure.

frame_2D_alg/agg_recursion.py
```diff
@@ -919,9 +920,9 @@ def fill_frame(_iy,_ix, elev, T):  # expand level_frame from pixel-level seed ti
     while T and elev < max_elev:
         tile_,C,R = fill_frame(iY,iX, elev, T)  # also prior 2D adjacents, project from seed tile
         if tile_:  # sparse, 2D higher-scope tile( oH( aH
-            Fr.Nt = sum2F(tile_); Fr.H += [Copy_(Fr.Nt)]  # min processed level
```
boris-kz: How do you get Fr.Nt for 925?

frame_2D_alg/agg_recursion.py
```diff
@@ -891,26 +891,27 @@ def frame_H(image, iY,iX, Ly,Lx, Y,X, rV, max_elev=4, ffb=0):
...
-                pv__ = PV__.copy(); pv__[frame != None] = 0
-                y,x = np.unravel_index(pv__.argmax(), PV__.shape)
-                if gv_(PV__[y,x] - ave):
-                    iy = _iy+ (y-cy)*(T.box[2]-T.box[0]); ix = _ix+ (x-cx)*(T.box[3]-T.box[1])
-                    T = frame_H(image, iy,ix, Ly,Lx, Y,X, rV, max_elev=elev)  # expand new tile to current level, no fb
-                else: break
-            else: break
-        if T_:
-            TT,C,R = sum_vt(T_, wTT=ttFrm); R += elev
+        _T_ = [[T,cy,cx]]; T_ = []; fT_ = []
```
boris-kz: `fT_ is __T_`: prior frontier?

frame_2D_alg/agg_recursion.py
kwcckw: We have that in sum2F?

    if typ==3: F.Nt.dTT = copy(TT); F.Nt.c = C; F.Nt.r = R
    
Each tile is CN, so Fr is CN `typ = 2 if fC else N.typ` , and their Nt's params are packed in the line above.

frame_2D_alg/agg_recursion.py
kwcckw: `fT_` packs all adjacent Ts from all while loops, computed from `frame_H`. We need another var here because `_T_` packs the prior frontier while `T_` packs the current frontier.

boris-kz: This can be checked in PV__: just mark all inspected locations with None?

frame_2D_alg/agg_recursion.py
boris-kz: Ok right

frame_2D_alg/agg_recursion.py
kwcckw: We need this `fT_` for the output, since we return `fT_` or  `tile_` to sum it into Fr? We use PV__ in proj_focus, and the checking is actually done with `frame[_y,_x]` now.

boris-kz: The output is T_, it should be appended from all loops. We don't need a separate T_ per loop?

kwcckw: The existing `T_` is the current frontier, so my `fT_` is actually the `T_` appended from all loops. We probably can change the name of the existing frontier `T_` and prior `_T_` to `front_` and `_front_`.

boris-kz: Ok I had "__T_"for prior frontier, "_T_" for new frontier, and "T_" for processed tiles

The damn phone can't even do proper formatting

kwcckw: > Ok I had `__T_` for prior frontier, `_T_` for new frontier, and `T_` for processed tiles
> 

Yes, `__T_` for prior frontier, `_T_` for new frontier, and `T_` for processed tiles, but i think it's clearer to have separate name for those frontiers.

boris-kz: I don't think it's clearer at all

frame_2D_alg/agg_recursion.py
```diff
@@ -891,26 +891,27 @@ def frame_H(image, iY,iX, Ly,Lx, Y,X, rV, max_elev=4, ffb=0):
...
+        while _T_:
+            for T,y,x in _T_:
+                if gv_(val_(T.dTT*T.wTT*ttFrm) * ((T.c+wFrm)/(T.r+cFrm)) - ave):
+                    dy,dx = T.box[2:] -T.box[:2]
+                    pTT, pc = proj_N(T, np.hypot(dy,dx), np.array([dy,dx]), elev,T.c)  # no proj r?
+                    if gv_(ave - val_(pTT*T.wTT*ttFrm) * ((pc+wFrm)/(T.r+elev+cFrm))):  # inverted val                    
+                        proj_focus(PV__,y,x,T)
+                        for _y,_x,A in ((y-1,x,(-dy,0)),(y+1,x,(dy,0)),(y,x-1,(0,-dx)),(y,x+1,(0,dx))):  # project in 4 directions
+                            if not (0<=_y<(Y-Ly) and 0<=_x<(X-Lx)) or frame[_y,_x] is not None: continue  # outside frame or checked
+                            if PV__[_y,_x]>ave:  # do we need gv_ here for each direction?
```
boris-kz: not sure, let's keep it for all directions for now

frame_2D_alg/agg_recursion.py
boris-kz: Kept your structure and the inverted gate; fixes, traced:

```python
def fill_frame(_iy,_ix, elev, T, Ly,Lx, image, rV):  # expand level_frame from pixel-level seed tile, similar to frame_blobs

    frame = np.full((Ly,Lx), None, dtype=object)  # level scope: Ly,Lx grid of Ly,Lx-pixel windows
    cy,cx = (Ly-1)//2,(Lx-1)//2  # seed cell = center
    PV__ = np.zeros([Ly,Lx])
    frame[cy,cx] = T  # mark seed cell, else adjacent tiles may re-form it
    T_ = []; __T_ = [(T,cy,cx)]  # output; eval wave: __T_ consumed, _T_ built, per cluster_N
    while __T_:
        _T_ = []  # next wave
        for T,y,x in __T_:  # eval all wave members
            if gv_(val_(T.dTT*T.wTT*ttFrm) * ((T.c+wFrm)/(T.r+cFrm)) - ave):  # viable tile, else dropped
                T_ += [T]
                dy,dx = T.box[2:] -T.box[:2]
                if gv_(ave - val_(pTT*T.wTT*ttFrm) * ((pc+wFrm)/(T.r+elev+cFrm))):  # inverted val: scan only if surround not predictable from T
                    proj_focus(PV__,y,x,T)  # radiate projected value
                    for _y,_x in ((y-1,x),(y+1,x),(y,x-1),(y,x+1)):  # fill 4 adjacent cells
                        if not (0<=_y<Ly and 0<=_x<Lx) or frame[_y,_x] is not None: continue  # outside grid or checked
                        if gv_(PV__[_y,_x] - ave):  # accumulated from all adjacent tiles, rechecked per new neighbor
                            iy = _iy+ (_y-cy)*dy; ix = _ix+ (_x-cx)*dx  # grid-to-image coords, pitch = parent span
                            if 0<=iy<(Y-Ly) and 0<=ix<(X-Lx):  # window inside image
                                frame[_y,_x] = _T or 0  # mark checked, failed _T excluded from waves and T_
                                if _T: _T_ += [(_T,_y,_x)]  # eval in next wave
        __T_ = _T_
    if T_:
            return T_,C,R
    return [],0,0
```

- Coordinates: _y,_x are grid indices — frame and PV__ are Ly×Lx arrays — but frame_H needs image coords. Your version passed grid coords to frame_H and bounded grid indices by Y-Ly, which would also crash frame[_y,_x] once coords exceed the grid. Restored the grid-to-image mapping and split the bounds: grid check for the arrays, your Y-Ly/X-Lx check on iy,ix. That image-bounds check was genuinely missing in all prior versions — tiles could walk off the image.

- List roles: your fragment appends formed-unvetted tiles to __T_ and clears T_ every wave, so the final sum would count nonviable tiles or an empty list. Restored: T_ = viable output, appended at pop inside the viability gate; __T_/_T_ = consumed/built wave, cluster_N naming. A formed tile is vetted once, when popped; nonviable pops are dropped, already marked.

- `[[T,_x,_y]]` transposes coords vs the `for T,y,x` unpack; and `T = frame_H(...)` inside `for T,y,x` shadows the parent mid-loop — renamed to _T.

- Dropped A from the direction tuples — unused, since proj_focus radiates radially as your accumulator; the direction loop only checks and fills the 4 cells.

Net semantics as you have it: expansion proceeds only through viable tiles whose direct projection is uncertain (inverted gate), into cells whose accumulated projected value clears ave — value high, not yet confirmed. PV-failed cells stay unmarked and get rechecked as each new neighbor radiates.

kwcckw: > T_ += [T]
 
I think this should be packing `_T` from frame_H only? Because the first T is actually the T from vect_ edge, but not the Fr from nested frame_H.

> Coordinates: y,x are grid indices 

Yes thanks, I overlooked this part. So I think we only need one check on the grid only, the coordinates check above should be redundant.

boris-kz: right, _T

frame_2D_alg/agg_recursion.py
```diff
@@ -508,7 +508,7 @@ def sum2F(N_, root=None, m_=[],d_=[], merge=0, froot=0, nF=None):  # -> CF/CL/CC
                 kern=n.kern*w; span=n.span*w; yx=n.yx*w; angl = copy(n.angl[0]) if n.angl is not None else None
```
boris-kz, line 551: on cross_comp:

**1. The fB path is dead at the gate.** `e.w` is only assigned in the else branch; boundary links keep default w → `sum(e.w for e in e_)` = 0 → gate always False. Boundary weight should be d-based anyway — you already have the precedent in cluster_C's `n.exe = (n.d if n.typ==1 else n.m)`. Same typ-conditional belongs here (`e.w = e.d if fB`) and in `get_exemplars`, which ranks by `Rt.m` — m-biased selection over contrast links picks the wrong seeds.

**2. Gate is still truthiness.** No `> ave` — carried over from before; any nonzero product passes.

**3. proj_L_'s output contract breaks both consumers.** `comp_C_` takes a node list (`combinations(C_,r=2)` internally) — feeding it pairs makes pairs-of-pairs. And `comp_N_` still does its own selection (sort, proj_V, prelink creation) — it would re-select the pre-selected. The clean version: proj_L_ returns typ=-1 prelinks carrying pTT (needed later for `Link.rTT`), comp_N_ iterates them, comps the gated ones, returns the rest as negatives; comp_C_ takes the same input.

**4. fC input should mirror fB.** If fC means centroid merging, `e_ = Ft.N_` (the Cs themselves), same as fB — the member-unpack + w path is only the default agg+. Currently fC shares the member path, so comp_C_'s merge would fuse member nodes as pseudo-centroids. (And still no caller passes fC=1.)

**5. Downstream is still on the old contract — this is where fB actually breaks.** Note the fB path is the first place the pool matches (`cluster_N(root,...)` with `root.N_ = e_` ✓), but: the reform block `rG = Ft.root` overwrites `G.Nt` with boundary clusters, and the Ct splice injects boundary sub-centroids into the core's `G.Ct` — both mislevels. Also `F2N(root)` runs *after* cluster_N, which already assumes CN attrs on root. For the default path the pool mismatch remains (`root.N_` ≠ e_) — set `root.Nt.N_ = e_` or pass the fork.

**6. Rim maintenance kills fB clustering.** `comp_N` appends positive links to rim only for `typ==3`; F2N'd links keep typ=1 → after comp, boundary nodes have only prelinks in rim → cluster_N's flood-fill (`__L_ = N.rim`) has nothing to traverse. Either F2N sets typ=3 or the rim-append condition loosens.

One reconciliation with last turn's xcomp argument: for fB, `combinations(e_,2)` + distance sort is redundant work — boundary adjacency is structural (shared endpoints). Let proj_L_ specialize pair *generation* per flag: endpoint-shared pairs for fB, distance/proj-selected combinations otherwise. Then the unified cross_comp keeps your gate discipline, fB stays cheap (O(n) pairs, and the gates stop recursion early on weak contours), and xcomp becomes unnecessary for Bt.

boris-kz: So with the current scheme we never need to cross_comp centroids, it's the highest-m_ medoids instead?

boris-kz: Claude:
Here are all edits as a diff — green lines added, red removed:

```diff
--- agg_recursion.py (yours)
+++ agg_recursion.py (edited)
@@ cross_comp @@
-def cross_comp(Ft, R, root, fC=0, fB=0):  # calls cluster_N, sub+, over exemplars spliced from C_
+def cross_comp(Ft, R, root, fB=0):  # over Ct: representative members, or Bt: contour links
+    # no C-to-C comp: co-located Cs merge by member overlap in cluster_P; trans-local C merge would be comp_C_(fall=0), feature-space
     G_ = []
-    if fB: e_=Ft.N_  # Bt
+    if fB:
+        e_ = Ft.N_  # Bt links, contiguity-paired in proj_L_
+        for e in e_: e.w = e.d  # contrast value
     else:  # selective for Ct only:
         e_ = {n for C in Ft.N_ for n in C.N_}
         for e in e_: e.w = sum(e.m_)  # combine memberships
-        e_ = [e for e in e_ if e.w]
-    if sum(e.w for e in e_) * ((Ft.c+wAgg) / (Ft.r+R+cAgg)) * ((len(e_)-1) **2 * wL):
+    e_ = [e for e in e_ if e.w]
+    if gv_(sum(e.w for e in e_) * ((Ft.c+wAgg) / (Ft.r+R+cAgg)) * ((len(e_)-1) **2 * wL) - ave):
         # select by dist and marginal proj m:
-        if pL_ := proj_L_(combinations(e_,2), root, R):
-            if Lt := comp_C_(pL_,R,fC=1) if fC else comp_N_(pL_, R):
+        if pL_ := proj_L_(combinations(e_,2), root, R, fB=fB):
+            if Lt := comp_N_(pL_, R):
                 L_,TT,c,r,cV = Lt
                 oF_[CoF.get().nF].V_ += [cV]  # combined comp_ results
                 root.L_ = L_  # val=m+d /clust, m/comp
                 if gv_(val_(TT*ttcN) * ((c+wcN)/(r+ccN)) * ((len(L_)-1)*wL) - ave):
+                    if not root.typ: F2N(root)  # promote at 1st sub+ or agg+, before cluster_N
+                    Nt = root.Nt; Nt.root = root; Nt.N_ = list(e_)  # pool fork for flood-fill
                     E_ = get_exemplars({N for L in L_ for N in L.N_}, r,c)
-                    G_,r = cluster_N(root, E_,r,c)  # -> sum2G
+                    G_,r = cluster_N(Nt, E_,r,c)  # -> sum2G
                     if G_:
-                        if not root.typ: F2N(root)  # promote at 1st sub+ or agg+
+                        root.Nt = sum2F(G_,root,froot=2)  #| C_? before H append: CN.H lives in Nt
                         root.H += [sum2F(L_,root,froot=1)]  # dLev per L_
-                        root.Nt = sum2F(G_,root,froot=2)  #| C_?
                         if Ct := root.Ct:
                             cross_comp(Ct, r, root) # sub+'agg+

@@ proj_L_ @@
-def proj_L_(pairs, root, r, max=20, fall=1):
+def proj_L_(pairs, root, r, max=20, fall=1, fB=0):
     ...
         if _N.sub != N.sub: continue  # comp x agg Lev?
         if N is _N: olp_ += [N]  # overlap = unit match, no miss
+        elif fB and not set(_N.N_) & set(N.N_): continue  # contour contiguity: shared endpoints only
         else:
     ...
                     pL = [dist, dy_dx, _N, N, lc, lr, pTT, m,d]
+                    for n in _N,N:
+                        if not hasattr(n,'prim'): n.prim = []
                     for l_ in pL_,_N.prim,N.prim: l_+=[pL]

@@ comp_N_ @@
         for N in set(N_):
             if N.rim: N.Rt = sum2F(N.rim)
-        cV = FV_(CoF.get(), *sum_vt(L_))  # +-ve Ls for oF, no oF.N_?
-        return L_,cV  # +ve only
+        TT,C,R = sum_vt(L_)
+        cV = FV_(CoF.get(), TT,C,R)  # +ve Ls for oF, no oF.N_?
+        return L_,TT,C,R, cV  # 5-tuple, unpacked in cross_comp | comp_F

@@ comp_N @@
-    if N.typ==3:
-        for n, _n in (_N,N),(N,_N):
-            n.rim += [L]; n.compared.add(_n)  # or unique comps?
+    for n, _n in (_N,N),(N,_N):
+        if isinstance(n, CN):  # incl F2N-converted links: rims for flood-fill in cluster_N
+            n.rim += [L]
+            if n.typ==3: n.compared.add(_n)  # or unique comps?

@@ comp_F @@
-                else: L_,TT,C,R,_ = comp_N_(Np_,r,nF,rL)
+                else:
+                    Lt = comp_N_( proj_L_(Np_,rL,r), r,nF,rL)  # select pairs first
+                    L_,TT,C,R = Lt[:4] if Lt else ([],np.zeros((2,9)),0,0)

@@ get_exemplars @@
-    N_= sorted(N_, key=lambda n: n.Rt.m*n.c, reverse=True); E_,Inh_ = set(),set()
+    N_= sorted(N_, key=lambda n: (n.d if n.typ==1 else n.Rt.m)*n.c, reverse=True); E_,Inh_ = set(),set()  # d-select contour links
     for rdn, N in enumerate(N_, start=1):  # strong-first
+        v = N.d if N.typ==1 else N.Rt.m
         inh_ = list(Inh_ & set(N.rim))  # stronger Es in N.rim
         oM = sum_vt(inh_,fm=1, wTT=ttE)[0] if inh_ else 0
-        oV = oM / (N.Rt.m or eps)  # relative olp V
-        if N.Rt.m * N.c * wE > ave* (_r+rdn+cE+oV):
+        oV = oM / (v or eps)  # relative olp V
+        if v * N.c * wE > ave* (_r+rdn+cE+oV):

@@ cluster_N @@
     for N in _N_:
-        N.fin =0; N.exe=1; sum2F(N.rim,N.Rt)  # only if N was added in trans-cluster?
+        N.fin =0; N.exe=1
+        if N.rim: N.Rt = sum2F(N.rim)  # replace, not accumulate: rim may get new Ls after last Rt update
     ...
-                G_ += [sum2G(ft_,ttcN, CN())]; Gt_+= [[tt,c,r]]  # eval sub+
+                G_ += [sum2G(ft_,ttcN, Ft.root)]; Gt_+= [[tt,c,r]]  # eval sub+; real up-chain: frame test, brrw, feedback

@@ cluster_C @@
-        if gv_(val_(DTT,Ft.root.wTT*ttcC) * (oc/(r+ccC)) * ((len(out_)-1)*wL) * wcC - ave):
-            Ct = sum2F(out_); Ft.root.Ct = Ct; Ct.root = Ft.root
+        rt = Ft if Ft.typ else Ft.root  # CN from frame_H | fork owner from sum2G
+        if gv_(val_(DTT, rt.wTT*ttcC) * (oc/(r+ccC)) * ((len(out_)-1)*wL) * wcC - ave):
+            Ct = sum2F(out_); rt.Ct = Ct; Ct.root = rt

@@ sum2G @@
-        if N.typ!=1 and bd*(wAgg*L) > avd*(br+cAgg*L): [F2N(L) for L in Bt.N_]
-        cross_comp(Bt, br, Bt)  # no ddfork; or xcomp only?
+        if N.typ!=1 and bd*(wAgg*L) > avd*(br+cAgg*L):  # strong contour only
+            [F2N(l) for l in Bt.N_]
+            cross_comp(Bt, br, Bt, fB=1)  # dd contour, contiguity-paired in proj_L_, no ddfork

@@ frame_H @@
-    elev, Fr = 0,[]
+    elev, Fr = 0, CN()
     ...
             if Ct := Fr.Ct:  # agg+ exemplars
                 cross_comp(Ct,R,Fr)
-                if elev and ffb:  # ffb =1 in main, no ffeedback in added tiles
+                if ffb:  # ffb =1 in main only: no ffeedback | elev expansion in added tiles
                     Fr,aTT,oTT,aH,oH = ffeedback(Fr, aTT,oTT,aH,oH)  # term,form oH(aH
                     elev +=1; T=Fr  # next-extension seed
                 else: break
             else: break
-        if Fr: FV_(CoF.get(), Fr.dTT, Fr.c, Fr.r)
-        return Fr  # intra-lev feedback
+        else: break  # empty tile_
+    if Fr: FV_(CoF.get(), Fr.dTT, Fr.c, Fr.r)
+    return Fr  # intra-lev feedback
```

The full file with these applied is already in the project doc; the diff above is every change, nothing else touched.

kwcckw: > So with the current scheme we never need to cross_comp centroids, it's the highest-m_ medoids instead?

But without comp in Cs, the seed in Cs will be selected based on their original m, and there's only the refinement from `dtt,_ = base_comp(_C,n) `?

Or C fork becomes a terminal fork now?

boris-kz: Cluster_C is now only to select the most representative Ns (medoids) for extended cross_comp.
I think that increases composition implicitly, through these N.root_?

boris-kz: Yes, but with a precise qualification on where the composition actually accrues. Trace the chain:

**What root_ does add.** After the reduction, cluster_C's product is the membership arrays, and those ride the explicit hierarchy for free: add_Nt splices `N.root_` → `G.C_` → G's Ct fork, so every composition level formed by cluster_N automatically inherits its type layer, re-indexed per rank, with zero comp cost. And C-to-C comparison didn't fully disappear — it survives exactly where it belongs: `comp_F(_N.Ct, N.Ct)` in comp_N's typ==3 fork spec. So type similarity contributes to G-to-G match, which feeds the next cluster_N. That's the implicit composition increase: root_ → per-G Ct → Ct term in higher links → stronger/wider next-level Gs. Overlap in root_ also preserves the un-comped C-to-C relation (shared-member mass), so nothing was lost by dropping direct centroid cross_comp.

**What it doesn't add.** Composition *depth* is still only explicit — cluster_N/sum2G wrapping. The type layer itself stays flat: no C-of-C, `w = Σm_` is rank-local by the reset discipline, and cross-rank generality exists only through shared members and the comp_F term. Selection raises composition *scale* rather than depth: representative members pass the distance/proj gates that average nodes fail, so links are longer and Gs wider per level. If you later want types-of-types, that's the comp_C_(fall=0) feature-space pass over Ct.N_ — taxonomy, orthogonal to composition.

Two consequences of the reduced role. First, `w = Σm_` is now load-bearing, which makes the cluster_P leftover a real bug rather than a nit: it resets only `root_` before the final out_ pass, so `m_/d_` keep last-cycle entries — w double-counts and the `zip(N.root_, N.m_, N.d_)` used at next cluster_C init misaligns. Fixed:```diff
@@ cluster_P @@
-    for N in N_: N.root_ = []  # replace with out_ Cs:
+    for N in N_: N.root_, N.m_, N.d_ = [],[],[]  # replace with out_ Cs: align with root_ for zip in cluster_C, w = sum(m_)
```

Second consequence: you now have two serial selections with different criteria — w = Σm_ (typicality, non-local) gates entry to cross_comp, then get_exemplars re-selects from linked nodes by Rt.m (local salience) to seed cluster_N. They're complementary, not redundant — but if you want one mechanism, fold w into the exemplar key (`n.w * n.Rt.m * n.c`) and let get_exemplars be the single selector. Relatedly, since membership precision now only matters near the selection threshold, cluster_P's full-fuzzy refine earns less per cycle than its cost model assumes — its entry gate (`(mat+dif)*(wcP*L) > Ave*(r+olp+ccP*L)`) could be tightened, or the refine skipped entirely below some pool size.

One terminological note: what you select are medoid-like *members* (high-w nodes near the C mean); the Cs themselves remain means. That distinction matters only if you ever use the C as a comparand — means blur multi-modal clusters where a medoid wouldn't — but as pure selection weight, Σm_ is indifferent to it.

kwcckw: > I think that increases composition implicitly, through these N.root_?

Yes, those root_ , m_ and d_ would be good to trace each soft membership, but we don't use N.root_ now anyway? 

And the most representative Ns will be the N with highest m in each C.N_?

boris-kz: I think it's the highest sum(N.m_), accross all of ther roots.
But I am still figuring it out
So I pushed partial edits, down comp_N_

kwcckw: > I think it's the highest sum(N.m_), accross all of ther roots.

Okay right, this is actually better, but i think we probably need the highest average m across all roots (Ns might have different number of clusters)? Also we probably can use d_ to select highest m and lowest d across all roots?

boris-kz: D is not meaningful in centroid clustering, and I see no point in averaging.
But we do get explicit agg+ from cluster_N, just pruned out in sub+. So the value of these clusters should be summed from both CC and CN: m_ and L_.

PR #215, opened 2026-07-24
7/24/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -895,36 +897,39 @@ def trace_edge(N_,_G_,_TT,_C, r,root):  # cluster contiguous shapes via PPs in e
...
-                            iy = _iy+ (_y-cy)*dy; ix = _ix+ (_x-cx)*dx  # frame-to-image coords, pitch = parent span
-                            if 0 <= iy < (Y-Ly) and 0 <= ix < (X-Lx):  # tile is inside image
+def frame_H(image, iY,iX, Ly,Lx, Y,X, rV, max_elev=4, ffb=0):
+
+    # move here? Else we need another oF_ for the gv_ in fill_frame
+    def fill_frame(_iy,_ix, elev, T, Ly,Lx, image, rV):  # expand level_frame from pixel-level seed tile, similar to frame_blobs
+    
+        nX, nY = Y//Ly,X//Lx  # total number of x and y in the tile grid
+        frame = np.full((nY,nX), None, dtype=object)  # level scope: Ly,Lx grid of Ly,Lx-pixel windows
+        cy,cx = int(_iy/Ly), int(_ix/Lx)  # seed should be based on input
```
kwcckw, line 907: I think there's some mistake in the prior code. Lx and Ly are the tile's x and y size (64 x 64), so we need to compute another nX and nY in order to get the tile grid size.

frame_2D_alg/agg_recursion.py
```diff
@@ -846,7 +847,8 @@ def PP2N(PP):
     if G_:
         FV_(CoF.get(), TT, C,1); L_,lTT, lc, lr = [],np.zeros((2,9)), 0, 1
         for G in G_: L_ += G.Lt.N_; lTT += G.Lt.dTT; lc += G.Lt.c; lr += G.Lt.r
-        return sum2G([[G_,'Nt',TT,C,1],[L_,'Lt',lTT,lc,lr]], TT+lTT, T) # c,R?
+        # actually L_ should be empty here? We don't have Ls of G_ yet
+        return sum2G([[[N for G in G_ for N in G.N_],'Nt',TT,C,1],[L_,'Lt',lTT,lc,lr]], TT+lTT, T) # c,R?
```
kwcckw, line 851: For vect_edge, we actually don't have L_ of G_ yet, so Lt here should be empty for tile?

frame_2D_alg/agg_recursion.py
boris-kz, line 86: Sorry, I was wrong yesterday, it still makes sense to cross-comp centroids because they are location constrained. So we need to revert a bunch of changes:

Yes — and the location constraint is what makes it valid, not just permissible:

1. C-to-C cross_comp needs ext params: dist, angle, proj in proj_L_/comp_N all presuppose meaningful yx, span, angl, kern. Weighted-sum ext params are only meaningful when members are co-located. Feature-space centroids would have undefined ext — there only medoid comp works. So constrained Cs are legitimate spatial nodes; comp them.

2. It's also the compressive choice: C is the denoised, higher-c comparand you already paid to form. Comping the medoid instead discards the summation and comps a noisy sample of it. Comp-summaries-not-samples is your general pattern at every other level (Ps→PPs→Gs); Cs shouldn't be the exception.

3. Trade-off to be aware of: CC is structurally flat — no H, no fork tree — so comp_N on a C pair either stays at base_comp depth or descends into the member product (215-216), which reintroduces member-level cost per C pair. If agg+ needs deep-structure derivatives, Cs need summed member H (sum2F currently builds H only for typ 3), or you accept the product branch.

4. This contradicts current code, which comps members, not Cs: e_ at 93 unpacks C.N_, and the 87 comment says "no C-to-C comp: co-located Cs merge by member overlap in cluster_P." That reason addresses merging — overlap-merge removes the need to comp Cs *for dedup*, not for forming agg+ derivatives between distinct co-level Cs. Different purposes; the comment conflates them.

5. One correction needed if you switch: fuzzy Cs share members, so C-to-C match is partially self-match. comp_C_ handles the `_C is C` case as unit overlap (171) but not partial sharing — discount by shared-member c fraction, into r or subtracted from m.

frame_2D_alg/agg_recursion.py
boris-kz: right now tile Lt is not used anyway, tile projection is not directional

frame_2D_alg/agg_recursion.py
```diff
@@ -942,8 +947,8 @@ def frame_H(image, iY,iX, Ly,Lx, Y,X, rV, max_elev=4, ffb=0):
                     elev +=1; T=Fr  # next-extension seed
                 else: break
             else: break
-        if Fr: FV_(CoF.get(), Fr.dTT, Fr.c, Fr.r)
-        return Fr  # intra-lev feedback
+    if Fr: FV_(CoF.get(), Fr.dTT, Fr.c, Fr.r)  # the indentation is wrong here? This should be after while loop
```
boris-kz, line 950: right

frame_2D_alg/agg_recursion.py
kwcckw: > it still makes sense to cross-comp centroids because they are location constrained

But location is only part of the positional params, rather than the core params?

frame_2D_alg/agg_recursion.py
kwcckw: Okay, so we not really need the L_ and their params here.

frame_2D_alg/agg_recursion.py
boris-kz: It doesn't matter, the point is that our centroids don't significantly overlap between G and frames.
If they did, similar C would already be merged

kwcckw: I see, and we don't really need N.w now? Unless we use it in the subsequent sub+ eval.

boris-kz: yes, right now it's replaced by N.m_. or N.root_ = [[C,w]...]?

kwcckw: I think we can remain the existing N.m_ or N.root_? Since it can be computed using N.root_'s m_ whenever we need to use it.

frame_2D_alg/agg_recursion.py
boris-kz: with a fixed grid size that we have now, is Lx | Ly ** elev

kwcckw: I see, with fixed grid size as Lx and Ly, we might get different tile size when the input image size is different.

frame_2D_alg/agg_recursion.py
boris-kz: N.root_ = [[C,m]...] is just clearer

frame_2D_alg/agg_recursion.py
kwcckw: But `cy,cx = (Ly-1)//2,(Lx-1)//2  # seed cell = center` always init from center, and `_iy` and `ix_` are no longer useful? So `cy` and `cx` should be computed from `_iy` and `_ix` instead?

boris-kz: right

frame_2D_alg/agg_recursion.py
boris-kz: Actually, N.root_ = [[C,m,d]...], replacing root_,m_,d_ in CC, but I think it should be in CF, all nodes can be clustered? Then we don't need class CC

kwcckw: I think we only need root_ in CN? We don't cluster CF anyway. Without CC, I guess we can use root_ to identify CC and CN. For CN clustered from CC, we still remain their m_ and d_?

boris-kz: root_ is for CC elements, not CCs. so it's probably optional param anyway?

frame_2D_alg/agg_recursion.py
```diff
@@ -895,36 +897,39 @@ def trace_edge(N_,_G_,_TT,_C, r,root):  # cluster contiguous shapes via PPs in e
...
-                if gv_(ave - val_(pTT*T.wTT*ttFrm) * ((pc+wFrm)/(T.r+elev+cFrm))):  # inverted val: scan only if surround not predictable from T
-                    proj_focus(PV__,y,x,T)  # radiate projected value
-                    for _y,_x in ((y-1,x), (y+1,x), (y,x-1), (y,x+1)):  # fill 4 adjacent cells
-                        if not (0<=_y<Ly and 0<=_x<Lx) or frame[_y,_x] is not None: continue  # outside frame or checked
-                        if gv_(PV__[_y,_x] - ave):  # accumulated from all adjacent tiles
-                            iy = _iy+ (_y-cy)*dy; ix = _ix+ (_x-cx)*dx  # frame-to-image coords, pitch = parent span
-                            if 0 <= iy < (Y-Ly) and 0 <= ix < (X-Lx):  # tile is inside image
+def frame_H(image, iY,iX, Ly,Lx, Y,X, rV, max_elev=4, ffb=0):
+
+    # move here? Else we need another oF_ for the gv_ in fill_frame
```
boris-kz, line 902: ok right

frame_2D_alg/agg_recursion.py
kwcckw: It's optional for CC, but when we sum CC into graphs, those graphs should preserve the root_?

frame_2D_alg/agg_recursion.py
```diff
@@ -551,18 +552,17 @@ def sum2F(N_, root=None, m_=[],d_=[], merge=0, froot=0, nF=None):  # -> CF/CL/CC
                 if typ==3: box=copy(n.box)
     F = (cls_[typ])(dTT=TT, c=C, r=R, nF=nF)
     if typ==3: F.Nt.dTT = copy(TT); F.Nt.c = C; F.Nt.r = R
+    # this should be depends on input m_ and d_?
```
boris-kz, line 555: not sure about this, lets leave it out for now

frame_2D_alg/agg_recursion.py
boris-kz: I don't see any attrs we need specifically fo CC, so this class is probably not needed, lets use CL instead

kwcckw: Okay, the it should be same as before where we still need to convert it into CN using F2N when we need to cluster CC.

frame_2D_alg/agg_recursion.py
```diff
@@ -616,6 +616,7 @@ def sum2G(ft_, fTT, root=None, init=1):  # core clustering function
             c = G.Lt.c; E_ = get_exemplars({N for link in L_ for N in link.N_}, r,c)
             if gv_(Vn * (wcC-wcN) * (mdecay(L_)-decay) - Av * (lr+1+(ccC-ccN)*L)):
                 r+=1; G_,r = cluster_C(G.Nt,E_,r,c)  # low decay: CC nodes
+                for n in G.Nt.N_: n.w += sum(l.m for l in n.L_)  # sum from L_
```
boris-kz, line 619: N.dTT already includes L.dTTs, so val_ already gets it

frame_2D_alg/agg_recursion.py
boris-kz: I didn't edit that yet.

frame_2D_alg/agg_recursion.py
boris-kz: we may actually need it for more precise projection, then L_ should be dPPs here

kwcckw: Alright, then we need some changes in sum2G since L's nt may not always packing the input N_ of sum2G now.

frame_2D_alg/agg_recursion.py
kwcckw: Alright, i will revert it later.

PR #216, opened 2026-07-25
7/25/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -574,13 +578,14 @@ def sum2G(ft_, fTT, root=None, init=1):  # core clustering function
     if G.Lt:  # sub+
         Lt = G.Lt; L_,lm,ld,lr = Lt.N_,Lt.m,Lt.d,Lt.r; L=len(L_)-1; Av = ave+avd
         if gv_(Vn := (lm+ld)*wcN - Av* (lr+1+ccN*L)):  # for cluster_N
-            c = G.Lt.c; E_ = get_exemplars({N for link in L_ for N in link.N_}, r,c)
+            c = G.Lt.c; E_ = get_exemplars(N_, r,c)  # why not using N_?
```
kwcckw, line 581: Actually why we use link_'s nt to get exemplars instead of N_ here? I think N_ should be good here?

boris-kz: right, thanks

frame_2D_alg/agg_recursion.py
```diff
@@ -889,7 +896,7 @@ def fill_frame(_iy,_ix, elev, T, Ly,Lx):  # expand level_frame from pixel-level
 
     global ave,avd; aTT=oTT=np.zeros((2,9)); aH,oH = [],[]  # regime refs across levs / ffeedback
     elev, Fr = 0,[]
-    T = vect_edge( frame_blobs_root( comp_pixel( image[iY:iY+Ly, iX:iX+Lx]), rV), iY,iX,Ly,Lx, rV)
```
kwcckw, line 892: I think previously Lx and Ly are meant for image size? But since it's the fixed tile grid size, we need to change the Ly and Lx to the tile size here, which is total pixel / total tile.

boris-kz: for image it's the same scaling as in fill_frame: `Ly**elev, Lx**elev`

kwcckw: Sorry, i'm a bit confused. So this means Lx and Ly are actually the tile image size, where it should scales according to elev?

boris-kz: Ly, Lx is fractal dimensions: 64,64, then Ly,Lx ** elev to get pixel-level dimensions

    def fill_frame(_iy,_ix, elev, T):  # expand level_frame from pixel-level seed tile, similar to frame_blobs

        frame = np.full((Ly,Lx),None, dtype=object)  # higer tile Ly,Lx = lower tile Ly,Lx **2
        cy,cx = int(_iy/Ly), int(_ix/Lx)  # seed is the center of frame
        PV__ = np.zeros([Ly,Lx])  # projected value map
        frame[cy,cx] = T; T_=[]; __T_=[(T,cy,cx)]  # output, eval wave
        while __T_:
            _T_ = []  # next wave
                    dy,dx = T.box[2:] -T.box[:2]
                                iy = _y**elev; ix = _x**elev
                                if _T := frame_H(image, iy,ix, Ly,Lx, Y,X, rV, max_elev=elev):  # expand new tile to current level, no fb
                                    T_ += [_T]; _T_ += [(_T,_y,_x)]; frame[_y,_x] = _T
                                else: frame[_y,_x] = 0
            __T_ = _T_
        if T_:
                return T_,C,R
        return [],0,0

kwcckw: > iy = _y**elev; ix = _x**elev

So this is shifting the starting grid x and y of the tile, but when we have fixed Ly, Lx , X and Y, the pixel dimension or the resolution will be the same?

boris-kz: Y,X are always at pixel resolution, y,x are new tile coords in a frame, at 64 x 64 resolution

kwcckw: Okay, so in this case, Lx and Ly are grid size, and tile size should be increased based on elev to get a larger scaling at higher elev? At higher elev, when we have higher tile image size but constant tile grid size, there will be overlapping between tiles?

boris-kz: No overlapping, higher tile is 64x64 lower tiles

boris-kz: for clarity, I will make Ly,Lx global and always pass elev to scale them for image mapping:

def frame_H(image, iY,iX, Y,X, rV, elev=1, max_elev=4, ffb=0):

    def fill_frame(_iy,_ix, elev, T):  # expand level_frame from pixel-level seed tile, similar to frame_blobs

        frame = np.full((Ly,Lx),None, dtype=object)  # higer tile Ly,Lx = lower tile Ly,Lx **2
        PV__ = np.zeros([Ly,Lx])  # projected value map
        T_ = []; frame[31,31] = T; __T_=[(T,31,31)]  # output, eval wave starts at higher tile center
        while __T_:
            _T_ = []  # next wave
                    dy,dx = T.box[2:] -T.box[:2]
                                iy = _y**elev; ix = _x**elev
                                if _T := frame_H(image, iy,ix, Y,X, rV, elev, max_elev=elev):  # expand new tile to current level, no fb
                                    T_ += [_T]; _T_ += [(_T,_y,_x)]; frame[_y,_x] = _T
                                else: frame[_y,_x] = 0
            __T_ = _T_
        if T_:
                return T_,C,R
        return [],0,0

    Fr = []
    while T and elev < max_elev:
        if tile_:
            Fr = sum2F(tile_); Fr.H += [Copy_(Fr.Nt)]  # minimally processed level
            Fr.N_ = [g for t in tile_ for g in t.N_]  # concat edge Gs
            cluster_C(Fr.Nt, get_exemplars(Fr.N_,R,C), R,C)
            if Ct := Fr.Ct:
                cross_comp(Ct, R, Fr)  # agg+ over CC-pruned exemplars
                    elev += 1; T=Fr  # next-extension seed
                else: break
            else: break
    if Fr: FV_(CoF.get(), Fr.dTT, Fr.c, Fr.r)

boris-kz: Fable:
Edited `frame_H` only; ffeedback/proj_focus/proj_N internals untouched.

```python
def frame_H(image, iY,iX, Y,X, rV, elev=1, max_elev=4, ffb=0):

    def fill_frame(_iy,_ix, elev, T):  # expand level frame from seed tile, similar to frame_blobs

        S = Ly**elev  # cell span in pixels, frame span = S*Ly
        frame = np.full((Ly,Lx),None, dtype=object)
        cy,cx = Ly//2,Lx//2; y0,x0 = _iy-cy*S, _ix-cx*S  # seed is the center of frame, y0,x0: pixel anchor
        PV__ = np.zeros([Ly,Lx])  # projected value map
        frame[cy,cx] = T; T_=[T]; __T_=[(T,cy,cx)]  # output, eval wave
        while __T_:
            _T_ = []  # next wave
                if gv_(val_(T.dTT, T.wTT*ttFrm) * ((T.c+wFrm)/(T.r+cFrm)) - ave):
                    dy,dx = T.box[2:]-T.box[:2]
                    pTT, pc = proj_N(T, np.hypot(dy,dx), np.array([dy,dx]), elev,T.c)  # aggregate proj, make directional per cell?
                    if gv_(val_(pTT, T.wTT*ttFrm) * ((pc+wFrm)/(T.r+elev+cFrm)) - ave):  # +ve, no uncertainty projection yet
                        proj_focus(PV__,y,x,T)  # radiate projected value
                        for _y,_x in ((y-1,x),(y+1,x),(y,x-1),(y,x+1)):  # fill 4 adjacent cells
                            iy,ix = y0+_y*S, x0+_x*S  # cell top-left, sub-frame centers on it: alignment deferred
                            if iy<0 or ix<0 or iy+S>Y or ix+S>X: frame[_y,_x] = 0; continue  # outside image
                                if _T := frame_H(image, iy,ix, Y,X, rV, 1, max_elev=elev):  # build new tile from pixels up to current level, no fb
                                    T_ += [_T]; _T_ += [(_T,_y,_x)]; frame[_y,_x] = _T
                                else: frame[_y,_x] = 0
            __T_ = _T_
        if len(T_) > 1:  # else no expansion, frame = seed
            TT,C,R = sum_vt(T_); R += elev
            if val_(TT, ttFrm) * ((C+wFrm)/(R+cFrm)) > ave:
                return T_,C,R
        return [],0,0

    aTT,oTT = np.zeros((2,9)),np.zeros((2,9)); aH,oH = [],[]  # regime refs across levs / ffeedback
    T = Fr = vect_edge( frame_blobs_root( comp_pixel( image[iY:iY+Ly**elev, iX:iX+Lx**elev]), rV), iY,iX,Ly,Lx, rV)  # base process
    while T and elev < max_elev:
        tile_,C,R = fill_frame(iY,iX, elev, T)  # seed tile -> sparse higher-scope tile( oH( aH
        if not tile_: break
        Fr = sum2F(tile_); Fr.H += [Copy_(Fr.Nt)]  # minimally processed level
        Fr.N_ = [g for t in tile_ for g in t.N_]  # concat edge Gs
        cluster_C(Fr.Nt, get_exemplars(Fr.N_,R,C), R,C)
        if not Fr.Ct: break
        cross_comp(Fr.Ct, R, Fr)  # agg+ over CC-pruned exemplars
        if ffb: Fr,aTT,oTT,aH,oH = ffeedback(Fr, aTT,oTT,aH,oH)  # top call only, term,form oH(aH
        elev += 1; T = Fr  # next-extension seed
    if Fr: FV_(CoF.get(), Fr.dTT, Fr.c, Fr.r)
    return Fr  # frame | base tile if no levels
```

Changes:

1. `T = Fr = vect_edge(...)`: base tile is the fallback return, so the recursive call at `max_elev=elev` (loop skipped) returns the built tile instead of `[]`. Fixes dead expansion.
2. Sub-call passes `elev=1`: new tiles build hierarchically from pixels to current level, instead of flat-processing a `Ly**elev` window. Termination: inner `max_elev` strictly decreases per nesting.
3. Level iteration decoupled from feedback: `elev += 1; T = Fr` unconditional, `ffeedback` gated by `ffb` alone. Otherwise sub-builds (ffb=0) could never climb past level 1. Also removes the vacuous `elev and`.
4. `if not tile_: break` and `if not Fr.Ct: break` — no infinite loop on empty level.
5. Coordinates: seed actually centered (`cy,cx = Ly//2,Lx//2`), pixel anchor `y0,x0 = _iy-cy*S`, cell→pixel `y0+_y*S`. One convention, invertible.
6. Image bounds via `Y,X` per cell; out-of-bounds cells marked 0.
7. Seed included: `T_=[T]`; expansion gate becomes `len(T_) > 1` — a seed-only frame returns empty, loop breaks, base tile propagates.
8. Weight convention: `val_(TT, wTT_arg)` instead of premultiplied `val_(TT*wTT)` in all three gates; dropped dead `wTT=ttFrm` from `sum_vt`.
9. Dropped vestigial `global ave,avd`; unaliased `aTT,oTT`.

Deferred, flagged in comments: per-cell directional projection (kept your aggregate box-diagonal proj_N call — its dist/A normalization is internally consistent, sign bias remains); sub-frame/cell alignment (sub-frame centers on cell top-left, so content overlaps neighbor cells by ~S/2); duplicate top-lev in H from `sum2F` + `Copy_(Fr.Nt)` — that one lives in sum2F, out of scope here.

kwcckw: I'm still trying to understand the process, and the fable version has this `cy,cx = Ly//2,Lx//2` and ` y0,x0 = _iy-cy*S, _ix-cx*S ` to shift back the coordinates based on the new S, is this relevant?

boris-kz: yes, that's correct. in my version it's cy = cx = 31, same as Ly//2,Lx//2

kwcckw: In your version:

   T_ = []; frame[31,31] = T; __T_=[(T,31,31)]

The starting tile always the same? The `_ix` and `_iy` becomes not relevant? 

It should be based on `_ix` and `_iy` since we will retrieve the image based on `_ix` and `_iy`:

      T = Fr = vect_edge( frame_blobs_root( comp_pixel( image[iY:iY+Ly**elev, iX:iX+Lx**elev]), rV), iY,iX,Ly,Lx, rV)  # base process

kwcckw: Also for this line:

    iy = _y**elev; ix = _x**elev

This is shifting the starting location, but is it intended to increase the image size based on elev?

If `_x` is 64 at elev=1, the next elev is `64**2 = 4096`?

boris-kz: `y0,x0 = _iy-cy*S, _ix-cx*S `

center is a center tile, y0,x0 are min pixel y,x in the tile

kwcckw: I see, so when elev is high, we may get off the frame coordinates, which make it no longer valid? For example:

When ix and iy coordinates are 500, from claude:

      S       = Ly**elev = 64**2          = 4096      # cell span, px
      cy,cx   = Ly//2                     = 32        # cell index of seed
      y0      = _iy - cy*S = 500 - 131072 = -130572
      x0      = _ix - cx*S                = -130572
      span    = Ly*S = 64*4096            = 262144  # frame extent, px
      
 We may actually get a very large span, with negative x0 and y0?

boris-kz: That's what max_elev is for. It doesn't matter, we need to focus on the core process. 
Right now there is no cross comp for Cs, we are still using CC only to select representative Ns for agg+. 
But as we discussed, Cs can also be compared. So these are two redundant forks, and I am not sure how to select between them.

boris-kz: Traced: `cross_comp(Ct,...)` at 906/110 always runs with fC=0, so `e_` = union of C members weighted by summed membership, then `comp_N_` over Ns. `comp_C_` is only reachable via the fC=1 fork, which is never taken. Two observations, then a selection criterion:

1. **The forks are informationally nested, not parallel.** C attrs are membership-weighted sums of member attrs, so comp_C_ ≈ comp of member aggregates: it recovers the between-C component of member-pair matches and discards the within-C component. comp_N_ over members recovers both, at ~(len(e_)/len(C_))² the cost. So the choice is resolution vs cost, and it's decidable per C, not globally.

2. **Per-C criterion: within-C coherence, which cluster_C already computes.** A C whose members converged (its D < Avd stability test, high M/D) is representative: its members add mostly noise beyond the prototype. A loose C is unrepresentative: its mean is a fiction. So form the agg+ comparand set as a mixture:
   - tight C → enters as one node (typ=2), members excluded or down-weighted by membership m,
   - loose C → skipped, its members enter individually with current w = summed membership.
   
   Then one comp pass over the mixed set. Member-level resolution for matched tight Cs is recovered by the existing specification recursion in comp_N (product over N_ on strong top-level match) — so you don't lose it, you just gate it by C-level match first. This collapses the two forks into the general principle already in comp_N: compare highest composition first, descend on match.

3. **Redundancy accounting if both run:** matches recovered by descent into C.N_ must be rdn-discounted against the C-level match that triggered them; overlapping membership goes into r (partially done via olp in cluster_C). Without this, the mixed set double-counts shared members across overlapping Cs.

4. **Ext derivatives constrain the C fork.** C.yx is a member mean; dist/angle between Cs is meaningful only for positionally compact Cs. proj_L_ decay partially self-corrects (loose C → large span → weak projection), but the coherence gate in (2) should include positional spread, not just dTT convergence — otherwise you compare prototypes whose "distance" is noise.

5. **Different projection value per fork, if you keep both hierarchies.** C-C match projects compression (one prototype covers both → c reduction, recognition axis); N-N match projects spatial continuation (extension axis). If clustering C-links (cluster_N over comp_C_ output) is worth doing at all, its gate should weigh compression gain, not projected external match — different valuation, so they're not redundant at the hierarchy level, only at the comparand level, and (2) resolves that level.

Net: no global either/or. Selection is per-C by convergence (D < Avd, plus positional compactness), producing a mixed comparand set; N-resolution returns selectively via comp_N's own descent.

kwcckw: > That's what max_elev is for. It doesn't matter, we need to focus on the core process. Right now there is no cross comp for Cs, we are still using CC only to select representative Ns for agg+. But as we discussed, Cs can also be compared. So these are two redundant forks, and I am not sure how to select between them.

Okay, and I think the comparison of Cs should be optional, added on top of n-fork's non stable clusters (using ds) for additional refinement? I think we had this similar process before.

PR #217, opened 2026-07-26
7/26/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -902,11 +900,11 @@ def fill_frame(_iy,_ix, elev, T):  # expand level frame from seed tile, similar
         if tile_:
             Fr = sum2F(tile_); Fr.H += [Copy_(Fr.Nt)]  # minimally processed level
             Fr.N_ = [g for t in tile_ for g in t.N_]  # concat edge Gs
-            cluster_C(Fr.Nt, get_exemplars(Fr.N_,R,C), R,C)
-            if Fr.Ct:
-                cross_comp(Fr.Ct, R, Fr)  # agg+ over CC-pruned exemplars
-                if ffb: Fr,aTT,oTT,aH,oH = ffeedback(Fr, aTT,oTT,aH,oH)  # top call only, term,form oH(aH
+            cross_comp(Fr.Nt, R, Fr)  # Fr.Ct should be added from sub+ in cluster_N
+            if Fr.Ct: [F2N(C) for C in Fr.Ct.N_]; cross_comp(Fr.Ct, R, Fr,fC=1)  # agg+ over CC, should be conditional based on sub+ Cs?
```
kwcckw: > But as we discussed, Cs can also be compared. So these are two redundant forks, and I am not sure how to select between them.

I think C fork should be conditional based on G's C, and probably we need to eval based on d so that the C fork refinement is applied on the high variance clusters?

frame_2D_alg/agg_recursion.py
```diff
@@ -575,7 +575,7 @@ def sum2G(ft_, fTT, root=None, init=1):  # core clustering function
     Ft_ += [sum2F(list(set(C_)), root.Ct) if C_ else CF()]  # add multiple root_ in Cs?
     G = comb_Ft(*Ft_, root, wTT=fTT)
     N_ = G.N_; N=N_[0]; G.sub = N.sub+1 if G.L_ else N.sub; r=G.r
-    if G.Lt:  # sub+
+    if G.Lt and any([N.rim for N in N_]):  # sub+ (skip Gs from vect_edge, they do not have rim anyway)
```
boris-kz: they should not be skipped, we just need to run them through proj_L_ in cross_comp, that's what it's for.
so we don't need exposed cluster_C and Ct check, that's also in cross_comp:

        if tile_:
            Fr = sum2F(tile_); Fr.H += [Copy_(Fr.Nt)]  # minimally processed level
            Fr.N_ = [g for t in tile_ for g in t.N_]  # concat edge Gs
            cross_comp(Fr.Ct, R, Fr)  # agg+ over CC-pruned exemplars
            if ffb: Fr,aTT,oTT,aH,oH = ffeedback(Fr, aTT,oTT,aH,oH)  # top call only, term,form oH(aH
            elev += 1; T = Fr  # next-extension seed

frame_2D_alg/agg_recursion.py
boris-kz: cluster_C is is actually supposed to prune Gs before comp_N_(G_). That's the point of sub+, initial Gs are tentative and over-extended: a chaining fault mode.
So it should be in sum2G only, and we probably don't need cluster_N there, cluster_C is already proximity constrained, we can use a combination of pairwise and groupwise similarity in sub+ refining?

frame_2D_alg/agg_recursion.py
kwcckw: Actually this is referring to the cluster_C or cluster_N of sub, and we can't cluster Gs before compare them? Because their rim is empty anyway, so how do we really cluster them here?

boris-kz: use prim that we get from proj_L_

frame_2D_alg/agg_recursion.py
boris-kz: Agreed on the restructure — it also dissolves the previous fork question: Cs stop being a parallel representation and become the refinement operator on Gs. sum2G already has the cluster_C | cluster_N sub+ fork (578-584, mdecay-gated), so this is mostly deletion plus rewiring. Traced issues:

1. **Dead wire, load-bearing for your proposal:** in sum2G, `G_,r = cluster_C(...)` at 583 — G_ is discarded. G's own attrs (dTT, c, span, angl, N_) stay those of the over-extended flood-fill cluster; sub-clusters are reachable only via Ct/roots. If CC's role is pruning G, the missing contract is: G attrs recomputed from surviving members, and splits replace G in the caller. That forces ordering change inside sum2G: CC must run on ft_ before comb_Ft (which currently sums forks first), or comb_Ft must rerun per split. And sum2G's return becomes a list (callers at cluster_N:379, trace_edge/vect_edge:864).

2. **Chaining fault is only fixed by multi-seed CC.** A bridged bimodal G splits only if ≥2 exemplars per tentative G survive; single-C fit against a drifting mean under-splits (mean absorbs the bridge). get_exemplars within G provides this, but its inhibition threshold was tuned for sparse frame-level seeding — within a tentative G it must be permissive enough to seed both modes, else CC inherits the chain.

3. **Contiguity is not guaranteed by groupwise fit.** Candidates arrive via rims of accepted members (proximity holds), but membership is fit-only: pruning a mid-chain member leaves accepted members disconnected within a C. Disconnected comparands corrupt ext (span, yx, angl → proj, comp derivatives). This is exactly what your pairwise term should do — but as a hard gate (reachability through currently-accepted members, the iM/span·decay norm from cluster_N:363 migrated into the membership loop), not a summed term. Summing pairwise + groupwise double-counts: C attrs already contain the member attrs that generated those rim links. Roles: pairwise = candidacy + contiguity gate, groupwise = inclusion value. Alternatively a cheap component split per C after convergence — but the gate is cleaner and iterative.

4. **Bt dies at sub-levels without replacement.** cluster_N's d>avd fork (367) is the only sub-level boundary source. CC replacement: after membership settles, links with endpoints in different Cs or d>avd → B_ per C. Arguably cleaner than flood-fill exclusion (boundary = between-cluster contrast by assignment), but it must be added, or core+contour exists only at the top.

5. **Exclusivity for comparands.** CC is fuzzy (root_ multi-membership); overlapping Gs as agg+ comparands double-count in comp_N_ and fork sums. Either exclusive assignment at sum2G output (max-m root, others → rdn/olp in r), or keep fuzzy and discount olp everywhere downstream. For a connectivity hierarchy exclusive is cleaner; fuzziness survives internally as m_ weights.

6. **Obsolete on closure:** cross_comp's else-branch (Ct-member selection, e.w) and the fC fork with it (comp_C_ then unreferenced); cluster_N's Ct-splice block (388-392); frame-level cluster_C at frame_H:904 (exemplars feed cross_comp directly); sum2G:589 `cross_comp(Ct,...)` — splits are already in Nt, so this call is redundant or repurposed. Open: cluster_P — root-constrained CC can't produce cross-G overlap, so its global-refine input disappears; either it runs over split outputs at frame level or it's dead too.

Net: the scheme is coherent — cluster_N shrinks to initial tentative flood-fill per level, CC-in-sum2G is the precision pass, sub+ recursion is divisive CC only. The two nontrivial design points are (3) contiguity gating and (4) boundary re-derivation; the rest is contract cleanup.

boris-kz: Yes, several families — none framed as your exemplar-seeded CC, but the pairwise/groupwise combination with an explicit knob is well published:

1. **CURE (Guha et al., 1998).** Closest in motivation: clusters represented by scattered points shrunk toward the centroid by tunable factor α. α=0 → all-points behavior (single-link, chaining-prone), α=1 → pure centroid. Explicitly designed to interpolate between chaining and compactness — same fault mode you're targeting.

2. **Lance–Williams flexible-β (1967).** Parametric recurrence over agglomerative merge criteria; β interpolates continuously across single-link (pairwise extreme) through average/centroid/Ward (groupwise extremes). The canonical tunable family; centroid and Ward variants are groupwise criteria executed in a pairwise merge loop.

3. **Chameleon (Karypis, Han, Kumar, 1999).** Merge score = relative interconnectivity × relative closeness^α: cross-cluster link density (aggregated pairwise) combined with a group-level closeness term, α tunable. Two-phase: fine graph partition first, then combined-criterion merging — structurally like your tentative-cluster-then-refine.

4. **MRF/energy segmentation (Boykov–Jolly 2001, GrabCut 2004).** Unary term = fit to cluster model (GMM ≈ centroid fit, groupwise) + λ·pairwise contrast term between neighbors. λ is the tunable balance. This is the largest literature on exactly this sum — and its known pathology is your double-counting point: λ is set ad hoc per dataset because the two terms share evidence.

5. **Spatially regularized fuzzy c-means (FCM_S, Ahmed 2002; FLICM, Krinidis–Chatzis 2010).** Membership = distance-to-centroid + α·neighborhood consistency term. Nearly your scheme in image domain: groupwise inclusion value plus local pairwise coherence. Notably FLICM's contribution was eliminating the tunable α by making the local term self-weighting — evidence that the free knob was a recognized liability.

6. **Constrained k-means (HMRF-KMeans, Basu–Bilenko–Mooney 2004).** Objective = point-to-centroid distortion + tunable pairwise link penalties. Semi-supervised framing, but the objective is literally weighted pairwise + groupwise.

7. **Contiguity-constrained clustering (regionalization: Murtagh 1985; SKATER; max-p).** Pairwise adjacency as hard gate, groupwise homogeneity as the optimized criterion — the asymmetric-roles version I argued for, standard in geographic clustering precisely because summed spatial terms distort the objective.

Pattern across all: weighted-sum versions (4, 5, 6) all struggle with setting the weight, because the terms aren't independent; hard-gate versions (7, and CURE's structural α) are more stable. Supports gate-not-sum for your case — with the difference that your "knob" wouldn't be a free hyperparameter but the same ave/decay machinery, i.e. feedback-tuned rather than hand-set.

frame_2D_alg/agg_recursion.py
kwcckw: Okay, so we probably need another step to convert the prim tuple into CL first, else it maybe add a lot more clutter in the cluster_N or cluster_C later.

boris-kz: Actually the way I framed it below, cluster_C is sub+ only, in sum2G. So the nodes it gets will have rim. That includes sum2G in vect_edge:

return sum2G([[G_,'Nt',TT,C,1],[L_,'Lt',lTT,lc,lr]], TT+lTT, T) # c,R?

kwcckw: I see, so `cross_comp(Fr.Ct, R, Fr)` is actually cross comparing those sub Cs instead.

>  So the nodes it gets will have rim. That includes sum2G in vect_edge:

    return sum2G([[G_,'Nt',TT,C,1],[L_,'Lt',lTT,lc,lr]], TT+lTT, T) # c,R?
    
But we are parsing G_ here, and the new clustered Gs will never get any rim before we comp them?

boris-kz: yes, the sum2G I was talking about is in trace_edge:

G_ += [sum2G([(n_,'Nt',ntt,nc,r)]+([(l_,'Lt',ltt,lc,r)] if l_ else []), ttTrc, root)]

frame_2D_alg/agg_recursion.py
```diff
@@ -89,7 +89,7 @@ def cross_comp(Ft, R, root, fC=0, fB=0):  # calls cluster_N, sub+, over exemplar
     if fB:
         e_= Ft.N_  # Bt
         for e in e_: e.w = e.d  # contrast value
-    else:  # select / Ct
+    elif fC:  # select / Ct
         e_ = {n for C in Ft.N_ for n in C.N_}
         for e in e_: e.w = sum(r[1] for r in e.root_)  # combine memberships
         e_ = [e for e in e_ if e.w]
```
boris-kz: So now we only need comp_N_ (no comp_C_), over the Gs refined and potentially decomposed by sub+. These Gs may or may not have centroid features?

Fable:
and the mixed population (refined | unrefined | singleton) should not be special-cased. 
Comparability is already structural: base_comp reads Et, kern, extT, dTT, which exist for all. The heterogeneity is a certainty difference, and the system already has the pricing mechanism: c/r. That reduces the question to three contracts:

1. **Refinement must be visible in r, or it bought nothing downstream.** A CC-refined G's attrs are variance-pruned (membership-weighted sums, converged D) — a tighter summary than a flat flood-fill sum of the same members. If sub+ rewrites dTT/kern/span/yx but leaves r as-is, comp_N_'s gates (proj_L_, gv_) treat refined and chained comparands as equally certain, and refinement cost was pure loss at this level. Contract: CC convergence (final D vs Avd, or Up trend) feeds G.r reduction — refined → lower r, unrefined → r carries the chaining uncertainty. Then no flags, no branching: selection prices the difference uniformly.

2. **comp_N over refined Gs *is* centroid comparison.** Refined G attrs are the prototype; the fC/comp_C_ fork stays subsumed. Where centroid features exist (m_, d_, root_ per member), they modulate the existing specification descent: member-pair comps in comp_N's product weighted by membership m on the refined side, 1 on the unrefined side. Asymmetric weighting is fine — weights are per-node, and root_ already stores [C,m,d].

3. **Decomposition must not discard bridge evidence.** Splitting a chained G partitions its L_ into within-split links (→ each split's Lt) and cross-split member links — which are precisely the evidence that caused the fault. Next-level comp_N_ over sibling splits would re-derive that relation from scratch via proj_L_. Instead: sum cross-split member links into a pre-link per split pair — a strong prior that gates/primes their next-level comp cheaply. Pruned bridge members (fitting no C) and their links → Bt or residue, per the earlier boundary re-derivation. Otherwise sub+ pays to isolate the bridge, then throws it away, then agg+ pays again to rediscover it.

4. Gs that never triggered sub+ (weak Lt, failed gv_) enter agg+ as-is — correct, since refinement is value-gated and forcing it inverts the cost logic. They're mostly filtered by exemplar/proj gates anyway; where they pass, their higher r discounts them.

Net: one comparand population, one comp path; heterogeneity handled by honest r accounting (1), membership-weighted descent (2), and link inheritance across the split (3). Only (1) is a new obligation — (2) and (3) reuse existing structures.

frame_2D_alg/agg_recursion.py
boris-kz: this should be external: if sum(e.w for e in e_) * ((Ft.c+wAgg) / (Ft.r+R+cAgg)) * ((len(e_)-1) **2 * wL):

def cross_comp(Ft, R, root):  # calls cluster_N, sub+, per B_ or spliced from C_-> N_ exemplars

    if Lt := comp_N_(Ft.N_, R):  # N_ may be a mix of typ 3 Ns and typ 2 Cs 
        L_, TT, c, r, V = Lt
        root.L_ = L_; oF_[CoF.get().nF].V_ += [V]  # comp V
        if gv_(val_(TT*ttcN) * ((c+wcN)/(r+ccN)) * ((len(L_)-1)*wL) - ave):  # return +ve, store -ve gate vals
            G_,r = cluster_N(root, E_,r,c)  # -> sum2G, sub+
            if G_:
                root.Nt = sum2F(G_,root,froot=2)  # | C_?
                if root.Ct:  # refined for agg+
                    cross_comp(root.Nt, r,root)
            return G_

frame_2D_alg/agg_recursion.py
```diff
@@ -584,9 +584,7 @@ def sum2G(ft_, fTT, root=None, init=1):  # core clustering function
             else:     G_,r = cluster_N(G.Nt,E_,r,c)  # CC sub-Gs, if any
     if root.root is None:  # when root is frame, their root is empty, and G is tile
```
boris-kz: do you remember what is this for?

frame_2D_alg/agg_recursion.py
kwcckw: > So now we only need comp_N_ (no comp_C_), over the Gs refined and potentially decomposed by sub+. These Gs may or may not have centroid features?

Actually I don't get this. If there's no more comp_C_
, how do we get the root.Ct above? 

        if root.Ct:  # refined for agg+
                cross_comp(root.Nt, r,root)

frame_2D_alg/agg_recursion.py
kwcckw: So when we sum Gs to tile, their Lt should be empty right? Since those Ls are per Glll.N_ instead of G_.

frame_2D_alg/agg_recursion.py
kwcckw: This is no longer relevant since we no longer need to pack each tile's directional edge N for merging purpose. It's done globally now

frame_2D_alg/agg_recursion.py
boris-kz: it's from cluster_C. 
we may still compare Cs, but that should done in comp_N_, comp_N already checks typ

frame_2D_alg/agg_recursion.py
boris-kz: Right.

frame_2D_alg/agg_recursion.py
boris-kz: So,

def get_exemplars(N_,_r,_c):  # multi-layer non-maximum suppression -> sparse seeds for diffusive clustering, cluster_N too?

    for n in N_: 
        # combined lateral and vertical match
    N_= sorted(N_, key=lambda n: n.w, reverse=True); E_,Inh_ = set(),set()
...

frame_2D_alg/agg_recursion.py
kwcckw: Okay, so that cross_comp of Cs should be remained in sum2G? Else this is actually comparing the G_ instead of C_ in the cross_comp:

         if root.Ct:  # refined for agg+
                cross_comp(root.Nt, r,root)

frame_2D_alg/agg_recursion.py
kwcckw: The root_ is optional,  i think we need to check with hasattr:

     n.w = (n.Rt.m + (sum([r[1] for r in n.root_]) if hasattr(n, 'root+') else 0) ) * n.c

boris-kz: make it default

frame_2D_alg/agg_recursion.py
boris-kz: yes,

So,

def get_exemplars(N_,_r,_c): # multi-layer non-maximum suppression -> sparse seeds for diffusive clustering, cluster_N too?

for n in N_: 
    n.w = (n.Rt.m + sum([r[1] for r in n.root_])) * n.c  # combined lateral and vertical match
N_= sorted(N_, key=lambda n: n.w, reverse=True); E_,Inh_ = set(),set()
...
then in sum2G:

```
if G.Lt:  # sub+
    Lt = G.Lt; L_,lm,ld,lr = Lt.N_,Lt.m,Lt.d,Lt.r; Av=ave+avd  # no levR = 1/len(L_): represented by c
    if gv_((lm+ld)* wcC - Av* (lr+1+ccC)):  # mdecay(L_)-decay?
        c = G.Lt.c; r+= 1
        G_,r = cluster_C(G.Nt, get_exemplars(N_,r,c), r,c)
     Et = sum2F( get_exemplars(N_,r,c))
     if gv_(Et.w*ttA * ((c+wAgg)/(r+cAgg)) * ((len(L_)-1)*wL) - ave):
         cross_comp(G.Nt, r, G)  # ultimate sub+

```

...

boris-kz: 1. **The load-bearing wire is still missing.** `G_,r = cluster_C(...)` — G_ discarded again. Nothing rebuilds G.Nt from the splits, so `cross_comp(G.Nt, r, G)` re-comps the *original* members — pairs already linked in G.Lt. The whole point (comp over refined/decomposed population) requires `G.Nt = sum2F(G_ + survivors)` (or in-place mutation by cluster_C, contract change) before the Et gate. Same applies to the gate inputs: second get_exemplars and the count term should run over the refined population, not stale N_ and old L_.

2. **Et.w is never set.** sum2F doesn't aggregate w (only Copy_ carries it, single-node). Et.w reads the CF default. Also sum2F on typ-3 exemplars triggers the deep branch — H concat, member-level sum2F, box — pure side-effect cost for one scalar. Use `sum(e.w for e in E_)` directly; no Et object needed.

3. **Sort-key / gate-key mismatch breaks NMS.** You sort by n.w (lateral+vertical) but the selection loop still gates on `N.Rt.m * N.c * wE` and `break`s on first failure. Break-on-fail assumes the sort key is monotone with the gate key; it isn't now — a high-membership/low-rim node sorts early, fails the lateral-only gate, and terminates selection over still-valid tails. The gate (and oV normalization, currently oM/Rt.m) must switch to n.w too.

4. **Scale mixing inside w.** Rt.m is intensive (val_ of c-weighted mean TT from sum2F); summed root_ memberships are extensive (grow with membership count). Their sum drifts with CC fuzziness — a node in 4 overlapping Cs outranks a strong-rim node regardless of fit quality. Either mean the membership term or c-weight both to the same order. Double-counting per se is tolerable for *ranking* (centrality in both axes is a fair exemplar criterion), but only at matched scale.

5. **Dimensional and count terms in the Et gate.** `Et.w*ttA` = scalar × (2,9) array into gv_ — meant wAgg or a val_ reduction? And the old top gate scaled by (len(e_)-1)² — pair count, matching O(n²) comp; new uses linear (len(L_)-1) over *old* links. Should be prospective comparand count (refined splits), superlinear.

6. **First call's vertical term is dead.** root_ is empty until cluster_C runs (and cluster_C resets it at entry), so seeding-call w = Rt.m·c — old behavior, fine, but the "combined lateral and vertical" semantics only exist in the post-CC call. Intentional layering, worth a comment; also means stale root_ from trans_cluster merges leaks into first-call w before the reset.

7. **Exactly one recursion site.** This sketch recurses per-G inside sum2G; your previous cross_comp sketch recursed via `if root.Ct: cross_comp(root.Nt,...)`. Both alive = double recursion (cross_comp → cluster_N → sum2G → cross_comp per G, *plus* root-level re-entry). Confirm the cross_comp-internal one is deleted; then `root.Ct` has no consumer and cluster_C:445 setting Ft.root.Ct is vestigial unless kept for Ct bookkeeping.

8. **Gate 1 is now purely intensive.** lm, ld, lr from sum2F are c-weighted means; with ccC unscaled by L and no count term, a 2-link Lt and a 200-link Lt gate identically on per-unit quality. Comment says count is "represented by c" but c doesn't appear in gate 1 — it's only extracted after. If the layering is intent (cheap intensive trigger, extensive gate 2), fine — but then gate 2 must be correct per (1), (2), (5), since it's the only place quantity binds.

9. Still open from last round: comp_N_ receiving raw N_ (proj_L_ relocation — max-dist, prim, pTT/rTT source) — this sketch inherits that unresolved contract via cross_comp.

kwcckw: I don't think we need this Et?

    Et = sum2F( get_exemplars(N_,r,c))

We could have just sum w from the exemplars?

And i think Et.w should be conditional based on G_ (if there's C fork), else N_?

boris-kz: this is not relevant

boris-kz: sum2F is for c-weighted summation.

kwcckw: Okay, so in the eval below, c and r should be from Et?

    gv_(Et.w*ttA * ((c+wAgg)/(r+cAgg)) * ((len(L_)-1)*wL) - ave):

kwcckw: So when N are Cs, we should pack their root_'s C into G.C_ only?

boris-kz: yes,

    if G.Lt:  # sub+
        Lt = G.Lt; L_,lm,ld,lr = Lt.N_,Lt.m,Lt.d,Lt.r; Av=ave+avd  # no levR = 1/len(L_): represented by c
        if gv_((lm+ld)* wcC - Av* (lr+1+ccC)):  # mdecay(L_)-decay?
            c = G.Lt.c; r += 1
            G_,r = cluster_C(G.Nt, get_exemplars(N_,r,c), r,c)
        Et = sum2F( get_exemplars(N_,r,c))  # or just sum_vt + w?
        if gv_(Et.w * ((c+wAgg)/(r+cAgg)) * ((len(L_)-1)*wL) - ave):
            cross_comp(Et, r, G)  # ultimate sub+

boris-kz: no, it's still G.N_, mixed-typ

kwcckw: So it will be:

    gv_(Et.w * ((Et.c+wAgg)/(Et.r+cAgg)) * ((len(L_)-1)*wL) - ave)

I can see your posted code is still using the default c and r above.

If there's G_, i think we should replace N_?

      if gv_((lm+ld)* wcC - Av* (lr+1+ccC)):  # mdecay(L_)-decay?
              c = G.Lt.c; r += 1
              N_,r = cluster_C(G.Nt, get_exemplars(N_,r,c), r,c)

Then Et should be summed from the Cs from cluster_C instead.

kwcckw: Previously we had:

    G.C_ += N.root_

So that G.C_ is their N_'s roots, that should be still relevant?

boris-kz: yes, but we get it directly from cluster_C

boris-kz: probably, I was thinking that we may have sub+ without cluster_C, not sure

kwcckw: Yes, that N_ replacement should works the same since if there's no cluster_C, it defaults to the original N_ anyway.

PR #218, opened 2026-07-27
7/27/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -101,13 +101,19 @@ def cross_comp(Ft, R, root):  # calls cluster_N, sub+, per B_ or spliced from C_
...
 
-    L_,N_ = [],[]
+    L_,N_,mrg_ = [],[],[]
     for dist, dy_dx, _N,N, lc,lr, pTT,m,d in pL_:
         if m > 0:
             if gv_(m*(lc/lr)*wN - ave*(r+cN)):  # comp if marginally predictable, update N.Rt pair eval, ave / proj surprise value?
                 Link = comp_N(_N,N, lr,lc, full = not tnF, A=dy_dx, span=dist, rL=root)
                 Link.rTT = np.abs(pTT - Link.dTT) / eps_(Link.dTT)  # relative prediction error to fit oF, direction-agnostic
                 L_+= [Link]; N_+= [_N,N]
+                if _N.root_ and gv_(Link.m*wF- ave*(Link.r+cF)):
```
kwcckw, line 111: When we use comp_N_ on both N and C, i think we only need this C merging under `fall`？

I think we no longer need the section under fall= 0?

            for C in C_: C.compared=set()
            for j in range( len(C_)-1):
                _C = C_[j]; C = C_[j+1]
                if _C in C.compared: continue
                tc += c; tr += r*c
                L = comp_N(_C,C,_r, c ,A=dy_dx,span=dist); L_+=[L]; N_ +=[_C,C]  # simplified for typ=2
                if L.m>ave: TTm+=L.dTT; cm+=L.c; rm+=L.r

boris-kz: Probably. If we have Cs and Gs mixed in Nt.N_, there is no point in separate Ct?
Or we can have separate Nt.N_ and Ct.N_, mixed in G.N_?

kwcckw: > If we have Cs and Gs mixed in Nt.N_, there is no point in separate Ct

But we won't have the combination of Cs and Gs unless we concatenate them? At this point i don't see how we can get both?

> Or we can have separate Nt.N_ and Ct.N_, mixed in G.N_?

This is actually not possible since G.N_ is Nt.N_, unless we pack it into another attr.

boris-kz: concat is not a problem, I am just not sure it makes sense

kwcckw: I see, I think it's fine since they are both clusters of Ns, and the difference is they are formed with different clustering method only.

boris-kz: We don't actually need to combine N_ and C_, just process them sequentially

kwcckw: Okay, and we still need to separate C and N in both Ct and Nt, since they should be packed separately in different forks, and right now they probably share the same comp_N_ only.

frame_2D_alg/agg_recursion.py
```diff
@@ -101,13 +101,19 @@ def cross_comp(Ft, R, root):  # calls cluster_N, sub+, per B_ or spliced from C_
 
 def comp_N_(pL_, r, tnF=None, root=2):  # incremental-distance cross_comp, max dist depends on prior match
 
-    L_,N_ = [],[]
+    L_,N_,mrg_ = [],[],[]
     for dist, dy_dx, _N,N, lc,lr, pTT,m,d in pL_:
         if m > 0:
```
boris-kz, line 106: def comp_N_(pL_, r, tnF=None, root=2, fall=0):  # incremental-distance cross_comp, max dist depends on prior match

    L_,N_,mrg_ = [],[],[]
    for dist, dy_dx, _N,N, lc,lr, pTT,m,d in pL_:
        if fall or m > 0:

kwcckw: Previously we have fall to compare within all pairs, so when fall == 0, it's sorting ddT using the max ttC_ index:

        for C in C_: C.compared=set()
        for j in range( len(C_)-1):
        
Now the fall is meant for default comp between the pairs? But when do we need this default comp? I think it should comp based on positive pL only?

boris-kz: comp_pL_  costs may not be worth the savings from selection

frame_2D_alg/meta_code.py
```diff
@@ -70,6 +70,7 @@ def __init__(f, **kw):
         f.wTT = kw.get('wTT',wTT)  # mean wTT = 1?
         f.typ = kw.get('typ',0)  # blocks sub_comp
         f.root = kw.get('root',None)  # convert to list in typ oFs?
+        f.root_ = kw.get('root_',[])  # C's roots
```
boris-kz, line 73: what is that for?

kwcckw: For default root_? As in get_exemplars:

      for n in N_:

boris-kz: Ok, right. Sorry I didn't have much time today

kwcckw: Alright, i will continue to check if there's any other issues.

PR #220, opened 2026-07-29
7/29/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -332,7 +332,7 @@ def cluster_C(Ft, E_,_r,_c):  # form centroids by clustering exemplar surround v
         _C_ += [C]
     out_ = []
     while True:  # reform C_
-        # need to add eval Rt: connectivity term
+        # need to add eval Rt: connectivity term
```
kwcckw, line 335: Right now all frontiers are default, so we need to add eval based on their Rt.m?

frame_2D_alg/agg_recursion.py
```diff
@@ -489,14 +489,17 @@ def add_H(H,h, root, fN=0):
             if Lev: add2F(Lev,lev,1)  # unpack nested Hs in add2H
             else: H.append(Copy_(lev, root, cls=(CF,CN)[fN]))
 
+# include xcomp in oF_? Else gv_ is pointless here
 def xcomp(root, G_, m, c, r):  # sub+|agg+: refine by CC and exe -> cross_comp
```
kwcckw, line 493: I guess we need to include xcomp in oF_? Else the gv_ below is pointless.

frame_2D_alg/agg_recursion.py
boris-kz: right now that eval is centroid-based:

`if M * (_C.m+wcC*len(n_)) > Ave * (r+_C.r+ ccC*len(n_)):`
to add connectivity term it's something like: 
`M + sum([l.m for l n.rim if l in _C.L_])`

frame_2D_alg/agg_recursion.py
```diff
@@ -370,7 +370,8 @@ def cluster_C(Ft, E_,_r,_c):  # form centroids by clustering exemplar surround v
             n.exe = (n.d if n.typ==1 else n.m) + sum(rt[1] for rt in n.root_) - ave
         _m,_d,_tt,_c,_r = sum_vt(out_, fm=1)
         FV_(CoF.get(),_tt,_c,_r)  # r/ deeper cross_comp?
-        if gv_(_m*_c * (Ft.root.wTT*ttcC) * (_c/(_r+ccC)) * ((len(out_)-1)*wL) * wcC - ave):
+        # should be using val_ here, along with _tt?
```
boris-kz, line 373: `_m` is already computed by that val_

frame_2D_alg/agg_recursion.py
boris-kz: right

frame_2D_alg/agg_recursion.py
```diff
@@ -514,11 +517,11 @@ def sum2G(ft_, fTT, root=None, init=1):  # core clustering function
     if G.Lt:  # sub+
         Lt = G.Lt; L_,lm,lc,lr = Lt.N_,Lt.m,Lt.c,Lt.r; Av=ave+avd  # no levR = 1/len(L_): represented by c
         if gv_(lm * lc * wcC - Av * (lr+1+ccC)):  # mdecay(L_)-decay?
-            xcomp(root, N_,lm,lc,lr)  # sub+
+            xcomp(G, N_,lm,lc,lr)  # sub+
     if G.Bt:
         Bt = G.Bt; bd,br,L = Bt.d,Bt.r,len(Bt.N_); rroot = root.root if root.root else 0
         if N.typ!=1 and bd*(wAgg*L) > avd*(br+cAgg*L): [F2N(L) for L in Bt.N_]
-        cross_comp(Bt, br, Bt)  # no ddfork; or xcomp too?
+        cross_comp(Bt, br, Bt)  # no ddfork; or xcomp too? I don't think we need centroid clustering on those B_?
```
boris-kz, line 524: it's the same logic, and CC is optional anyway

frame_2D_alg/agg_recursion.py
```diff
@@ -489,14 +489,17 @@ def add_H(H,h, root, fN=0):
...
-    Ct, cm, cc, cr = cluster_C(root, get_exemplars(G_,r,c), r,c)  # refine: root.Nt = Ct?
-    if gv_((m+cm) * ((c+cc+wAgg) / (r+cr+cAgg)) * ((len(G_)-1)*wL) - ave):
-        root.H += [Copy_(root)]  # lower agg lev
-        e_ = get_exemplars(Ct.N_, r,c)
-        cross_comp(e_, r,root)  # agg+
+    for G in G_: G.Rt = sum2F(G.rim)  # add Rt.m for get_exemplars below
+    if (out:= cluster_C(root, get_exemplars(G_,r,c), r,c)):  # refine: root.Nt = Ct?
+        Ct, cm, cc, cr = out
+        if gv_((m+cm) * ((c+cc+wAgg) / (r+cr+cAgg)) * ((len(G_)-1)*wL) - ave):
+            root.H += [Copy_(root)]  # lower agg lev (why not root.Nt?)
```
boris-kz, line 500: why root.Nt?

frame_2D_alg/agg_recursion.py
```diff
@@ -489,14 +489,17 @@ def add_H(H,h, root, fN=0):
             if Lev: add2F(Lev,lev,1)  # unpack nested Hs in add2H
             else: H.append(Copy_(lev, root, cls=(CF,CN)[fN]))
 
+# include xcomp in oF_? Else gv_ is pointless here
 def xcomp(root, G_, m, c, r):  # sub+|agg+: refine by CC and exe -> cross_comp
 
     proj_L_(combinations(G_,2), root.Nt, r, frim=1)  # add pL into rim
```
boris-kz, line 495: this prim is actually not needed, cluster_C should be per G, not root?

frame_2D_alg/agg_recursion.py
kwcckw: I see, thanks for explaining.

frame_2D_alg/agg_recursion.py
kwcckw: Okay, so instead of `_m*_c * (Ft.root.wTT*ttcC)`, we need to use wcC? Because `(Ft.root.wTT*ttcC)` gives an array instead of single value.

frame_2D_alg/agg_recursion.py
kwcckw: Okay, then we should add xcomp the same way here.

frame_2D_alg/agg_recursion.py
kwcckw: The `root.Nt` packs the current level's clusters? Using `Copy_(root)` may copy some unrelated params such as H or Lt?

frame_2D_alg/agg_recursion.py
kwcckw: We actually need prim in frame_H:

            if gv_(m * c * wcC - ave * (r+1+ccC)):
                xcomp(Fr, N_,m,c,r)  # sub+
                
N_ is concatenated from all tiles, so their rim is empty and we need to add their prim here. I guess we need to have frim=1 when call xcomp in frame_H.

frame_2D_alg/agg_recursion.py
boris-kz: right

frame_2D_alg/agg_recursion.py
boris-kz: cluster_C should be per G, not root
that means exemplars will be selected from G.N_, and all Ns have real rim

frame_2D_alg/agg_recursion.py
boris-kz: I think they are related, lower lev is a nested representation

frame_2D_alg/agg_recursion.py
kwcckw: > that means exemplars will be selected from G.N_, and all Ns have real rim
 
 Yes, this is true only for sub+. For frame_H:
 
            N_ = [g for t in tile_ for g in t.N_]  # concat edge Gs
            Fr = sum2F(N_); m,c,r = Fr.m,Fr.c,Fr.r
            Fr.H += [sum2F(tile_)]  # minimally processed level
            if gv_(m * c * wcC - ave * (r+1+ccC)):
                xcomp(Fr, N_,m,c,r)  # sub
                
When we concatenate tiles' N_, we don't have real rim between them (cross tiles) yet?

boris-kz: doesn't matter, that concat N_ is the G_ I was talking about, each n in N.N_ has rim

kwcckw: Yes, each G.N_ has existing rim, but the C clustering is per G_ instead of G_'s N_?

     cluster_C(root, get_exemplars(G_,r,c), r,c)

boris-kz: I just told you that cluster_C should be per G, not root, several times

kwcckw: Okay, sorry, i was thinking it will be different in frame_H.

So for tiles in frame_H, we should skip the centroid clustering now? Since each tile.N_ is G and each of them should have evaluated for cluster_C before.

boris-kz: yes, tiles will be unpacked

kwcckw: If we unpack them to the G.N_ level and concatenate them, each of their n.rim is still within their own G, and we need extra comps or proj_L to get their cross-tiles actual rim or prim?

boris-kz: No, cluster_C is supposed to refine / split individual Gs, that's sub+

frame_2D_alg/agg_recursion.py
```diff
@@ -370,7 +370,8 @@ def cluster_C(Ft, E_,_r,_c):  # form centroids by clustering exemplar surround v
             n.exe = (n.d if n.typ==1 else n.m) + sum(rt[1] for rt in n.root_) - ave
         _m,_d,_tt,_c,_r = sum_vt(out_, fm=1)
         FV_(CoF.get(),_tt,_c,_r)  # r/ deeper cross_comp?
-        if gv_(_m*_c * (Ft.root.wTT*ttcC) * (_c/(_r+ccC)) * ((len(out_)-1)*wL) * wcC - ave):
+        # should be using val_ here, along with _tt?
+        if gv_(_m*_c * val_(_tt, ttcC) * (_c/(_r+ccC)) * ((len(out_)-1)*wL) * wcC - ave):
             Ct = sum2F(out_); Ft.root.Ct = Ct; Ct.root = Ft.root
```
boris-kz, line 375: so all that should be in sum2F:

        if gv_((_m*_c *wcC) / (_r+ccC) * ((len(out_)-1)*wL) - ave):
            return sum2F(out_, Ft.root, nF='Nt')  # also pass _m,_c,_r?

frame_2D_alg/agg_recursion.py
boris-kz: def xcomp(root, G_, m, c, r):  # sub+|agg+: refine by CC and exe -> cross_comp

    if Ct := cluster_C(root, get_exemplars(G_,r,c), r,c):  
        # cluster_C refines, splits each G -> Ct.N_
        if gv_((m+Ct.m) * ((c+Ct.c+wAgg) / (r+Ct.r+cAgg)) * ((len(G_)-1)*wL) - ave):
            root.H += [Copy_(root)]  # lower agg
            cross_comp( get_exemplars(Ct.N_, r,c), r,root)  # agg+

frame_2D_alg/agg_recursion.py
boris-kz: if G.Bt:
        Bt = G.Bt; B_,bd,bc,br = Bt.N_,Bt.d,Bt.c,Bt.r; rroot = root.root if root.root else 0
        if N.typ!=1 and gv_(bd*bc * wcC - Av * (br+1+ccC)):
            [F2N(L) for L in Bt.N_]
            xcomp(G, N_,bd,bc,br, nF='Bt')  # sub+, no ddfork?
?

frame_2D_alg/agg_recursion.py
boris-kz: Traced xcomp (491), callees get_exemplars (234), cluster_C (323), cross_comp (86), and callers sum2G (515, 520), frame loop (858).

1. **Signature mismatch, closure-blocking**: 492 calls `cluster_C(root, get_exemplars(G_,r,c), r,c, nF)` — 5 args; `cluster_C(Ft, E_,_r,_c)` takes 4. TypeError on every xcomp entry.

2. **Agg+ branch is dead even after fixing 1**: `get_exemplars(Ct.N_, r,c)` at 496 runs rim-based NMS on centroids, but centroids have no rim (Rt is only set in comp_N_ 112 from rim; centroid `root_` is empty — sum2F's fC branch appends to member `N.root_`, not to C). So every `n.w = 0`, the first eval fails, fallback yields a single seed, `combinations(E_,2)` is empty, comp_N_ returns None. Agg+ never executes. Conceptually rim-NMS is the wrong redundancy test for centroids anyway — their overlap is shared membership (root_ intersections), and cluster_P's merge-redundant-Cs step already performs that suppression. Passing Ct.N_ directly to cross_comp is probably right.

3. **nF is dead weight through the whole chain**: cluster_C 378 hardcodes `sum2F(out_, Ft.root, nF='Nt')`; cluster_N's reform writes `rG.Nt` (315); cross_comp writes `root.L_` (90). So the Bt-fork xcomp at 520 routes its results into the same G.Nt/G.L_/G.H as the Nt-fork call at 515 — second call clobbers the first's L_ and appends a second H snapshot. Fork identity is lost immediately after the call site.

4. **Bt fork passes the wrong node set**: 518-520 unpacks `B_ = Bt.N_`, converts via `[F2N(L) for L in Bt.N_]`, then calls `xcomp(G, N_, bd,bc,br)` — N_, not B_. The F2N conversion is pointless and the fork re-processes Nt nodes under Bt valuation. Should be B_.

5. **H snapshot at 497 is wrong twice**: (a) `Copy_(root)` with init=0 deep-copies `N.H` (592) — each appended level nests all prior levels, so H grows super-linearly per agg+ iteration, breaking the flat-level convention used by add_H/sum2F typ==3. (b) It's unconditional: appended before cross_comp, whose cluster_N reform of root is separately gated (313) and can fail — leaving a spurious level identical to the current root. Snapshot and reform must share one gate.

6. **Count-basis mismatch in the 494 gate**: `(len(G_)-1)*wL` projects comparison combinatorics over input Gs, but the actual comparands are (exemplars of) Ct.N_, whose count differs after split/merge. Should be len(Ct.N_).

7. **r not propagated**: the gate charges `r+Ct.r+cAgg`, but cross_comp at 496 receives bare `r`. Centroid-formation cost is evaluated but not inherited by the next operation's accounting.

8. **Bt call mixes d into a match gate raw**: at 520, m=Bt.d, so the 494 gate computes `(bd + Ct.m)`. Per your own convention, d's predictive worth is borrowed (`|D|*rM`), not added raw to match.

9. Minor: `_c` is unused in both get_exemplars and cluster_C (shadowed by the local `_c` at 372); `m` is unused in xcomp outside the gate. Dead parameters if 7/8 aren't the intended fix.

frame_2D_alg/agg_recursion.py
kwcckw: Yes, so this should replace Nt, and i think we need to pack existing N_ to H before the Nt replacement.

frame_2D_alg/agg_recursion.py
kwcckw: > No, cluster_C is supposed to refine / split individual Gs, that's sub+

This actually has the same issue as tiles.N_ because L.rim is empty, and we use rim to select frontier in cluster_C. So I guess we should use all to all instead of rim as frontier in cluster_C now?

boris-kz: no, read again. all we need is N.rim for N in G.N_ in G_

kwcckw: But in this line:

    xcomp(G, N_,bd,bc,br, nF='Bt')  # sub+, no ddfork?

It's parsing N_, rather than the B_ for Bt fork? And B_ is link instead of G.N_.

boris-kz: Ok right. So we need something like this:

def xcomp(root, G_, m, c, r, nF):  # sub+|agg+: refine by CC and exe -> cross_comp

    C__ = []; M = C = R = 0
    for G in G_:
        if Ct := cluster_C(root, get_exemplars(G.N_, r,c), r,c):  # refines, splits each G
            C_, _m,_c,_r = Ct; C__+=C_; M+=_m; C+=c; R+=r
    if C__:  # spliced refined sub_G_s
        sum2F(C__, root, nF)  # also pass M,C,R?
        if gv_((m+M) * ((c+C+wAgg) / (r+ R/len(C__)+ cAgg)) * ((len(G_)-1)*wL) - ave):
            root.H += [Copy_(root)]  # lower agg
            cross_comp( get_exemplars(Ct.N_, r,c), r,root)  # agg+

kwcckw: I see, so this is actually splicing each G's C_. For tiles, this line should be redundant:

    Ct := cluster_C(root, get_exemplars(G.N_, r,c), r,c, nF):
    
Since tile's G_ should have run sub+ in sum2G before, so C__ could be just concatenated from tile.G_'s C_ (tile.G_'s Nt.N_), which we only can identify it via tile.G_'s Nt.typ.

boris-kz: The tiles are simply unpacked, G_ is spliced from all tiles

kwcckw: Okay sorry, i misunderstood earlier. So if we use the same xcomp in sum2G, it's actually splicing G.N_'s N_ since the input G_ is G.N_ in sum2G.

boris-kz: Right

kwcckw: So I will go through them again later.

PR #221, opened 2026-07-30
7/30/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -671,11 +673,8 @@ def proj_V(_N, N, dist, dy_dx, dec, r):  # _N x N induction
                 m, d = val_(pTT, ttN_, 1)
                 if fall or m > ave:
                     lc = min(_N.c, N.c); lr = r + (N.r + _N.r) / 2  # +|-match certainty
-                    pL = [dist, dy_dx, _N, N, lc, lr, pTT, m, d]
-                    for l_ in pL_, _N.prim, N.prim: l_ += [pL]
-                    if frim:  # for frame_H's cluster_C
-                        pL = CL(typ=-1, N_=[_N,N],dTT=pTT,m=m,d=d,c=lc,r=lr, angl=[dy_dx,1],span=dist)
-                        _N.rim += [pL]; N.rim += [pL]
+                    pL_ += [[dist, dy_dx, _N, N, lc, lr, pTT, m, d]]
```
kwcckw, line 676: So the prim is no longer relevant and we only need the pL here.

frame_2D_alg/agg_recursion.py
```diff
@@ -488,16 +489,18 @@ def add_H(H,h, root, fN=0):
...
             else: H.append(Copy_(lev, root, cls=(CF,CN)[fN]))
 
-def xcomp(root, G_, m, c, r, nF):  # sub+|agg+: refine by CC and exe -> cross_comp
+def xcomp(root, G_, m, c, r, nF='Nt'):  # sub+|agg+: refine by CC and exe -> cross_comp
 
     C__ = []; M = C = R = 0
     for G in G_:
-        if Ct := cluster_C(root, get_exemplars(G.N_, r,c), r,c):  # refines, splits each G
-            C_, _m,_c,_r = Ct; C__+=C_; M+=_m; C+=c; R+=r
+        # skip when it's called from traced-edge, where G is PP, and PP.N_ is empty
```
boris-kz, line 496: it will be from vect_edge, where G is edge and G.N_ is PPs. So it's fine, we just need to assign rim to PPs

boris-kz: unpack:
def cross_comp(root, G_, m, c, r, nF='Nt'):  # sub+|agg+: refine by CC and exe -> cross_comp

    C__ = []; M = C = R = 0
    for G in G_:
        if Ct := cluster_C(G, get_exemplars(G.N_, r,c), r,c):  # refines, splits each G
    if C__:
        sum2F(C__, root, nF=nF)  # also pass M,C,R?
        if gv_((m+M) * ((c+C+wAgg) / (r+ R/len(C__)+ cAgg)) * ((len(G_)-1)*wL) - ave):
            E_ = get_exemplars([F2N(C) for C in Ct.N_], r,c)
            if Lt := comp_N_( proj_L_(combinations(E_,2), root, R), R):
                L_, TT, c, r, V = Lt
                if gv_(val_(TT,ttcN) * ((c+wcN)/(r+ccN)) * ((len(L_)-1)*wL) - ave):  # return +ve, store -ve gate Vs
                    e_ = get_exemplars({N for L in L_ for N in L.N_}, r,c)  # +ve Ls only
                    return cluster_N(root, e_,r,c)  # sum2G -> agg+

frame_2D_alg/agg_recursion.py
```diff
@@ -488,16 +489,18 @@ def add_H(H,h, root, fN=0):
...
+        if G.N_ and (Ct := cluster_C(G, get_exemplars(G.N_, r,c), r,c)):  # refines, splits each G
+            C_, _m,_c,_r = Ct.N_,Ct.m,Ct.c,Ct.r; C__+=C_; M+=_m; C+=_c; R+=_r
     if C__:  # spliced refined sub_G_s
-        sum2F(C__, root, nF)  # also pass M,C,R?
-        if gv_((m+M) * ((c+C+wAgg) / (r+ R/len(C__)+ cAgg)) * ((len(G_)-1)*wL) - ave):
+        sum2F(C__, root, nF=nF)  # also pass M,C,R?
+        if gv_((m+M) * ((c+C+wAgg+wX) / (r+ R/len(C__)+ cAgg+cX)) * ((len(G_)-1)*wL) - ave):
```
boris-kz, line 501: just wAgg, each function should be evaluated separately

frame_2D_alg/agg_recursion.py
```diff
@@ -512,16 +515,16 @@ def sum2G(ft_, fTT, root=None, init=1):  # core clustering function
...
     Ft_ += [sum2F(list(set(C_)), root.Ct) if C_ else CF()]  # add multiple root_ in Cs?
     G = comb_Ft(*Ft_, root, wTT=fTT)
-    N_ = G.N_; N=N_[0]; G.sub = N.sub+1 if G.L_ else N.sub; r=G.r
+    N_ = G.N_; N=N_[0]; G.sub = N.sub+1 if G.L_ else N.sub; r=G.r; Av=ave+avd
     if G.Lt:  # sub+
-        Lt = G.Lt; L_,lm,lc,lr = Lt.N_,Lt.m,Lt.c,Lt.r; Av=ave+avd  # no levR = 1/len(L_): represented by c
+        Lt = G.Lt; L_,lm,lc,lr = Lt.N_,Lt.m,Lt.c,Lt.r  # no levR = 1/len(L_): represented by c
         if gv_(lm * lc * wcC - Av * (lr+1+ccC)):  # mdecay(L_)-decay?
             xcomp(G, N_,lm,lc,lr)  # sub+, cross_comp
     if G.Bt:
```
boris-kz, line 523: if G.Bt:
        Bt = G.Bt; bd,bc,br = Bt.d,Bt.c,Bt.r; rroot = root.root if root.root else 0
        if N.typ!=1 and gv_(bd*bc * wcC - Av * (br+1+ccC)):  # no ddfork, eval len?
            cross_comp(Bt, [F2N(L) for L in Bt.N_],bd,bc,br, nF='Bt')

frame_2D_alg/agg_recursion.py
kwcckw: > it will be from vect_edge, where G is edge and G.N_ is PPs. So it's fine, we just need to assign rim to PPs

It's actually could be from trace_edge too when we sum PPs into G:

    G_ += [sum2G([(n_,'Nt',ntt,nc,r)]+([(l_,'Lt',ltt,lc,r)] if l_ else []), ttTrc, root)]

Those n_ are PPs, so their N_ is empty.

frame_2D_alg/agg_recursion.py
kwcckw: Okay right, thanks, so this should be cX and wX in sum2G:

        if gv_(lm * lc * wX - Av * (lr+1+cX)):  # mdecay(L_)-decay?
            xcomp(G, N_,lm,lc,lr)  # sub+, cross_comp
            
In frame_H:

            if gv_(m * c * wX - ave * (r+1+cX)):
                xcomp(Fr, N_, m,c,r)  # sub+, cross_comp

frame_2D_alg/agg_recursion.py
kwcckw: So this `gv_(bd*bc * wcC - Av * (br+1+ccC))` should be `gv_(bd*bc * wX - Av * (br+1+cX))` now.

boris-kz: right

frame_2D_alg/agg_recursion.py
boris-kz: that should be sum2F?

frame_2D_alg/agg_recursion.py
boris-kz: so that's replacing wAgg, cAgg, we don't need them?

frame_2D_alg/agg_recursion.py
kwcckw: I think it may not be fully compatible since we need Lt in sum2G, while sum2F sums only N_, probably there's some other issues too.

frame_2D_alg/agg_recursion.py
kwcckw: Yes, so cross_comp's w and c is no longer needed.

boris-kz: but this is a root_, not comp_: it calls both comp_ and clust_

frame_2D_alg/agg_recursion.py
boris-kz: There is no L_ in PPs?

kwcckw: Their L is actually added from boundary? So it's actually not empty in trace_edge.

            for N in N_:
                _N_ = [rN for B in N.B_ for rN in B.root_ if rN is not N]   # + node-mediated
                for _N in list(set(_N_)):  # share boundary or cores if lG with N, same val?
                    cT = tuple(sorted((N.id,_N.id)))
                    if cT in cT_: continue
                    cT_.add(cT)
                    L = comp_N(_N,N, r,_C,A=dy_dx, span=dist)  # current L is dPP
                    if val_(L.dTT,ttTrc,1)[1] * ((L.c+wTrc)/(r+cTrc)) > ave:
                        L_+=[L]; N.rim+=[L]; _N.rim+=[L]  # or PP typ=3 in F2N
                    if L_: lTT,lc,_ = sum_vt(L_,wTT=ttTrc)

boris-kz: G_ += [sum2G([(n_,'Nt',ntt,nc,r)]+([(l_,'Lt',ltt,lc,r)] if l_ else []), ttTrc, root)]

this is adding edge, not PP. PP may have rim but not L_. just add rim to CL (PP)?

kwcckw: > L = comp_N(_N,N, r,_C,A=dy_dx, span=dist)  # current L is dPP

From comp_N, it did add L into the PP.rim, and PP is a CN here since we converted it before parsing them to trace_edge.

boris-kz: PP should be CL, then in comp_N:

    if N.typ>1 and gv_(m* (c/r)* wN_ - ave*(r+cN_)):  # skip PPs, may comp Nts?

kwcckw: > PP should be CL, then in comp_N:
> 
> ```
> if N.typ>1 and gv_(m* (c/r)* wN_ - ave*(r+cN_)):  # skip PPs, may comp Nts?
> ```

Alright, so i guess we can try this first, but to add rim into CL, we need to add the entire Rt into CL now.

frame_2D_alg/agg_recursion.py
kwcckw: > but this is a root_, not comp_: it calls both comp_ and clust_

I guess we can remain the wAgg and cAgg, since this is another comp eval within the xcomp, and wAgg and cAgg should be meaningful here.

boris-kz: it's just a name, X is fine

frame_2D_alg/agg_recursion.py
boris-kz: so just fold sum2F in sum2G?
If there is no arg Lt, L_= set(N.rim for N in N_)

frame_2D_alg/agg_recursion.py
kwcckw: Okay, i thought we are talking about the nested cross_comp gv_ eval in xcomp since it's not a problem to have a nested eval here and  i don't see any problem if we remain the eval for that.

frame_2D_alg/agg_recursion.py
boris-kz: Feasible, with corrections:

**1. Fold sum_vt, not sum2F.** What callers duplicate is caller-side `sum_vt` + tuple packing (`(n_,nF,tt,c,r)`), not sum2F. sum2F has ~10 other roles (Rt from rims 122/284, H levs 152/465/576, centroids 362/402/428, Bt 807, Xt 536), and its typ-dispatch builds CN with root wiring + H concat (456-471) — wrong side effects for fork wrappers, which must stay bare CF. Keep sum2F as is; sum2G sums fork lists internally.

**2. Gating moves in or doubles.** Both callers eval on the sums *before* calling: cluster_N 310-314, trace_edge 761. If sum2G owns summation, it must own the eval and return None on fail — consistent with your 317 comment "or the below is in sum2G?". trace_edge's fail-branch reset (765: fin=0, root=root) then keys off None. Otherwise sums get computed twice.

**3. Rim-derived Lt: exact in trace_edge, needs filtering as a generic default.** In trace_edge, union of member rims = l_ exactly: 737 adds only +ve links to rims, and flood-fill closes over L_, so no member rim link exits the Gt. All l_ bookkeeping (742, 751, 753, 756) can go. But generically, comp_N 162 appends links to typ-3 rims regardless of sign, and rims hold links to non-members. So the default must be:
`L_ = {l for n in N_ for l in n.rim if l.m>0 and l.N_[0].root is l.N_[1].root}` (or both-in-N_ check). Your sketch `set(N.rim for N in N_)` is a set of lists — needs the union comprehension. The excluded remainder (external or -m, +d links) is the natural Bt default — answers your 762 "add Bt?" for free. cluster_N keeps passing explicit L_/B_: its L_ is density-filtered (300), rims contain rejected links.

**4. Fold exposes an Lt-scaling inconsistency.** trace_edge pre-scales `ltt *= lc/nc` (756) before packing; cluster_N stores raw fork sums and c-weights only in the gate (313). One convention must win — raw fork sums (cluster_N's) is the consistent one; if lc/nc weighting is principled it belongs in the gate, not in stored Lt.dTT.

Folded top section:

```python
def sum2G(N_, fTT, root=None, L_=None, B_=None, init=1):  # core clustering function, sums internal

    if L_ is None:  # default from rims, exact in trace_edge
        L_ = list({l for n in N_ for l in n.rim if l.m>0 and all(e in N_ for e in l.N_)})
    if not init:
        N_ += root.N_; L_ += root.L_
    Ft_ = []
    for f_, nF in zip((N_,L_,B_ or []), ('Nt','Lt','Bt')):
        if f_: tt,c,r = sum_vt(f_, wTT=fTT); Ft_ += [CF(N_=f_,nF=nF,dTT=tt,m=(vt:=val_(tt,wTT,1))[0],d=vt[1],c=c,r=r)]
        else:  Ft_ += [CF()]
    # gate here (from cluster_N 310-314 form), return None on fail
    # rest unchanged: Ct splice via sum2F, comb_Ft, sub+ evals
```

Caller diffs: trace_edge 763 → `sum2G(n_, ttTrc, root)` (drop l_ tracking); cluster_N 315 → `sum2G(N_, ttcN, CN(), L_, B_)` with 306-314 deleted, handle None; vect_edge 814 → `sum2G(G_, TT, T)` — note resumming from G_ yields post-comb_Ft dTT rather than the accumulated pre-comb TT from 767; marginally different, arguably more correct, and _TT/_C accumulation in trace_edge returns becomes gate-only.

boris-kz: anyway, vect_edge is fine, that sum2G is still for edges, not PPs

kwcckw: > so just fold sum2F in sum2G? If there is no arg Lt, L_= set(N.rim for N in N_)

The arg Lt related params is already inside sum2G's ft_ (ft_[1])? But why we need to fold sum2F into sum2G? This `L_= set(N.rim for N in N_)` could be done when we check with empty ft_[1]?

boris-kz: > anyway, vect_edge is fine, that sum2G is still for edges, not PPs

kwcckw: > anyway, vect_edge is fine, that sum2G is still for edges, not PPs

It's clustering PPs via the boundary B_? Because the input N_ here is PPs.

    def trace_edge(N_,_G_,_TT,_C, r,root):  # cluster contiguous shapes via PPs in edge blobs or lGs in boundary/skeleton?

boris-kz: yes, that sum2G is per cluster, not PP. and sum2G in vect_edge is per tile, not cluster

kwcckw: > yes, that sum2G is per cluster, not PP. and sum2G in vect_edge is per tile, not cluster

Yes, so the existing sum2G should be fine.

frame_2D_alg/agg_recursion.py
boris-kz: I unpacked it:

def cross_comp(root, G_, m, c, r, nF='Nt'):  # sub+|agg+: refine by CC and exe -> cross_comp

    C__ = []; M = C = R = 0
    for G in G_:
        if Ct := cluster_C(G, get_exemplars(G.N_, r,c), r,c):  # refines, splits each G
    if C__:
        sum2F(C__, root, nF=nF)  # also pass M,C,R?
        if gv_((m+M) * ((c+C+wN_) / (r+ R/len(C__)+ cN_)) * ((len(G_)-1)*wL) - ave):
            E_ = get_exemplars([F2N(C) for C in Ct.N_], r,c)
            if Lt := comp_N_( proj_L_(combinations(E_,2), root, R), R):
                L_, TT, c, r, V = Lt
                if gv_(val_(TT,ttcN) * ((c+wcN)/(r+ccN)) * ((len(L_)-1)*wL) - ave):  # return +ve, store -ve gate Vs
                    e_ = get_exemplars({N for L in L_ for N in L.N_}, r,c)  # +ve Ls only
                    return cluster_N(root, e_,r,c)  # sum2G -> agg+

kwcckw: I think we shouldn't need to get the exemplars before comp_N_ here? Because the rest of the Ns will be skipped, and comp should be default for all Ns?

        E_ = get_exemplars([F2N(C) for C in Ct.N_], r,c)
        if Lt := comp_N_( proj_L_(combinations(E_,2), root, R), R):

boris-kz: ok, those exemplars are for clustering, not comp_N_
also, we need gv_ per cluster_C too:

    for G in G_:
        if gv_(G.m * ((G.c*wcC) / (G.r*ccC)) * ((len(G_)-1)*wL) - ave):
            if Ct := cluster_C(G, get_exemplars(G.N_, r,c), r,c):  # refines, splits each G

kwcckw: So we can remove the xcomp's oF, and the existing xcomp should be replaced with cross_comp now.

boris-kz: 1. **Closure blocker on this path: centroid comparands never acquire rims.** C__ elements are typ=2 (sum2F 442: `fC → typ=2`; F2N reclasses but doesn't touch typ). comp_N appends rim only at 160-162 (`if N.typ==3`), and comp_N_ 121 (`if N.rim:`) skips Rt for empty rims. Chain: after comp, every N in L_ has rim=[] and Rt.m=0 → get_exemplars 245: n.w = vertical-only; 253 gate `N.Rt.m * N.c * wE > ...` fails for all → fallback 258 returns one arbitrary seed → cluster_N 284: `sum2F(N.rim, N.Rt)` on empty list → IndexError at sum2F 440 (`N = N_[0]`); even guarded, flood at 289 has no rims to walk. Pre-existing gap, but the old `Ct.N_` leak meant this path barely executed; your fix to C__ makes it the main line. Fix direction: comp_N 160 gate → `typ >= 2`, or append rims in comp_N_ for its comparands regardless of typ.

2. **Per-G gate: wrong count, inconsistent form.**
- `(len(G_)-1)*wL` is batch size — constant across the loop, and category-mismatched: the gated op is cluster_C over G's *internal* nodes. As written, a lone G with 100 nodes gets multiplier 0 (never refined) while 10 weak Gs all get a 9·wL boost. The relevant count is `len(G.N_)-1`.
- `(G.c*wcC)/(G.r*ccC)` is multiplicative; the convention everywhere else (95, 101, 314, 761, 766) is additive `(c+w)/(r+cost)`. Multiplicative collapses wcC/ccC to a constant ratio absorbable into ave, discarding the oF cost-offset semantics. Consistent form: `G.m * ((G.c+wcC)/(G.r+ccC)) * ((len(G.N_)-1)*wL) - ave`.

3. **len(G_) is stale in the second gate.** With the per-G gate, contributors ⊆ G_, and the op being gated scales with len(C__)² pairs. The wL count convention is comparand count (cf. 101: len(L_)-1), so this should be `(len(C__)-1)*wL`. The mismatch was latent before; the new gate makes it real.

4. **Dropping pre-comp get_exemplars: right call, one cost note.** The old exemplars-of-centroids was double selection — centroids are already exemplar-seeded compression. Per-pair selection now rests on proj_L_'s `dist < max=20` cap and comp_N_'s projection gate (109), which is finer. But `combinations(C__,2)` enumeration itself is O(n²) in len(C__) with no bound; fine if C__ per root stays small. Also correct: this is the first version where cross-G centroid comp actually runs over the full set — the old leaked `Ct` compared only the last G's centroids.

5. Minor, pre-existing: (a) second gate's `R/len(C__)` — R is a sum of per-Ct r's divided by centroid count; c-weighted mean (as at 312, 320) if you care. (b) `L_, TT, c, r, V = Lt` rebinds c,r to link-fork totals, so get_exemplars and cluster_N downstream receive comp-derived c,r rather than node totals — confirm intentional.

boris-kz: I didn't review these edits yet:

 **cross_comp:**
```python
def cross_comp(root, G_, m, c, r, nF='Nt'):  # sub+|agg+: refine by CC and exe -> cross_comp

    C__ = []; M = C = R = 0
    for G in G_:
        if gv_(G.m * ((G.c+wcC) / (G.r+ccC)) * ((len(G.N_)-1)*wL) - ave):  # additive costs, count = G members
            if Ct := cluster_C(G, get_exemplars(G.N_, r,c), r,c):  # refines, splits each G
                C__ += C_; M+=_m; C+=_c; R+=_r*_c  # splice refined sub_G_s
    if C__:
        R /= C  # c-weighted mean
        sum2F(C__, root, nF=nF)  # also pass M,C,R?
        if gv_((m+M) * ((c+C+wN_) / (r+R+cN_)) * ((len(C__)-1)*wL) - ave):  # count = comparands
            if Lt := comp_N_( proj_L_(combinations([F2N(C) for C in C__],2), root, R), R):
                L_, TT, c, r, V = Lt
                if gv_(val_(TT,ttcN) * ((c+wcN)/(r+ccN)) * ((len(L_)-1)*wL) - ave):  # return +ve, store -ve gate Vs
                    e_ = get_exemplars({N for L in L_ for N in L.N_}, r,c)  # +ve Ls only
                    return cluster_N(root, e_,r,c)  # sum2G -> agg+
```

**comp_N_ 111-113** — rim maintenance moves here (fixes the typ=2 gap; comparands here are always F2N'd or native CN):
```python
            L_+= [Link]; N_+= [_N,N]
            for n,_n in zip(Link.N_, Link.N_[::-1]): n.rim += [Link]; n.compared.add(_n)  # was typ=3 only in comp_N
```

**comp_N 160-162** — delete:
```python
    if N.typ==3:
        for n, _n in (_N,N),(N,_N):
```

**add_Lt 559** — keep its rims via manual append (its comp_N calls bypass comp_N_):
```python
            for L in pL_:
                l = comp_N(*L.N_, G.r,L.c,1, L.angl[0], L.span); L_ += [l]
                for n,_n in zip(l.N_, l.N_[::-1]): n.rim += [l]; n.compared.add(_n)
```

Consequences of the move, verify intent:
1. comp_N 145 (recursive member comp, typ<3) no longer appends parent-comp dLs to *member* rims — rims become strictly same-level: comp_N_ products + manual appends in trace_edge/add_Lt. I read the old member-rim pollution as accidental; if intended, keep a rL-gated append instead.
2. cluster_P 411 comp_N on raw CL centroids: unchanged (typ=2 never appended before) — this is why the fix can't be a typ>=2 gate in comp_N.
3. trace_edge 737 manual append: unchanged, no double-append. `compared` is write-only throughout, moved along for symmetry.
4. R passed to proj_L_/comp_N_ is now a mean, not a sum of per-Ct r's — old sum inflated lr in proj gates roughly by number of contributing Gs; gates loosen accordingly.
5. Left untouched: `L_, TT, c, r, V = Lt` rebinding c,r to link totals before get_exemplars/cluster_N — pending your confirmation.

kwcckw: Thanks, i will go through the code again later.

PR #222, opened 2026-07-31
7/31/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -712,8 +703,6 @@ def proj_TT(L, cos_d, dist, r, pTT, wTT, dec=1, fdec=0):     # accumulate L|N' p
     wTT = CoF.get().wTT*ttPrj
     for L in N.L_+ N.B_:  # proj C-to-N links:
         proj_TT(L, cos_d, dist, L.r+_r, iTT, wTT, dec); c+=L.c  # accum iTT internally
-    if hasattr(N.Ct,'Lt') and N.Ct.Lt:
-        for L in N.Ct.Lt.N_: proj_TT(L,cos_d,dist,L.r+_r,iTT,wTT,dec); c+=L.c
```
kwcckw, line 716: So now we replaced Nt with Ct, the projection from Ct.Lt is no longer needed right? Since it will be from N.L_ anyway.

frame_2D_alg/meta_code.py
```diff
@@ -102,6 +102,7 @@ def __init__(n, **kw):
         n.fin = kw.get('fin',0)  # clustered, temporary
         n.compared = kw.get('compared', set())
         n.typ = kw.get('typ',3)  # full comp
+        n.C_ = kw.get('C_',[])  # C roots
```
boris-kz, line 105: use root_?

frame_2D_alg/agg_recursion.py
boris-kz: right

frame_2D_alg/agg_recursion.py
```diff
@@ -157,7 +157,7 @@ def comp_H(_Nt,Nt, Link):  # tentative pre-comp, before comp N_?
         box = np.array([min(_y,y),min(_x,x),max(_y,y),max(_x,x)])
         angl = [A, np.sign(TT[1] @ ttN_[1])]
         L.yx=yx; L.box=box; L.span=span; L.angl=angl; L.kern=(_N.kern+N.kern)/2
-    if N.typ==3:
+    if isinstance(N, CN):  # all CN should add their rim here, that's including the converted PP or CL
```
boris-kz, line 160: so what are we excluding here?

kwcckw: I think mostly Ls since we may call comp_N in comp_F.

frame_2D_alg/meta_code.py
kwcckw: That `root_` is for Cs? When G is summed from Cs, we need `C_` to pack Cs' root_?

frame_2D_alg/agg_recursion.py
boris-kz: they should be adding rim too?

frame_2D_alg/meta_code.py
boris-kz: it works the same, C will have G roots instead of C roots?

kwcckw: This C may have C roots when there's subsequent cluster_C (when G.N_ are Cs)?

frame_2D_alg/agg_recursion.py
kwcckw: We only need that comp to get tFt's params, and why we need that rim?

frame_2D_alg/meta_code.py
boris-kz: cluster_C is replacing Nt, those Cs are refined / split Ns, which will be replacing root_ for their nodes?

kwcckw: Okay right thanks. Then we can reuse the same root_ here.

frame_2D_alg/agg_recursion.py
```diff
@@ -87,20 +87,20 @@ def cross_comp(root, G_, m, c, r, nF='Nt'):  # sub+|agg+: refine by CC and exe -
 
     C__ = []; M = C = R = 0
```
boris-kz, line 88: def cross_comp(root, G_, m, c, r, nF='Nt'):  # sub+|agg+: refine by CC and exe -> cross_comp

    C__ = []; M = C = R = 0
    for G in G_:
        if gv_(G.m * ((G.c*wcC) / (G.r*ccC)) * ((len(G.N_)-1)*wL) - ave):
            if Ct := cluster_C(G, get_exemplars(G.N_, r,c), r,c):  # refines, splits each G
    if C__:
        root.Nt = sum2F(C__,root, nF=nF, froot=2); L=len(C__); R/=L  # also pass M,C,R?
        if gv_((m+M) * (c*(C+wN_) / (r*(R+cN_))) * ((L-1)*wL) - ave):
            if Lt := comp_N_( proj_L_(combinations([F2N(C) for C in C__],2), root,R), R):
                L_, TT, c, r, V = Lt
                if gv_(val_(TT,ttcN) * (c*wcN /(r*ccN)) * ((L_-1)*wL) - ave):  # return +ve, store -ve gate Vs
                    e_ = get_exemplars({N for L in L_ for N in L.N_}, r,c)  # +ve Ls only
                    return cluster_N(root.Nt, e_,r,c)  # sum2G -> agg+

kwcckw: In sum2F:

    if root is not None:
        F.wTT = root.wTT
        
Those Cs are clustered from the original G_, do we need to sum them into root again (add2F(root,F,2) )?

boris-kz: the sum should actually replace the the root, how would you do that?

kwcckw: In sum2G:

    cross_comp(G, N_,lm,lc,lr)
    
This should be parsing G.Nt?

To replace the root (if root is Nt), we can use rroot = root.root; rroot.Nt = new Nt?

Or if we need to replace the actual root, i think we can return the new root from cross_comp function to replace it.

boris-kz: I guess it's Nt if there is one, is root always CN?

kwcckw: Before this the root of cross_comp is Ft, and we can replace it via root.root's nF assignment.

boris-kz: that doesn't work for Bt, so we need to pass nF:

cross_comp(G, N_,lm,lc,lr, nF='Nt')

setattr(root,nF, sum2F(C__,root, nF=nF, froot=2))
?

kwcckw: Why that's not working for Bt? If we parse Ft, the input nF could be Ft.nF?

Then the last line in cross_comp, it should be root (Ft) instead:

    return cluster_N(root, e_,r,c)

frame_2D_alg/agg_recursion.py
boris-kz: I guess we don't but it doesn't hurt?

frame_2D_alg/agg_recursion.py
boris-kz: we don't actually need to pass Ft, it will be replaced anyway?

kwcckw: Okay right, so we can parse root and nF, and the replacement can be done with the nF.

frame_2D_alg/meta_code.py
boris-kz: so we need to add this root_ replacement process?

kwcckw: > so we need to add this root_ replacement process?

Yes, we still need this root_ to pack G's C_'s root_, but the replacement should be in cluster_C only (when G becomes N), and that reset root_ anyway:

frame_2D_alg/agg_recursion.py
kwcckw: Okay, rim is default to CL, so we don't need F2N on those L, so it can be default. It shouldn't work on CF only.

frame_2D_alg/meta_code.py
boris-kz: Ok please edit

PR #223, opened 2026-08-01
8/1/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -99,9 +99,8 @@ def __init__(n, **kw):
         n.box = kw.get('box',np.array([np.inf, np.inf, -np.inf, -np.inf]))  # y0, x0, yn, xn
         n.sub = kw.get('sub',0)  # composition depth relative to top-composition peers?
         n.exe = kw.get('exe',0)  # exemplar, temporary
-        n.fin = kw.get('fin',0)  # clustered, temporary
-        n.compared = kw.get('compared', set())
```
kwcckw, line 103: I don't see we use compared anymore, and fin will be init as 0 each time before the clustering, so i think we can remove them.

frame_2D_alg/agg_recursion.py
```diff
@@ -312,9 +311,9 @@ def trans_cluster(G):
             r = (nr*nc + lr*lc + br*bc) /c  # br includes overlap?
             tt= (nt*nc + lt*lc + bt*bc) /c  # tentative
             if gv_(val_(tt*Ft.root.wTT*ttcN) * (c*wcN /(r*ccN)) * ((len(N_)-1)*wL) - ave):  # apply Fw_ and Fc_ in every eval_?
-                G_ += [sum2G(ft_,ttcN, CN())]; Gt_+= [[tt,c,r]]  # eval sub+
+                G_ += [sum2G(ft_,ttcN,Ft)]; Gt_+= [[tt,c,r]]  # eval sub+  (why we need empty CN root here?)
```
kwcckw, line 314: So the root should be Ft here? Why it's an empty CN?

frame_2D_alg/meta_code.py
boris-kz: ok, great

frame_2D_alg/meta_code.py
```diff
@@ -99,9 +99,8 @@ def __init__(n, **kw):
...
         n.typ = kw.get('typ',3)  # full comp
+        n.C_ = kw.get('C_',[])  # C roots
```
boris-kz, line 103: this is root_?

frame_2D_alg/agg_recursion.py
boris-kz: yes, empty CN is probably obsolete

frame_2D_alg/agg_recursion.py
```diff
@@ -88,7 +88,7 @@ def cross_comp(root, G_, m, c, r, nF='Nt'):  # agg+: refine by CC,exe -> cross_c
     C__ = []; M = C = R = 0
```
boris-kz, line 88: Trying Opus 5 on a conceptual level:  
https://claude.ai/share/4f6e1f5e-1ad4-4a08-ba41-09555ced8179

frame_2D_alg/agg_recursion.py
```diff
@@ -429,10 +428,10 @@ def cluster_P(_C_, _c, root):  # multi-seed mean shift: parallel centroid refine
         FV_(CoF.get(), dCt.dTT, dCt.c, dCt.r)
     return out_
 
-def sum2F(N_, root=None, m_=[],d_=[], merge=0, froot=0, nF=None):  # -> CF/CL/CC/CN
+def sum2F(N_, root=None, m_=None,d_=None, merge=0, froot=0, nF=None):  # -> CF/CL/CC/CN
 
     c_ = np.array([n.c for n in N_], dtype=float); C = c_.sum(); w_ = c_/C; N = N_[0]
-    fC = np.any(m_)
+    fC = m_ is not None  # very unlikely , but if all input m_ is zeros, fC maybe false
```
boris-kz, line 434: if all input m_ is zeros, sum2F should not be called?

frame_2D_alg/agg_recursion.py
```diff
@@ -372,7 +371,7 @@ def cluster_C(Ft, E_,_r,_c):  # form centroids by clustering exemplar surround v
         else: out_+=C_; break  # converged
     if out_:
         for n in [N for C in out_ for N in C.N_]:  # exemplar V + sum n match_dev to Cs, m* ||C rvals:
```
boris-kz, line 373: so we don't actually need reassign n.root_ here, it should always be empty before cluster_C?

frame_2D_alg/agg_recursion.py
```diff
@@ -372,7 +371,7 @@ def cluster_C(Ft, E_,_r,_c):  # form centroids by clustering exemplar surround v
...
-            n.exe = (n.d if n.typ==1 else n.m) + sum(rt[1] for rt in n.root_) - ave
+            n.exe = ((n.d if n.typ==1 else n.m) + sum(rt[1] for rt in n.root_) - ave)>0
```
boris-kz, line 374: n.exe = (n.d if n.typ==1 else n.m) + sum(rt[1] for rt in n.root_) > ave

frame_2D_alg/agg_recursion.py
kwcckw: Do you mean this init?

I was thinking if the n could be G summed from prior Cs? So their root_ maybe packing their n_'s root_.

frame_2D_alg/agg_recursion.py
kwcckw: I think that's possible in cluster_P because it's not conditional?

frame_2D_alg/meta_code.py
kwcckw: Yes, sorry, this should be removed.

frame_2D_alg/agg_recursion.py
boris-kz: These Cs won't be directly clustered, only their roots

frame_2D_alg/agg_recursion.py
boris-kz: it's prunned in the end:

    for i, _C in enumerate(C_):

frame_2D_alg/agg_recursion.py
kwcckw: Okay, but we need to init `_root_` anyway, so we can init `root_` here?

frame_2D_alg/agg_recursion.py
kwcckw: Okay right thanks. We need that empty C to preserve the index in cluster_P.

frame_2D_alg/agg_recursion.py
boris-kz: I meant that yesterdays discussion about root_ replacement, it's not relevant

frame_2D_alg/agg_recursion.py
kwcckw: > data destruction before evaluation.

So that's why we have projection to perform the cheap eval before the expensive selective process right?

boris-kz: That's a more coarse process selection, we have a much finer grain cluster selection before that

frame_2D_alg/agg_recursion.py
kwcckw: I tracked it with claude:

image

So the Gs input in cross_comp might be summed from Cs, where their root_ is not empty, which mean we should reset them.

boris-kz: It doesn't matter, those Cs and their root are preserved as individual nodes in the new G, so their root_ is not affected.

frame_2D_alg/agg_recursion.py
kwcckw: Those cluster selections are referring to the ave-based gv_ or non gv_ eval?

boris-kz: cluster selection is specifically in cluster_N and cluster_C:

            if gv_(val_(tt*Ft.root.wTT*ttcN) * (c*wcN /(r*ccN)) * ((len(N_)-1)*wL) - ave):  # apply Fw_ and Fc_ in every eval_?
                G_ += [sum2G(ft_,ttcN, Ft)]; Gt_+= [[tt,c,r]]  # eval sub+

this is an eval for resolution reduction, which is selectively destructive

kwcckw: > That's a more coarse process selection, we have a much finer grain cluster selection before that

But from the code, the more coarse projection is before the finer clustering process? It should be coarse to fine selection?  Why the finer is before the coarse selection?

boris-kz: More coarse is tile selection, but we need a seed tile for that, so overall process is search scope expansion: fine to coarse.

PR #224, opened 2026-08-02
8/2/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -819,7 +821,7 @@ def fill_frame(_iy,_ix, elev, T):  # expand level_frame from pixel-level seed ti
                             if not (0<=_y<Ly and 0<=_x<Lx) or frame[_y,_x] is not None: continue  # outside frame or checked
                             if gv_(PV__[_y,_x] - ave):  # accumulated from all adjacent tiles
                                 iy = _y**elev; ix = _x**elev
-                                if _T := frame_H(image, iy,ix, Y,X, rV, max_elev=elev):  # agg+ new tile to the current level
+                                if _T := frame_H(image, iy,ix, Y,X, rV, max_elev=elev):  # agg+ new tile to the current level
```
kwcckw, line 824: I think we should restructure the code in frame_H, specifically here because there are several bugs.

When the max_elev is elev and when elev==1, the subsequent `frame_H`  always return empty `_T` because `while T and elev < max_elev:` in frame_H always false (elev is not < max_elev).

kwcckw: The second problem is we pack `_T` into `T_` only if the `_T` is not empty from frame_H. So if `_T` is not empty, the recursion is endless, but if `_T` is empty, `T_` is always empty since we never fill it.

A better explanation from claude:

`The call site is its own only supplier. T_ is fed solely by _T := frame_H(...), and frame_H returns non-empty solely when its fill_frame returned non-empty. So:`

`_T non-empty requires the child to have succeeded, which requires a grandchild to have succeeded, … — an unbounded chain of successful descendants. Unsatisfiable.`
`_T empty is therefore the only reachable outcome, so T_ stays empty at every level, forever.`

frame_2D_alg/agg_recursion.py
```diff
@@ -92,7 +92,7 @@ def cross_comp(root, G_, m, c, r, nF='Nt'):  # agg+: refine by CC,exe -> cross_c
                 C_, _m,_c,_r = Ct.N_,Ct.m,Ct.c,Ct.r
                 C__ += C_; M+=_m; C+=_c; R+=_r  # splice refined sub_G_s
     if C__:
-        setattr(root,nF, sum2F(C__,root, nF=nF, froot=2)); L=len(C__); R/=L  # also pass M,C,R?
+        setattr(root,nF, sum2F(C__,root, nF=nF, froot=2)); L=len(C__); R/=L  # also pass M,C,R? (existing root.H will be replaced by C_'s H?)
```
boris-kz, line 95: yes, we have this H packing in sum2F:

        if typ==3:  # frame?

frame_2D_alg/agg_recursion.py
```diff
@@ -718,8 +718,7 @@ def trace_edge(N_,_G_,_TT,_C, r,root):  # cluster contiguous shapes via PPs in e
             cT_.add(cT)
             dy_dx = _N.yx-N.yx; dist = np.hypot(*dy_dx)  # Rc = r+ (N.r+_N.r)/2
             L = comp_N(_N,N, r,_C,A=dy_dx, span=dist)  # current L is dPP
-            if val_(L.dTT,ttTrc,1)[1] * ((L.c+wTrc)/(r+cTrc)) > ave:
-                L_+=[L]; N.rim+=[L]; _N.rim+=[L]  # or PP typ=3 in F2N
+            if val_(L.dTT,ttTrc,1)[1] * ((L.c+wTrc)/(r+cTrc)) > ave: L_+=[L]
```
boris-kz, line 721: we need rim in PPs too

frame_2D_alg/agg_recursion.py
boris-kz: That makes no sense to me. We always have T:

    while T and elev < max_elev:

frame_2D_alg/agg_recursion.py
kwcckw: Yes,  rim packing is default in comp_N now, so it should be already added?

frame_2D_alg/agg_recursion.py
kwcckw: Okay right, thanks. They should be the same N.H, so the whole H will be reconstructed.

frame_2D_alg/agg_recursion.py
boris-kz: ok right

frame_2D_alg/agg_recursion.py
```diff
@@ -797,7 +796,10 @@ def PP2N(PP):
                         G_,TT,c,R = trace_edge([F2N(N) for N in N_], G_,TT,c,3,T); C += c  # flatten B_-mediated Gs
     if G_:
         FV_(CoF.get(), TT, C,1)
-        return sum2G([[G_,'Nt',TT,C,1]], TT, T) # c,R?
+        T = sum2G([[G_,'Nt',TT,C,1]], TT, T) # c,R?
+        A = np.sum([l.angl[0] for G in G_ for l in G.L_], axis=0)
+        T.angl = [A, np.sign(T.dTT[1] @ ttVct[1])]  # we need angl for projection
```
boris-kz, line 801: thanks

frame_2D_alg/agg_recursion.py
kwcckw: > We always have T:

This T is from vect_edge, but not from frame_H, and T from frame_H is actually Fr:

        if _T := frame_H(image, iy,ix, Y,X, rV, max_elev=elev):  # agg+ new tile to the current level 
            T_ += [_T]; _T_ += [(_T,_y,_x)]; frame[_y,_x] = _T
            
And we only pack `_T` into T_, but not T.

boris-kz: Who cares, we always have seed T

kwcckw: Okay, so we need to change the T packing into default, so that the returned tile_ includes the input T.

boris-kz: It should not be default

kwcckw: I think seed tile T should be included in tile_? Since it's part of the grid's T too.

boris-kz: it's conditional:

        frame[cy,cx] = T; T_=[]; __T_=[(T,cy,cx)]  # output, eval wave
        while __T_:
            _T_ = []  # next wave

kwcckw: But we never pack this seed tile `T` in the loop? We only pack `_T` from the nested frame_H:

      if _T := frame_H(image, iy,ix, Y,X, rV, max_elev=elev):  # agg+ new tile to the current level 
           T_ += [_T]; _T_ += [(_T,_y,_x)]; frame[_y,_x] = _T
           
So the conditional eval is for adjacent tiles instead of T.

boris-kz: Adjacents are inititialized as seed. We've been through this a bunch of times

frame_2D_alg/agg_recursion.py
boris-kz: So angle etc should be added to G in sum2G:

    if G.Lt:  # sub+
        add_Lt(G, Lt, wTT)
            cross_comp(G, N_,lm,lc,lr, nF='Nt')  # sub+, cross_comp

frame_2D_alg/agg_recursion.py
kwcckw: Yes, but the problems are really due to different reasons:
 
1. When elev ==1, and we parse max elev = elev to adjacents' frame_H,  this `while T and elev < max_elev` will never be true for adjacents.  If this is intended, then 2. below must be resolved, else we won't get any `_T` from adjacents' frame_H.

2. The tile_ doesn't pack the seed tile T in fill_frame,which causes it always return empty Fr from adjacents.

I verified with Claude and it points the the similar problems.

frame_2D_alg/agg_recursion.py
kwcckw: This is a special case because we don't have Ls between Gs here, so the resulting T shouldn't get any Lt. But we need the T.angl, so it should be summed externally here?

boris-kz: ok but we need the above in sum2G anyway, just add for G in G_ loop if fT?

frame_2D_alg/agg_recursion.py
boris-kz: 2 is bullshit, you should know better than that

frame_2D_alg/agg_recursion.py
kwcckw: This is a special case, i think we shouldn't change the core sum2G? So i think the G_ loop to get angl from G.L_ should be done externally here.

boris-kz: G needs angl and mang regardless

def add_Lt(G, Lt,wTT, fG=0):  # fG if G/tile has no L_

    L_ = Lt.N_
    angl, mang = np.zeros(2), 0; L_ = [l for n in G.N_ for l in n.L_] if fG else L_
    for l in L_: angl += l.angl
    G.mang = np.mean(comp_A(angl, l.angl[0])[0] for l in L_)
    G.angl = [angl, np.sign(G.dTT[1] @ ttcN[1])]

kwcckw: We already have add_Lt in comb_Ft：

      def comb_Ft(Nt, Lt, Bt, root,wTT):  # from sum2G, default Nt
      
          G = CN(Nt=Nt,Lt=Lt,Bt=Bt,root=root); Nt.root=G; Lt.root=G; Bt.root=G
          dF_ = []
              if Ft: dF_ += [comp_F(T, Ft, root.r,G)]; T.dTT,T.c,T.r = sum_vt([T,Ft],wTT=wTT)  # Bt*brrw?
              else:  dF_ += [CF()]
          add2F(G,T, merge=2)
          if Lt: add_Lt(G, Lt,wTT)
          G.m,G.d = val_(G.dTT,wTT,1)  # recompute m and d after add_Nt and add_Lt above
          return G
          
 And this doesn't resolve the issue since we didn't parse any Lt to form T, so `if G.Lt: ` is false and the additional add_Lt won't be executed.

boris-kz: ok that add_Lt is actually in comb_Ft

boris-kz: so we need to have this fG fork in comb_Ft instead of add_Lt

frame_2D_alg/agg_recursion.py
kwcckw: Yes, that's the prior agreed structure because seed T is not a Fr. But there's a problem with the frame_H recursion here because we get Fr only if `T_` is not empty, but `T_` only packs Fr, so we will never get any `_T` here.

frame_2D_alg/agg_recursion.py
kwcckw: So if fG, it add N_'s L_ params such as angl and mang into G?

frame_2D_alg/agg_recursion.py
boris-kz: it starts with T, not T_. LLMs can find problems just fine, try solving them

frame_2D_alg/agg_recursion.py
boris-kz: yes, that should be in comb_Ft

boris-kz: if Lt: add_Lt(G, Lt,wTT)
    else: 
        L_ = [l for n in G.N_ for l in n.L_]; angl, mang = np.zeros(2), 0
        for l in L_: angl += l.angl
        G.mang = np.mean(comp_A(angl, l.angl[0])[0] for l in L_)
        G.angl = [angl, np.sign(G.dTT[1] @ ttcN[1])]

kwcckw: Thanks, this looks better.

frame_2D_alg/agg_recursion.py
kwcckw: Yes, but that T is not included in T_, so that's why there's the suggestion to pack  seed T into T_ above, but you declined it.

boris-kz: it's included in `__T_` that's all we need

        frame[cy,cx] = T; T_=[]; __T_=[(T,cy,cx)]  # output, eval wave
        while __T_:
            _T_ = []  # next wave
                    dy,dx = T.box[2:] -T.box[:2]
                                iy = _y**elev; ix = _x**elev
                                if _T := frame_H(image, iy,ix, Y,X, rV, max_elev=elev):  # agg+ new tile to the current level
                                    T_ += [_T]; _T_ += [(_T,_y,_x)]; frame[_y,_x] = _T
                                else: frame[_y,_x] = 0
            __T_ = _T_
        if T_:
                return T_,C,R
        return [],0,0

    Fr = []
    while T and elev < max_elev:
        if tile_:

frame_2D_alg/agg_recursion.py
boris-kz: we actually need comp_pL_ instead of add_Lt now:

    if Lt: comp_pL_(G, Lt,wTT)
    L_ = Lt.N_ if Lt else [l for n in G.N_ for l in n.L_]
    angl, mang = np.zeros(2), 0  # in all Gs or Ts only?
    for l in L_: angl += l.angl
    G.mang = np.mean(comp_A(angl, l.angl[0])[0] for l in L_)
    G.angl = [angl, np.sign(G.dTT[1] @ ttcN[1])]

frame_2D_alg/agg_recursion.py
kwcckw: We include it in `__T_` to get the adjacent Fr from this line `_T := frame_H(image, iy,ix, Y,X, rV, max_elev=elev)`, but in that adjacent call, it faces the same problem since it needs tile_ to be filled from `tile_,C,R = fill_frame(iY, iX, elev, T)` to return Fr. This is actually a cyclic issue.

boris-kz: It is the first adjacent. The cycle you are talking about is the order of expansion, and we always have the seed

PR #225, opened 2026-08-03
8/3/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -833,6 +834,7 @@ def fill_frame(_iy,_ix, elev, T):  # expand level_frame from pixel-level seed ti
     global ave,avd; aTT=oTT=np.zeros((2,9)); aH,oH = [],[]  # regime refs across levs / ffeedback
     Fr = []
     T = vect_edge( frame_blobs_root( comp_pixel( image[iY:iY+Ly**elev, iX:iX+Lx**elev]), rV), iY,iX,Ly,Lx, rV)  # base process
+    Fr = T  # i guess we need this as base, and replace with the expansion below?
```
kwcckw, line 837: > It is the first adjacent. The cycle you are talking about is the order of expansion, and we always have the seed

Since we always have the seed, Fr should be init as T from vect_edge(as base), and to be replaced by subsequent expanded tile_?

frame_2D_alg/agg_recursion.py
```diff
@@ -813,7 +814,7 @@ def fill_frame(_iy,_ix, elev, T):  # expand level_frame from pixel-level seed ti
             for T, y,x in __T_:  # eval all wave tiles
                 if gv_(val_(T.dTT*T.wTT*ttFrm) * ((T.c+wFrm)/(T.r+cFrm)) - ave):
                     dy,dx = T.box[2:] -T.box[:2]
-                    pTT, pc = proj_N(T, np.hypot(dy,dx), np.array([dy,dx]), elev,T.c)  # no proj r?
+                    pTT, pc = proj_N(T, np.hypot(dy,dx), np.array([dy,dx]), elev,T.c)  # no proj r?  (We need L_ in T for the projection here)
```
kwcckw, line 817: Looks like we need T.L_ in proj_N:

    for L in N.L_+ N.B_:  # proj C-to-N links:
        proj_TT(L, cos_d, dist, L.r+_r, iTT, wTT, dec); c+=L.c  # accum iTT internally
    for L in N.rim:
        proj_TT(L, cos_d,dist,L.r+_r,eTT,wTT,dec); c+=L.c

Which we need to fill G with Ns' L_, as in my line 532 above?

        if not Lt: G.L_ = L_  # for T projection

frame_2D_alg/agg_recursion.py
```diff
@@ -501,7 +501,6 @@ def sum2G(ft_, fTT, root=None, init=1):  # core clustering function
     N_ = G.N_; N=N_[0]; G.sub = N.sub+1 if G.L_ else N.sub; r=G.r; Av=ave+avd
     if G.Lt:  # sub+
         Lt = G.Lt; L_,lm,lc,lr = Lt.N_,Lt.m,Lt.c,Lt.r  # no levR = 1/len(L_): represented by c
-        comp_pL_(G, Lt, wTT, fG=1)
```
kwcckw, line 504: This is actually redundant now since we already have comp_pL in comb_Ft.

Also in comp_pL:

      def comp_pL_(G, Lt,wTT):
      
This eval ` Lt.m * wN > ave*(cN+Lt.r)` should be called before calling the function?

boris-kz: we probably just need to use gv_() for each comp_N here, with some coef per pL_
so it can be unpacked in comb_Ft

boris-kz: if Lt:
        L_ = Lt.N_; L_,pL_ = [],[]; [L_.append(L) if L.typ==1 else pL_.append(L) for L in L_]
    else: L_ = [l for n in G.N_ for l in n.L_]; G.L_ = L_  # no sum2F?
    angl = np.zeros(2)  # in all Gs or Ts only?
    for l in L_: angl += l.angl[0]
    G.mang = np.mean([comp_A(angl, l.angl[0])[0] for l in L_])
    G.angl = [angl, np.sign(G.dTT[1] @ ttcN[1])]
    G.m,G.d = val_(G.dTT,wTT,1)  # recompute after add_Nt and comp pL_
    return G

frame_2D_alg/agg_recursion.py
```diff
@@ -525,10 +524,12 @@ def comb_Ft(Nt, Lt, Bt, root,wTT):  # from sum2G, default Nt
     add_Nt(G)  # add kern,ext, doesn't affect comp_F
     if Lt: comp_pL_(G, Lt,wTT)
     L_ = Lt.N_ if Lt else [l for n in G.N_ for l in n.L_]
-    angl, mang = np.zeros(2), 0  # in all Gs or Ts only?
-    for l in L_: angl += l.angl
-    G.mang = np.mean(comp_A(angl, l.angl[0])[0] for l in L_)
-    G.angl = [angl, np.sign(G.dTT[1] @ ttcN[1])]
+    if L_:  # skip singleton G from trace_edge because both Lt and N.L_ are empty
```
boris-kz, line 527: we don't need to skip, it works anyway

frame_2D_alg/agg_recursion.py
boris-kz: it really doesn't matter, Fr is always recomputed below:

Fr = sum2F(N_); m,c,r = Fr.m,Fr.c,Fr.r

frame_2D_alg/agg_recursion.py
kwcckw: > we don't need to skip, it works anyway

Okay, and If there's no L_, `G.mang = np.mean([comp_A(angl, l.angl[0])[0] for l in L_])` returns NaN since `np.mean([])` gets NaN, so it's better to skip them?

frame_2D_alg/agg_recursion.py
kwcckw: That Fr is computed only if elev < max_elev, and there's a deeper nested frame_H in fill_frame, so it may not be true all the time:

      while T and elev < max_elev:
              if tile_:

frame_2D_alg/agg_recursion.py
boris-kz: ok

frame_2D_alg/agg_recursion.py
boris-kz: right

frame_2D_alg/agg_recursion.py
boris-kz: how exacly it won't be true?

kwcckw: This is currently true for the first frame_H call only (when ffb == 1) because in the first adjacent:

    if _T := frame_H(image, iy,ix, Y,X, rV, max_elev=elev)
    
When elev = 1, max_lev = elev = 1, this `elev < max_elev` will never be true, and this led to empty T_. When the returned tile_ (or T_) is returned empty from fill_frame, no new Fr will be summed.

boris-kz: Ok, that's because the seed tile elev after processing should be +=1. That means:

if _T := frame_H(image, iy,ix, Y,X, rV, max_elev=elev+1)

this is a real bug, not all this bullshit we talked about in the last two days.

kwcckw: Yes, and there's another bug that i mentioned earlier on the structure of `_T`:

`_T` is always empty because it needs the nested call to have non empty `_T`, but that's not possible without the base Fr:

from claude:

The single reason, in both: Fr is non-empty only if some child was non-empty. Nothing in the function can produce a non-empty return without a successful child, so the deepest call always returns [] and it propagates all the way up. Recursion depth is irrelevant to that induction — only a base case that returns something without consulting a child breaks it.

boris-kz: >Fr is non-empty only if some child was non-empty.

that's how it should be. 
the base Fr is seeded by the first tile that we get from vect_edge.
You and Claude seem to be unable to comprehend that this process is bottom-up:
There is no Fr before seed T, by definition

kwcckw: Yes, but the current problem is the bottom packing is relies on Fr instead of seed T, which that's why i suggest to init Fr as T above?

boris-kz: It doesn't depend on Fr

boris-kz: Ok, that T_ inclusion should be from last wave, not next wave:

        while __T_:
            _T_ = []  # next wave
            for _T, y,x in __T_:  # last wave
                if gv_(val_(_T.dTT*_T.wTT*ttFrm) * ((_T.c+wFrm)/(_T.r+cFrm)) - ave):
                    T_ += [_T]; dy,dx = _T.box[2:] -_T.box[:2]
                    pTT, pc = proj_N(_T, np.hypot(dy,dx), np.array([dy,dx]), elev,_T.c)  # no proj r?
                    if gv_(val_(pTT*T.wTT*ttFrm) * ((pc+wFrm)/(_T.r+elev+cFrm)) - ave):  # +ve, no uncertainty projection yet
                        proj_focus(PV__,y,x,_T)  # radiate projected value, make directional per L_?
                                iy = _y**elev; ix = _x**elev
                                if T := frame_H(image, iy,ix, Y,X, rV, max_elev=elev+1):  # agg+: incr to the current level
                                    _T_ += [(T,_y,_x)]; frame[_y,_x] = T
                                else: frame[_y,_x] = 0

frame_2D_alg/agg_recursion.py
```diff
@@ -813,7 +814,7 @@ def fill_frame(_iy,_ix, elev, T):  # expand level_frame from pixel-level seed ti
             for T, y,x in __T_:  # eval all wave tiles
                 if gv_(val_(T.dTT*T.wTT*ttFrm) * ((T.c+wFrm)/(T.r+cFrm)) - ave):
                     dy,dx = T.box[2:] -T.box[:2]
-                    pTT, pc = proj_N(T, np.hypot(dy,dx), np.array([dy,dx]), elev,T.c)  # no proj r?
+                    pTT, pc = proj_N(T, np.hypot(dy,dx), np.array([dy,dx]), elev,T.c)  # no proj r?  (We need L_ in T for the projection here)
                     if gv_(val_(pTT*T.wTT*ttFrm) * ((pc+wFrm)/(T.r+elev+cFrm)) - ave):  # +ve, no uncertainty projection yet
                         proj_focus(PV__,y,x,T)  # radiate projected value, make directional per L_?
```
boris-kz, line 819: In proj_focus, we need V * dec * 1|1.4 * mangl(tile.angl, cross-tile angl)?

kwcckw: Before this we are using a `# a = aspect = dy/dx, affects axial directions only`, so mang should replace that?

frame_2D_alg/agg_recursion.py
kwcckw: Yes, this should resolve the issue, but seed tile could be included in T_ if the gv_ eval is true now

frame_2D_alg/agg_recursion.py
boris-kz: something like this, Chat:

`rA_`, `rd_`, axial `abs(cos)`, and the mean-preserving blend are right. But the current upstream contract for `mang` is not.

Use these changes:

```python
def proj_focus(PV__, y,x, tile, elev):

    H,W = PV__.shape
    Dec = decay ** (
        np.hypot(*(tile.box[2:] - tile.box[:2]))
        / np.hypot(H**elev, W**elev)
    )
    ...
    A = tile.angl[0] if tile.angl else None
    if A is not None and (An := np.hypot(*A)) > eps:
        mA_ = np.abs(rA_ @ A) / (An * rd_)
        mang = max(0, 2 * tile.mang - 1)  # temporary, with current directed mang
        w_ = 1 + mang * (mA_ / mA_.mean() - 1)
    else:
        w_ = np.ones(8)
    ...
    pV__ = V * Dec ** (n * rd_) * rd_ * w_
```

And pass `elev` from `fill_frame`:

```python
proj_focus(PV__, y,x, _T, elev)
```

Specific issues:

* `elev` is not currently in `proj_focus` scope.
* `np.hypot(Ly,Lx)**elev` is not the diagonal of an elevation-scaled frame. That is `np.hypot(Ly**elev, Lx**elev)`.
* `rd_` belongs in the decay exponent because diagonal rays travel farther. The separate `* rd_` then represents the extra projected span.
* A zero resultant angle must produce isotropic weights, not `1 - tile.mang`.

The important mismatch is that current `G.mang` is computed from directed `comp_A`, whose incoherent baseline is roughly `0.5`; your formula assumes incoherence is `0`. Without the `2*mang-1` conversion, a directionally incoherent tile still gets substantial anisotropy, which propagates through `pV__ → PV__ → next-tile selection`.  

That conversion is only provisional. Since this projection is axial, the clean solution is to compute both `G.angl` and `G.mang` axially upstream; the current signed sum of link vectors can cancel opposite directions that belong to the same axis.

kwcckw: So `rA_` and `rd_` are radial angle and distance? Each rA_ is the 8 directional angle (0,45,90,135,180..) and rd_ is their distance (1,1.4,1,1.4...)?

boris-kz: right

kwcckw: So rA_ should be in radian right?

Also this eval `if A is not None and (An := np.hypot(*A)) > eps:` should be just ` (An := np.hypot(*A)) > eps`, and if this eval is false, i think the rest should be skipped?

boris-kz: Opus 5: 
Three issues:

**1. The Dec ratio is inverted.** Numerator is the tile's content span, denominator is the tile pitch at elev — so a small blob in a big cell gets exponent ≪1, `Dec≈1`, slow decay; a pattern filling the cell gets exponent 1, fastest decay. Backwards: predictive range scales with the pattern's own extent. One step is `pitch` pixels, so in units of the pattern's coherence length that step costs `pitch/span`:

```python
    Dec = decay ** (np.hypot(H**elev, W**elev) / (np.hypot(*(tile.box[2:]-tile.box[:2])) + eps))
```
Elev-consistency is now right either way — both terms scale as `**elev`, ratio ~1 for a full-span tile.

**2. The trailing `* rd_` double-counts distance.** 1.4 is already in `Dec**(n*rd_)`, correctly, as longer distance. Multiplying by `rd_` again re-adds the original 40% diagonal boost you're removing. Drop it:
```python
        pV__ = V * Dec**(n*rd_) * w_
```

**3. `mang` is still directed while `mA_` is axial.** `max(0, 2*mang-1)` = clamped `mean(cos)` over links. A tile with a coherent axis but sign-split links scores ~0 → isotropic projection, exactly when the axis is most reliable. The sign-arbitrariness that forces `abs()` on `rA_ @ A` forces it here too:
```python
    G.mang = np.mean([abs(2*comp_A(angl, l.angl[0])[0] - 1) for l in L_])  # axial
```
Then `mang = tile.mang` directly, no re-centering.

Deferred, but it's what makes (3) only a partial fix: `angl` is a plain sum of arbitrarily-signed link angles, so the reference axis A is itself deflated by cancellation — `mA_` and `mang` are both measured against a possibly-degenerate vector, and `An > eps` catches only total cancellation. A true axial mean needs doubled-angle summation (accumulate `cos2θ, sin2θ`, then halve), which changes `angl`'s representation everywhere it's used.

```python
def proj_focus(PV__, y,x, tile, elev):  # radial accum of projected focus value in PV__

    m,d = val_(tile.dTT, tile.wTT*ttFrm, 1); c = tile.c  # m,d,n = tile.m, tile.d, tile.c  # add r?
    V = (m-ave + d-avd) * c
    H,W = PV__.shape  # = win__, tile pitch = H|W **elev
    Dec = decay ** (np.hypot(H**elev,W**elev) / (np.hypot(*(tile.box[2:]-tile.box[:2])) +eps))  # per-step decay, in units of tile span
    rA_ = np.array([
    (-1,-1), (-1,0), (-1,1),
    ( 0,-1),         ( 0,1),
    ( 1,-1), ( 1,0), ( 1,1)
    rd_ = np.hypot(rA_[:,0], rA_[:,1])  # rel dist: 1|1.4
    A = tile.angl[0] if tile.angl else None
    if A is not None and (An := np.hypot(*A)) > eps:
        mA_ = np.abs(rA_ @ A) / (An * rd_)  # axial alignment of rim dir to link_ orientation, 0:1
        w_ = 1 + tile.mang * (mA_ / mA_.mean() - 1)  # mean-preserving, isotropic if mang=0
    else:
        w_ = np.ones(8)
    n = 1  # radial distance
    while y-n>=0 and x-n>=0 and y+n<H and x+n<W:  # rim is within frame
        pV__ = V * Dec**(n*rd_) * w_  # diag dist is 1.4 * axial
        if np.max(pV__) < ave:
            break  # no direction is worth extending
        rim_coords = np.array([
        (y-n,x-n), (y-n,x), (y-n,x+n),
        (y, x-n),           (y, x+n),
        (y+n,x-n), (y+n,x), (y+n,x+n)
        ], dtype=int)
        row,col = rim_coords[:,0], rim_coords[:,1]
        PV__[row,col] += pV__  # in-place accum pV to rim
        n += 1
```

Call site: `proj_focus(PV__, y,x, _T, elev)`.

And line 534, so `mang` is axial like `mA_`:
```python
        G.mang = np.mean([abs(2*comp_A(angl, l.angl[0])[0] -1) for l in L_])  # |cos|, link angl signs are arbitrary
```

kwcckw: On this part:

    rA_ = np.array([
        (-1,-1), (-1,0), (-1,1),
        ( 0,-1),         ( 0,1),
        ( 1,-1), ( 1,0), ( 1,1)
        
This is actually the dy and dx ?

kwcckw: Also w is computed with rA_ and rd_, and rA_ and rd_ are computed with single unit distance, but it's applied in n++ radial distance?

boris-kz: not sure, I pushed some edits, will check tomorrow

kwcckw: Alright, I will check again later.

PR #226, opened 2026-08-04
8/4/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -664,29 +664,23 @@ def proj_focus_opus(PV__, y,x, tile, elev):  # radial accum of projected focus v
...
-        (y-n,x-n), (y-n,x), (y-n,x+n),
-        (y, x-n),           (y, x+n),
-        (y+n,x-n), (y+n,x), (y+n,x+n)
-        ], dtype=int)
-        row,col = rim_coords[:,0], rim_coords[:,1]
-        PV__[row,col] += pV__  # in-place accum pV to rim
+        k_ = np.arange(-n, n+1); e_ = np.full(2*n+1, n); i_ = k_[1:-1]; f_ = np.full(2*n-1, n)
+        rA_ = np.concatenate([np.stack([-e_,k_],1), np.stack([e_,k_],1),
+                              np.stack([i_,-f_],1), np.stack([i_,f_],1)]).astype(float)  # 8n dy,dx
+        rd_ = np.hypot(rA_[:,0], rA_[:,1])  # true distance, n:n*1.41
```
kwcckw: I think rA_ and rd_ should be computed within the n loop, because the 8 directions is only valid for n == 1? For higher n, it actually has higher number of directions, such as 16 when n == 2 or 24 when n == 3.

For example, the existing 8 directions scheme skips some PV__'s directions in higher n:

Image

I think it should cover all directions in PV__?

Image

boris-kz: it's the same 8 directions, the same 64x64 grid on all levels, we only yx of the center of current-level tile

kwcckw: > it's the same 8 directions, the same 64x64 grid on all levels, we only yx of the center of current-level tile

But in higher n, with larger rim, the 8 directions may not limited to 8 tiles only? The rim size should be incremented with the n?

boris-kz: That rim is potential tiles of the same size as the projecting tile. Each will start with a bottom-level tile and expand to the size of the projecting tile.
Sorry I won't have much time today. Actually this whole month, I am moving.

kwcckw: > That rim is potential tiles of the same size as the projecting tile. Each will start with a bottom-level tile and expand to the size of the projecting tile.

I see, since it's the same size, we always get 8 new tiles even in the larger rim (higher n) right? But i think we still need to compute rd_ within the loop since larger rim has longer real distance, which we can't use the same rd_ across all ns.

> Sorry I won't have much time today. Actually this whole month, I am moving to Tennessee.

I see, that's not really near, all the best with your move!

boris-kz: relative distance is the same on all levels

kwcckw: I see, so that `rd_` and `rA_` are relative distance and angle instead. Then i think we can use opus version since i don't see any other issue now.

Also there's some problems in the ffeedback, specifically the injection after the clust_oF_ and split_oF_, i will go through that again later.

boris-kz: ok, so I won't be pushing tonight

kwcckw: Alright, and i will include the opus version's proj_focus later.

PR #227, opened 2026-08-05
8/5/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -183,10 +183,10 @@ def clust_oF_():  # simplified oF rim-mediated centroid clustering
     T_ = []
     for F in F_:
         F.w = sum([w * F.w / (F.w+_F.w) for _F, w in F.rim])
-        if F.w > ave:
+        if F.w > ave:  # so if F.w is weak, skip the F totally? I think we still need to recycle them?
```
kwcckw, line 186: So we form T only if F.w is > ave, and i think we need to recycle those <ave's Fs? Since they still might be called by the code? This F.w eval should be meant for initializing new cluster only?

boris-kz: We already have tile_ in Fr.H[0], that's the backup

kwcckw: Actually this is referring to the oF in clust_oF_. By looking at the current scheme, we form initial Ts with oF and their rim only if oF.w > ave, and the rest of <ave's oFs will be skipped. Those skipped oFs should be still relevant and we should recycle them back into the output oF_?

frame_2D_alg/agg_recursion.py
```diff
@@ -528,11 +528,11 @@ def comb_Ft(Nt, Lt, Bt, root,wTT):  # from sum2G, default Nt
             for L in pL_: L_ += [comp_N(*L.N_, G.r,L.c,1, L.angl[0], L.span)]
             sum2F(L_,Lt); add2F(G, Lt, merge=2)
     else: L_ = [l for n in G.N_ for l in n.L_]; G.L_=L_  # no sum2F?
+    angl = np.zeros(2)  # in all Gs or Ts only?
```
boris-kz, line 531: that's a default

frame_2D_alg/agg_recursion.py
```diff
@@ -657,56 +657,28 @@ def proj_V(_N, N, dist, dy_dx, dec, r):  # _N x N induction
...
+    H,W = PV__.shape  # = win__
     Dec = decay ** (np.hypot(H**elev,W**elev) / (np.hypot(*(tile.box[2:]-tile.box[:2])) +eps))  # per-step decay, in units of tile span
-    rad_A_ = np.array([
+    rA_ = np.array([
     (-1,-1), (-1,0), (-1,1),
     ( 0,-1),         ( 0,1),
     ( 1,-1), ( 1,0), ( 1,1)
     ], dtype=float)  # rim dirs, same order as rim_coords, n-invariant
-    rel_dist_ = np.hypot(rad_A_[:,0], rad_A_[:,1])  # 1|1.4
+    rd_ = np.hypot(rA_[:,0], rA_[:,1])  # 1|1.4
```
boris-kz, line 672: I don't like this renaming

frame_2D_alg/agg_recursion.py
kwcckw: The default angle is None, and we might get singleton G from trace_edge, so this is to fill that singleton G's angl with 0s rather than None.

Also in proj_focus, when there's no angle, w_ is default to 1s, so when there's no specific angle from the tile, their w_ is spread evenly? Or we should skip their projection entirely?

     if np.hypot(*A): ...
     else:  w_ = np.ones(8)

frame_2D_alg/agg_recursion.py
kwcckw: Alright, i will revert them.

frame_2D_alg/agg_recursion.py
boris-kz: ok right. yes, it should be spread evenly without angle, so angle should only apply to asymmetric projected V: gradient of PV = mean(L.d for L in L_)?

frame_2D_alg/meta_code.py
boris-kz: ok so the diffs below are to that effect?
we just need else: `_F_+=[F]`?

frame_2D_alg/agg_recursion.py
kwcckw: Right now the gradient is computed using the lines below:

    m,d = val_(tile.dTT, tile.wTT*ttFrm, 1); c = tile.c 
    V = (m-ave + d-avd) * c
    
We already using d from their accumulated dTT, which is better than mean of Ls' d ?

frame_2D_alg/meta_code.py
kwcckw: Yes, so if oF not in any of the existing T_'s N_, they should be added back to the output.

frame_2D_alg/meta_code.py
```diff
@@ -206,9 +206,13 @@ def clust_oF_():  # simplified oF rim-mediated centroid clustering
             else: T.fin = 1  # converged | weak, filtered below
         _w__ = w__
     out_ = []
-    for i,T in enumerate([t for t in T_ if t.w > ave]):
-        T.nF = i; out_ += [T]  # rename by index
-    oF_ = out_
+    for T in [t for t in T_ if t.w > ave]: out_ += [T]
```
boris-kz, line 209: then here it's
else: out_ += T.N_
`out _ += _F_`
?

frame_2D_alg/agg_recursion.py
boris-kz: this should be from Ls only, and tile.m is a non-directional part?

frame_2D_alg/meta_code.py
boris-kz: T_,_F_ = [],[]
    for F in F_:
        F.w = sum([w * F.w / (F.w+_F.w) for _F, w in F.rim])
        if F.w > ave:
            T = CoF(N_= [F]+[f for f,_ in F.rim], L_= [0 for _ in F_]); T.fin=0  # L_: dense prior w_, aligned with F_ in all Ts
            form_body(T); T_ += [T]
        else: _F_ += [F]
    _w__ = [[0 for f in F_] for t in T_]
    fR = 1  # refine
    while fR:
        fR = 0
        w__ = [[0 for F in T.N_] for T in T_]
        for i,T in enumerate(T_):
            if T.fin: continue
            Dw, N_,w_ = 0, [],[]
                if F.root_: w/=sum(_T.w for _T in F.root_)
                Dw += abs(w - _w__[i][j]); w__[i][j] = w
                if w>ave: N_ += [F]; w_ += [w]; F.root_.add(T)
            T.N_ = N_; T.L_ = w_
            if w_: T.w = np.mean(w_)
            if sum(w_) > ave and Dw > ave:
        _w__ = w__
    _out_,out_ = [],[]
    for T in T_: 
        if T.w > ave: out_ += [T]
        else: out_ += T.N_
    for i,F in enumerate(list(set(out_+_F_))):
        F.nF = i; out_ += [F]  # rename by index
    oF_ = out_

kwcckw: I think some `_F_` might be in T.N_ (added from rim), so `_F_` must be recomputed after we finalized out_.

frame_2D_alg/agg_recursion.py
kwcckw: Okay right, mean Ls' d looks better if we need only their direction.

frame_2D_alg/meta_code.py
kwcckw: > ```
> T_,_F_ = [],[]
> for F in F_:
>     F.w = sum([w * F.w / (F.w+_F.w) for _F, w in F.rim])
>     if F.w > ave:
>         T = CoF(N_= [F]+[f for f,_ in F.rim], L_= [0 for _ in F_]); T.fin=0  # L_: dense prior w_, aligned with F_ in all Ts
>         form_body(T); T_ += [T]
>     else: _F_ += [F]
> _w__ = [[0 for f in F_] for t in T_]
> fR = 1  # refine
> while fR:
>     fR = 0
>     w__ = [[0 for F in T.N_] for T in T_]
>     for i,T in enumerate(T_):
>         if T.fin: continue
>         Dw, N_,w_ = 0, [],[]
>         for j,F in enumerate(T.N_):  # rim-local candidates: members prune, never join
>             w = (comp_body(T.body, F.body) / min(T.fc, F.fc) if F else 0)
>             if F.root_: w/=sum(_T.w for _T in F.root_)
>             Dw += abs(w - _w__[i][j]); w__[i][j] = w
>             if w>ave: N_ += [F]; w_ += [w]; F.root_.add(T)
>         T.N_ = N_; T.L_ = w_
>         if w_: T.w = np.mean(w_)
>         if sum(w_) > ave and Dw > ave:
>             form_body(T); fR = 1  # rebuild from remaining members, refine
>         else: T.fin = 1  # converged | weak, filtered below
>     _w__ = w__
> _out_,out_ = [],[]
> for T in T_: 
>     if T.w > ave: out_ += [T]
>     else: out_ += T.N_
> for i,F in enumerate(list(set(out_+_F_))):
>     F.nF = i; out_ += [F]  # rename by index
> oF_ = out_
> ```

I still think `_F_` should be recomputed after out_ is finalized. I will include that later.

boris-kz: ok, that's replacing clustered callees?

frame_2D_alg/agg_recursion.py
boris-kz: Actually that L_ is concat between the nodes anyway, we can just use T.m for flat and T.d for directional projection? Sorry I didn't edit agg_recursion yet

kwcckw: > T.m for flat and T.d for directional projection

But all projection are directional in 8 directions?

boris-kz: I mean anisotropic element of projection

frame_2D_alg/meta_code.py
kwcckw: Some `_F_` might be in the clustered T.N_, added via their rim, so we can't pack them in the else since it maybe clustered via rim anyway.

PR #228, opened 2026-08-06
8/6/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -670,42 +669,14 @@ def proj_focus_opus(PV__, y,x, tile, elev):  # radial accum of projected focus v
...
-        dec = Dec ** n
-        pV__= np.array([
-        # this should be V * dec * 1|1.4 * mangl(tile.angl, cross-tile angl)
-        V * dec * 1.4, V * dec * a, V * dec * 1.4,  # a = aspect = dy/dx, affects axial directions only
-        V * dec / a,                V * dec / a,
-        V * dec * 1.4, V * dec * a, V * dec * 1.4
-        ], dtype=float)
-        if np.sum(pV__) < ave * 8:
-            break  # < min adjustment
+        pV__ = (Vm + Vd*w_) * Dec**(n*rel_dist_)  # diag dist is 1.4 * axial (w_ is applied on the directional Vd only?)
```
kwcckw, line 678: > I mean anisotropic element of projection

So something like this? Where we apply w_ on the directional Vd only?

frame_2D_alg/agg_recursion.py
```diff
@@ -670,42 +669,14 @@ def proj_focus_opus(PV__, y,x, tile, elev):  # radial accum of projected focus v
...
-        n += 1
-
-def proj_focus(PV__, y,x, tile):  # radial accum of projected focus value in PV__
-
-    m,d = val_(tile.dTT, tile.wTT*ttFrm, 1); c = tile.c  # m,d,n = tile.m, tile.d, tile.c  # add r?
-    V = (m-ave + d-avd) * c
-    dy,dx = tile.angl[0]; a = dy / (dx or eps)  # average link_ orientation, projection
-    Dec = decay ** (np.hypot(*(tile.box[2:]-tile.box[:2])) / np.hypot(Ly,Lx))
-    H,W = PV__.shape  # = win__
+        w_ = 1 + tile.mang * (mA_ / mA_.mean() - 1)  # 1 + to scale w_ into positive? But why -1 after normalize mA_?
```
kwcckw, line 674: The `mang` is ranged from -1 to 1, and `1 +` to convert it into positive? But do you know why we need -1 here?

frame_2D_alg/meta_code.py
```diff
@@ -175,7 +175,7 @@ def parse_funcs(paths):
 
 def clust_oF_():  # simplified oF rim-mediated centroid clustering
 
-    global oF_;  F_ = [F for F in oF_ if F]  # called Fs only: uncalled don't cluster, deleted at swap
+    global oF_;  F_ = [F for F in oF_ if F]  # called Fs only: uncalled don't cluster, deleted at swap (we still need the uncalled in the next iteration? )
```
boris-kz, line 178: Maybe we need it if the last loop was not representative, then uncalled Fs should be in `_F_`

frame_2D_alg/agg_recursion.py
boris-kz: -1 because mang is defined as deviation from average, not an absolute value. 
so it should be a signed value here too?

frame_2D_alg/meta_code.py
```diff
@@ -206,13 +205,16 @@ def clust_oF_():  # simplified oF rim-mediated centroid clustering
                 form_body(T); fR = 1  # rebuild from remaining members, refine
             else: T.fin = 1  # converged | weak, filtered below
         _w__ = w__
-    _out_,out_ = [],[]
+    out_ = []
     for T in T_:
         if T.w > ave: out_ += [T]
-        else: out_ += T.N_
+    _F_ = [F for F in oF_ if F not in (_F for T in out_ for _F in T.N_)]  # recycle input oF only if it's not in any of the T.N_
```
boris-kz, line 211: thanks

frame_2D_alg/agg_recursion.py
```diff
@@ -656,12 +656,11 @@ def proj_V(_N, N, dist, dy_dx, dec, r):  # _N x N induction
...
     return pL_
 
-# not fully reviewed:
-def proj_focus_opus(PV__, y,x, tile, elev):  # radial accum of projected focus value in PV__
+def proj_focus(PV__, y,x, tile, elev):  # radial accum of projected focus value in PV__
 
     m,d = val_(tile.dTT, tile.wTT*ttFrm, 1); c = tile.c  # m,d,n = tile.m, tile.d, tile.c  # add r?
-    V = (m-ave + d-avd) * c
-    H,W = PV__.shape  # = win__, tile pitch = H|W **elev
+    Vm, Vd = (m-ave)*c, (d-avd)*c
```
boris-kz, line 662: this is abs(d) - avd?

kwcckw: If we use abs here, the direction will be determined by w only?

frame_2D_alg/meta_code.py
kwcckw: > Maybe we need it if the last loop was not representative, then uncalled Fs should be in _F_

Yes, I'm using oF_ above so uncalled Fs should be recycled too.

frame_2D_alg/agg_recursion.py
kwcckw: The expected average value is 0? But the deviation could be 1 or -1? 

And if we use abs d when computing V, I think we need signed value here, which mean that 1+ should be removed.

frame_2D_alg/agg_recursion.py
boris-kz: actually no, that abs d was for comp d, here we need to use signed d

frame_2D_alg/agg_recursion.py
boris-kz: right

frame_2D_alg/agg_recursion.py
```diff
@@ -670,42 +669,14 @@ def proj_focus_opus(PV__, y,x, tile, elev):  # radial accum of projected focus v
     ], dtype=float)  # rim dirs, same order as rim_coords, n-invariant
     rel_dist_ = np.hypot(rad_A_[:,0], rad_A_[:,1])  # 1|1.4
     A = tile.angl[0]
-    if np.hypot(*A) > eps:
+    if np.hypot(*A):
         mA_ = np.abs(rad_A_ @ A) / rel_dist_  # axial alignment, scale-free: |A| cancels in w_
```
boris-kz, line 673: mA doesn't depend on rel_dist?

kwcckw: I think that's for normalization?

boris-kz: it doesn't affect angle?

frame_2D_alg/agg_recursion.py
boris-kz: I think it's just

    if np.hypot(*A):
        mA_ = np.abs(rim_A_ @ A) * tile.mang  # axial alignment, scale-free: |A| cancels in w_, signed

frame_2D_alg/agg_recursion.py
kwcckw: Okay right, then we shouldn't use dist here.

frame_2D_alg/agg_recursion.py
kwcckw: Okay thanks, this looks better, that dist is actually not relevant in angle.

frame_2D_alg/agg_recursion.py
boris-kz: probably, I will check tomorrow

PR #229, opened 2026-08-07
8/7/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -142,6 +142,7 @@ def gv_(v, i=None):
     if v > 0: return v  # if +ve Fv triggers oF recursion, invert sign at the caller?
     else: oF_[CoF.get().nF].gV_[i] -= v  # ave Fv - neg v: +ve val of eval
 
+# always pack node along with typ for rebuild?
```
kwcckw, line 145: So this line in clust_oF_ is not fully working yet:

    F.fdef = ast.FunctionDef(name=f'oF{F.nF}',args=ast.arguments(),body=F.body)  

The problem is we need to rebuild the exact ast body using the new F.body, and the line above assign body using F.body, which is wrong. To do that i think we need pack the original ast node in `build` so that we can retrieve the real ast node via F.body during the rebuild process? Right now the real ast nodes is skipped if it's a nested func call with sub.

boris-kz: I don't get this, how is build not rebuilding the body?
also need to review clustering:  https://claude.ai/share/1521c450-421c-446b-a5ca-751be7886324

kwcckw: We actually may convert the forking ast nodes into `(('gv_',l), sub_)` or ` (type(node), sub_)` in body, which we need the actual ast node in the fdef's child nodes.

> cluster_P centroids are identical. All Cs share the same N_ ("all Ns are in all Cs"), and sum2F accumulates TT += n.dTT * (n.c/C) — m_,d_ are stored but never weight the sum. So every refined centroid gets the same dTT/c/r, differing only in bookkeeping m_

Actually this is corret. In cluster_P, we sum all N_'s params into all C in the same way without considering their m_ and d_ weightage, so that w_ in sum2F should be based on m_ instead?

          fC = any(m_)

kwcckw: For the weightage in cluster_P, i think we need to weight c based on their m, something like this?

    if fC: w_ *= m_/sum(m_)

boris-kz: why?

kwcckw: The existing cluster_P sums all params with the same weight w_, and that w_ should be scaled based on their m_ or the soft membership value?

boris-kz: ok try that. sorry, I am not quite following this ATM

kwcckw: Okay, i will include some additional cluster_P related updates first, and i will include some "pending review" comment as a reminder so that we can come back to this later.

PR #230, opened 2026-08-08
8/8/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -354,10 +354,10 @@ def cluster_C(Ft, E_,_r,_c):  # form centroids by clustering exemplar surround v
                 C._N_ = list(set(N__) - set(n_))  # new frontier
                 if D < Avd: out_+=[C]  # output if stable
                 else:       C_ += [C]  # reform
-                DTT+=dTT; mat+=M; dif+=D; cnt+=T; rdn+=R; Up+=up
+                DTT+=dTT; cnt+=T; rdn+=R; Up+=up
         r = _r+ rdn/(cnt or eps)
         L = len(out_+ C_); olp = sum([len(N.root_) for N in N_])  # rdn+=olp, prioritize stronger?
-        if gv_((mat+dif)* (wcP*L) - Ave* (r+olp+ ccP*L)):
+        if gv_(sum(val_(DTT,fd=1))* (wcP*L) - Ave* (r+olp+ ccP*L)):
```
kwcckw, line 360: So i think it's better to recompute and m and d using DTT here? That `mat` and `dif` are accumulated from M and D, which may not be accurate?

frame_2D_alg/agg_recursion.py
```diff
@@ -422,15 +422,19 @@ def cluster_P(_C_, _c, root):  # multi-seed mean shift: parallel centroid refine
                     L.N_ = [N,C]; C.L_ += [L]
                 out_ += [C]
     if out_:
-        dCt = sum2F(list(set(_C_)-set(out_)))  # compress, out_ for CoF?
-        FV_(CoF.get(), dCt.dTT, dCt.c, dCt.r)
+        # dCt = sum2F(list(set(_C_)-set(out_)))  # compress, out_ for CoF?
+        # FV_(CoF.get(), dCt.dTT, dCt.c, dCt.r)
+        # pending review: the difference should be input's params - output's params? We will never get same Cs from _C_ and out_ anyway
+        iTT,iC,iR = sum_vt(iC_); oTT,oC,oR = sum_vt(out_)
+        FV_(CoF.get(), iTT-oTT, iC-oC, iR-oR)
```
kwcckw, line 429: If we need to get the difference, it should be input's params - out's params here?

That `_C_` maybe replaced in the loop, and `out_` will never be the same as input `_C_` anyway.

frame_2D_alg/agg_recursion.py
boris-kz: yes, that's more accurate

frame_2D_alg/agg_recursion.py
boris-kz: `_C_` is never replaced, we only update each C?

frame_2D_alg/meta_code.py
```diff
@@ -209,7 +210,7 @@ def clust_oF_():  # simplified oF rim-mediated centroid clustering
     for T in T_:
         if T.w > ave: out_ += [T]
     _F_ =[F for F in oF_ if F not in (_F for T in out_ for _F in T.N_)]  # recycle singletons
-    for i,F in enumerate(list(set(out_+_F_))):
+    for i,F in enumerate(_F_+out_):  # we shouldn't get duplicated oFs here
```
boris-kz, line 213: So out_ is now oF clusters only??

frame_2D_alg/agg_recursion.py
kwcckw: Yes, but their member may change, and `_C_` is from prior iteration, so the difference should be between the iC_, the input Cs' params and the final out_ ?

frame_2D_alg/meta_code.py
kwcckw: Yes, out_ is Ts , which is init from F and their rim, and then we refine them in the loop, so it's always the oF clusters.

frame_2D_alg/agg_recursion.py
boris-kz: Ok right, dTTs are different.

frame_2D_alg/meta_code.py
```diff
@@ -142,6 +142,7 @@ def gv_(v, i=None):
     if v > 0: return v  # if +ve Fv triggers oF recursion, invert sign at the caller?
     else: oF_[CoF.get().nF].gV_[i] -= v  # ave Fv - neg v: +ve val of eval
 
+# always pack node along with typ for rebuild?
```
boris-kz, line 145: it should be a new typ after rebuild anyway?

frame_2D_alg/meta_code.py
boris-kz: so,

    oT_ = [T for  T in T_ if T.w > ave]
    oF_ = [F for F in oF_ if F not in (_F for T in oT_ for _F in T.N_)]  # recycle singletons
    for i,F in enumerate(oF_ + oT_):
        F.nF=i; F.fdef = ast.FunctionDef(name=f'oF{F.nF}',args=ast.arguments(),body=F.body)  # reinit the fdef: ast node
        oT_ += [F]  # rename by index
    return oT_

frame_2D_alg/agg_recursion.py
```diff
@@ -422,15 +422,19 @@ def cluster_P(_C_, _c, root):  # multi-seed mean shift: parallel centroid refine
...
     return out_
 
 def sum2F(N_, root=None, m_=[],d_=[], merge=0, froot=0, nF=None):  # -> CF/CL/CN
 
     c_ = np.array([n.c for n in N_], dtype=float); C = c_.sum(); w_ = c_/C; N = N_[0]
     fC = any(m_)
     typ = 2 if fC else N.typ
+    if fC: w_ *= m_/sum(m_)  # pending review: to update weights with soft membership via m_
```
boris-kz, line 437: this is a differential initialization so we don't get identical Cs?

frame_2D_alg/meta_code.py
kwcckw: Yes, but their body is different? We may merge their body, and we need the actual ast node in new typ oF's fdef (or nF_)?

That fdef's child nodes will be used in call_sites or inject_oF_ later for the rebuilding process:

    def inject_oF_(oF_, g):  # inject AST in g, recompile g[name]
    
        for oF in oF_:
            oF.fdef = nF_[oF.nF]; oF.g = g
            oF.body=[]; oF.fc=0; oF.g_=[]; oF.gV_=[]
                if r := build(oF,node): t,fc = r; oF.body += [t]; oF.fc += fc
            exec( compile( ast.Module(body=[oF.fdef], type_ignores=[]), '<oF>', 'exec'), oF.g)

frame_2D_alg/agg_recursion.py
kwcckw: In cluster_P, all Cs have the same members, but each with different membership value (m_ or d_), so they shouldn't be identical.

When we sum Ns into C in cluster_P, we need to consider their membership value, which is m_?

Else all Cs have the same w_ since they are all init from the same set of N_:

    c_ = np.array([n.c for n in N_], dtype=float); C = c_.sum(); w_ = c_/C;

boris-kz: that m_ should maps to C.N_, so it has 0s for Ns that are not compared to C yet?

frame_2D_alg/meta_code.py
boris-kz: sorry, I am not following this ATM

frame_2D_alg/agg_recursion.py
kwcckw: Yes, but we compare them all anyway in cluster_P:

        for j,N in enumerate(N_):
            for i,C in enumerate(_C_): md__[j,i] = val_(base_comp(C,N)[0], ttcP)
            
So each w is mapped to `m_[i]/sum(m_)` after we apply `w_ *= m_/sum(m_)`.

frame_2D_alg/meta_code.py
kwcckw: Okay, i will include some draft first, we can come back to this later.

frame_2D_alg/meta_code.py
boris-kz: _T_ = [T for  T in T_ if T.w > ave]
    _F_ = [F for F in oF_ if F not in (_F for T in _T_ for _F in T.N_)]; out_ = []  # recycle singletons
    for i,F in enumerate(_F_ + _T_):
        F.nF=i; F.fdef = ast.FunctionDef(name=f'oF{F.nF}',args=ast.arguments(),body=F.body)  # reinit the fdef: ast node
        out_ += [F]  # rename by index
    return out_

PR #231, opened 2026-08-09
8/9/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -397,13 +397,23 @@ def cluster_P(_C_, _c, root):  # multi-seed mean shift: parallel centroid refine
...
                     if C in removed: continue
                     l = comp_N(C,_C,(C.r+_C.r)/2, min(C.c,_C.c), A=(a:=_C.yx-C.yx), span=np.hypot(*a))
                     if l.m*wF > ave*(l.r+cF):
-                        add2F(_C,C,1); removed += [C]
-        new_ = [Copy_(N,root,init=1,cls=CL) for j,N in enumerate(N_)
-                if np.sort(md__[j,:,0])[-2:].min()*wcP > ave*(N.r+ccP)]  # seed overlap Ns
+                        add2F(_C,C,2); removed += [C]  # merge should be 2 here, all Cs have the same N_
+                        _C.m_ += C.m_; _C.d_ += C.d_  # not sure, sum their m_ and d_ when we merge _C?
+                        for N in N_: 
+                            N.root_[i][1] += N.root_[j][1]; N.root_[i][2] += N.root_[j][2]  # sum m and d reference of merged _C
```
kwcckw, line 407: I think we missed the section to update C's m_,d_ and N_'s root_ when we merge Cs here. 

For C.m_ and d_, we can just sum it from C to `_C`? And it should be the same for root_ where we sum N_'s C.m and C.d into `_C`.m and `_C.`d in root_?

frame_2D_alg/agg_recursion.py
```diff
@@ -397,13 +397,23 @@ def cluster_P(_C_, _c, root):  # multi-seed mean shift: parallel centroid refine
...
-        new_ = [Copy_(N,root,init=1,cls=CL) for j,N in enumerate(N_)
-                if np.sort(md__[j,:,0])[-2:].min()*wcP > ave*(N.r+ccP)]  # seed overlap Ns
+                        add2F(_C,C,2); removed += [C]  # merge should be 2 here, all Cs have the same N_
+                        _C.m_ += C.m_; _C.d_ += C.d_  # not sure, sum their m_ and d_ when we merge _C?
+                        for N in N_: 
+                            N.root_[i][1] += N.root_[j][1]; N.root_[i][2] += N.root_[j][2]  # sum m and d reference of merged _C            
+                            N.root_.pop(j)
+        new_ = []
+        for j,N in enumerate(N_):
+            if md__[j,:,0].min()*wcP > ave*(N.r+ccP):  # should be weakest > ave to init new C?
```
kwcckw, line 411: This section should be init new C based on strong N? So if the weakest C_'s m of N > ave, we init new C based on this N?

And after the init, i think we need to pack this new C to all N.root_ again?

frame_2D_alg/agg_recursion.py
```diff
@@ -388,7 +388,7 @@ def cluster_P(_C_, _c, root):  # multi-seed mean shift: parallel centroid refine
         Lc = len(_C_); L = Lc*Ln  # Lc,L,md__ per cycle since _C_ varies
         md__ = np.zeros((Ln,Lc,2)); O = 0
         for j,N in enumerate(N_):
-            for i,C in enumerate(_C_): md__[j,i] = val_(base_comp(C,N)[0], ttcP)
+            for i,C in enumerate(_C_): md__[j,i] = val_(base_comp(C,N)[0], ttcP,fd=1)  # we need fd = 1 here to assign both m and d, else the same m is assign to both
```
boris-kz, line 391: thanks

frame_2D_alg/agg_recursion.py
```diff
@@ -397,13 +397,23 @@ def cluster_P(_C_, _c, root):  # multi-seed mean shift: parallel centroid refine
...
-                for C in C_[i+1:]:
+                for j, C in enumerate(C_[i+1:],start=i+1):
                     if C in removed: continue
                     l = comp_N(C,_C,(C.r+_C.r)/2, min(C.c,_C.c), A=(a:=_C.yx-C.yx), span=np.hypot(*a))
                     if l.m*wF > ave*(l.r+cF):
-                        add2F(_C,C,1); removed += [C]
-        new_ = [Copy_(N,root,init=1,cls=CL) for j,N in enumerate(N_)
-                if np.sort(md__[j,:,0])[-2:].min()*wcP > ave*(N.r+ccP)]  # seed overlap Ns
+                        add2F(_C,C,2); removed += [C]  # merge should be 2 here, all Cs have the same N_
+                        _C.m_ += C.m_; _C.d_ += C.d_  # not sure, sum their m_ and d_ when we merge _C?
```
boris-kz, line 405: if C is merged, we need to recompute m_,d_ relative to new C
or get the average of the old ones, less acurate

                        _C.m_ = [(_m+m)/2 for _m,m in zip(_C.m_,C.m_)]; _C.d_ = [(_d+d)/2 for _d,d in zip(_C.d_,C.d_)]

frame_2D_alg/agg_recursion.py
boris-kz: same thing for root_, m,d s should be recomputed

frame_2D_alg/agg_recursion.py
kwcckw: > same thing for root_, m,d s should be recomputed

Okay right, so instead of average, i think we can recompute their m_ and d_:

    m_,d_ = [],[]
    for j,N in enumerate(N_):
        m,d = val_(base_comp(_C,N)[0], ttcP,fd=1) 
        m_+=[m]; d_+=[d]

boris-kz: that's in the next loop, we already have it:

        for j,N in enumerate(N_):
            for i,C in enumerate(_C_): md__[j,i] = val_(base_comp(C,N)[0], ttcP, fd=1)

frame_2D_alg/agg_recursion.py
boris-kz: so that will be in the next loop too

frame_2D_alg/agg_recursion.py
kwcckw: > so that will be in the next loop too

Yes, that should be assigned in the next sum2F, and i think we need to make sure there's next loop whenever we merge Cs, which we can set conv = 0 when we merge Cs.

frame_2D_alg/agg_recursion.py
boris-kz: Ns don't change, we we just need to select >ave Cs?
and they will be packed in root_ in base_comp

kwcckw: That > ave's Cs is already selected in the final line:

     _C_ = [c for c in C_ if c not in removed and c.m*wcP > ave*c.r*ccP]
     
I thought we need to reinit new cluster based on strong N here? Else this section is actually redundant.

frame_2D_alg/agg_recursion.py
boris-kz: this merge Cs should be before convergence eval?

frame_2D_alg/agg_recursion.py
boris-kz: then we need to use get_exemplars here?

kwcckw: That get_exemplars is per N, but we need to select based on N's m_ (N's m across all Cs), which is different with the process in get_exemplars?

boris-kz: yes, add a fork in get_exeplars?

kwcckw: I'm not sure if we need the same process in get_exemplars, because it doesn't involve rim, and we only need to init new C when weakest N's m across all Cs > ave?

PR #232, opened 2026-08-10
8/10/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -397,14 +397,19 @@ def cluster_P(_C_, _c, root):  # multi-seed mean shift: parallel centroid refine
         if gv_(O*wcP - ave*(root.r+ccP*L)):  # merge redundant Cs
             for i,_C in enumerate(C_):
                 if _C in removed: continue
-                for C in C_[i+1:]:
+                for j, C in enumerate(C_[i+1:],start=i+1):
```
kwcckw, line 400: Sorry, this is no longer needed.

frame_2D_alg/agg_recursion.py
```diff
@@ -397,14 +397,19 @@ def cluster_P(_C_, _c, root):  # multi-seed mean shift: parallel centroid refine
...
                     if C in removed: continue
                     l = comp_N(C,_C,(C.r+_C.r)/2, min(C.c,_C.c), A=(a:=_C.yx-C.yx), span=np.hypot(*a))
                     if l.m*wF > ave*(l.r+cF):
-                        add2F(_C,C,1); removed += [C]
-        new_ = [Copy_(N,root,init=1,cls=CL) for j,N in enumerate(N_)
-                if np.sort(md__[j,:,0])[-2:].min()*wcP > ave*(N.r+ccP)]  # seed overlap Ns
-        _C_ = [c for c in C_ if c not in removed and c.m*wcP > ave*c.r*ccP] + new_  # survive+prune, +seeds
+                        add2F(_C,C,2); removed += [C]  # merge should be 2 here, all Cs have the same N_ (conv should be false since we fill removed here)
```
kwcckw, line 404: > this merge Cs should be before convergence eval?

I think it's fine since we fill removed here, and convergence will false anyway.

frame_2D_alg/agg_recursion.py
```diff
@@ -397,14 +397,19 @@ def cluster_P(_C_, _c, root):  # multi-seed mean shift: parallel centroid refine
...
+                        _C.m,_C.d = val_(_C.dTT, ttcP, fd=1)  # update m and d for the next loop's base comp
+        for j,N in enumerate(N_):
+            if np.sort(md__[j,:,0])[-2:].min()*wcP > ave*(N.r+ccP):  # seed overlap Ns
```
kwcckw, line 407: > yes, add a fork in get_exeplars?

Sorry, i misunderstood on this section earlier. This section is actually init new C based on strong seed overlapped Cs. 

So we should init new independent C if there's strong overlapping Cs, and recompute this new C's params by comparing it to N_ again, as below.

frame_2D_alg/agg_recursion.py
```diff
@@ -446,8 +451,8 @@ def sum2F(N_, root=None, m_=[],d_=[], merge=0, froot=0, nF=None):  # -> CF/CL/CN
                 kern=n.kern*w; span=n.span*w; yx=n.yx*w; angl = copy(n.angl[0]) if n.angl is not None else None
                 if typ==3: box=copy(n.box)
     F = (cls_[typ])(dTT=TT, c=C, r=R, nF=nF); F.N_ = n_
-    if np.any(m_): F.m_,F.d_ = m_,d_; F.m, F.d = sum(m_),sum(d_)  # m_, d_ may empty here
-    else:          F.m, F.d = val_(TT, fd=1)   # consolidate all val_(TT) with a flag like FV_?
+    if np.any(m_): F.m_,F.d_ = m_,d_
+    F.m, F.d = val_(TT, fd=1)   # consolidate all val_(TT) with a flag like FV_?  (m and d should be always compute from dTT?)
```
boris-kz, line 455: that m_,d_ is precomputed membership vals for centroids, so we don't need to recompute them? 
but then we also need c_, so

`C = sum(c_); F.c=C`
`F.m, F.d = sum(m*c for m,c in zip(m_,c_))/C, sum(d*c for d,c in zip(d_,c_))/C`

same for membership vals from root_, except we don't need d_?
if root_:
`c_ = [rt[0].c for rt in root_]; C = sum(rt[0].c for rt in root_); F.wc = C`
`F.w = sum(rt[1] * c for rt,c in zip(m_,c_))/C`

kwcckw: > that m_,d_ is precomputed membership vals for centroids, so we don't need to recompute them?

Yes, but those ms and ds are precomputed from each n's comparison dTT anyway:

      m,d = val_(dtt,ttcC,1); dTT+=dtt
      
I was thinking if it's better to compute the final m and d using the summed dTT, then we can skip the additional `/C` process.

boris-kz: this is redundant, we already have these m_,d_

kwcckw: Okay right, and actually `val_` doesn't include C, we still need to include the C.

> same for membership vals from root_, except we don't need d_?
> if root_:
> `c_ = [rt[0].c for rt in root_]; C = sum(rt[0].c for rt in root_); F.wc = C`
> `F.w = sum(rt[1] * c for rt,c in zip(m_,c_))/C`

So this is for the existing exemplar selection section only?

    for n in N_:

frame_2D_alg/agg_recursion.py
boris-kz: ok thanks

frame_2D_alg/agg_recursion.py
boris-kz: right

frame_2D_alg/agg_recursion.py
boris-kz: sorting is not actually used here?
also no overlap between Ns, only between root Cs / N?

kwcckw: We need sorting to get the last two highest val Cs?

I think the overlap is referring to the same N has high Val in at least two Cs, so we need to init new independent C based on this N?

PR #233, opened 2026-08-11
8/11/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -406,10 +407,10 @@ def cluster_P(_C_, _c, root):  # multi-seed mean shift: parallel centroid refine
             if np.sort(md__[j,:,0])[-2:].min()*wcP > ave*(N.r+ccP):  # seed overlap Ns
                 C = Copy_(N,root,init=1,cls=CL); C.N_ = N_
                 md_ = np.array([val_(base_comp(C,n)[0], ttcP,fd=1) for n in N_])
-                C.m_,C.d_ = md_[:,0],md_[:,1]; _C_ += [C]
+                C.m_,C.d_ = md_[:,0],md_[:,1]; C_ += [C]
```
kwcckw, line 410: > sorting is not actually used here?
also no overlap between Ns, only between root Cs / N?

Yes, it's a same N between multiple Cs, but it has strong val in at least 2 Cs. So that strong N should init new independent C here?

Is the logic explained by Claude below correct?

`np.sort(md__[j,:,0])[-2:].min()` is the second-largest membership for node j. So the gate reads: at least two Cs claim this N above ave. That's the minimal definition of "N sits in an overlap zone", and seeding Copy_(N, init=1) puts a prototype at the overlap point so mean shift can pull out a mode that the existing seeds straddled but neither captured. Standard split-on-overlap in multi-seed mean shift, which is what the docstring says cluster_P is.

boris-kz: It really makes no difference, we should not have new seeds, just recycle strong and merged Cs.
N.root_ overlap should be used for prune / merge Cs, that's a C redundancy indicator

frame_2D_alg/meta_code.py
boris-kz, line 345: def split_oF(t, oF):  # pack sub gate from gates

        and oF.w * sum(get_fc(p) for p in t[1]) > ave):
        sub = CoF(root=oF, fc=get_fc(t), body=[t], caller_={oF})
        oF_.append(sub); nF_.append(None); sub.nF = len(oF_)-1
        return sub
    if isinstance(t,tuple) and t[1]:  # return body and split nested node recursively
        return (t[0], tuple(split_oF(s,oF) for s in t[1]),*t[2:])  # t[2] could be prior ast node
    return t

in ffeedback:

            for oF in copy(oF_): oF.body = [split_oF(t, oF) for t in oF.body]
            oF_ = clust_oF_()

but that adds nesting, we need to unpack the oF, so it's

            oF_ = [split_oF(t, oF) for _oF in copy(oF_) for t in _oF.body]
?

frame_2D_alg/agg_recursion.py
kwcckw: Okay, the pruning is solely based on m now:

But there's always will be overlapping since all new Cs are init with the same N_? So using m alone should be sufficient to prune N from C?

frame_2D_alg/meta_code.py
kwcckw: > but that adds nesting, we need to unpack the oF, so it's
>  oF_ = [split_oF(t, oF) for _oF in copy(oF_) for t in _oF.body]

Yes, so this is actually unpacking the following code in split_oF:

      for oF in copy(oF_):  # copy because we append new sub during
              oF.body = [split(t,oF) for t in oF.body]

frame_2D_alg/agg_recursion.py
boris-kz: Not N_, just recycle strong and merged Cs

kwcckw: Right now we actually have 2 pruning stages:

                N_,m_,d_ = [],[],[]
                    
Both C and their N_ will be pruned. For Cs, their m is actually the olp since higher olp should get higher m, so we don't need to add olp in the eval again?

boris-kz: N_ should not be pruned, I don't know where that came from.
`C olp should be stronger _C'ms in their N_' N.root_ `

boris-kz: This whole thing is wrong, and I don't have the time to fix it

kwcckw: > C olp should be stronger _C'ms in their N_' N.root_ 

I see, so that whole rebuilding section with N_ below should be removed. 

The new pruning section will be something like this? The prune eval should be added with wcP and ccP?

    for i, _C in enumerate(C_):
        _C.olp = sum(rt[1] for n in C.N_ for rt in getattr(n, 'root_', []) if rt[0] is not C and rt[1] > _C.m)     
        if _C.m * wcP > ave * (_C.r + _C.olp + ccP):  # prune, add olp as stronger ms?
            out_ += [C]

boris-kz: more like 

            if _C.m * wcP > ave * (_C.r + _C.olp + ccP):  # prune, add olp as stronger ms?
                _C_ += [C]  

and this after merge and excluding converged Cs?

kwcckw: Right now this is for converged C_, after the refining section? And the merge is actually within the refinement iteration loops.

kwcckw: Or do you mean this pruning should be within the refinement iteration too?

boris-kz: yes, pruning should be within the refinement

PR #234, opened 2026-08-12
8/12/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -182,8 +182,9 @@ def clust_oF_():  # simplified oF rim-mediated centroid clustering
         if (w := comp_body(_F.body, F.body) / min(_F.fc, F.fc)) > ave:
             _F.rim += [(F,w)]; F.rim += [(_F,w)]; _F.w += w; F.w += w
     T_,_F_ = [],[]
-    for F in F_:
-        F.w = sum([w * F.w / (F.w+_F.w) for _F, w in F.rim])
```
kwcckw, line 186: So i think we need to precompute all oF.w here? 

Because if we compute it in the loop, some `_F.w` is already summed from their rim, while some are not (depends on the sequence)?

frame_2D_alg/meta_code.py
```diff
@@ -197,7 +198,7 @@ def clust_oF_():  # simplified oF rim-mediated centroid clustering
             Dw, N_,w_ = 0, [],[]
             for j,F in enumerate(T.N_):  # rim-local candidates: members prune, never join
                 w = (comp_body(T.body, F.body) / min(T.fc, F.fc) if F else 0)
-                if F.root_: w/=sum(_T.w for _T in F.root_)
+                if F.root_: w/=sum(_T.w for _T in F.root_ if _T in T_ or T.fin)  # only current iteration T or converged T?
```
kwcckw, line 201: The redundancy computation here should be decided by the current iteration Ts and the converged Ts? Because those Ts in prior loops are no longer relevant if they are not converged.

Or we can remove all non converged's F.root_ in the beginning of the iteration?

boris-kz: Right, current and converged, and remove root_s. 
But overlap may change the value of converged Fs too, so we the covergence test should be for the whole T_ (let's rename it to FC_)?

frame_2D_alg/meta_code.py
boris-kz: I don't think this w * F.w / (F.w+_F.w) is correct anyway

frame_2D_alg/meta_code.py
kwcckw: > But overlap may change the value of converged Fs too, so we the covergence test should be for the whole T_

But how the whole T_ helps here? I think we need a 2 steps process to pack initial round of Ts into F.root_ first, and then only apply this `w/=sum(_T.w for _T in F.root_ if _T in T_ or T.fin)` to each T in the second pass?

boris-kz: yes, but Ts can't converge individually, the process is competitive

frame_2D_alg/meta_code.py
kwcckw: To sum their rim's w, we need something like `sum of rim w / sum of rim c`?

frame_2D_alg/meta_code.py
kwcckw: Since it's competitive and it's pruning here, i think we need a priority sequence? So the T with lowest w should check for the pruning first?

boris-kz: It's competitive on the level of rs in N.root_, so rdn = ave stronger rs / ave weaker rs?

frame_2D_alg/meta_code.py
boris-kz: I forgot the logic behind this, will check tomorrow

frame_2D_alg/meta_code.py
kwcckw: >  rdn = ave stronger rs / ave weaker rs

Is this mean the average of strong roots' w divide by the average value of weaker roots' w?

But how do we decide strong or weak roots here? Using root.w > ave?

boris-kz: Stronger or weaker than root ref to current C.

PR #235, opened 2026-08-13
8/13/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -177,42 +177,51 @@ def parse_funcs(paths):
...
-                if w>ave: N_ += [F]; w_ += [w]; F.root_.add(T)
-            T.N_ = N_; T.L_ = w_
-            if w_: T.w = np.mean(w_)
-            if sum(w_) > ave and Dw > ave:
-                form_body(T); fR = 1  # rebuild from remaining members, refine
-            else: T.fin = 1  # converged | weak, filtered below
+                w = (comp_body(T.body, F.body) / min(T.fc, F.fc) if F else 0)  
+                F.root_ += [T]; w__[i][j] = w
+        Dw = 0; Rdn = 0; FC_ = []
+        for i,T in enumerate(T_):  # second pass
```
kwcckw, line 199: So i restructured clust_oF_, and i think we need two steps here? Because we need to add the comparison w of all roots in the first step, and compute the rdn based on stronger and weaker root's w in the second step?

frame_2D_alg/meta_code.py
```diff
@@ -177,42 +177,51 @@ def parse_funcs(paths):
...
+                Dw += abs(w - _w__[T_.index(T)][T.N_.index(F)])
+                if w > ave * r:  N_ += [F]; w_ += [w]
+                else:              F.root_.remove(T)
+            if N_:
+                T.N_ = N_; T.L_ = w_; T.w = np.mean(w_)
+                if T.w > ave * Rdn: FC_ += [T] 
+            Rdn += rdn  
+        for FC in FC_: form_body(FC)
+        if Dw > ave * Rdn: T_ = FC_
+        else:              break  # convergence over the whole FC_'s Dw and Rdn?
```
kwcckw, line 218: If we need the convergence eval based on the whole FC_, other than Dw, we need FC_'s w and their rdn? Something like this:

     sum of FC_'s w * Dw > ave * Rdn?

frame_2D_alg/meta_code.py
boris-kz: right

frame_2D_alg/meta_code.py
boris-kz: I think this should be dw per oF, then summed in Dw

kwcckw: My Dw above is already summed from dw per oF, that's all we need?

    Dw += abs(w - _w__[T_.index(T)][T.N_.index(F)])

boris-kz: I think it's Dv:

            Dv, N_,w_ = 0, [],[]
                if F.root_: w/=sum(_T.w for _T in F.root_ if _T in T_ or T.fin)  # only current iteration T or converged T?
                Dv += abs(w - _w__[i][j]) - ave*F.r; w__[i][j] = w
                if w>ave: N_ += [F]; w_ += [w]; F.root_.add(T)
            T.N_ = N_; T.L_ = w_
            if w_: T.w = np.mean(w_)
            if sum(w_)*T.r * Dv > ave:

kwcckw: I see, but this Dv is per T, but not FC_? We need another DV to eval the whole FC_?

boris-kz: right:

    while True:  # refine Ts
        Dv = 0; w__ = [[0 for F in T.N_] for T in T_]
        for i,T in enumerate(T_):
            if T.fin: continue
            N_,w_ = [],[]
                if F.root_: w/=sum(_T.w for _T in F.root_ if _T in T_ or T.fin)  # only current iteration T or converged T?
                Dv += abs(w -_w__[i][j]) - ave*F.r; w__[i][j] = w
                if w>ave: N_ += [F]; w_ += [w]; F.root_.add(T)
            T.N_ = N_; T.L_ = w_
            if w_: T.w = np.mean(w_)
            if sum(w_) > ave*T.r:
        if Dv > 0:  _w__ = w__
        else: break

kwcckw: Okay, so this is not updated right?

     if F.root_: w/=sum(_T.w for _T in F.root_ if _T in T_ or T.fin)  # only current iteration T or converged T?
 
 We need to use the rdn computed from stronger and weaker roots?

boris-kz: right. this should actually be the same in cluster_C, and I will rename T_ to FC_

kwcckw: In cluster_C:

    n_+=[n]; m_+=[m]; d_+=[d]; c=n.c; T+=c; M+=m*c; D+=abs(d)*c; R+=n.r*c 

R is just accumulated from node's r, and then accumulate into rdn when we form C: `rdn+=R` , then eventually normalize by cnt: `r = _r+ rdn/(cnt or eps)`

It should be the same as the above? I thought we need the stronger and weaker roots earlier since it's competitive?

boris-kz: Probably, please edit

PR #236, opened 2026-08-14
8/14/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -191,27 +191,39 @@ def clust_oF_():  # simplified oF rim-mediated centroid clustering
...
-                Dv += abs(w -_w__[i][j]) - ave*F.r; w__[i][j] = w
-                if w>ave: N_ += [F]; w_ += [w]; F.root_.add(C)
+                w = w__[i][j]  
+                if F.root_:
+                    rw_ = [w__[FC_.index(r)][r.N_.index(F)] for r in F.root_ if r is not C]  
+                    # stronger ws and weaker ws
+                    sw_ = [sw for sw in rw_ if sw > w]
+                    ww_ = [ww for ww in rw_ if ww <= w] + [w]  # add current comparison w to prevent empty divisor?
+                    r = np.mean(sw_)/np.mean(ww_) if sw_ else 1
+                    F.r = r; C.r += r
```
kwcckw, line 210: So i updated the code to updte F.r with the r computed via stronger and weaker roots here.

Also actually it should be `F.r += r` since the same F might be in multiple Cs.

Then, the r will be summed in the same way into C.r.

frame_2D_alg/meta_code.py
```diff
@@ -191,27 +191,39 @@ def clust_oF_():  # simplified oF rim-mediated centroid clustering
...
+                    Dv += abs(w -_w__[i][j]) - ave*F.r
+                    w *= r; w__[i][j] = w
```
kwcckw, line 212: Right now r is a ratio, range from 0 to 1, i think we should update w with this ratio?

frame_2D_alg/meta_code.py
```diff
@@ -191,27 +191,39 @@ def clust_oF_():  # simplified oF rim-mediated centroid clustering
...
             C.N_ = N_; C.L_ = w_
             if w_: C.w = np.mean(w_)
-            if sum(w_) > ave*C.r:
-                form_body(C); fR = 1  # rebuild from remaining members, refine
+            if sum(w_) > ave*C.r: form_body(C)  # rebuild from remaining members, refine
             else: C.fin = 1  # converged | weak, filtered below
         if Dv > 0:  _w__ = w__
         else: break
-    _C_ = [C for  C in FC_ if C.w > ave]
+    _C_ = [C for  C in FC_ if C.w > ave and C.fin]  # only converged clusters?
```
boris-kz, line 221: I would use all Cs, converged is always tentative

frame_2D_alg/meta_code.py
boris-kz: yes, so it's sr_ and wr_

frame_2D_alg/meta_code.py
boris-kz: then we also need to sum F.fc, eval is for w*fC?

frame_2D_alg/meta_code.py
```diff
@@ -191,27 +191,39 @@ def clust_oF_():  # simplified oF rim-mediated centroid clustering
...
-                w = (comp_body(C.body, F.body) / min(C.fc, F.fc) if F else 0)
-                if F.root_: w/=sum(_C.w for _C in F.root_ if _C in FC_ or C.fin)  # only current iteration T or converged T?
-                Dv += abs(w -_w__[i][j]) - ave*F.r; w__[i][j] = w
-                if w>ave: N_ += [F]; w_ += [w]; F.root_.add(C)
+                w = w__[i][j]  
+                if F.root_:
+                    rw_ = [w__[FC_.index(r)][r.N_.index(F)] for r in F.root_ if r is not C]  
+                    # stronger ws and weaker ws
+                    sw_ = [sw for sw in rw_ if sw > w]
+                    ww_ = [ww for ww in rw_ if ww <= w] + [w]  # add current comparison w to prevent empty divisor?
```
boris-kz, line 208: it's + .5?

                if F.root_:
                    rw_ = [w__[FC_.index(r)][r.N_.index(F)] for r in F.root_ if r is not C]; sr_,wr_ = [.5],[]  # self
                    for rw in rw_: 
                        if rw > w: sr_ = [rw]  # stronger or weaker alt root
                        else: wr_ += [rw]
                    r = np.mean(wr_)/np.mean(sr_)

frame_2D_alg/meta_code.py
kwcckw: Okay, actually it's better to use w*fC, but then we need to change the r so that it becomes a numerical value instead of ratio? Since r usually is numerical rather than a ratio.

frame_2D_alg/meta_code.py
kwcckw: But we always skip converged Cs (using C.fin), so they are actually permanent?

frame_2D_alg/meta_code.py
kwcckw: Why it's 0.5 here? When there's no weaker roots, the stronger root's w will be *2?

boris-kz: this is self value, neutral

frame_2D_alg/meta_code.py
boris-kz: that's for refinement only, this is after all that

frame_2D_alg/meta_code.py
boris-kz: yes it should be total redundancy.
sorry I will traveling for the next two weeks, will try to check when I can

kwcckw: Okay, and right now we are using ratio:

    r = np.mean(wr_)/np.mean(sr_)

For total redundancy, instead of ratio, i think it should be the `stronger w * fc` - `weaker w * fc`?

frame_2D_alg/meta_code.py
kwcckw: Okay right, so this C.fin shouldn't be used to pack the final clusters here.

frame_2D_alg/meta_code.py
boris-kz: Normally we use r = n_stronger alternatives, so the weaker ones don't really matter.
using ratio it's sum_stronger_alternatives / current_root

kwcckw: > using ratio it's sum_stronger_alternatives / current_root

So we are gonna stick with the ratio here? But it's the ratio of sum of stronger roots' w to the current w?

boris-kz: yes, r += sum_stronger_alternatives / current_root, so it can be 0 and >1

kwcckw: Alright, I will include that later.

PR #237, opened 2026-08-15
8/15/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -182,11 +182,12 @@ def clust_oF_():  # simplified oF rim-mediated centroid clustering
         if (w := comp_body(_F.body, F.body) / min(_F.fc, F.fc)) > ave:
             _F.rim += [(F,w)]; F.rim += [(_F,w)]; _F.w += w; F.w += w
     FC_,_F_ = [],[]
-    w_ = [sum([w * F.w / (F.w+_F.w) for _F, w in F.rim]) for F in F_]  # need to review
+    w_ = [sum([w * F.w / (F.w+_F.w) for _F, w in F.rim]) for F in F_]  # need to review
```
kwcckw, line 185: I think the F.w here should be just the average between F.w and F.rim's w? The existing `product/sum` is always scale down their value.

frame_2D_alg/meta_code.py
```diff
@@ -198,23 +199,18 @@ def clust_oF_():  # simplified oF rim-mediated centroid clustering
...
-                        if rw > w: sr_ = [rw]  # stronger or weaker alt root
-                        else: wr_ += [rw]
-                    r = np.mean(wr_)/np.mean(sr_)
-                    F.r = r; C.r += r
-                    Dv += abs(w -_w__[i][j]) - ave*F.r
-                    w *= r; w__[i][j] = w
-                if w>ave: N_ += [F]; w_ += [w]
+                    r = sum([w__[FC_.index(r)][r.N_.index(F)] for r in F.root_ if r is not C and r.w >w])/w  # sum of stronger roots' w/w
+                    F.r += r; C.r += r
+                    Dv += abs(w -_w__[i][j]) - (ave+r) * min(C.fc, F.fc)
```
kwcckw, line 208: I think the eval should be `w difference - (ave +r ) * fc`?

But we have that `min(_F.fc, F.fc)` when compute w above:

    (w := comp_body(_F.body, F.body) / min(_F.fc, F.fc))

So that `min(_F.fc, F.fc))` should be removed if we include fc in the eval?

frame_2D_alg/meta_code.py
boris-kz: just a product, F.w should be in 0:1?
sorry, I won't be pushing today

frame_2D_alg/meta_code.py
boris-kz: so it's just C.fc?

frame_2D_alg/meta_code.py
kwcckw: Okay, and F.w is already summing their rim's w when we have comp_body, why we need to multiply them again? At this point, i think w shouldn't be range from 0 to 1 since it's should be accumulated from all rim?

frame_2D_alg/meta_code.py
kwcckw: I think it F.fc or C.fc. The Dv here is per F and C pair, so it should be F.fc, and it should be C.fc when we eval C to reform their body:

    if sum(w_) > (ave+C.r)*C.fc: form_body(C)  # rebuild from remaining members, refine

PR #238, opened 2026-08-16
8/16/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -179,43 +179,37 @@ def clust_oF_():  # simplified oF rim-mediated centroid clustering
     global oF_;  F_ = [F for F in oF_ if F]  # uncalled Fs, don't cluster, pack in _F_ if the loop is not representative?
     for F in F_: F.rim = []; F.w = 0; F.root_ = []
     for _F, F in combinations(F_,2):  # w = relative compression: shared / min cost, ave-commensurate
-        if (w := comp_body(_F.body, F.body) / min(_F.fc, F.fc)) > ave:
+        if (w := comp_body(_F.body, F.body)) > ave * min(_F.fc, F.fc):
             _F.rim += [(F,w)]; F.rim += [(_F,w)]; _F.w += w; F.w += w
     FC_,_F_ = [],[]
-    w_ = [sum([w * F.w / (F.w+_F.w) for _F, w in F.rim]) for F in F_]  # need to review
```
kwcckw, line 185: > just a product, F.w should be in 0:1?

Actually i think it should be compression value summed from all rims instead of a ratio of 0-1? Which they are already summed right after comp_body above:

    _F.w += w; F.w += w

frame_2D_alg/meta_code.py
```diff
@@ -179,43 +179,37 @@ def clust_oF_():  # simplified oF rim-mediated centroid clustering
...
+            N_ = [F]+[f for f,_ in F.rim]; fc=sum(N.fc for N in N_)  # fc is sum of N_'s fc? But then min(C.fc, F.fc) is always F.fc
+            C = CoF(N_= N_, L_= [0 for _ in F_],fc=fc); C.fin=0  # L_: dense prior w_, aligned with F_ in all Ts
             form_body(C); FC_ += [C]  # function clusters
     _w__ = [[0 for f in F_] for t in FC_]
     while True:  # refine Ts
         Dv = 0; w__ = [[0 for F in C.N_] for C in FC_]
         for i,C in enumerate(FC_):  # first pass to compute w and assign F's root_
             if C.fin: continue
             for j,F in enumerate(C.N_):  # rim-local candidates: members prune, never join
-                w = (comp_body(C.body, F.body) / min(C.fc, F.fc) if F else 0)
```
kwcckw, line 197: I think the division using `min(C.fc, F.fc)` should be removed, where w should be the actual value, and fc should be used only in the eval, in the format of `saving > (ave + redundancy) * function cost`?

    if sum(w_) > (ave+C.r)*C.fc: form_body(C) 

And we need to add weight on the left hand side of the eval too?

frame_2D_alg/meta_code.py
boris-kz: probably, sorry I don't have much time these days

frame_2D_alg/meta_code.py
boris-kz: not sure right now, probably be able to focus on these thing for a few days

kwcckw: Okay, then i will continue to update the workflow in split_oF, clust_oF, and get_fdef since the added get_fdef is not fully review yet.

PR #239, opened 2026-08-17
8/17/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -358,9 +355,11 @@ def get_fc(n):
 # divisive clustering:
 def split_oF(t, oF):  # pack sub gate from gates
     if (isinstance(t,tuple) and t[0] in (ast.If,ast.IfExp) and isinstance(h:=t[1][0],tuple) and isinstance(h[0],tuple) and h[0][0]=='gv_'
+        and not any(isinstance(n, (ast.Break, ast.Continue)) for n in ast.walk(t[2]))  # skip if contain break or continue, not relevant in new split oF
```
kwcckw, line 358: So there's another problem when we split forks in split_oF because some block of code may contain break or continue, which may not be relevant if we split it into new func. I think we should skip the splitting process if the block of code contain break or continue since it breaks the logic?

PR #240, opened 2026-08-18
8/18/2026

kwcckw:


frame_2D_alg/meta_code.py
```diff
@@ -299,12 +296,12 @@ def form_forks(bod_):
             _fork[1] = (ast.IfExp, *form_forks(_ifnode[1:]))  # recursively form_forks in t1,t2.. of ([f],t,t2...)
             bod_[bod_.index(forks[0])] = tuple(_fork)  # reassign the merged fork
         return bod_  # bod_ is [([f], t),([f], t),...]
-    _body_ = [f.body for f in F.N_ if f]
+    _body_ = [f.body for f in F.N_]  # pending review: should include the empty oF because those split oF has empty call_
```
kwcckw, line 299: So I added "pending review" in the sections that needed further discussion later.

frame_2D_alg/meta_code.py
```diff
@@ -358,9 +355,11 @@ def get_fc(n):
 # divisive clustering:
 def split_oF(t, oF):  # pack sub gate from gates
     if (isinstance(t,tuple) and t[0] in (ast.If,ast.IfExp) and isinstance(h:=t[1][0],tuple) and isinstance(h[0],tuple) and h[0][0]=='gv_'
+        and not any(isinstance(n, (ast.Break, ast.Continue)) for n in ast.walk(t[2]))  # pending review: skip if block of code contains break or continue, not relevant in new split oF
         and oF.w * sum(get_fc(p) for p in t[1]) > ave):
-        sub = CoF(root=oF, fc=get_fc(t), body=[t], caller_={oF})
-        oF_.append(sub); nF_.append(None); sub.nF = len(oF_)-1
+        sub = CoF(root=oF, fc=get_fc(t), body=[t], caller_={oF}); oF_.append(sub);
+        sub.fdef = get_fdef(sub, name=f'oF{len(oF_)-1}')   # reform fdef
+        nF_.append(sub.fdef); sub.nF = len(oF_)-1  # pending review: the split oFs have issues on the input and output argument
```
kwcckw, line 362: So the existing split_oF is actually more complicated than expected. When we split block of code into new oF, we need to handle their input arguments (to parse the variables) and returned outputs too. Here's some suggestions from claude:

Image

I think we should go for C, where we will parse the required inputs + returns, but we should probably include the number of required inputs + returns as part of the cost?

boris-kz: These costs per oF don't change. The inputs are always passed and modified internally?

kwcckw: > These costs per oF don't change. 

But the number of inputs and outputs to the split oF may be varied according to the requirement of the split oF's code, so that number of inputs and outputs should matter too?

> The inputs are always passed and modified internally?

It's not default for the split oF, which we actually need to add that to the split oF's fdef (the split oF ast node).

boris-kz: I don't think the number matters. Anyway, we need to work on a higher level

kwcckw: Alright, then it should just parse all the inputs and outputs.

Another issue in split_oF is the usage of break or continue, which becomes not relevant if it's inside the split's oF. At this point, i think we should skip and shouldn't split the section of code if it contains break or continue? Since that actually break the logic of the function.

PR #241, opened 2026-08-19
8/19/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -403,17 +403,19 @@ def cluster_P(_C_, _c, root):  # multi-seed mean shift: parallel centroid refine
                     l = comp_N(C,_C,(C.r+_C.r)/2, min(C.c,_C.c), A=(a:=_C.yx-C.yx), span=np.hypot(*a))
                     if l.m*wF > ave*(l.r+cF):
                         add2F(_C,C,2); removed += [C]; _C.m,_C.d = val_(_C.dTT,ttcP,fd=1)  # for next-loop base comp
+                        for N in C.N_:
+                            for rt in N.root_: rt[0] = _C  # update root from C to _C, for olp computation below
```
kwcckw, line 407: In cluster_P, looks like we still need to update the N.root_ from `C` to `_C` after the merging and before the next loop because we need this new root reference to compute the olp below.

PR #242, opened 2026-08-20
8/20/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -393,6 +393,7 @@ def cluster_P(_C_, _c, root):  # multi-seed mean shift: parallel centroid refine
             m_ = md__[j,:,0]; O += m_.sum() - m_.max()  # cross-C ambiguity, gates split/merge
         C_ = [sum2F(N_, root, md__[:,i,0], md__[:,i,1]) for i in range(Lc)]  # mean shift, aligned to md__
         Mt = md__[:,:,0].sum()  # total V
+        # pending review: for the case of diverging, we need to break if diverge?
         conv = md__.shape==_md__.shape and Mt* np.abs(md__-_md__).sum()* (wcP*L) <= ave*(root.r+ccP*L)  # convergence
```
kwcckw, line 397: So for the case of diverging, this may run indefinitely, i think we need to check for divergence and break?

That may possible from d, the analysis from claude:

    C.d is written from d_ (sum2F) and read back as _D (base_comp) — self-multiplying loop, gain wD/2·⟨vm⟩/n >> 1.

frame_2D_alg/agg_recursion.py
```diff
@@ -119,7 +119,7 @@ def comp_N_(pL_, r, tnF=None, root=2, fall=0):  # incremental-distance cross_com
     if mrg_: N_ = list(set(N_) - set(mrg_))
     if L_:
         for N in set(N_):
-            if N.rim: N.Rt = sum2F(N.rim,root=N,froot=2)
+            if N.rim: N.Rt = sum2F(N.rim); N.Rt.root = N  # pending review: we shouldn't sum N.rim's params into N? Those rim params should be summed to G only (N.s root)
```
boris-kz, line 122: I guess so, the rim is external

frame_2D_alg/agg_recursion.py
boris-kz: I guess we can project the trend, stop if there is no improvement

kwcckw: For trend projection, it will be something like an extrapolation of all prior iteration's md__ sum?

PR #243, opened 2026-08-21
8/21/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -393,7 +394,16 @@ def cluster_P(_C_, _c, root):  # multi-seed mean shift: parallel centroid refine
...
         C_ = [sum2F(N_, root, md__[:,i,0], md__[:,i,1]) for i in range(Lc)]  # mean shift, aligned to md__
         Mt = md__[:,:,0].sum()  # total V
-        conv = md__.shape==_md__.shape and Mt* np.abs(md__-_md__).sum()* (wcP*L) <= ave*(root.r+ccP*L)  # convergence
+        # pending review: for the case of diverging, we need to break if diverge?
+        conv = 0
+        dM = np.abs(md__[:,:,0]-_md__[:,:,0]).sum() if md__.shape==_md__.shape else 0
+        if dM:
+            if _dM:  # converge always false when _dM is 0 in the first iteration
+                rM = dM / _dM
+                if rM >= 1 and abs(_rM - rM)>ave: conv = -1  # Diverging  (ratio of dM to _dM is growing larger)
```
kwcckw: > I guess we can project the trend, stop if there is no improvement

I think we can use the ratio of prior `_dM` to current `dM` -> `rM`, and if that ratio is increasing (>ave), it means divergence? For convergence, difference between ms should be decreasing over the iterations.

frame_2D_alg/agg_recursion.py
```diff
@@ -363,7 +363,7 @@ def cluster_C(Ft, E_,_r,_c):  # form centroids by clustering exemplar surround v
             out_ = cluster_P(out_,T, Ft)  # refine all memberships in parallel by global backprop
             break
         if Up * (wcC*len(C_)) > Avd * (r+ (ccC*len(C_))):
-            for C in C_: C._m_ = C.m_; C._d_ = C.d_
+            # for C in C_: C._m_ = C.m_; C._d_ = C.d_  # pending review: we don't need _m_ and _d_ now?
```
boris-kz: right, it's all in root_

frame_2D_alg/agg_recursion.py
boris-kz: Probably just _dM-dM > ave, maybe summed over a few loops. We only need total ddM to justify the costs

frame_2D_alg/agg_recursion.py
```diff
@@ -403,17 +413,21 @@ def cluster_P(_C_, _c, root):  # multi-seed mean shift: parallel centroid refine
...
-                _C_ += [C]  # also after merge and converge
+                _C_ += [_C]; i_ += [i]  # also after merge and converge
+        md__ = md__[:,i_]  # pending review: remove md__ of removed Cs, to be used in the next convergence test in the next iteration    
         if not _C_ or (conv and (not removed) and len(_C_)==Lc):  # skip if all _C_ failed the  c.m*wcP > ave*c.r*ccP eval above
             break
         _md__ = md__; cnt += 1
     out_ = []
     for N in N_: N.root_ = []  # replace with out_ Cs:
-    for i, _C in enumerate(C_):
+    for i, _C in enumerate(_C_):  # should be _C_ here to get the converged and merged Cs
```
boris-kz: thanks

frame_2D_alg/meta_code.py
```diff
@@ -175,54 +176,50 @@ def parse_funcs(paths):
                 nF_[iF_.get(node.name)] = node
 
 def clust_oF_():  # simplified oF rim-mediated centroid clustering
-
-    global oF_;  F_ = [F for F in oF_ if F]  # uncalled Fs, don't cluster, pack in _F_ if the loop is not representative?
+    # pending review: include the prior split new oFs although their call_ is empty?
```
boris-kz: right

frame_2D_alg/agg_recursion.py
```diff
@@ -376,14 +376,15 @@ def cluster_C(Ft, E_,_r,_c):  # form centroids by clustering exemplar surround v
         if gv_((_m *_c *wcC) / (_r+ccC) * ((len(out_)-1)*wL) - ave):
             Ft.root.H += [Copy_(Ft)]
             return sum2F(out_, Ft.root, nF='Nt')
-
+# pending review: we not really need this input _c or the total _C_'s c?
```
boris-kz: what do you loop, if not `_C_`?

frame_2D_alg/agg_recursion.py
kwcckw: So it will be the summed of ` _dM-dM ` over a few `n` loops > ave? If the summed of difference is positive and > ave, that's mean it's diverging? I guess we should have that `n` as one of the filter too?

frame_2D_alg/agg_recursion.py
kwcckw: Actually i meant the parameter T here, which is summed from Cs' c:

            out_+=C_; T=sum([f.c for f in out_])
            out_ = cluster_P(out_,T, Ft)  # refine all memberships in parallel by global backprop
            
I don't see we use it in the cluster_P now?

boris-kz: ok, I guess we don't

frame_2D_alg/agg_recursion.py
boris-kz: if _dM-dM > ave, that's converging

kwcckw: `dM` is computed from `np.abs(md__[:,:,0]-_md__[:,:,0]).sum()`. So dM is expected to be very small if it's converging? Because both md__ s should get similar values when it's converging. So `_dM-dM` should be even smaller if it's converging?

boris-kz: If converging, '_dM' 
should be bigger than dM

kwcckw: Okay right, thanks. I was confused earlier.

PR #244, opened 2026-08-22
8/22/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -393,7 +392,15 @@ def cluster_P(_C_, _c, root):  # multi-seed mean shift: parallel centroid refine
...
         C_ = [sum2F(N_, root, md__[:,i,0], md__[:,i,1]) for i in range(Lc)]  # mean shift, aligned to md__
         Mt = md__[:,:,0].sum()  # total V
-        conv = md__.shape==_md__.shape and Mt* np.abs(md__-_md__).sum()* (wcP*L) <= ave*(root.r+ccP*L)  # convergence
+        # pending review: for the case of diverging, we need to break if diverge?
+        conv = 0
+        dM = np.abs(md__[:,:,0]-_md__[:,:,0]).sum() if md__.shape==_md__.shape else 0
+        if dM:
+            if _dM:  # first loop doesn't have prior _dM
+                ddM_ += [_dM - dM]; ddM_ = ddM_[-5:]  # summed over last 5 loops
+                conv = Mt*(wcP*L)/sum(ddM_) <= ave*(cnt+root.r+ccP*L)  # add cnt as redundancy?
```
kwcckw, line 401: So i think the sum of `ddM_` should be added as denominator, where the convergence is more likely to happen when the sum  is large. Also `cnt` should be added to the redundancy?

frame_2D_alg/agg_recursion.py
```diff
@@ -119,8 +119,8 @@ def comp_N_(pL_, r, tnF=None, root=2, fall=0):  # incremental-distance cross_com
     if mrg_: N_ = list(set(N_) - set(mrg_))
     if L_:
         for N in set(N_):
-            if N.rim: N.Rt = sum2F(N.rim,root=N,froot=2)
-        TT, C, R = sum_vt(L_)
+            if N.rim: N.Rt = sum2F(N.rim); N.Rt.root = N  # pending review: we shouldn't sum N.rim's params into N? Those rim params should be summed to G only (N.s root)
```
boris-kz, line 122: These comments are a week old? Why do they show as new?

kwcckw: All the prior week's edits will be shown here unless I pull your latest update again. I didn't do that since that will remove all the pending review's edits.

For your convenience, you actually can check the edits under pending review since those reviewed edits will have their 'pending review' removed. Then the rest of the edits are still useful as a reference to update your local repo later.

boris-kz: I can't really review all that now. The ones we resolved in the discussion don't need comments

kwcckw: Alright, i will remove the additional comments later, but the edits will be still present in the upcoming PR.

frame_2D_alg/agg_recursion.py
boris-kz: L is not really relevant here, it's only needed when we form clusters because that adds discrete cost. Refinement stops if the costs exceed the gains, and divergence means negative gains, so it should work the same

PR #245, opened 2026-08-23
8/23/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -318,10 +318,10 @@ def trans_cluster(G):
         for tt,c,gr in Gt_: w=c/C; TT+=tt*w; R+=gr*w
         if gv_(val_(TT*Ft.root.wTT*ttcN) * (C*wcN /(_r+R+ccN)) * ((len(G_)-1)*wL) - ave):  # reform root,Nt, no other forks yet:
             rG = Ft.root; Nt=rG.Nt; Nt.N_=G_; Nt.dTT=TT; Nt.c=C; Nt.r=R
-            rG.dTT=TT; rG.c=C; rG.r=R
+            rG.dTT=TT; rG.c=C; rG.r=R; rG.m, rG.d = val_(TT, ttcN)
```
kwcckw, line 321: So after replacing TT, i think we should recompute their m and d using the new TT here?

frame_2D_alg/agg_recursion.py
```diff
@@ -794,18 +806,17 @@ def PP2N(PP):
                 if gv_(edge.G * (wVct*L) - sum([P.latT[4] for P in edge.P_]) * (cVct*L)):
                     PPm_ = comp_slice(edge, rV, ttVct)  # add comp_slice's weights?
                     N_ = [PP2N(PPm) for PPm in PPm_]
-                    c = sum([PPm.c for PPm in N_]); C += c
```
kwcckw, line 797: Actually this `c = sum([PPm.c for PPm in N_]); C += c` is redundant since we will compute the same c in line 814 below:

    tt,c,r = sum_vt(N_)

We could have just sum c from the c of sum_vt?

PR #246, opened 2026-08-24
8/24/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -392,8 +395,14 @@ def cluster_P(_C_, _c, root):  # multi-seed mean shift: parallel centroid refine
...
             m_ = md__[j,:,0]; O += m_.sum() - m_.max()  # cross-C ambiguity, gates split/merge
         C_ = [sum2F(N_, root, md__[:,i,0], md__[:,i,1]) for i in range(Lc)]  # mean shift, aligned to md__
-        Mt = md__[:,:,0].sum()  # total V
-        conv = md__.shape==_md__.shape and Mt* np.abs(md__-_md__).sum()* (wcP*L) <= ave*(root.r+ccP*L)  # convergence
+        conv = 0
+        dM = np.abs(md__[:,:,0]-_md__[:,:,0]).sum() if md__.shape==_md__.shape else 0
+        if dM:
+            if _dM:  # first loop doesn't have prior _dM
+                ddM_ += [_dM - dM]; ddM_ = ddM_[-5:]  # summed over last 5 loops
+                conv = wcP*sum(ddM_) <= ave*(cnt+root.r+ccP)
```
kwcckw, line 403: > L is not really relevant here, it's only needed when we form clusters because that adds discrete cost. Refinement stops if the costs exceed the gains, and divergence means negative gains, so it should work the same

So it should be something like this, and Mt shouldn't be relevant now? We only need the sum of ddM_ and it should converge when the gain is small or negative?

boris-kz: If negative, we need to return to last maximum, so it should be recorded?

frame_2D_alg/agg_recursion.py
```diff
@@ -91,7 +91,7 @@ def cross_comp(root, G_, m, c, r, nF='Nt'):  # agg+: refine by CC,exe -> cross_c
             if Ct := cluster_C(G.Nt, get_exemplars(G.N_,r,c),r,c):  # refines, splits each G
                 C_, _m,_c,_r = Ct.N_,Ct.m,Ct.c,Ct.r
                 C__ += C_; M+=_m; C+=_c; R+=_r  # splice refined sub_G_s
-    if C__:
+    if C__:  # pending review: C__ is actually converted PPs, since G_ are graphs from trace_edge, so trans comp in comp_N will never be true
```
boris-kz, line 94: only in the first call, cross_comp can be called recursively

frame_2D_alg/agg_recursion.py
```diff
@@ -141,6 +142,9 @@ def comp_H(_Nt,Nt, Link):  # tentative pre-comp, before comp N_?
     if N.typ > 1 and gv_(m* (c/r)* wN_ - ave*(r+cN_)):  # skip PPs, Nts?
         L.H = [Copy_(L)]  # lev0 to preserve resolution before adding deeper tLevs, min len H = 2
         dn_ = []  # cross_comp N_| Ft_ -> top tLev
+        if _N.H and N.H:  # pending review: So this comp_H should be default first when both Ns have H?
```
boris-kz, line 145: this would be secondary to comp(N_) below, but before recursive spec in each sub_N
not sure how to condition it, that's why I left it out for now

kwcckw: If it needs before recursive spec, i think it should be before the comp, but that's only true in the deeper recursive comp_N (skip the first comp_N)?

frame_2D_alg/agg_recursion.py
kwcckw: Okay right, thanks for clarifying.

frame_2D_alg/agg_recursion.py
kwcckw: I think we can break and return and assign `C_` as `_C_` when the sum is negative?

frame_2D_alg/agg_recursion.py
boris-kz: Yes, we have to do spec breadth first

frame_2D_alg/agg_recursion.py
boris-kz: Something like that

frame_2D_alg/agg_recursion.py
kwcckw: Alright, then i think we need additional flag, something like fH to identify that since we didn't record the depth of the spec recursion now.

boris-kz: no flag, just if H. then we can call batched specs

kwcckw: Okay, so this should be moved to `comp_N_`, so that we can have breadth-first spec before the subsequent recursion, or i think it could be in proj_L_, so that we can run the breadth_first spec per N pair before comp_N.

boris-kz: I think it's in comp_N, this is still specification, even if we do it in batches 
There is no H per comp_N_.

kwcckw: > I think it's in comp_N, this is still specification, even if we do it in batches

But comp_N always per N pair, it can't be in batches?

boris-kz: Each of the Ns has N.N_, that's the batch

kwcckw: I see, thanks for claryifng, let me change that later.

PR #248, opened 2026-08-25
8/25/2026_2

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -105,22 +105,24 @@ def cross_comp(root, G_, m, c, r, nF='Nt'):  # agg+: refine by CC,exe -> cross_c
 def comp_N_(pL_, r, tnF=None, root=2, fall=0):  # incremental-distance cross_comp, max dist depends on prior match
 
     L_,N_,mrg_ = [],[],[]
-    for dist, dy_dx, _N,N, lc,lr, pTT,m,d in pL_:  # pL is L = dist, dy_dx, _N,N if fall: not selective
+    for dist, dy_dx, _N,N, lc,lr, pTT,m,_ in pL_:  # pL is L = dist, dy_dx, _N,N if fall: not selective
         if fall or (m>0 and gv_(m * (lc*wN / (lr*cN)) - ave*(r+cN))):  # marginal -gV
         # comp if marginally predictable: proj surprise value?
             Link = comp_N(_N,N, lr,lc, full = not tnF, A=dy_dx, span=dist, rL=root)
+            # pending review: we no longer using rTT now? 
             Link.rTT = np.abs(pTT - Link.dTT) / eps_(Link.dTT)  # relative prediction error/oF, direction-agnostic
```
kwcckw, line 113: > I think rTT is still relevant, it's logically different

Right now we only pack Link.rTT, but not for oFs. So it should be selective for certain functions, for example in cluster_P:

    FV_(CoF.get(), iTT-oTT, iC-oC, iR-oR)

We need to pack an additional oF.rTT as `(iTT-oTT)/iTT`?

frame_2D_alg/agg_recursion.py
```diff
@@ -392,8 +397,15 @@ def cluster_P(_C_, _c, root):  # multi-seed mean shift: parallel centroid refine
...
             m_ = md__[j,:,0]; O += m_.sum() - m_.max()  # cross-C ambiguity, gates split/merge
         C_ = [sum2F(N_, root, md__[:,i,0], md__[:,i,1]) for i in range(Lc)]  # mean shift, aligned to md__
-        Mt = md__[:,:,0].sum()  # total V
-        conv = md__.shape==_md__.shape and Mt* np.abs(md__-_md__).sum()* (wcP*L) <= ave*(root.r+ccP*L)  # convergence
+        conv = 0
+        dM = np.abs(md__[:,:,0]-_md__[:,:,0]).sum() if md__.shape==_md__.shape else 0
+        if dM:
+            if _dM:  # first loop doesn't have prior _dM
+                ddM_ += [_dM - dM]; ddM_ = ddM_[-5:]  # summed over last 5 loops
+                if sum(ddM_)<0: break  # break and preserve the last Cs before the divergence
```
kwcckw, line 405: > where is the reversal we talked about?

Sorry, i forgot to include it in the prior edit. So we should break when sum of ddM_ is negative, and then pruning after the refinement will be using the last Cs before the divergence.

frame_2D_alg/agg_recursion.py
boris-kz: oF.dTT is actually the same as rTT: values added by the function. So that rTT probably is dTT of some other object, I need to refresh on that.

kwcckw: Okay, previously we had that rTT computed from similar ratio of `(output-input)/input`:

        oF = CoF.get(); oF.N_=copy(G_); oF.rTT=sel_TT

boris-kz: So this should probably be a generic oF value method, something like vTT_, with options for selection (clustering) and match (comparison) values?

kwcckw: > with options for selection (clustering) and match (comparison) values?

But for each function, they usually either perform either clustering or comparison or other? And this could be identified from their typ: 'root_','comp_','clus_','fbac_','prep_'

I'm not sure if we need vTT_ since the TT for each option is mutually exclusive.

PR #249, opened 2026-08-26
8/26/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -475,19 +494,18 @@ def sum2F(N_, root=None, m_=[],d_=[], merge=0, froot=0, nF=None):  # -> CF/CL/CN
...
     if F.c:
         C=F.c+n.c; _w,w = F.c/C, n.c/C; F.r = F.r*_w + n.r*w; F.c = C
         if isinstance(F,CoF) and isinstance(n,CoF): F.fw += n.m  # sum subtree gain, fixed cost, extensive: no weighting?
-        setattr(F, a, getattr(F,a)*_w + getattr(n,a)*w)
+        setattr(F, 'dTT', getattr(F,'dTT')*_w + getattr(n,'dTT')*w)
     else:
-        setattr(F,a,getattr(n,a)); F.c=n.c; F.r=n.r
+        setattr(F,'dTT',getattr(n,'dTT')); F.c=n.c; F.r=n.r
         if isinstance(F,CoF) and isinstance(n,CoF): F.fw=n.m
+    F.m, F.d = val_(F.dTT,fd=1)
```
kwcckw, line 506: So i think we need to recompute F.m and Fd after the addition of dTT in add2F.

frame_2D_alg/agg_recursion.py
```diff
@@ -697,8 +715,10 @@ def proj_focus(PV__, y,x, tile, elev):  # radial accum of projected focus value
 def proj_N(N, dist, A,_r,_c, dec=1):  # arg rc += N.rc+Nw, recursively specify N projection val, add pN if comp_pN?
 
     def proj_TT(L, cos_d, dist, r, pTT, wTT, dec=1, fdec=0):     # accumulate L|N' pTT with iTT|eTT internally
+        # pending review: we reassign ave in ffeedback: ave, avd = val_(aTT, fd=1), so ave can't be negative else we won't get a valid decay here
         dec = dist if fdec else ave** (1 + dist * dec / L.span)  # not fully revised, ave = match decay rate / unit distance
         TT = np.array([L.dTT[0] * dec, L.dTT[1] * cos_d * dec])  # IxE angle alignment * decay?
+        # pending review: decay cancel out each other in val_:  (m_/ eps_(m_+ad_))
         cert = abs(val_(TT* wTT*ttPrj) * ((L.c+wPrj)/(r+cPrj)) - ave)  # approximation
```
kwcckw, line 722: The decay is actually cancel out each other in val_:

   vm = (m_/ eps_(m_+ad_)) @ wTT[0] 

It's actually `(L.dTT[0] * dec) / ([L.dTT[0] * dec + L.dTT[1] * cos_d * dec)` 

and it becomes `(L.dTT[0]) / ([L.dTT[0]  + L.dTT[1] * cos_d) * dec/dec `

I guess decay should be applied to cert only?

frame_2D_alg/agg_recursion.py
```diff
@@ -697,8 +715,10 @@ def proj_focus(PV__, y,x, tile, elev):  # radial accum of projected focus value
 def proj_N(N, dist, A,_r,_c, dec=1):  # arg rc += N.rc+Nw, recursively specify N projection val, add pN if comp_pN?
 
     def proj_TT(L, cos_d, dist, r, pTT, wTT, dec=1, fdec=0):     # accumulate L|N' pTT with iTT|eTT internally
+        # pending review: we reassign ave in ffeedback: ave, avd = val_(aTT, fd=1), so ave can't be negative else we won't get a valid decay here
         dec = dist if fdec else ave** (1 + dist * dec / L.span)  # not fully revised, ave = match decay rate / unit distance
```
kwcckw, line 719: We may update ave in ffeedback:

    ave, avd = val_(aTT, fd=1)

Which the `ave` may become negative. So i think we can't use ave as the base here `ave** (1 + dist * dec / L.span)` since that may lead to complex number.

frame_2D_alg/agg_recursion.py
```diff
@@ -105,26 +105,31 @@ def cross_comp(root, G_, m, c, r, nF='Nt'):  # agg+: refine by CC,exe -> cross_c
 def comp_N_(pL_, r, tnF=None, root=2, fall=0):  # incremental-distance cross_comp, max dist depends on prior match
```
boris-kz, line 105: Reconsidering hexagonal tiling, with a good LLM we can convert from the rectagonal automatically:

me:
in ANNs, context and content segmentation is by 2D rectangular matrices, but hexogonal matrices are more "convex" or compact: lesser average distance from the center?

Fable:
Yes, and it's provable: among all equal-area partitions of the plane, the regular hexagonal tiling minimizes mean squared distance to cell center (Fejes Tóth; it's the optimal 2D quantizer). But the intra-cell margin over squares is small, and the substantive differences are elsewhere.

1. Cell compactness, unit area. Mean distance to center: square .383, hex .377, disk .376 — hex is 1.4% better than square, 0.3% short of the disk. Mean squared distance: .167 / .160 / .159 (hex −3.8%). Perimeter at equal area: hex −7%. "Convex" isn't the term (both are convex); this is circularity / second moment. So the positional loss from representing a cell by its center improves by a few percent. Not a reason to switch.

2. Adjacency is where the difference is real. A square grid has two neighbor classes: 4 at distance 1 sharing an edge, 4 at 1.41 sharing a point. Everything downstream is a patch for that: the .25/.5/.25 diagonal weights in comp_pixel, the 1.4 diagonal factor in proj_focus rim, and the mismatch between the 4-adjacent wave in fill_frame (line 816) and the 8-adjacent rim in proj_focus — two connectivities that don't agree (a closed 4-path doesn't separate under 8-connectivity). Hex has one class: 6 neighbors, distance 1, full-edge contact, one connectivity. Ring k is 6k cells and near-circular vs 8k with corners at k×1.41, so a square context window over-samples diagonals at distance. Gradient from three axes at 60° is isotropic without weighting fudges.

3. Sampling: for isotropic band-limited input the hex lattice needs 13.4% fewer samples than square at equal resolution (densest packing of spectral replicas). Largest measurable win, and it's a lattice property, not a cell property.

4. Cost, specific to frame_H: square tiles nest exactly (Ly**elev, lines 804, 819). Hexagons don't tile hexagons; hierarchical hex schemes (GBT 7-cell aggregates, H3) rotate per level and have ragged supertile boundaries. So hex trades isotropy for exact hierarchical closure. Cheap middle: brick offset, alternate tile rows shifted by half a tile on the same rectangular array — 6 edge-sharing neighbors at distances 1 and 1.12 instead of 1 and 1.41, nesting error half a tile per level (1/128 of supertile). Storage stays a matrix either way: hex grids are always stored as skewed rectangular arrays, which is also why ANNs use rectangles — memory layout and matmul, not geometry. "Hexagonal matrix" is a neighbor-offset function, not a data structure (HexaConv is masked 7-cell kernels on rectangular tensors).

5. In your scheme the grid only matters where the grid is the adjacency: comp_pixel kernel and the tile wave in frame_H / proj_focus. Above that, nodes carry yx, box, angl and links carry real distance and angle, so neighborhoods are already radial — hex isotropy is what link-based comparison gives for free, at the cost of irregularity. So the question applies to those two places, not to the clustering levels.

Aside, on-topic: comp_pixel lines 173, 178, 180 mix axes. 173 is BL−TR (should be BL−TL, same column), 178 is TR−BL (should be TR−TL), 180 is BR−TR, a vertical difference inside dx (should be BR−BL). As written dy and dx share an anti-diagonal term with opposite signs.

kwcckw: I see, so using hexagonal tiling is more compact since each neighbor is nearer to the center, but i think it will be more complicated when we dealing with the geometries, such as coordinates, direction and distance related params since we will need an additional func to map the hexagonal coordinates.

But as you mentioned above, i think we probably can try that now since LLM may resolve most of the problem in converting the code now.

boris-kz: Yes so we don't need to do this now. 
But it says we have a bug:

Aside, on-topic: comp_pixel lines 173, 178, 180 mix axes. 173 is BL−TR (should be BL−TL, same column), 178 is TR−BL (should be TR−TL), 180 is BR−TR, a vertical difference inside dx (should be BR−BL). As written dy and dx share an anti-diagonal term with opposite signs.

frame_2D_alg/agg_recursion.py
```diff
@@ -138,25 +143,29 @@ def comp_H(_Nt,Nt, Link):  # tentative pre-comp, before comp N_?
     TT = base_comp(_N,N)[0] if full else comp_derT(_N.dTT[1],N.dTT[1])
     m,d = val_(TT, ttN_,1)
     L = CL(N_=[_N,N], dTT=TT,m=m,d=d,c=c,r=r, root=rL)
+    if full:  # pending review: add geometry before Copy_(L) below
```
boris-kz, line 146: this geometry adding should be in Copy_?
if full is redundant to
if N.typ > 1 and gv_(m* (c/r)* wN_ - ave*(r+cN_)):

kwcckw: We init L above without the geometry:

So we need to add it first:

Then the copied `L.H = [Copy_(L)] ` will be having those params.

frame_2D_alg/agg_recursion.py
kwcckw: > Aside, on-topic: comp_pixel lines 173, 178, 180 mix axes. 173 is BL−TR (should be BL−TL, same column), 178 is TR−BL (should be TR−TL), 180 is BR−TR, a vertical difference inside dx (should be BR−BL). As written dy and dx share an anti-diagonal term with opposite signs.

Yes, i think the suggestion makes more sense since it should be in orthogonal direction. The existing code has diagonal elements:

    dy__ = (
            (i__[2:, :-2] - i__[:-2, 2:]) * 0.25 +   # Line 117: BL - TR  (Anti-diagonal)
            (i__[2:, 1:-1] - i__[:-2, 1:-1]) * 0.50 + # Line 118: Bottom - Top (Vertical)
            (i__[2:, 2:] - i__[:-2, 2:]) * 0.25      # Line 119: BR - TR  (Right-column vertical diff)
    )
    dx__ = (
            (i__[:-2, 2:] - i__[2:, :-2]) * 0.25 +   # Line 122: TR - BL  (Anti-diagonal)
            (i__[1:-1, 2:] - i__[1:-1, :-2]) * 0.50 + # Line 123: Right - Left (Horizontal)
            (i__[2:, 2:] - i__[:-2, 2:]) * 0.25      # Line 124: BR - TR  (Vertical diff duplicated into dx)
    )
    
 The suggestion using only the orthogonal directional differences:
 
    dy__ = (
            (i__[2:, :-2] - i__[:-2, :-2]) * 0.25 +  # BL - TL (Left vertical)
            (i__[2:, 1:-1] - i__[:-2, 1:-1]) * 0.50 + # B  - T  (Center vertical)
            (i__[2:, 2:]   - i__[:-2, 2:])   * 0.25   # BR - TR (Right vertical)
    )
    dx__ = (
            (i__[:-2, 2:] - i__[:-2, :-2]) * 0.25 +  # TR - TL (Top horizontal)
            (i__[1:-1, 2:] - i__[1:-1, :-2]) * 0.50 + # R  - L  (Center horizontal)
            (i__[2:, 2:]   - i__[2:, :-2])   * 0.25   # BR - BL (Bottom horizontal)
    )

frame_2D_alg/agg_recursion.py
boris-kz: Ok, we can add those in initialization?

kwcckw: We can do that, but right now it's optional only if full is true:

    if full:  # pending review: add geometry before Copy_(L) below
        angl = [np.zeros(2) if A is None else A, np.sign(TT[1] @ ttN_[1])]

And the full is still relevant since it's true when tnF is false?

boris-kz: ok just "and" those ifs

frame_2D_alg/agg_recursion.py
boris-kz: That initial connectivity clustering in our scheme is actually optional, we can start CC with initial exemplars?  

Yes, mechanically and conceptually, with one real loss.

1. Mechanically: cluster_C consumes only rims — E.rim for the initial frontier (332), n.rim for expansion (352), N.Rt from comp_N_ (122) for exemplar weights (245). Nothing from cluster_N. And G scope isn't enforced anyway: 340 admits any rim neighbor into N_, so CC already runs on the flat link graph. G only supplies the exemplar pool (91), the density gate (90), and the root (376-377). cluster_N is itself exemplar-seeded below top level (285), so both stages start from the same seeds; e_ at 102 can go straight to cluster_C(root.Nt, e_).

2. Conceptually: the seed group a centroid needs is E plus E.rim — a radius-1 connectivity cluster, which CC already builds (330-332). So the precedence claim reduces to: links precede centroids; the flood-fill pass need not. CC-from-exemplars is itself a rim-constrained flood-fill, differing from CN only in acceptance criterion — match to the running mean (341) vs pairwise / rim-density match (293-298) — and in membership: soft overlapping root_ weights (353) vs exclusive fin. Merging stages doesn't skip connectivity, it collapses two criteria into one pass under the centroid criterion.

3. What the separate CN pass provides that merged CC doesn't:
- Chains. Pairwise acceptance keeps gradual drift (A~B~C, A≁C) as one cluster; centroid acceptance splits it into overlapping segments around exemplars, since a chain has no mean that matches its ends. Contiguous shapes are chains, categories are not. So CN-first is right where the input is dominated by contiguity, CC-first where by repetition. That's the discriminant, and it's per input: gate CN on summed link m being high while summed match-to-exemplar is low.
- Exclusive partition and the contrast fork: Lt, Bt (301, 305), root.Nt reform (319). CC has no negative fork and no boundary; without CN before CC the level has no Bt until the CN over centroids at 103.
- trans_cluster (269) only exists in CN.
- Cost: pairwise m is already computed, CN acceptance is free; CC pays one base_comp per node per iteration. The gate at 90 is the only surviving argument for CN as pre-filter, and it's cost, not validity.

4. Closure if dropped: cross_comp takes flat N_ (already what frame_H passes at 839), loop 89-93 collapses to one cluster_C call, gate 90 is replaced by 252 + 351. cluster_N stays at 103 over centroids, but nothing then consumes G_ as CC scope, so CN becomes a terminal per-level fork — partition, boundary, trans-cluster, feedback — not a seeding stage. The docstring's four stages become exemplars → CC → CP with CN as a parallel contour fork, and the README's "connectivity clustering seeds centroid formation" becomes "cross-comp seeds centroid formation". Note this scoping role is already not exercised across elevs: cross_comp's return is discarded at 839, G_ reaches the next elev only via Fr.Nt, and 835 takes t.N_, the raw edge Gs.

If by initial CN you mean trace_edge rather than cluster_N: same analysis, its unique output is the contour (Bt at 792).

Chat:

Yes. The mandatory precursor is **local cross-comparison**, not hard connectivity clustering.

Cross-comp forms the rims/links and node-level match density needed by `get_exemplars`. Those sparse exemplars can initialize centroid clustering directly. `cluster_C` already copies each exemplar into an initial centroid and expands it through the rim frontier, so it does not structurally require a prior `cluster_N` result.  

The sequence can therefore be:

`cross_comp -> get_exemplars -> cluster_C -> optional cluster_P -> higher cross_comp`

with connectivity clustering as a conditional pre-pass:

`cross_comp -> get_exemplars -> optional cluster_N -> cluster_C`

The distinction is important: you are removing **hard connectivity-cluster formation**, not connectivity. Both exemplar selection and centroid expansion still consume the local link graph. The exemplar is a local maximum of connectivity value, while membership is determined by direct match to the evolving centroid.

`cluster_N` remains useful where similarity is transitive but not centroid-coherent: curved chains, elongated contours, irregular/non-convex regions, or cases where hard boundaries should restrict competition among centroids. Direct exemplar-seeded centroid clustering is preferable for compact distributions where match to the evolving centroid remains strong.

At the `agg_recursion` entry this is especially plausible because the initial nodes are already PPs or comparable composed patterns, not raw pixels. The prior connectivity G currently serves mainly as a scope around `cluster_C`; it can be replaced by a frame-level `Ft` over the initial nodes, provided the same `Ct` output is passed to higher cross-comp. 

So the README claim should become roughly:

> Local connectivity selects exemplars. Connectivity clustering optionally precompresses or delimits subsequent centroid clustering.

rather than saying connectivity clusters necessarily seed centroid formation.

boris-kz: So we need to combine get_exemplars, cluster_N, and cluster_C in cluster_(), with incrementally added inclusion and merge criteria?

Fable:

The principle that unifies them: inclusion value of a node = lateral match (its links into the cluster) + vertical match (to the cluster's centroid), each term used once it's reliable, with their overlap discounted via r as everywhere else, since it's the same node content matched twice. Your commented-out 343-344 is exactly this. The stages fall out of term availability, not depth:

1. Seed: lateral only, single link to E — the centroid is E, so vertical would duplicate lateral.
2. Contiguity: lateral only, links into the growing cluster; nt_vt (293) and iM at 297 are already this. Exclusive, monotone: comparand doesn't move, nothing to revise.
3. Centroid: vertical term added once centroid support clears cost (351 gate). Two consequences follow, not choices: the comparand now moves with membership, so members must be re-tested (revision), and vertical match is graded, so membership is soft. Reachability still by rims.
4. Parallel: vertical only, for nodes with no links into the cluster — reachability dropped when global overlap warrants it (360 gate). That's cluster_P, the same ascent one step further, so it belongs in the same function as the last stage.

Merge is the same two terms cluster-to-cluster: lateral = shared links and members (the O at 392 is this), vertical = centroid comp (402). Absorption in stage 2 is merge on lateral alone; trans_cluster's trans-link m (274) is a lateral merge term, so it becomes a criterion inside rather than a post-pass. Merge candidates should ascend the same way: adjacent Cs first, all pairs only in stage 4.

Exclusivity resolves spatially: hard core = nodes included on lateral terms in stage 2 (fin), soft halo = later inclusions. Bt = contrast links out of the core, collected at every stage from members' negative rim links, no closure needed.

Closure of the contract chain:
- sum2G and sum2F merge into one weighted summation, weight 1 for core members; CN and CL outputs unify into one class carrying Nt, Lt, Bt plus membership weights.
- cross_comp calls cluster_() once per level; the per-G loop and the two get_exemplars calls disappear; stage gates are 252, 294, 351, 360, 364, 397 relocated inside.
- The vertical term needs a reliability gate beyond count: centroid not dominated by the candidate itself, so at least two members and centroid M above ave before it's added.

Chat:

It should be one staged `cluster_()` engine with a shared cluster state and progressively richer criteria. The possible paths are:

`E -> CN`
`E -> CC`
`E -> CN -> CC`

That preserves the option to start centroid clustering directly from extended exemplars rather than first completing connectivity clusters. This matches the intended sparse-to-fuzzy expansion from connectivity toward centroid clustering. 

1. **Exemplar/core stage**

   * Select the strongest uninhibited seed.
   * Grow by the strictest local connectivity criterion.
   * Produce an extended exemplar with `N_`, `L_`, `_N_` frontier, and provisional `B_`.
   * A collision with a stronger core inhibits the weaker seed; this is not yet cluster merging.

2. **Connectivity stage**

   * Add cluster-level connectivity density, span normalization, and possibly boundary support to the inclusion value.
   * Membership remains crisp.
   * When two fronts meet, apply an inter-cluster connectivity merge criterion; otherwise the connecting links become boundary evidence.
   * Exhausting the frontier completes `B_`; only here should `sum2G`, `trans_cluster`, and root replacement run.

3. **Centroid stage**

   * Start from the whole extended exemplar `E.N_`, not just its original seed.
   * Add centroid match and overlap cost to the inherited core-connectivity evidence.
   * Permit multiple memberships.
   * Frontier collisions normally create overlap; merge only when the centroids are redundant.

So `get_exemplars` becomes the seed ordering/core-initialization part of `cluster_`; `cluster_N` and `cluster_C` become stage-specific inclusion, collision-resolution, and finalization policies. The common traversal should carry the same proto-cluster state through all stages instead of each function restarting from raw exemplar nodes. The earlier code already treated exemplar selection, connectivity clustering, and centroid expansion as successive clustering stages, but implemented them as separate passes. 

More precisely, the incremental structure is:

```python
include = core_connectivity

if fN:
    include += cluster_connectivity
    merge   += inter_cluster_connectivity

if fC:
    include += centroid_match - overlap
    merge   += centroid_redundancy
```

`cluster_P` should remain separate: it changes the process from local frontier diffusion to global parallel competition among all centroids.

boris-kz: We do need separate get_exemplars for parallelization, but cluster_N and cluster_C can be merged?

kwcckw: > That initial connectivity clustering in our scheme is actually optional, we can start CC with initial exemplars?

We actually start with CC now? In cross_comp, we actually split each G and get global CC before the cluster_N in the last line.


> We do need separate get_exemplars for parallelization, but cluster_N and cluster_C can be merged?

So i guess it will become something like a sequential pipeline to pack those functions? But cluster_N and cluster_P could be optional, the possible paths will be something like this?

       cluster_N
       cluster_N -> cluster_C
       cluster_N- > cluster_C -> cluster_P
       cluster_C
       cluster_C -> cluster_P

boris-kz: >In cross_comp, we actually split each G and get global CC before the cluster_N in the last line.

That's not a start, the start is vect_edge, and PPs are connectivity clusters
That won't change, but in cross_comp it will be cluster_N->cluster_C
Main difference is that CNs will not have to terminate before centroids kick in

PR #250, opened 2026-08-27
8/27/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -259,6 +281,125 @@ def get_exemplars(N_,_r,_c):  # multi-layer non-maximum suppression -> sparse se
...
+    G_ = cluster_n(Ft, E_, r+1, c)
+    if G_:
+        CC_ = cluster_c(Ft, G_,r+2,c)
+        if CC_: G_ = CC_     
+    _m,_d,_tt,_c,_r = sum_vt(G_, fm=1)
+    FV_(CoF.get(),_tt,_c,_r)  # r per deeper cross_comp?
+    if gv_((_m *_c *(wcC+wcN)) / (_r+(ccN+ccC)) * ((len(G_)-1)*wL) - ave):
+        rG = Ft.root; rG.H += [Copy_(Ft)]; Nt=rG.Nt; Nt.N_=G_; Nt.dTT=_tt; Nt.c=_c; Nt.r=_r
+        rG.dTT=_tt; rG.c=_c; rG.r=_r; rG.m, rG.d = val_(_tt, (ttcN+ttcC))
+        return Ft
```
kwcckw: > That's not a start, the start is vect_edge, and PPs are connectivity clusters
That won't change, but in cross_comp it will be cluster_N->cluster_C

So if we have cluster_N -> cluster_C, that cluster_N should be default and cluster_C should be optional, as in my code above?

    E_ = get_exemplars(_N_,r,c)
    G_ = cluster_n(Ft, E_, r+1, c)
    if G_:
        CC_ = cluster_c(Ft, G_,r+2,c)

I added `cluster_` to run `get_exemplars`, `cluster_N` and `cluster_C` all in a single function. In cross_comp, we can skip the existing initial cluster_C and run only this `cluster_` after the comps? Something like this:

    def cross_comp(root, N_, rr):  # agg+: refine by CC,exe -> cross_comp, nF: core|contour?
    
        if Lt := comp_N_( proj_L_(combinations(N_,2), root,rr), rr):
            L_, TT, c, r, V = Lt
            if gv_(val_(TT,ttcN) * (c*wcN /(r*ccN)) * ((len(L_)-1)*wL) - ave):  # return +ve, store -ve gate Vs、
                Ft = cluster_(root.Nt,list(set([n for L in L_ for n in L.N_])),r,c)  # sum2G-> agg+
                L = len(root.Nt.N_)
                if Ft and gv_(val_(TT,ttX) * (c*wX /(r*cX)) * ((L-1)*wL) - ave):
                    cross_comp(root, root.Nt.N_, root.r)

PR #251, opened 2026-08-28
8/28/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -259,6 +281,123 @@ def get_exemplars(N_,_r,_c):  # multi-layer non-maximum suppression -> sparse se
...
+                _C_ = C_
+            else: out_+=C_; break  # converged
+        if out_:
+            for n in [N for C in out_ for N in C.N_]:  # exemplar V + sum n match_dev to Cs, m* ||C rvals:
+                n.exe = (n.d if n.typ==1 else n.m) + sum(rt[1] for rt in n.root_) > ave
+        return out_
+
+    E_ = get_exemplars(_N_,r,c)
+    N__ = cluster_n(Ft, E_, r+1, c)
+    G_ = cluster_c(Ft, N__, r+2, c)
```
kwcckw, line 381: > Main difference is that CNs will not have to terminate before centroids kick in

So i updated the code to output N__ (group of Ns) from the connectivity clustering in cluster_n, then the termination is done in cluster_C. For Bt(B_) and Lt(L_), i guess we only can retrieve them via C.N_ now, which i pack them in N.LB_ (list of (L_,B_)). Each terminated G from centroid clustering will compute their Lt and BT in the final step, after the convergence, as in my code below:

        if G_:
            for G in G_:  # form Lt and Bt from C.N_'s L_ and B_
                L_, B_ = [], []
                for N in G.N_: L_ += N.LB_[0]; B_ += N.LB_[1]
                for (F_,nF) in zip((L_, B_),('Lt','Bt')):
                    F_ = list(set(F_)) or []; tt,fc,fr = sum_vt(F_,wTT=ttcN) if F_ else (np.zeros((2,9)),0,0)
                    Ft = CF(N_=F_,nF=nF,dTT=tt,m=(vt:=val_(tt,wTT,1))[0],d=vt[1],c=fc,r=fr)
                    setattr(G,nF,Ft)  # set Lt and Bt
                trans_cluster(G)

boris-kz: Thanks, sorry I haven't checked that yet. I should be back to work in a few days

kwcckw: Great, I will continue to improve the new cluster_func since it may be still buggy

PR #252, opened 2026-08-29
8/29/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -259,6 +281,124 @@ def get_exemplars(N_,_r,_c):  # multi-layer non-maximum suppression -> sparse se
...
+                _C_ = C_
+            else: out_+=C_; break  # converged
+        if out_:
+            for n in [N for C in out_ for N in C.N_]:  # exemplar V + sum n match_dev to Cs, m* ||C rvals:
+                n.exe = (n.d if n.typ==1 else n.m) + sum(rt[1] for rt in n.root_) > ave
+        return out_
+
+    E_ = get_exemplars(_N_,r,c)
+    N__ = cluster_n(Ft, E_, r+1, c)
+    G_ = cluster_c(Ft, N__, r+2, c)
```
kwcckw: So i think we probably can combine the connectivity clustering and centroid clustering in a single traversal instead of the existing 2 steps process?

After we cluster all rim mediated Ns, that group of Ns will be eval for centroid clustering to further refine their membership? And the centroid clustering should be optional for those group of non-stable Ns ?

kwcckw: Alright, i will combine them next. Also for instability, it will be measured by N_'s d?

boris-kz: I don't know if we need d for convergence, dM is better

kwcckw: Right now we are using both dM and dD, but not dD alone:

          if _C in _root_:
              i =_root_.index(_C); up += abs(n._root_[i][1]-m) + abs(n._root_[i][2]-d)
          else: up += m+abs(d)
          
 So we should use M only for convergence?
 
 That D is for the stable test only?
 
      if D < Avd: out_+=[C]  # output if stable

PR #253, opened 2026-08-30
30-8-2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -259,6 +281,111 @@ def get_exemplars(N_,_r,_c):  # multi-layer non-maximum suppression -> sparse se
...
+            for _C in _C_:  # C.m,d /rTT? sort / sum(_C.m_)?
+                N__,n_,m_,d_,M,D,T,R,dTT,up = [],[],[],[],0,0,0,0, np.zeros((2,9)),0  # /C
+                for n in _C.N_+_C._N_:  # current + frontier
+                    if n not in N_: N_+=[n]; n.root_,n._root_ = [],[]
+                    dtt,_ = base_comp(_C,n)  # or comp_N, decay?
+                    m,d = val_(dtt,ttcC,1); dTT+=dtt
+                    lm= lc= lr= 0  # lateral connectivity term                
+                    for l in n.rim:  
+                        vt = nt_vt(*l.N_)
+                        lm+=vt[0]; lr+=vt[2]; lc+=vt[3]  # connected to existing N
```
kwcckw, line 330: So i merged both cluster_N and cluster_C, and most of the code structure is actually from cluster_C.

The main difference is during the addition of N to C in the refinement step, where it is based on the connectivity eval (lateral connectivity's m via rim and the comp m via base comp) instead of the original default addition in cluster_C.

boris-kz: cluster() should be a core loop, not a helper?
also:

1. `G_ = cluster_(Ft, E_, r+1, c)` calls the outer function, not inner `cluster`. Self-recursion: get_exemplars of a shrinking E_ bottoms out at the single-N fallback branch, which returns the same set forever. The one-underscore name distance between `cluster_` and `cluster` invites exactly this.
2. `nt_vt`: `if l.m > 0 or l.m > 0` — second disjunct duplicated, should be `l.d > 0`.
3. `cluster_P(out_, Ft)` — signature is `(_C_, _c, root)`; Ft binds to `_c`, `root` missing.
4. `r = _r + R/T`: T=0 is now reachable — the new per-n gate can reject all of `_C.N_+_C._N_` (impossible in cluster_C, which had no gate).
5. Registry closure. `cluster_` is not in meta_code `_names`, and `cluster_N`/`cluster_C` are: (a) parse_funcs finds no defs for them → `nF_[7],nF_[8]` stay None → meta_code's import-time build loop crashes on `ast.iter_child_nodes(None)`; (b) even if registered names are patched, `cluster_` won't be recompiled by inject_oF_, so its gv_ calls run with i=None and the first failed gate crashes at `gV_[None]`; (c) FV_(CoF.get()) attributes value to the caller's oF since untraced functions don't push a CoF context; (d) the 14-constant unpacks at agg_recursion:53-55 change arity. The merge is exactly the operation clust_oF_ exists to perform at runtime; doing it in source requires consolidating one oF entry with one weight set. The draft instead hardcodes sums (`wcC+wcN`, `ttcN+ttcC`): each ttc* has fixed mean, so `val_(_tt, ttcN+ttcC)` doubles the scale of rG.m,d relative to every other val_ call.

**Connectivity term**

6. `for l in n.rim: vt = nt_vt(*l.N_)` — each call sums the union of both endpoints' rims, and n.rim is inside that union for every l, so n's own rim is re-counted |rim| times. lm, lr, lc inflate roughly quadratically with degree.
7. lm is not "connected to existing N": nt_vt has no membership filter, so it measures generic neighborhood density, not link strength into `_C`. Structural adjacency to the cluster exists via frontier construction, but its weight doesn't enter the gate — a node hanging off the cluster by one weak link inside an unrelated dense clique passes. The commented cluster_C predecessor (`n.rim & _C.L_`) was membership-restricted; the only restricted term left is iM, active only past len(n_)>9.
8. Cost form `ave*(r + lr*lc)`: product of summed r and summed c on the cost side violates the project convention (c multiplies value, r adds to cost). lr*lc grows quadratically vs lm's linear growth, so dense nodes are net penalized — inverse of the term's purpose. Also m (intensive similarity) + lm (extensive sum) with no normalization: lm dominates.
9. r leakage: `r = _r + R/T` is assigned per _C after its n-loop, and the next _C's per-n gate reads it — _C[i]'s admission threshold depends on _C[i-1]'s redundancy. Harmless in cluster_C only because nothing per-n consumed r there.
10. The span/iM refinement is flood-fill logic where sequential growth is the semantics; in a membership sweep over `_C.N_+_C._N_` iteration order is arbitrary, so admission is order-dependent — contrary to the parallel-refine contract cluster_P exists to enforce.

**Fork formation and trans_cluster on centroids**

11. G's are CL (sum2F, typ=2). CL has no Lt/Bt sub-forks and its `L_` is a plain empty attr, not the CN property over Lt. `setattr(G,'Lt',ft)` creates an orphan; `trans_cluster` iterates `G.L_ == []` — the whole block is dead. cluster_N's Gs were CN via sum2G.
12. Even wired up, trans_cluster presumes exclusive roots: it resolves `tL.N_[i].root.root` and merges via `rt0.N_ += rt1.N_; add_Nt`. The centroid path never sets member `.root` (no add_Nt/sum2G), so those chains are stale from the previous level; fuzzy membership lives in root_ lists, where exclusive-root merge is the wrong operation.
13. Lt/Bt classification is by link m/d sign only, no internal/external test — L_ admits links whose far endpoint is outside G.N_. cluster_N's L_ were internal by flood-fill construction, B_ external by contrast; the topology defining core vs contour is dropped.

**Contract and return path**

14. If this replaces cluster_N, stage 2 of the docstring (exclusive connectivity partition, boundary clustering, divisive sub+, sum2G→comb_Ft→cross_comp per G, Xt) has no implementation — only its residue grafted dead onto centroids. Currently the draft is neither a cluster_C replacement nor a full merge; pick one, then the cross_comp call sites: both do their own get_exemplars, while cluster_ selects internally — passing pre-selected E_ double-applies NMS.
15. Return: Ft mutated in place but Ft.m,d never recomputed from _tt — cross_comp unpacks stale Ct.m. Also `rG.H += [Copy_(Ft)]` stores a CF snapshot where cross_comp stores CN root snapshots; H levels become type- and content-heterogeneous under add_H/comp_H.
16. Seeding via `sum2F([N],Ft,m_=[1],d_=[0])` appends `[[seedC,1,0]]` to N.root_ (fC branch) besides the manual _root_ entry; on first-cycle exit those stale entries survive and inflate n.exe and next-level exemplar w. It also replaces E.m with placeholder 1, so exemplar strength no longer biases the first C-eval — compounded by the eval change from cluster_C's multiplicative `M*(_C.m+…)` to additive `(_C.m+M)*…`.
17. `up` drops the d-drift term but convergence is still gated by Avd; either restore d or gate by Ave.
18. The cluster_P trigger uses `val_(DTT,fd=1)` where dTT accumulates over all evaluated nodes including gate-rejected ones (`dTT+=dtt` precedes the gate), and is un-c-weighted, replacing c-weighted mat+dif — trigger inflated by rejected mass. Minor: `D/T < Avd` makes stability size-invariant where cluster_C's extensive D made large C's harder to stabilize — fine if intended, but it interacts with 18 (both gates now diverge from cluster_C in opposite directions).

Vestigial: `n.fin` set, never read; Rt recompute inside `cluster` after get_exemplars already consumed it; `_n not in n_` in the frontier comprehension is redundant with the set subtraction and O(n²).

kwcckw: Thanks, the feedback is useful, the function mostly is still buggy. I will go through them later.

> cluster() should be a core loop, not a helper?

Yes, so i guess we can unpack the function.

PR #254, opened 2026-08-31
8/31/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -259,6 +281,113 @@ def get_exemplars(N_,_r,_c):  # multi-layer non-maximum suppression -> sparse se
...
+            for N in C.N_:
+                for L in N.rim:
+                    m,d,_,_ = nt_vt(*L.N_)
+                    if m > ave * (r-1):  L_ += [L]   
+                    elif d > avd * (r-1): B_ += [L]    
+            ft_ = [[C.N_,'Nt',C.dTT, C.c, C.r]]
+            for i,(F_,nF) in enumerate(zip((L_,B_),('Lt','Bt'))):
+                F_ = list(set(F_)) or []; tt,fc,fr = sum_vt(F_,wTT=ttcN) if F_ else (np.zeros((2,9)),0,0)
+                ft_+= [[F_,nF,tt,fc,fr]]
+            G = sum2G(ft_, ttcN, root=Ft)
```
kwcckw, line 378: So i unpacked cluster and added sum2G in the termination. I think the terminated C from centroid clustering should be served as the G.Nt, and we need additional L_ and B_ from C.N_ to form G.Lt and G.Bt via the sum2G here.

boris-kz: Yes,  L_ and B_ should be from N_ rims

kwcckw: Okay, and the current process is a bit different because centroid clustering packs only Nt, and C is a CL, so we would need additional steps to retrieve L_, B_ through  C.N_, and then compute G using the sum2G from the converged C's params, L_ and B_.

boris-kz: Actually in this scheme Cs are just extended Ns, with L_ and B_, so they should be CNs?

kwcckw: Now the cluster_C is backwards because C is a CL, which need additional extension to CN later. So in cluster_C or cluster_P, we should sum C into G with the same process in sum2G (using comb_ft)?

My current code is summing them in the final stage, after all Cs are converged because i don't think we need C as CN during the refinement stage?

boris-kz: Okay we don't need full CN for refining, I guess that final conversion is fine. But the base_comp for refining should be on Lt, not Nt? Because that Lt is produced by cross_comp, which determines all clustering.

kwcckw: Right now it's comparing n to C:

I think it's still relevant to comp n since we are clustering n instead of L? And there will be overlapping values if we need to sum Lt from n.rim before comp them.

Their rim or L could be useful for the lateral connectivity?

                lm= lc= lr= 0  # lateral connectivity term                
                for l in n.rim:
                    if l.N_[0] in n_ or l.N_[1] in n_:  # connected to existing N?
                        vt = nt_vt(*l.N_)
                        lm+=vt[0]; lr+=vt[2]; lc+=vt[3]

boris-kz: It doesn't matter, that Lt is per n, not per L. It's just the value of top composition level of the n.

kwcckw: Okay, then we actually can compare C to n.Rt since n.Rt is summed from their rim. I guess we should comp Lt that sum from both positive and negative rim right?

PR #255, opened 2026-09-01
9/1/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -259,6 +262,109 @@ def get_exemplars(N_,_r,_c):  # multi-layer non-maximum suppression -> sparse se
...
+        C._N_= list({n for l in N.rim for n in l.N_ if n is not N})  # init frontier
+        _C_ += [C]
+    out_ = []; cr = r +_r
+    while True:  # reform C_
+        C_, cnt,rdn,DTT,Up = [],0,0,np.zeros((2,9)),0; Ave = ave*(cr+ccC); Avd = avd*(cr+ccC)  # both Ave and Avd should be computed with loop-local r?
+        for _C in _C_:  # C.m,d /rTT? sort / sum(_C.m_)?
+            N__,n_,m_,d_,M,D,T,R,dTT,up = [],[],[],[],0,0,0,0, np.zeros((2,9)),0  # /C
+            for n in _C.N_+_C._N_:  # current + frontier
+                if n not in N_: N_+=[n]; n.root_,n._root_ = [],[]; n.Rt = sum2F(n.rim); n.Rt.root = n
+                dtt,_ = base_comp(_C,n.Rt)  # or comp_N, decay?
```
kwcckw, line 304: > It doesn't matter, that Lt is per n, not per L. It's just the value of top composition level of the n.

So i think we can just comparing `_C` with `n.Rt` here since Rt is the top composition of n after the new comp. Also with this Rt comp, we no longer need the lateral connectivity values from n.rim since the rim params should be included in Rt that used in the comp now.

boris-kz: Rt is external to n, nothing to do with composition. Top composition is actually an evaluation term, not a comparand. So base_comp(_C,n) should be fine, sorry.

kwcckw: I see, so the top composition is from Lt, the internal links, but it sums to n's params anyway, so using n should be fine here.

Also for the lateral connectivity via n.rim, the idea is that we need to add additional match when there's any mediating links between n and the existing clustered n_?

boris-kz: not sure what you mean, rim is mediating links

kwcckw: In your existing cluster_C, this commented section right after the base_comp:

     # lm= lc= lr= 0  # lateral connectivity term
     # for l in (n.rim & _C.L_): lm+=l.m; lc for l.c; lr+=l.r
     
So this lm should be added to m as the connectivity match?

boris-kz: this is actually redundant to base_comp(_C,n), it's a more detailed version that's only feasible for relatively ns. Current sequence is switching to incrementally more complex clustering criteria: single linkage -> average linkage -> link-constrained centroid match -> tile-constrained centroid match. That rim overlap is for average linking

kwcckw: I see, the existing `n` inclusion is not default after we merge it with cluster_N, so it's actually just using the comp between C and n: `dtt,_ = base_comp(_C,n)` to add n into the cluster C.

For `link-constrained centroid match`, it's the existing centroid clustering process, and the frontier is determined by n's rim?

boris-kz: Right

kwcckw: Okay, so in this merged version, n inclusion is determined by m, and we have another connectivity refinement via span:

                    span = np.sqrt(len(n_))  # approx span
                        iM = sum([l.m for l in n.rim if (l.N_[0] if l.N_[1] is n else l.N_[1]) in n_])
                        if iM / (span*decay) < ave * (cr-1): continue  # normalized N-to-N_ match, sum for sum2G?
                        
I think this is redundant to the base_comp between C and n too? Since it's eval using the same rim.

PR #256, opened 2026-09-02
9/2/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -532,11 +642,12 @@ def comb_Ft(Nt, Lt, Bt, root,wTT):  # from sum2G, default Nt
     if any(dF_): sum2F(dF_,G.Xt)  # cross-fork covariance
     add_Nt(G)  # add kern,ext, doesn't affect comp_F
     if Lt:
-        L_ =Lt.N_; L_,pL_ = [],[]; [L_.append(L) if L.typ==1 else pL_.append(L) for L in L_]
+        Link_ =Lt.N_; L_,pL_ = [],[]; [L_.append(L) if L.typ==1 else pL_.append(L) for L in Link_]
         if pL_ and sum_vt(pL_,fm=1,wTT=wTT)[0] *wN > ave*(cN * np.mean([L.r for L in pL_])):
             for L in pL_: L_ += [comp_N(*L.N_, G.r,L.c,1, L.angl[0], L.span)]
             sum2F(L_,Lt); add2F(G, Lt, merge=2)
-    else: L_ = [l for n in G.N_ for l in n.L_]; G.L_=L_  # no sum2F?
+        G.Lt = Lt
```
kwcckw, line 649: I'm not sure if this is intended, but i think we missed the assignment of Lt? We still need to update G.Lt here in comb_ft?

frame_2D_alg/agg_recursion.py
```diff
@@ -259,6 +264,102 @@ def get_exemplars(N_,_r,_c):  # multi-layer non-maximum suppression -> sparse se
...
+        for _C in _C_:  # C.m,d /rTT? sort / sum(_C.m_)?
+            N__,n_,m_,d_,M,D,T,R,dTT,up = [],[],[],[],0,0,0,0, np.zeros((2,9)),0  # /C
+            for n in _C.N_+_C._N_:  # current + frontier
+                if n not in N_: N_+=[n]; n.root_,n._root_ = [],[]
+                dtt,_ = base_comp(_C,n)  # or comp_N, decay?
+                m,d = val_(dtt,ttcC,1); dTT+=dtt
+                if m > ave * cr:
+                    # pending review: refine by rim is redundant?
+                    span = np.sqrt(len(n_))  # approx span
+                    if span > 3:  # refine by rim connectivity / norm span
```
kwcckw, line 311: > this is actually redundant to base_comp(_C,n), it's a more detailed version that's only feasible for relatively ns. Current sequence is switching to incrementally more complex clustering criteria: single linkage -> average linkage -> link-constrained centroid match -> tile-constrained centroid match. That rim overlap is for average linking

So i guess the connectivity refinement via span is also redundant since it's using the rim too?

                span = np.sqrt(len(n_))  # approx span
                    iM = sum([l.m for l in n.rim if (l.N_[0] if l.N_[1] is n else l.N_[1]) in n_])
                    if iM / (span*decay) < ave * (cr-1): continue  # normalized N-to-N_ match, sum for sum2G?

boris-kz: this is from new Fable 5.1:

Read against current agg_recursion/meta_code and the prior review.

Resolved from prior: self-call, nt_vt disjunct, T=0 guard, nt_vt overcount and lm cost form (lm dropped), per-n r leakage (cr), Gs as CN via sum2G, Ft.m/d recompute.

**Closure-blocking**

1. `cluster_P(out_, Ft)` still binds Ft to `_c`. `_c` is unused inside cluster_P — drop it there.

2. trans_cluster: `.root` of any not-yet-packaged node is Ft (set by comb_Ft when the node was formed). Ft is a truthy CF, so `rt1 = Ft` → `rt0.N_ += Ft.N_` merges the whole level into G, and `add_Nt(rt0)` re-roots every node to it; in the other direction `add_Nt(Ft)` fails on `.box`. Latent in cluster_N for unclustered nodes; here certain, since trans_cluster runs per G inside the packaging loop before later Cs are packaged. Independently, under graded membership the op is wrong: shared members trans-link to both Cs, so all overlapping Cs collapse into one — cluster_P's O-merge done exclusively. Drop it, or restate as frontier extension inside the loop (trans-linked sub-nodes' parents join N__) and leave redundancy to cluster_P.

3. Up-branch clears root_ for all of `_C.N_+_C._N_`, which includes members of Cs sent to out_ this cycle. Those memberships survive only in `_root_`, overwritten next cycle; cluster_P's `_md__` init and next-level get_exemplars see none of them. Clear only entries whose C is in `_C_`. Same for the 0-cluster fallback `out_ = _C_`: restore from `_root_`. Seeding: sum2F(fC) already appends `[C,1,0]` to root_ besides your manual `_root_` line; replace it with `N._root_ = N.root_; N.root_ = []`.

4. add_Nt writes bare Cs into G.root_; node-side consumers (get_exemplars, cluster_P:384, cross_comp:117) expect [C,m,d]. Next-level get_exemplars on G_ subscripts a CL. G is a new node, root_ should start empty; member centroids belong in G.C_ if wanted.

**Loop semantics**

5. Stability. `D/T < Avd` is homogeneity, not convergence: a seed plus matching neighbors is homogeneous on cycle 1, exits to out_, and its frontier is never used — the Cs that keep extending are the incoherent ones. Exit should be per-C drift: `up` below Ave → out_. D/T should gate extension (build N__ only while coherent), so coherent Cs grow and incoherent ones contract through the m gate. The global Up test then only early-stops.

6. `up` counts admits only: a prior member failing `m > ave*cr` vanishes without registering, so a C shrinking to nothing reads as converged. Accumulate for every evaluated n (rejected prior member = full loss). Also Up is m-only but gated by Avd → Ave.

7. iM ("redundant?"): not as a criterion — frontier is scope, this is the only link-strength term, and centroid match is not pairwise match. Redundant as a separate hard gate with sqrt-len normalization: fold it into m (mean link match into `_C.N_`, added or multiplied) so one gate decides. Either way test against `_C.N_`, not the growing `n_`: as written admission depends on sweep order, and the first 9 admits per cycle skip it.

8. r enters each gate 2–3 times: Ave = ave*(cr+ccC), then ×(r+_C.r+ccC·len) where r = _r+mean(n.r) and _C.r ≈ mean(n.r). Same at the cluster_P trigger and Up test (cr in Ave and in the multiplier). One r per gate. Init `cr = r+_r` = 2r+1 vs `_r+mean(n.r)` after cycle 1.

9. `_C.m + M`: _C.m = sum(prior m_) ≈ prior M, so the term is ≈2M; and seed _C.m = 1 drops E's strength, which cluster_C's multiplicative form kept.

10. sum2F(fC) sets C.m = sum(m_); base_comp then compares that membership total against n.m, a node's own match. Keep C.m = val_(TT) for comparison, store the sum separately.

11. cluster_P trigger: DTT includes gate-rejected nodes (`dTT+=dtt` precedes the gate); `val_(DTT,fd=1)` uses default wTT while everything else here uses ttcC; and olp on the cost side means more overlap → less likely to trigger the stage whose job is resolving overlap (docstring: "if significant global overlap").

**Packaging**

12. Lt/Bt by nt_vt sign only: every link classifies identically from every C, so there is no C-specific core/contour. Internal = far endpoint in C.N_ (or its root_ entry for C above threshold); Bt = external with d > avd·r. nt_vt was the flood-fill growth criterion; with membership decided by centroid, the link's own m/d is the classifier and nt_vt can go.

13. `ave*(r-1)` thresholds use r = last _C's loop-local value. Use C.r.

14. Extension outside Ft (`if n not in N_`) plus add_Nt's exclusive `.root` re-roots nodes of other Gs/Fts into this G, and fuzzy members get root = last packaged G. If .root stays exclusive, set it once after all Cs are packaged, by strongest membership.

**Untraced contract changes**

15. cross_comp: the line-91 per-G refine loop is now redundant with sum2G's sub+ (cross_comp → cluster_ on G.Nt); lines 102–103 become `cluster_(root.Nt, N_, r, c)` with get_exemplars removed. cross_comp reduces to comp_N_ over G_ → cluster_.

16. meta_code `_names`/`typ_`: 'cluster_N','cluster_C' → 'cluster_' (13 entries), unpacks at 53–55 to match. Inside cluster_: wcN/ccN (trans_cluster) and ttcN (packaging sums, sum2G fTT) → the single merged set; wcP/ccP at the trigger is fine as callee cost.

17. Docstring 21–26 and 77–84 describe 4 stages; the code has 3, with connectivity folded into frontier+iM. Minor: reset root_ over Ft.N_, not just _N_ (stale prior-level entries reach olp and add_Nt); the Rt recompute before get_exemplars duplicates comp_N_:122.

kwcckw: Thanks, some of them are useful.

> up counts admits only: a prior member failing m > ave*cr vanishes without registering, so a C shrinking to nothing reads as converged. Accumulate for every evaluated n (rejected prior member = full loss). Also Up is m-only but gated by Avd → Ave.

That C will be removed since n_ is empty, but it will be considered as converged because the else here will be true:

        if Up * (wcC*len(C_)) > Avd * (cr+ (ccC*len(C_))):
            for n in N_: n._root_ = n.root_
            for n in set([_n for _C in _C_ for _n in _C.N_ + _C._N_]): n.root_ = []
            _C_ = C_
        else: out_ = merged if (merged:= out_ +C_) else _C_; break  # converged
        
So i guess we need a same process in cluster_P, where we need to identify the case of divergence?

> iM ("redundant?"): not as a criterion — frontier is scope, this is the only link-strength term, and centroid match is not pairwise match. Redundant as a separate hard gate with sqrt-len normalization: fold it into m (mean link match into _C.N_, added or multiplied) so one gate decides. Either way test against _C.N_, not the growing n_: as written admission depends on sweep order, and the first 9 admits per cycle skip it.

This is actually the connectivity refinement via span i mentioned above. So if this is redundant now, the merged function is actually just the cluster_C, but with additional `m>ave * r` eval when we include new n.

boris-kz: I guess it's cluster_C because it can cover all criteria, we just need to make them selective and refinement optional.

On related research:
Mapped to the elements of cluster_ under review, closest first.

**1. Seed set expansion (overlapping community detection).** Structurally the nearest match to exemplars → rim-frontier growth → overlapping Cs → parallel refine. Whang, Gleich and Dhillon find good seeds, then expand each seed set in parallel using personalized PageRank to optimize conductance; the later version adds neighborhood inflation, i.e. seeding with the ego-net rather than the node, which is your E + rim init. Their criterion is purely structural (conductance), no attribute centroid; useful mainly for the seeding question, since seed quality is the axis they studied. Local expansion goes back to Andersen, Chung and Lang's local graph partitioning with PageRank vectors.

**2. SNIC (Achanta & Süsstrunk 2017)** is the hard version of link-constrained centroid growth and answers the iM question. A priority queue is initialized with the K seeds and always pops the element with the smallest distance to its centroid; the popped pixel takes that label, the centroid is updated with it, and its unlabeled neighbors are pushed with their distance to the centroid. Adjacency is scope only; centroid distance is the sole criterion; connectivity is enforced by construction, with no post-processing, and it is non-iterative. Global best-first admission across all (C, frontier n) pairs is also the standard fix for sweep-order dependence: Mehnert and Jackway showed that seeded region growing gives different segmentations under different processing orders, and removed the dependency by parallel processing with re-examination.

**3. FCM termination** supports drift over homogeneity as the exit test. Bezdek's FCM stops on change in the membership matrix between iterations, with termination on successive cluster centers as the cheaper alternative when the data count greatly exceeds the cluster count. Nobody stops on within-cluster dispersion.

**4. Spatially constrained FCM** is the graded form of the link constraint. Zhang and Chen add a penalty term with spatial neighborhood information to the objective, biasing toward piecewise-homogeneous labeling; Chuang's sFCM builds a new membership from the neighbors' memberships, which for a homogeneous region leaves the membership unchanged and for an outlier pixel reduces it. That is "fold iM into m", with neighbor membership in C rather than raw link match.

**5. Exemplars.** Density peaks: centers have locally higher density and are relatively far from any object with higher density, which is get_exemplars' Rt.m plus rim inhibition. Known failure is the one cluster_P exists for: multiple peaks in one cluster split it into subclusters.

**6. Lt/Bt.** Felzenszwalb–Huttenlocher's boundary predicate is the missing internal/external test: two components merge if the edge between them is small compared to their minimum internal difference. For link-side clustering, Ahn, Bagrow and Lehmann cluster links rather than nodes and show overlap and hierarchy are two faces of the same issue — the closest precedent for Bt as a link cluster and for overlap arising from links.

**7. Merge in cluster_P.** Mode-seeking over-fragments and merges after: medoid shift can fail to cluster points of the same mode, and quick shift trades off under- and over-fragmentation explicitly; Paris and Durand build the hierarchy by topological persistence. Persistence (merge when the saddle between modes is shallow relative to the modes) is a principled replacement for the ad hoc O-gate.

**8. Linkage ladder.** Single, average and centroid are Lance–Williams parameter settings, but centroid and median methods violate reducibility, so the hierarchy can contain inversions — the formal version of your 2→3 criterion-validity shift. Contiguity-constrained clustering (SKATER/REDCAP) is the hard-constraint relative: REDCAP merges units by contiguity-constrained agglomeration, converts the result to a spanning tree, then partitions it.

No precedent found for trans-level links merging or extending clusters at the parent level, nor for the core+contour object as the unit of the next comparison; the nearest analogs are multilevel coarsening schemes, which carry no derivative structure.

kwcckw: > I guess it's cluster_C because it can cover all criteria, we just need to make them selective and refinement optional.

We actually have several refinements now：

1. refine by m (n inclusion in C)
2. refine by rim connectivity  (the iM)
3. refine by cluster_P (all membership)

The rim connectivity and cluster_P are optional, but they are done in each main iteration. I guess cluster_P should be done once, right after the main refinement?

kwcckw: > Adjacency is scope only; centroid distance is the sole criterion

I guess this is what we are doing now? We use rim to extend the frontier, and the refinement is done via the comp to the centroid only. So that eval using the rim connectivity can be removed?

              span = np.sqrt(len(n_))  # approx span
                  iM = sum([l.m for l in n.rim if (l.N_[0] if l.N_[1] is n else l.N_[1]) in n_])
                  if iM / (span*decay) < ave * (cr-1): continue  # normalized N-to-N_ match, sum for sum2G?

PR #257, opened 2026-09-03
9/3/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -532,11 +642,11 @@ def comb_Ft(Nt, Lt, Bt, root,wTT):  # from sum2G, default Nt
     if any(dF_): sum2F(dF_,G.Xt)  # cross-fork covariance
     add_Nt(G)  # add kern,ext, doesn't affect comp_F
     if Lt:
-        L_ =Lt.N_; L_,pL_ = [],[]; [L_.append(L) if L.typ==1 else pL_.append(L) for L in L_]
+        Link_ =Lt.N_; L_,pL_ = [],[]; [L_.append(L) if L.typ==1 else pL_.append(L) for L in Link_]
         if pL_ and sum_vt(pL_,fm=1,wTT=wTT)[0] *wN > ave*(cN * np.mean([L.r for L in pL_])):
             for L in pL_: L_ += [comp_N(*L.N_, G.r,L.c,1, L.angl[0], L.span)]
-            sum2F(L_,Lt); add2F(G, Lt, merge=2)
-    else: L_ = [l for n in G.N_ for l in n.L_]; G.L_=L_  # no sum2F?
+            sum2F(L_,Lt); add2F(G, Lt, merge=2); Lt.N_ = L_
```
kwcckw, line 648: Sorry, my prior edit is wrong, it should be G.L_=L_  here, or Lt.N_ = L_. 

But i think after the pL_'s recomp, we can have `G.Lt = sum2F(L_, 'Lt')` without the additional `sum2F(L_,Lt)` and `G.L_=L_` reassignment?

frame_2D_alg/agg_recursion.py
```diff
@@ -259,6 +264,102 @@ def get_exemplars(N_,_r,_c):  # multi-layer non-maximum suppression -> sparse se
...
+        cr = _r+ rdn/(cnt or eps)  # for evals below and subsequent iterations
+        L = len(out_+ C_); olp = sum([len(N.root_) for N in N_])  # rdn+=olp, prioritize stronger?
+        if gv_(sum(val_(DTT,fd=1))* (wcP*L) - Ave* (cr+olp+ ccP*L)):
+            out_+=C_; out_ = cluster_P(out_, Ft)  # refine all memberships in parallel by global backprop
+            break
+        if Up * (wcC*len(C_)) > Avd * (cr+ (ccC*len(C_))):
+            for n in N_: n._root_ = n.root_
+            for n in set([_n for _C in _C_ for _n in _C.N_ + _C._N_]): n.root_ = []
+            _C_ = C_
+        else: out_ = merged if (merged:= out_ +C_) else _C_; break  # converged or diverged (when out_ + C_ is empty)
```
kwcckw, line 336: When we have empty `out_` and `C_` after the refinement, that's mean no new C is formed, i think that's the case of divergence and we should return the last good `_C_`?

boris-kz: right

frame_2D_alg/agg_recursion.py
boris-kz: right

frame_2D_alg/agg_recursion.py
kwcckw: So for the case of divergence in cluster_C, it's different with cluster_P because we may get empty C_, which led to a hard cutoff during the refinement here.

frame_2D_alg/agg_recursion.py
```diff
@@ -259,6 +264,102 @@ def get_exemplars(N_,_r,_c):  # multi-layer non-maximum suppression -> sparse se
     else:  E_ = set([N_[0]]); N_[0].exe=1  # no gain, no inhibition, any N can be seed
     return E_
 
+
+def cluster_(Ft, _N_, r, c):
```
boris-kz, line 268: Sorry, this whole merged cluster_ is probably wrong, CN and CC are different representations, with separate clustering criteria: https://claude.ai/share/e66d36ab-b4df-4c85-b14d-35d6cd12ed87

kwcckw: Previously we had that sequential clustering process, and eventually led to the merged clustering function.

For separate clustering process, cluster_C should be per G.N_ only, as in your existing code?

        C__ = []; M = C = R = 0
        for G in G_:
            if gv_(G.m * ((G.c*wcC) / (G.r*ccC)) * ((len(G.N_)-1)*wL) - ave):
                if Ct := cluster_C(G.Nt, get_exemplars(G.N_,r,c),r,c):  # refines, splits each G

boris-kz: Yes, the two sub+ forks. Sub+ because we need to select the nodes for clustering, based on their initial cluster / tile. CN and CC behave differently, we shouldn't mix them

kwcckw: Alright, so this will be reverted to your existing version. Also for sub+, do we need to follow the root's nF?

In sum2G:

    if G.Lt:  # sub+
            cross_comp(G, N_,lr,'Nt')  # sub+, cross_comp
            
This `Nt` in `cross_comp(G, N_,lr,'Nt')` should be root.nF? Because we may form G with Bt fork too.

PR #258, opened 2026-09-04
9/4/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -97,34 +96,42 @@ def cross_comp(root, G_, m, c, r, nF='Nt'):  # agg+: refine by CC,exe -> cross_c
             root.H += [Copy_(root)]  # lower agg lev
             if Lt := comp_N_( proj_L_(combinations([F2N(C) for C in C__],2), root,R), R):
                 L_, TT, c, r, V = Lt
-                root.L_ = L_; oF_[CoF.get().nF].V_ += [V]  # +-/ comp
+                if nF == "Nt":  # optional for Nt only?
+                    root.Lt.N_=L_; root.Lt.dTT=TT; root.Lt.c=c; root.Lt.r=r; root.Lt.m,root.Lt.d=val_(TT,ttX,fd=1)
```
kwcckw, line 100: So i reverted the code. 

Also for this L_ assignment, it should be valid when root nF is Nt? Since we don't get Bt.Lt anyway?

frame_2D_alg/agg_recursion.py
```diff
@@ -332,17 +349,15 @@ def cluster_C(Ft, E_,_r,_c):  # form centroids by clustering exemplar surround v
         E._root_+=[[C,1,0]]  # self (C,m,d)
         C._N_= list({n for l in E.rim for n in l.N_ if n is not E})  # init w for first loop eval
         _C_ += [C]
-    out_ = []
+    out_ = []; r = _r
     while True:  # reform C_
-        C_, cnt,rdn,DTT,Up = [],0,0,np.zeros((2,9)),0; Ave = ave*(_r+ccC); Avd = avd*(_r+ccC)
+        C_, cnt,rdn,DTT,Up = [],0,0,np.zeros((2,9)),0; Ave = ave*(r+ccC); Avd = avd*(r+ccC)   # should be r instead of _r for loop local?
```
kwcckw, line 354: I think the r here should be using the loop local `r`?

frame_2D_alg/agg_recursion.py
```diff
@@ -374,16 +387,31 @@ def cluster_C(Ft, E_,_r,_c):  # form centroids by clustering exemplar surround v
...
+                for N in out.N_:
+                    for L in N.rim:
+                        m,d = nt_vt(*L.N_)
+                        if m > ave * _r:  L_ += [L]   
+                        elif d > avd * _r: B_ += [L]    
+                ft_ = [[C.N_,'Nt',C.dTT, C.c, C.r]]
+                for i,(F_,nF) in enumerate(zip((L_,B_),('Lt','Bt'))):
+                    F_ = list(set(F_)) or []; tt,fc,fr = sum_vt(F_,wTT=ttcC) if F_ else (np.zeros((2,9)),0,0)
+                    ft_+= [[F_,nF,tt,fc,fr]]
+                G = sum2G(ft_, ttcC, root=Ft); G_ += [G]
```
kwcckw, line 402: In cluster_C, i added section to convert the final converged C to CN(G) since we need them as CN for the next iteration.

Those converted Cs will be assigned as root.N_ in cross_comp:

    if C__:
        setattr(root,nF, sum2F(C__,root, nF=nF, froot=2)); L=len(C__); R/=L  # pass M,C,R in sum2F?

frame_2D_alg/agg_recursion.py
boris-kz: Thanks. Yes, only for Nt

frame_2D_alg/agg_recursion.py
boris-kz: That local r is summed from Cs?

frame_2D_alg/agg_recursion.py
boris-kz: thanks

frame_2D_alg/agg_recursion.py
kwcckw: That r is summed below, from Rdn of Cs:

    r = _r+ rdn/(cnt or eps)
    
But the first `r` is init from input `_r`.

boris-kz: right

frame_2D_alg/agg_recursion.py
```diff
@@ -84,7 +84,6 @@ def cent_TT(dTT, r):  # EM-like weight attr matches | diffs by their match to th
 - feedback filter updates
```
boris-kz, line 84: I started checking out GPT 6:  https://chatgpt.com/share/6a9b6961-1858-83ea-8288-04173b3c9277

kwcckw: I can see somehow it's still not thinking deeper enough, that's already in thinking mode right? I guess the perks are only better speed and lesser token usage for simpler task.

boris-kz: It's probably better for conventional tasks. I had it on max effort, and it's a lot faster than 5.6

kwcckw: I see, but this probably useful for multi agent mode, where we probably can use fable for planning and gpt for execution, but at this point i think it's not better than claude for cogalg yet.

PR #259, opened 2026-09-05
9/5/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -321,7 +321,7 @@ def trans_cluster(G):
                 F_ = list(set(F_)) or []; tt,fc,fr = sum_vt(F_,wTT=ttcN) if F_ else (np.zeros((2,9)),0,0)
                 ft_+= [[F_,nF,tt,fc,fr]]
             (_,_,nt,nc,nr),(_,_,lt,lc,lr),(_,_,bt,bc,br) = ft_
-            c = nc + lc + bc
+            c = nc + lc + bc  # lc is redundant? Since L.c is min of N.c and _N.c
```
kwcckw: I think this lc is redundant?

Because l.c is the min of both `_N.c` and `N.c`, which is part of nc?

frame_2D_alg/agg_recursion.py
```diff
@@ -337,16 +337,14 @@ def trans_cluster(G):
...
 
 def cluster_C(Ft, E_,_r,_c):  # form centroids by clustering exemplar surround via rims of new member nodes, within root
 
     N_= copy(Ft.N_); _C_=[]  # revert if 0 clusters?
     for n in N_: n.root_, n._root_ = [],[]
     for i,E in enumerate(E_):
-        C = Copy_(E, Ft,init=1,cls=CL)
-        C.N_,C.L_,C.m_,C.d_ = [E],[],[1],[0]
-        E._root_+=[[C,1,0]]  # self (C,m,d)
+        C = sum2F([E], Ft, m_=[1], d_=[0])  # use sum2F for consistency?
```
kwcckw: For consistency, i guess we can init the first C with sum2F? Then we can skip the additional step to assign N_, m_ and d_.

frame_2D_alg/agg_recursion.py
```diff
@@ -382,6 +379,8 @@ def cluster_C(Ft, E_,_r,_c):  # form centroids by clustering exemplar surround v
             _C_ = C_
         else: out_+=C_; break  # converged
     if out_:
+        CN_ = [N for C in out_ for N in C.N_]
+        rN_ = [N for N in N_ if N not in CN_]  # recycle non clustered Ns? or they will be skipped entirely?
```
kwcckw: I think there might be leftover Ns that's not in any of the final converged Cs. In that case, we should recycle those Ns into the output? Else they will not be accessed again,

frame_2D_alg/agg_recursion.py
```diff
@@ -355,7 +353,6 @@ def cluster_C(Ft, E_,_r,_c):  # form centroids by clustering exemplar surround v
         for _C in _C_:  # C.m,d /rTT? sort / sum(_C.m_)?
             N__,n_,m_,d_,M,D,T,R,dTT,up = [],[],[],[],0,0,0,0, np.zeros((2,9)),0  # /C
             for n in _C.N_+_C._N_:  # current + frontier
```
kwcckw, line 356: Right now we have `n_+=[n]` below, which mean frontier always added to C.N_ by default once C is formed. Should we at least eval the frontier's m before adding it to C.N_, same as the prior `cluster_`?

frame_2D_alg/agg_recursion.py
boris-kz: we still need to select it?

frame_2D_alg/agg_recursion.py
boris-kz: was that you or LLM? there is no summing needed here.

frame_2D_alg/agg_recursion.py
boris-kz: evaluation should be by new C, which can't be computed until the loop is complete

frame_2D_alg/agg_recursion.py
boris-kz: CNs are separate, they are mosly needed on next level as exemplars, if sufficiently similar to strong CC:
The main difference is that CN is structured / transitive / interaction-based, while CC is purelly compressive and used only to select representative exemplars?

Astra:

**Selecting exemplars through centroids is a sensible shortcut.** The centroid estimates shared content; the best-matching actual member preserves a concrete configuration, including its internal hierarchy and relations.

Strictly:

* **Medoid:** the member with the lowest total dissimilarity to the other members.
* **Centroid-selected exemplar:** the member that best matches their centroid.

These coincide for summed squared Euclidean distance, but **your relative-match criterion does not guarantee equivalence**. Averaging can change which member appears most representative.

For your scheme, the natural sequence is: refine the centroid and memberships, then select the member with the strongest representative match. Keep the centroid’s aggregate statistics alongside that exemplar: the exemplar supplies concrete structure, while the centroid records what actually generalizes across the group.

Fable:

Yes, and the code currently does the opposite on both counts.

**The distinction.** CN is a transitive closure over admitted links: identity by interaction, so chaining is legitimate, membership exclusive, structure retained (Lt, Bt, H) and compared. CC is direct member-to-centroid, non-transitive, no internal structure, overlapping because it makes no identity claim — its output is a prediction of members' params and the residual, i.e. compression. Exemplars are nodes, not centroids; CC's contribution is the typicality term in w, which raises nodes that are well-predicted by a kind and lets the rim inhibition in get_exemplars do the selection.

**Where the code disagrees.**
1. C__ replaces root.Nt and the Cs become the comparison operands. Under this role centroids are never nodes; cross_comp compares Gs, and F2N goes back to links only.
2. The memberships CC forms never reach get_exemplars today: F2N(C) resets root_ = [], the second get_exemplars at cross_comp:102 runs over those Cs, and the first (per G, line 91) runs before that G's cluster_C. So the typicality term is dead in the current flow. It becomes live only when nodes keep their identity through the level.

**What CC sheds under this role.** Packaging into Lt/Bt, H append (`rG.H += Copy_(Ft)` existed only because Cs were replacing Nt), `Ft.N_ = G_`, frontier growth beyond G, exclusivity of the split (my point 3 last turn — unnecessary once parts aren't nodes). What remains: seeds from exemplars, membership reform with convergence by drift, redundancy merge in cluster_P, value to FV_, root_ entries on members, optionally the C list on Ct for inspection.

**Scope follows the seeding it serves.** Intra-G CC over G.N_ weights exemplars for sub+ within G. Level-wide seeding of cluster_N currently has only Rt.m, which is fine as a first version; a level-wide CC over Gs (the Ct slot) would add typicality there, with the same role, just wider scope and N·E cost.

**One cheap second consumer, still compressive.** Shared root_ entries between two nodes are a match term for comp_N — same form as oM in get_exemplars — and cost nothing once memberships persist. That doesn't make centroids operands; it makes kind co-membership evidence for a link.

So: connectivity forms things, centroid rates their members' typicality, typicality seeds the next connectivity. The chain-decomposition role I gave cluster_C last turn is then done by exemplar-seeded sub+ instead, with the wave vote deciding where fills stop.

boris-kz: So centroids should be packed in separate N.Ct, not N.Nt, as before?

frame_2D_alg/agg_recursion.py
kwcckw: Yes, Ls are selective from N pair, mainly from proj_L and comp_N_ , so this shouldn't be redundant, thanks.

frame_2D_alg/agg_recursion.py
kwcckw: Okay right thanks. I was thinking to remove the usage of Copy_ to simplify the Copy_ by removing the init, I guess Copy_ is more efficient here.

frame_2D_alg/agg_recursion.py
kwcckw: >  CC is purelly compressive and used only to select representative exemplars

So that's mean exemplar should be selected from C.N_ only?  But for the recycled Ns, they are not in any C, so if they are strong , they should form single N's C with them? 

> So centroids should be packed in separate N.Ct, not N.Nt, as before?

If we need them for exemplar selection, N.Ct.N_ and N.N_ should work the same? Ct is the compressed version of G, and we only need it only if we need to access Ct params separately, which we don't ?

frame_2D_alg/agg_recursion.py
kwcckw: If it's solely by new C eval, we either add all frontier and existing Ns to C or remove it all. Which means the frontier will never be included individually.

boris-kz: the eval is by similarity of each N to new C. This is the most basic operation of centroid clustering

frame_2D_alg/agg_recursion.py
boris-kz: initial get_exemplars doesn't need Cs, they can be used for refinement

>Ct is the compressed version of G,

you obviously have no intuition of the process

kwcckw: > CN is structured / transitive / interaction-based, while CC is purelly compressive and used only to select representative exemplars?

Do you mean CC should be more light-weighted, and we only need their CC.N_ for exemplar selection. while the comp and cluster should always use the CN from C.N_? But i think C is clustered from N anyway, we actually should use CN to represent them since they will be comp and clustered again in higher level.

frame_2D_alg/agg_recursion.py
kwcckw: Alright, so this is actually following the conventional centroid clustering, where similarity M is summed from all Ns, and we eval using that summed M to reform the new C. There's actually no selection in each individual frontier.

boris-kz: selection is per C:

            if M * (_C.m+wcC*len(n_)) > Ave * (r+_C.r+ ccC*len(n_)):  # else: Up+= sum(_C._m_)+ sum([abs(d) for d in _C._d_])?

frame_2D_alg/agg_recursion.py
boris-kz: C is a compressed representation of C.N_, Cs overlap, N.Ct is all Cs that N is an element of

kwcckw: > initial get_exemplars doesn't need Cs, they can be used for refinement
> 
> > Ct is the compressed version of G,
> 
> you obviously have no intuition of the process

Ct is computed via centroid clustering in sub+, which is a more refined version of G? So Ct.N_ should replace G.N_?

kwcckw: > C is a compressed representation of C.N_, Cs overlap

Okay right, Cs overlap, which may not be a good choice to be packed as G.N_. I guess it's better to pack them in a separate Ct, rather than replacing Nt, as in the current scheme.

PR #261, opened 2026-09-07
9/7/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -87,24 +87,23 @@ def cross_comp(root, G_, m, c, r, nF='Nt'):  # agg+: refine by CC,exe -> cross_c
...
-                    oF_[CoF.get().nF].V_ += [V]  # +-/ comp
-                    if gv_(val_(TT,ttcN) * (c*wcN /(r*ccN)) * ((len(L_)-1)*wL) - ave):  # return +ve, store -ve gate Vs
-                        e_ = get_exemplars({N for L in L_ for N in L.N_}, r,c)  # +ve Ls only
-                        cluster_N(getattr(root,nF), e_,r,c)  # sum2G -> agg+
+        if gv_(G.m * ((G.c*wcC) / (G.r*ccC)) * ((len(G.N_)-1)*wL) - ave) and (Ct := cluster_C(G.Nt, get_exemplars(G.N_,r,c),r,c)):   # prune G_ before call?
+            C__+=Ct.N_; M+=Ct.m; C+=Ct.c; R+=Ct.r  # centroids / root
+        else: g_ += [G]  # fall back to G if not forming any Cs
+    med_ = list(set([C.N_[np.argmax(C.m_)] for C in C__]))  # x-comp medoids: nodes with highest|>ave match to given C
+    # extend rng only, skip N pairs compared within their G:
+    pairs = [(_N, N) for _N, N in combinations(med_,2) if not any(_N in L.N_ and N in L.N_ for L in _N.rim)]
```
kwcckw, line 95: The existing code get the new pairs of Ns by checking if the N in the pair is in any of the other N.rim. But this extension doesn't include any distance for rng eval?

frame_2D_alg/agg_recursion.py
```diff
@@ -87,24 +87,23 @@ def cross_comp(root, G_, m, c, r, nF='Nt'):  # agg+: refine by CC,exe -> cross_c
...
-                if Lt := comp_N_( proj_L_(pairs, root,R), R):
-                    L_,TT,c,r,V = Lt
-                    if nF=="Nt": root.Lt.N_=L_; root.Lt.dTT=TT; root.Lt.c=c; root.Lt.r=r; root.Lt.m,root.Lt.d=val_(TT,ttX,fd=1)
-                    oF_[CoF.get().nF].V_ += [V]  # +-/ comp
-                    if gv_(val_(TT,ttcN) * (c*wcN /(r*ccN)) * ((len(L_)-1)*wL) - ave):  # return +ve, store -ve gate Vs
-                        e_ = get_exemplars({N for L in L_ for N in L.N_}, r,c)  # +ve Ls only
-                        cluster_N(getattr(root,nF), e_,r,c)  # sum2G -> agg+
+        if gv_(G.m * ((G.c*wcC) / (G.r*ccC)) * ((len(G.N_)-1)*wL) - ave) and (Ct := cluster_C(G.Nt, get_exemplars(G.N_,r,c),r,c)):   # prune G_ before call?
+            C__+=Ct.N_; M+=Ct.m; C+=Ct.c; R+=Ct.r  # centroids / root
+        else: g_ += [G]  # fall back to G if not forming any Cs
```
kwcckw, line 92: So i think we should pack G when c fork is false, and the med_ will never be empty here if we repack g_ back into med_ for the subsequent agg+.

frame_2D_alg/agg_recursion.py
```diff
@@ -87,24 +87,23 @@ def cross_comp(root, G_, m, c, r, nF='Nt'):  # agg+: refine by CC,exe -> cross_c
...
+    med_ = list(set([C.N_[np.argmax(C.m_)] for C in C__]))  # x-comp medoids: nodes with highest|>ave match to given C
+    # extend rng only, skip N pairs compared within their G:
+    pairs = [(_N, N) for _N, N in combinations(med_,2) if not any(_N in L.N_ and N in L.N_ for L in _N.rim)]
+    med_ = list(set(N for P in pairs for N in P)) + g_
+    setattr(root,nF, sum2F(med_,root, nF=nF,froot=2)); L=len(med_); R/=L  # pass M,C,R?
```
kwcckw, line 97: Sorry, this Ft update should be valid only if there's any Ct is formed, which we need something like this:

    med_ = list(set([C.N_[np.argmax(C.m_)] for C in C__]))
    pairs = [(_N, N) for _N, N in combinations(med_,2) if not any(_N in L.N_ and N in L.N_ for L in _N.rim)]
    med_ = list(set(N for P in pairs for N in P))
    if med_: setattr(root,nF, sum2F(med_,root, nF=nF,froot=2)); L=len(med_); R/=L  # pass M,C,R?
    med_ += g_

frame_2D_alg/agg_recursion.py
boris-kz: it's not into med_, they should be compared separately

frame_2D_alg/agg_recursion.py
boris-kz: rng is extended in comp_N_ indirectly, we should have higher m here:

        if _N != N and fall or (m>0 and gv_(m * (lc*wN / (lr*cN)) - ave*(r+cN))):  # marginal -gV

frame_2D_alg/agg_recursion.py
boris-kz: this should only be done for med_, not for g_

frame_2D_alg/agg_recursion.py
kwcckw: Okay, so that m> 0 should be updated, and we probably can use N.sub here? Also we need G.sub = N.sub+1 in sum2G too.

frame_2D_alg/agg_recursion.py
kwcckw: > this should only be done for med_, not for g_

For g_, we need another similar comp-cluster sections to compare and cluster them? Are those g_ and C__ here are consider the same fork? If yes we still need to merge them at some point of the process?

boris-kz: it should be the same process, but only same-composition pairs are compared

frame_2D_alg/agg_recursion.py
boris-kz: why sub, it's the same-composition Ns

boris-kz: yes, that's in all agg+

frame_2D_alg/agg_recursion.py
kwcckw: Are the G in g_ here is consider the same composition level as C? Since both of them are clusteree by the same level of Ns, but with different clustering method.

boris-kz: composition level is represented by N.sub, which is higher for g_. The clustering should be largely the same

boris-kz: we need to use the same helper for med_ and g_?

chat:

```python
# L88: insert inside cross_comp, before C__,g_ initialization
    def xcomp(N_, pairs, M,C,R):
        if gv_((m+M) * (c*(C+wN_) / (r*(R+cN_))) * ((len(N_)-1)*wL) - ave):
                lev = Copy_(root if nF=="Nt" else getattr(root,nF))
                if nF!="Nt": lev.H = copy(getattr(root,nF).H)
                TT,nc,nr = sum_vt(N_)  # fork values from this population
                Ft = CF(N_=N_,dTT=TT,c=nc,r=nr,nF=nF,root=root,wTT=root.wTT,H=[lev])
                Ft.m,Ft.d = val_(TT,ttX,fd=1); setattr(root,nF,Ft)
                L_,TT,lc,lr,V = Lt
                if nF=="Nt":
                    root.Lt = CF(N_=L_,dTT=TT,c=lc,r=lr,nF="Lt",root=root,wTT=root.wTT)
                    root.Lt.m,root.Lt.d = val_(TT,ttX,fd=1)
                root.dTT,root.c,root.r = sum_vt([root.Nt,root.Lt,root.Bt])
                root.m,root.d = val_(root.dTT,root.wTT,fd=1)
                oF_[CoF.get().nF].V_ += [V]
                    e_ = get_exemplars({N for L in L_ for N in L.N_}, lr,lc)
                    for N in N_: N.fin=0
                    cluster_N(Ft,e_,lr,lc)
```

```python
# Previous L97-129: replace both passes
        if pairs := [(_N,N) for _N,N in combinations(med_,2) if not any(_N in L.N_ and N in L.N_ for L in _N.rim)]:
            xcomp(med_,pairs,M,C,R/len(med_))
    if len(g_) > 1:
        TT,C,R = sum_vt(g_)
        xcomp(g_,combinations(g_,2),val_(TT,ttX),C,R)
```

kwcckw: > composition level is represented by N.sub, which is higher for g_. The clustering should be largely the same

I see, then there will be two resulting forks here under the same root? And only one of them should replace the root.Nt and update root.Lt, while the other fork should be packed separately in a new param?

boris-kz: They both form top-level Gs, but with different internal depth (sub)

PR #262, opened 2026-09-08
9/8/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -85,19 +85,21 @@ def cent_TT(dTT, r):  # EM-like weight attr matches | diffs by their match to th
 '''
 def cross_comp(root, G_, m, c, r, nF='Nt'):  # agg+: refine by CC,exe -> cross_comp, nF: core|contour?
     # draft:
-    def xcomp(N_, pairs, M,C,R):
+    def xcomp(N_, pairs, M,C,R,nF):
         if gv_((m+M) * (c*(C+wN_) / (r*(R+cN_))) * ((len(N_)-1)*wL) - ave):
             if Lt := comp_N_(proj_L_(pairs,root,R),R):
-                lev = Copy_(root if nF=="Nt" else getattr(root,nF))
-                if nF!="Nt": lev.H = copy(getattr(root,nF).H)
+                ft = getattr(root,nF); lev = Copy_(ft)  # root is CN, we should copy Nt too?
```
kwcckw, line 91: The root is G here, so we should copy their Nt? Why we need to copy root when nF == "Nt"?

Also this `if nF!="Nt": lev.H = copy(getattr(root,nF).H)` shouldn't be needed since H will be copied in Copy_ anyway.

frame_2D_alg/agg_recursion.py
```diff
@@ -153,7 +155,7 @@ def comp_N_(pL_, r, tnF=None, root=2, fall=0):  # incremental-distance cross_com
 
     L_,N_,mrg_ = [],[],[]
     for i, (dist, dy_dx, _N,N, lc,lr, pTT,m,_) in enumerate(pL_):  # pL is L = dist, dy_dx, _N,N if fall: not selective
-        if _N != N and fall or (m>0 and gv_(m * (lc*wN / (lr*cN)) - ave*(r+cN))):  # marginal -gV
+        if _N != N and fall or gv_(m * (lc*wN / (lr*cN)) - ave*(r+cN+N.sub+_N.sub)):  # marginal -gV
```
kwcckw, line 158: The `m>0` here is redundant? That gv_ won't be true anyway if m <0 since ave, cN and sub won't be negative.

frame_2D_alg/agg_recursion.py
```diff
@@ -85,19 +85,21 @@ def cent_TT(dTT, r):  # EM-like weight attr matches | diffs by their match to th
...
-                Ft.m,Ft.d = val_(TT,ttX,fd=1); setattr(root,nF,Ft)
+                Ft = CF(N_=N_,dTT=TT,c=nc,r=nr,nF=nF,root=root,wTT=root.wTT,H=[lev]+ft.H)
+                Ft.m,Ft.d = val_(TT,ttX,fd=1); 
                 L_,TT,lc,lr,V = Lt
-                if nF=="Nt":
+                if nF=="Nt": 
                     root.Lt = CF(N_=L_,dTT=TT,c=lc,r=lr,nF="Lt",root=root,wTT=root.wTT)
-                    root.Lt.m,root.Lt.d = val_(TT,ttX,fd=1)
-                root.dTT,root.c,root.r = sum_vt([root.Nt,root.Lt,root.Bt])
+                if nF == 'Ct': setattr(root.Nt,'Ct',Ft)  # nF == Ct only when call xcomp with med_? or merge both
```
kwcckw, line 99: So now we have 2 forks in cross_comp, from both medoid of centroids and clustered g_.

We need to repack them back to root, and higher composition clusters from g_ should be packed into root.Nt, while the lower composition clusters via medoids should be added to the root.Nt.H?

My current edit pack clusters from medoid into root.Nt.Ct, which is temporary and not correct.

frame_2D_alg/agg_recursion.py
kwcckw: Another thing is the sub+ using the same set of N_:

    if G.Lt:  # sub+
            cross_comp(G, N_,lm,lc,lr, 'Nt')  # sub+, cross_comp

That N_ has the same sub value in every deeper sub+, so every G in deeper sub+ has the same sub value. We should parse the new sub value instead?

frame_2D_alg/agg_recursion.py
boris-kz: Astra:

`Copy_()` copies `H` only for `CN`.

For `Nt`, we copy the `CN` root, so its history is included. For `Bt`, we copy a `CF`, whose `H` would be empty. That line preserves its earlier history inside `lev`, which becomes the new fork’s `H=[lev]`.

It compensates for missing `CF.H` copying in `Copy_()`.

kwcckw: Okay right thanks, but we probably can include this additional line to copy H instead;

    if nF == 'Nt': lev.H = [Copy_(level, root=Ft) for level in root.Nt.H]

frame_2D_alg/agg_recursion.py
kwcckw: > They both form top-level Gs, but with different internal depth (sub)

Alright, so we should return the clustered Gs from both xcomp (medoids + g_), combine them before repack them into root's Ft.

boris-kz: So it's the difference between scope and depth of composition: scope determines if the top levels are comparable between the Gs, depth aligns the comparison of the lower levels
That means we don't need separate N.sub, it's just len H, with H also representing missing layers

frame_2D_alg/agg_recursion.py
boris-kz: m>0 is just the first light filter.
sub is not needed anyway, see above

frame_2D_alg/agg_recursion.py
kwcckw: Okay right, so from base level, Gs clustered from g_ should have an additional depth compare to Gs clustered from medoid. In proj_L:

The above should be checking len(N.H) now, same as the eval in comp_N too.

boris-kz: right

kwcckw: So we shouldn't have any overlapping between the Gs of medoids and g_ right? Since there shouldn't be any overlapping between input G_ in cross_comp, and we won't have any cross_comp with nF == 'Ct' now.

boris-kz: right. and we need to add empty level in medoid clusters

kwcckw: > and we need to add empty level in medoid clusters

Actually i don't quite get this. Could you explain on the reason on add empty level in medoid clusters? If we add empty level, those medoid clusters will have the same H length as the g_'s clusters?

boris-kz: yes, because their top level has the same scope

kwcckw: If we fill their H, we only can check their actual depth by checking the number of non empty H level, but they will be compared since their H length is the same?

boris-kz: the top level is compared, lower levels must be non-empty and aligned

frame_2D_alg/agg_recursion.py
```diff
@@ -106,17 +108,17 @@ def xcomp(N_, pairs, M,C,R):
                     cluster_N(Ft,e_,lr,lc)
     C__,g_ = [],[]; M= C= R= 0
     for G in G_:  # or prune G_ before call?
-        if gv_(G.m * ((G.c*wcC) / (G.r*ccC)) * ((len(G.N_)-1)*wL) - ave and (Ct := cluster_C(G.Nt, get_exemplars(G.N_,r,c),r,c))):
+        if gv_(G.m * ((G.c*wcC) / (G.r*ccC)) * ((len(G.N_)-1)*wL) - ave) and (Ct := cluster_C(G.Nt, get_exemplars(G.N_,r,c),r,c)):  # gv_ should exclude Ct
             C__+=Ct.N_; M+=Ct.m; C+=Ct.c; R+=Ct.r  # centroids / root
         else: g_ += [G]  # fall back to G if not forming any Cs
-
+    if len(g_) > 1:
```
boris-kz, line 114: So each G in g_ should be refined G.N_ similarity to G.C_?

Yes—refine each node’s contribution to `G` using its matches to the relevant centroids in `G.C_`. Those matches already exist in `C.m_`, aligned with `C.N_`; reuse them rather than compare again.

Each node should be supported by its corresponding type, not necessarily every centroid. Overlapping memberships need redundancy accounting.

This refines `G`’s content summary. Its links still define composition and connectivity; weak centroid similarity alone doesn’t justify removing a connecting node.

# cross_comp: replace former L93-95
                G.Ct=Ct; Ct.root=G; Ct.nF="Ct"
                m__ = {N: [0,0] for N in G.N_}  # strongest, total positive match
                for _C in Ct.N_:
                    for N,nm in zip(_C.N_,_C.m_):
                        if N in m__:
                            m__[N][0] = max(m__[N][0],nm)
                            m__[N][1] += max(nm,0)
                w_ = np.array([N.c*m__[N][0] for N in G.N_], dtype=float)
                if W := w_.sum():
                    w_ /= W
                    r_ = [N.r + (s/v-1 if v else 0) for N,(v,s) in m__.items()]  # fractional overlap beyond strongest match
                    Nt = G.Nt; Nt.c = sum(N.c for N in G.N_)
                    Nt.dTT = sum(N.dTT*w for N,w in zip(G.N_,w_))
                    Nt.r = sum(nr*w for nr,w in zip(r_,w_))
                    Nt.m,Nt.d = val_(Nt.dTT,Nt.wTT,fd=1)
                    G.kern = sum(N.kern*w for N,w in zip(G.N_,w_))
                G.dTT,G.c,G.r = sum_vt([G.Nt,G.Lt,G.Bt])
                G.m,G.d = val_(G.dTT,G.wTT,fd=1)
            g_ += [G]

kwcckw: > So each G in g_ should be refined G.N_ similarity to G.C_?

So we should refine each G params again based on the similarity of clustered G.C_ to existing G.N_?

Or it's just replacing G's params with the new updated G.C_? But that could be done with sum_vt of G.C_ to get those tt, c, r, m and d?

boris-kz: Actually it's not needed, CN is defined differently

kwcckw: Okay, and right now we assign Ct as N.Ct, and the code actually already summing Ct's params to N in comb_Ft:

    add2F(G,T, merge=2); add2F(G,Bt,merge=2); add2F(G,Ct,merge=2)  # add Bt and Ct's params to G?

boris-kz: that summing make sense, but it should account for redundancy in C_ to N_ mapping

kwcckw: But there's always redundancy since Cs are clustered from the same N?

A better way is re-sum G params from Bt, Lt, Ct and the non C-clustered Ns?

boris-kz: the Cs also represent Ns outside current G. This accounting is pretty complex, we probably need to use averages

kwcckw: Okay, that's possible when we have frontier added from N's rim, and some of the rim is not within G.

Average is something like the average value of N's params in both Ct and Nt?

boris-kz: something like that. I won't be pushing tonight, could you replace that N.sub s with len H, etc?

kwcckw: Alright, but since we fill the H level, this checking:

Becomes:

     if len(_N.H) != len(N.H): continue
     
Which will never be false? Or the comp still need to be checked with the actual depth that with non empty level?

boris-kz: The level is only filled in new G, not in its nodes

kwcckw: Okay thanks, so in the next level, the combined Gs from medoid and g_ will be compared.

PR #263, opened 2026-09-09
9/9/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -85,38 +85,37 @@ def cent_TT(dTT, r):  # EM-like weight attr matches | diffs by their match to th
...
     if med_ := list(dict.fromkeys(C.N_[np.argmax(C.m_)] for C in C__)):
          if pairs := [(_N, N) for _N, N in combinations(med_, 2) if not any(_N in L.N_ and N in L.N_ for L in _N.rim)]:
             med_ = list(dict.fromkeys(N for P in pairs for N in P))
-            xcomp(med_, pairs, M, C, R / len(med_))
+            if (out:= xcomp(Ft, med_, pairs, M, C, R / len(med_))): mG_, mL_ = out; G_ += mG_; L_ += [mL_]     
     if len(g_) > 1:
         TT, C, R = sum_vt(g_)
-        xcomp(g_, combinations(g_, 2), val_(TT, ttX), C, R)
+        if (out := xcomp(Ft, g_, combinations(g_, 2), val_(TT, ttX), C, R)): gG_, gL_ = out; G_ += gG_; L_ += [gL_] 
+    if G_:  # combine mG_ and gG_
```
kwcckw, line 110: So i restructured the code here so that we merge both Gs and L_ from medoids and g_ before packing them again into the root's Ft and root.Lt.

frame_2D_alg/agg_recursion.py
```diff
@@ -85,38 +85,37 @@ def cent_TT(dTT, r):  # EM-like weight attr matches | diffs by their match to th
...
+                L_,TT,lc,lr,V = Lt; oF_[CoF.get().nF].V_ += [V]
                 if gv_(val_(TT,ttcN) * (lc*wcN /(lr*ccN)) * ((len(L_)-1)*wL) - ave):
                     e_ = get_exemplars({N for L in L_ for N in L.N_}, lr,lc)
                     for N in N_: N.fin=0
-                    cluster_N(Ft,e_,lr,lc)
+                    return cluster_N(Ft,e_,lr,lc), L_
     C__,g_ = [],[]; M= C= R= 0
     for G in G_:  # or prune G_ before call?
-        if gv_(G.m * ((G.c*wcC) / (G.r*ccC)) * ((len(G.N_)-1)*wL) - ave and (Ct := cluster_C(G.Nt, get_exemplars(G.N_,r,c),r,c))):
+        if gv_(G.m * ((G.c*wcC) / (G.r*ccC)) * ((len(G.N_)-1)*wL) - ave) and (Ct := cluster_C(G.Nt, get_exemplars(G.N_,r,c),r,c)):  # gv_ should exclude Ct
```
boris-kz, line 98: I think G vals should exclude G.Ct too, these are alternative representations, only relevant to to G nodes.
or it may concat in higher GG.Ct.H[-2]? 

Astra:
Both fit: exclude `G.Ct` from `G`’s direct values, while carrying its representation upward through the centroid hierarchy.

* `GG.Ct.N_`: centroids of the **Gs** in `GG.N_`.
* The inherited lower level: concatenated `G.Ct.N_`, whose centroids describe those Gs’ constituent **Ns**. Preserve their membership and originating-G associations.

That inherited level is `GG.Ct.H[-2]` **if `H[-1]` contains the current G-level centroids**. If the current level exists only in `GG.Ct.N_`, the inherited level is `H[-1]`.

Packing that history should update the centroid hierarchy’s summaries without adding them to `GG`’s structural values.

frame_2D_alg/agg_recursion.py
```diff
@@ -619,7 +620,9 @@ def comb_Ft(Nt, Lt, Bt, Ct, root,wTT):  # from sum2G, default Nt
     for Ft in Lt, Bt:  # connectivity forks
         if Ft: dF_ += [comp_F(T, Ft, root.r,G)]; T.dTT,T.c,T.r = sum_vt([T,Ft],wTT=wTT)  # Bt*brrw?
         else:  dF_ += [CF()]
-    add2F(G,T, merge=2); add2F(G,Bt,merge=2); add2F(G,Ct,merge=2)  # add Bt and Ct's params to G?
+    add2F(G,T, merge=2)
+    if Ct: G.dTT = (G.dTT+Ct.dTT)/2; G.r = (G.r+Ct.r)/2; G.c = (G.c+Ct.c)/2  # not sure, average value of Nt and Ct?
```
boris-kz, line 624: so I don't think this is needed

frame_2D_alg/agg_recursion.py
```diff
@@ -598,7 +598,8 @@ def sum2G(ft_, fTT, root=None, init=1):  # core clustering function
     C_= [c for N in Ft_[0].N_ for c in N.C_]  # splice centroids
     Ft_ += [sum2F(list(set(C_)), root.Ct,nF='Ct') if C_ else CF(nF='Ct')]  # add multiple root_ in Cs?
     G = comb_Ft(*Ft_, root, wTT=fTT)
-    N_ = G.N_; N=N_[0]; G.sub = N.sub+1 if G.L_ else N.sub; r=G.r; Av=ave+avd
+    N_ = G.N_; N=N_[0]; r=G.r; Av=ave+avd
+    for n in N_: add_H(G.H, n.H, G)
```
boris-kz, line 602: this should be in add_Nt(G), in comb_Ft?

def add_Nt(G):  # in sum2G and trans_cluster

    N_ = G.N_; c_ = np.array([N.c for N in N_]); C = c_.sum()
    G.kern, G.yx = np.zeros(4),np.zeros(2); yx_ = []
    for N in N_:
        G.root_ += [rt for rt in N.root_]  # Ct || Nt  (use rt for consistency, to be used in get_exemplars:  rc = sum(r[0].c for r in n.root_))
        G.kern += N.kern*w; yx = N.yx; yx_+=[yx]  # * w?
        add_H(G.H, N.H, G)
    G.yx = np.mean(yx_,axis=0); G.span = (c_ @ np.hypot(*(np.array(yx_)-G.yx).T)) / C if len(N_)>1 else N_[0].span

frame_2D_alg/agg_recursion.py
kwcckw: Okay, but i don't think we will get Ct.H since we never run cross_comp with nF == Ct? So for each level G, we are at most getting a single level Ct, unless we need to sum nodes' existing Ct.N_ into new level Ct.H first before packing the new Ct level.

frame_2D_alg/agg_recursion.py
kwcckw: Okay right, thanks. But we need to add another line to fill the empty layer when N is medoid. After the comb_Ft section:

    if N.root_: G.H += [[]]

frame_2D_alg/agg_recursion.py
boris-kz: yes, this concat is in cluster_C(CN)

frame_2D_alg/agg_recursion.py
```diff
@@ -153,7 +152,7 @@ def comp_N_(pL_, r, tnF=None, root=2, fall=0):  # incremental-distance cross_com
 
     L_,N_,mrg_ = [],[],[]
     for i, (dist, dy_dx, _N,N, lc,lr, pTT,m,_) in enumerate(pL_):  # pL is L = dist, dy_dx, _N,N if fall: not selective
-        if _N != N and fall or (m>0 and gv_(m * (lc*wN / (lr*cN)) - ave*(r+cN))):  # marginal -gV
+        if _N != N and fall or m>0 and gv_(m * (lc*wN / (lr*cN)) - ave*(r+cN+len(N.H)+len(_N.H))):  # marginal -gV
```
boris-kz, line 155: why len(N.H)+len(_N.H)?

frame_2D_alg/agg_recursion.py
kwcckw: Okay, and I think we need to change add_Nt:

   add_H(G.H, N.H, G)

So that when nF is Ct, we need to add another line:

     add_H(G.Ct.H, N.Ct.H, G)

frame_2D_alg/agg_recursion.py
kwcckw: The prior sub level or H level should be an additional cost to the m comparison eval?

boris-kz: comp is selective per level

frame_2D_alg/agg_recursion.py
boris-kz: this will be a separate add_Ct?

frame_2D_alg/agg_recursion.py
boris-kz: this should be in cluster_N?

kwcckw: This should be added to the new G when the nodes are medoids? Why it should be in cluster_N?

frame_2D_alg/agg_recursion.py
kwcckw: But for rng extension in each new level, we need to increase the cost relative to the rng?

frame_2D_alg/agg_recursion.py
kwcckw: Not sure if we need add_Ct since we only need to deal with the H? We don't add their params to the G anyway.

frame_2D_alg/agg_recursion.py
boris-kz: ok in sum2G:

    if fmed: G.H += [[]]  # empty top lev

frame_2D_alg/agg_recursion.py
boris-kz: ok right

frame_2D_alg/agg_recursion.py
boris-kz: the cost and gain are per comp, rng+ only increases n of comps

kwcckw: Okay, thanks for the clarification.

PR #264, opened 2026-09-10
9/10/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -85,35 +85,36 @@ def cent_TT(dTT, r):  # EM-like weight attr matches | diffs by their match to th
...
     G_,L_ = [],[]
-    ft = getattr(root,nF); Ft = CF(nF=nF,root=root,wTT=root.wTT,H=[Copy_(ft)]+copy(ft.H))
+    ft = getattr(root,nF); Ft = CF(nF=nF,root=root,wTT=root.wTT,H=[Copy_(ft)]+copy(ft.H))  # top level first? Which mean the other section need to use insert at 0 index?
     if med_ := list(dict.fromkeys(C.N_[np.argmax(C.m_)] for C in C__)):
          if pairs := [(_N, N) for _N, N in combinations(med_, 2) if not any(_N in L.N_ and N in L.N_ for L in _N.rim)]:
             med_ = list(dict.fromkeys(N for P in pairs for N in P))
-            if out:= xcomp(Ft, med_, pairs, M,C, R/len(med_), fmed=1): mG_, mL_ = out; G_ += mG_; L_ += [mL_]
+            if out:= xcomp(Ft, med_, pairs, M,C, R/len(med_)): 
+                mG_, mL_ = out; G_ += mG_; L_ += mL_
+                for G in mG_: G.H += [[]]  # empty top lev
```
kwcckw, line 107: Actually i think we can just insert the empty level here? Then we can skip the usage of fmed.

frame_2D_alg/agg_recursion.py
```diff
@@ -85,35 +85,36 @@ def cent_TT(dTT, r):  # EM-like weight attr matches | diffs by their match to th
...
-                    return cluster_N(Ft,e_,lr,lc, fmed), L_
+                    return cluster_N(Ft,e_,lr,lc), L_
     C__,g_ = [],[]; M= C= R= 0
     for G in G_:  # or prune G_ before call?
         if gv_(G.m * ((G.c*wcC) / (G.r*ccC)) * ((len(G.N_)-1)*wL) - ave) and (Ct := cluster_C(G.Nt, get_exemplars(G.N_,r,c),r,c)):
             C__+= Ct.N_; M+=Ct.m; C+=Ct.c; R+=Ct.r  # centroids / root
         else: g_ += [G]  # fall back to G if not forming any Cs
     G_,L_ = [],[]
-    ft = getattr(root,nF); Ft = CF(nF=nF,root=root,wTT=root.wTT,H=[Copy_(ft)]+copy(ft.H))
+    ft = getattr(root,nF); Ft = CF(nF=nF,root=root,wTT=root.wTT,H=[Copy_(ft)]+copy(ft.H))  # top level first? Which mean the other section need to use insert at 0 index?
```
kwcckw, line 101: Sorry, the top level index should be the last, so actually H should be `copy(ft.H)+ [Copy_(ft)]` here.

frame_2D_alg/agg_recursion.py
```diff
@@ -316,7 +317,7 @@ def trans_cluster(G):
             _L_ = []
             for L in set(__L_) - in_:  # flood-fill via frontier links
                 _N = L.N_[0] if L.N_[1].fin else L.N_[1]; in_.add(L)
-                if not _N.fin and _N in Ft.N_:
```
kwcckw, line 319: This `_N in Ft.N_` is obsolete now since we parse the empty Ft to cluster_N, and `_N_` is always in G since it's retrieved from L's nt now.

frame_2D_alg/agg_recursion.py
boris-kz: right

boris-kz: if GL_:= xcomp(Ft, med_, pairs, M,C, R/len(med_)):
                G_,L_ = GL_
                for G in G_: G.H += [[]]  # top lev is unpacked

frame_2D_alg/agg_recursion.py
boris-kz: ok

frame_2D_alg/agg_recursion.py
kwcckw: I think it's better to pack G_ and L_ from the xcomp here to the combined Gs and Ls from both clustering process?

boris-kz: it works the same

kwcckw: Okay, i just noticed the updated code, let me check through them first.

PR #265, opened 2026-09-11
9/11/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -90,7 +90,8 @@ def xcomp(Ft, N_, pairs, M,C,R):  # core
             if Lt := comp_N_(proj_L_(pairs,root,R),R):
                 L_,TT,lc,lr,V = Lt; oF_[CoF.get().nF].V_ += [V]
                 if gv_(val_(TT,ttcN) * (lc*wcN /(lr*ccN)) * ((len(L_)-1)*wL) - ave):
-                    e_ = get_exemplars({N for L in L_ for N in L.N_}, lr,lc)
+                    N_ = {N for L in L_ for N in L.N_}; e_ = get_exemplars(N_, lr,lc)
+                    for N in N_: N.fin = 0; N.Rt = sum2F(N.rim);N.Rt.root = N  # only if N was added in trans-cluster?
```
kwcckw, line 94: So we should reset fin here since the prior reset in cluster_N below is wrong:

    for N in _N_:
        N.fin =0; N.exe=1; N.Rt = sum2F(N.rim);N.Rt.root = N  # only if N was added in trans-cluster?

That's only the for exemplars, while we need to do it for all Ns here.

frame_2D_alg/agg_recursion.py
```diff
@@ -569,7 +569,7 @@ def sum2G(ft_, fTT, root=None, init=1):  # core clustering function
     if G.Lt:  # sub+
         Lt = G.Lt; L_,lm,lc,lr = Lt.N_,Lt.m,Lt.c,Lt.r  # no levR = 1/len(L_): represented by c
         if gv_(lm*lc*wX - Av* (lr+1+cX)):  # mdecay(L_)-decay?
-            cross_comp(G, N_,lm,lc,lr, 'Nt')  # sub+, cross_comp
+            cross_comp(G, N_,lm,lc,lr, 'Nt')  # sub+, cross_comp  (may get endless recursion sum2G -> cross_comp -> cluster_N)
```
kwcckw, line 572: Although this is very rare, it's possible to get endless recursion here since we may cluster a same set of `N_` into a new `G` in the next recursion' xcomp's cluster_N, which will run a same cross_comp with the same set of N_ again.

frame_2D_alg/agg_recursion.py
```diff
@@ -589,6 +589,7 @@ def comb_Ft(Nt, Lt, Bt, Ct, root,wTT):  # from sum2G, default Nt
     add2F(G,T, merge=2); add2F(G,Bt,merge=2)  # exclude Ct
     if any(dF_): sum2F(dF_,G.Xt)  # cross-fork covariance
     add_Nt(G)  # add kern,ext, doesn't affect comp_F
+    if (n_ := [n for N in G.N_ for n in N.N_]): G.H += [sum2F(n_, G)]
```
kwcckw, line 592: So i noticed the existing g_'s clusters doesn't have any H, while we pad empty layer on med clusters. I think we should new H level in sum2G for those g_'s clusters?

frame_2D_alg/agg_recursion.py
boris-kz: ok, but why only if trans-clusters?
N.Rt = sum2F(N.rim,root=N)

frame_2D_alg/agg_recursion.py
boris-kz: the threshold is incremented above:  lr+1

frame_2D_alg/agg_recursion.py
boris-kz: it's in comb_Ft:

if (n_ := [n for N in G.N_ for n in N.N_]): G.H += [sum2F(n_, G)]

frame_2D_alg/agg_recursion.py
kwcckw: Sorry, i copied that comment from prior code, shouldn't be relevant now. 

Also the Rt computation should be before the `get_exemplars` since we use Rt.m in getting the exemplars:

            N_ = {N for L in L_ for N in L.N_}
            for N in N_: N.fin = 0; N.Rt = sum2F(N.rim);N.Rt.root = N
            e_ = get_exemplars(N_, lr,lc)

frame_2D_alg/agg_recursion.py
kwcckw: Yes, I added this line in my current commit, else G.H remains empty without the first H append.

boris-kz: this is already in sum2F, I think we need to move it to add_Nt:

        if typ==3:  # frame?

frame_2D_alg/agg_recursion.py
boris-kz: root assign is in sum2F:

N.Rt = sum2F(N.rim,root=N)

frame_2D_alg/agg_recursion.py
boris-kz: I think this nested cross_comp was to extend rng, previously compared pairs should not be re-compared

frame_2D_alg/agg_recursion.py
kwcckw: The G.H is actually Nt.H, so we still need it in sum2F when we sum Nt? It's just we didn't use sum2F in sum2G but we init the Fts manually, I guess we should use sum2F in sum2G to compute Fts and it should be simpler that way?

frame_2D_alg/agg_recursion.py
kwcckw: Okay right, we have that N pair rim checkkng to prevent the same pair comparison again, something must be wrong, let me check again later.

frame_2D_alg/agg_recursion.py
kwcckw: From sum2F, we need froot=2, but we shouldn't sum Rt into root? 

    if root is not None:
        F.wTT = root.wTT
    if froot == 1:

frame_2D_alg/agg_recursion.py
boris-kz: add_Nt is called from comb_Ft, which is called from sum2G

frame_2D_alg/agg_recursion.py
boris-kz: make Rt typ=2?

kwcckw: Alright, will change that later.

frame_2D_alg/agg_recursion.py
kwcckw: Alright, this should work since C type shouldn't be 3 here .

frame_2D_alg/agg_recursion.py
boris-kz: or we can check nF instead, it's clearer

frame_2D_alg/agg_recursion.py
boris-kz, line 100: this exclusive else is not quite right, there should be an option to cross_comp both, just increment r

frame_2D_alg/agg_recursion.py
kwcckw: I guess we can add both? Something like:  If type != 2 and nF !='Rt'?

frame_2D_alg/agg_recursion.py
kwcckw: Those g_ will.be compared anyway? Or do you mean combine them with C__?

PR #266, opened 2026-09-12
9/12/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -101,14 +105,14 @@ def xcomp(Ft, N_, pairs, M,C,R):  # core
...
-         if pairs := [(_N,N) for _N,N in combinations(med_,2) if not any(_N in L.N_ and N in L.N_ for L in _N.rim)]:
+         if pairs := get_pairs(g_):
             med_ = list(dict.fromkeys(N for P in pairs for N in P))
             if GL_:= xcomp(Ft, med_, pairs, M,C, R/len(med_)):
                 G_,L_ = GL_
                 for G in G_: G.H+=[[]]  # medoids: top lev is unpacked
     if len(g_) > 1:
         TT,C,R = sum_vt(g_)
-        if GL_:= xcomp(Ft,g_,combinations(g_,2), val_(TT,ttX),C,R): G_ += GL_[0]; L_ += GL_[1]
+        if GL_:= xcomp(Ft,g_,get_pairs(g_), val_(TT,ttX),C,R): G_ += GL_[0]; L_ += GL_[1]
```
kwcckw, line 115: So we need to get check g_'s rim here to get their non duplicated pairs since g_ maybe compared before if we call them via sub+:

    cross_comp(G, N_,lm,lc,lr+1, 'Nt')  # sub+, cross_comp

frame_2D_alg/agg_recursion.py
```diff
@@ -299,7 +303,7 @@ def trans_cluster(G):
                 for tFt in lev.N_:  # Lt doesn't form trans-links
                     for tL in tFt.N_:
                         if tL.m*wcN > ave*ccN:  # merge trans_link.N_.roots
-                            rt0 = getattr(tL.N_[0].root,'root',None).root; rt1 = getattr(tL.N_[1].root,'root',None).root  # CNs
+                            rt0 = getattr(tL.N_[0].root,'root',None); rt1 = getattr(tL.N_[1].root,'root',None)  # CNs
```
kwcckw, line 306: Sorry i was confused earlier, that additional `.root` is redundant since we are already getting `.root.root` here.

frame_2D_alg/agg_recursion.py
```diff
@@ -582,7 +589,7 @@ def comb_Ft(Nt, Lt, Bt, Ct, root,wTT):  # from sum2G, default Nt
     T = Copy_(Nt)  # temporary accumulator
     dF_ = []
     for Ft in Lt, Bt:  # connectivity forks
-        if Ft: dF_ += [comp_F(T, Ft, root.r,G)]; T.dTT,T.c,T.r = sum_vt([T,Ft],wTT=wTT)  # Bt*brrw?
+        if Ft: dF_ += [comp_F(T, Ft, root.r,G)]; T.dTT,T.c,T.r = sum_vt([T,Ft],wTT=wTT)  # Bt*brrw? (Bt here is redundant to add2F below? add2F(G,Bt,merge=2))
```
kwcckw, line 592: We sum Bt to T after the trans comp here, which partly redundant to the `add2F(G,Bt,merge=2)` below since T will be summed to G? I guess we shouldn't sum it here?

frame_2D_alg/agg_recursion.py
boris-kz: this is for med_, g_ is new Gs, they were not compared yet

frame_2D_alg/agg_recursion.py
boris-kz: this is comparison between forks, we are summing fork matches, not forks themselves

kwcckw: Okay, then this should be a typo since we are summing Ft, which is the Bt or Lt and it's wrong.

It should be:

        if Ft: dF = comp_F(T, Ft, root.r,G); dF_ += [dF]; T.dTT,T.c,T.r = sum_vt([T,dF],wTT=wTT)  # Bt*brrw? 
        else:  dF_ += [CF()]

frame_2D_alg/agg_recursion.py
kwcckw: For sub+, we recompare the Ns within G, so we actually should compare them? Those N_ should've compared before the cross_comp of sub+:

      if G.Lt:  # sub+
              cross_comp(G, N_,lm,lc,lr+1, 'Nt'

boris-kz: in sub+ there is no g_, that's only for agg+

frame_2D_alg/agg_recursion.py
boris-kz: that's summing in T, which is a temporary accumulator, doesn't matter if it's redundant

frame_2D_alg/agg_recursion.py
kwcckw: Yes, the problem is we are parsing G.N_ as G_ in sub+:

            cross_comp(G, N_,lm,lc,lr+1, 'Nt')  # sub+, cross_comp
            
Which it may be packed into `g_` if there's no centroid clustering between them:

    for G in G_:  # or already pruned?
        if gv_(G.m * ((G.c*wcC)/(G.r*ccC)) * ((len(G.N_)-1)*wL) - ave) and (Ct:= cluster_C(G.Nt, get_exemplars(G.N_,r,c),r,c)):
            C__+= Ct.N_; M+=Ct.m; C+=Ct.c; R+=Ct.r  # centroids / root
        else: g_ += [G]  # no Cs, or eval rdn?

boris-kz: So this added unpack should only be done if root is tile, else the medoids are the Gs

kwcckw: Okay, so for sub+, it needs the refinement via centroid clustering only, rather than recluster the same set of N_ again.

boris-kz: You got it backwards, centroid clustering *is* reclustering the same N_s

kwcckw: Okay right, thanks, so it needs centroid clustering before the reclustering, while for Gs, centroid clustering is actually optional.

boris-kz: at this point top Gs don't go through CC at all, it's all in sub+. 
but I am not sure about that, maybe CN|CC should be a top fork in cross_comp

kwcckw: > at this point top Gs don't go through CC at all, it's all in sub+.

Right now the top CC run through G_'s N_, rather than root's G_. For CN|CC top fork, we probably can make this cluster_N in xcomp conditional? So that it returns from either cluster_N or cluster_C?

    return cluster_N(Ft, get_exemplars(N_,lr,lc), lr,lc), L_

boris-kz: CC|CN selection should depend on each exemplar. But then we will have both clustering forks in the same loop, which is very messy

boris-kz: astra:

**Your last comment identifies the right granularity: selection per exemplar, not one CN/CC choice for the entire `cross_comp` call. But that does not require interleaving both algorithms in one large loop.**

### 1. Separate scheduling from cluster growth

I would use a common exemplar loop to select and dispatch the operation, while keeping connectivity expansion and centroid refinement separate. Accumulate their results and finalize the containing forks once.

However, **calling the existing `cluster_N` or `cluster_C` once per exemplar is not sufficient**. They currently own batch-level state: `cluster_N` marks membership through `N.fin` and updates `Ft`; `cluster_C` clears centroid memberships across `Ft.N_`, optionally invokes joint `cluster_P`, and replaces the containing `Ct`. Those responsibilities must remain outside the per-exemplar dispatch; otherwise a later seed can erase earlier results.

The substantive decision is **which criterion best explains the exemplar’s surrounding nodes**, not merely an intrinsic property of the exemplar. My proposed distinction: favor CN where predictive value depends on transitive relationships; favor CC where direct similarity to a representative accounts for enough of that value more cheaply. The initial selection can be provisional as the neighborhood expands.

### 2. The discussion mixes two hierarchy levels

`cluster_C(G.Nt, get_exemplars(G.N_, ...))` clusters **members within each G**, not **the Gs within their parent**. The current `cross_comp` does the former; kwcckw correctly identifies that distinction late in the thread.

Consequently, making the final `cluster_N(...)` conditional would introduce a different operation, not simply relocate the existing CC operation. It would also choose once for the whole batch, contrary to your per-exemplar requirement.

I would first make the working collection explicit: **the exemplars, candidate nodes, and resulting membership must refer to the same hierarchy level**. Unpacking `G.N_` is a separate descent decision. Whether the root is a tile is an indirect way of expressing that distinction.

### 3. The duplicate-comparison objection was valid for the implemented path

The claim that `g_` contains only new graphs does not hold for the current shared function: sub+ passes existing `G.N_`, and the fallback branch places those same objects into `g_`. kwcckw eventually traces precisely that path. Checking existing comparisons there is therefore justified unless the routing is changed.

But the PR introduces a separate error: **line 108 calls `get_pairs(g_)` inside the medoid branch; it should use `med_`.** When CC succeeds for every input graph, `g_` is empty, so the selected medoids are never cross-compared.

### 4. `T` being temporary does not resolve the `Bt` duplication

On this point, kwcckw’s objection is correct, but their proposed replacement with `dF` is not.

The propagation is explicit: `T` accumulates `Nt`, `Lt`, and `Bt`; `add2F(G,T,merge=2)` transfers those weighted totals into persistent `G`; then `add2F(G,Bt,merge=2)` includes `Bt` again. `merge=2` suppresses membership concatenation, **not** accumulation of `dTT`, `c`, and `r`. The resulting graph values are subsequently recomputed and consumed by `sum2G`.

Replacing `Ft` with `dF` changes another contract: the second comparison would use `Nt` plus the first comparison’s derivatives, rather than accumulated `Nt + Lt`. Keep the fork accumulation in `T`, keep comparison results in `Xt`, and remove the second addition of `Bt`.

**The architectural priority is to separate hierarchy descent, per-exemplar method selection, and batch-level state updates. Their current entanglement—not the existence of two clustering criteria—is what makes the proposed loop messy.**

kwcckw: > CC|CN selection should depend on each exemplar. But then we will have both clustering forks in the same loop, which is very messy

Yes, but it should be not much changes from the existing process?

Both forks should be optional, but both forks' G should have the same depth now and we can simply merge them for the next recursion?

boris-kz: we will have a mixture of CNs and CCs in each expansion loop, they will overlap so we need to increment their r accordingly. CCs don't add composition depth, they are only used to select next-level exemplars: medoids

boris-kz: worked for 10 min:

Draft core using the PR’s helpers. `fC(E)` leaves the per-exemplar selector explicit; shared-member match is a provisional overlap measure. Caller wiring, trans-link merging, and `cluster_P` are not included.

```python
def cluster_(Ft, E_,_r,_c, fC):  # fC(E): select CC, else CN

    def reform(T):  # temporary summary, no H / member-root updates
        N_ = T.N_; c_ = np.array([n.c for n in N_]); c = c_.sum(); w_ = c_/c
        T._r = _r + sum(n.r*w for n,w in zip(N_,w_))
        if T.fC:
            T.c=c; T.w = c_ @ np.array(T.m_) / c  # membership value, separate from centroid data
            w_ = c_*T.m_; w_ /= w_.sum()
        for a in ('dTT','kern','yx','span','m','d'):
            setattr(T,a, sum(getattr(n,a)*w for n,w in zip(N_,w_)))
        if not T.fC:
            T.dTT,T.c,r = sum_vt(N_+T.L_+T.B_); T._r = _r+r
            T.m,T.d = val_(T.dTT,ttcN,1); T.w=T.m
        A_ = [n.angl[0] for n in N_ if n.angl is not None]
        T.angl = [sum(A_), np.sign(T.dTT[1] @ ttcC[1])] if A_ else None
        T.r = T._r + T.olp

    def overlap(T_):  # provisional: shared member match across current alternatives
        for T in T_:
            m_ = dict(zip(T.N_,T.m_)); M = sum(max(m,0)*n.c for n,m in m_.items())
            if not T.fC: M += sum(max(L.m,0)*L.c for L in T.L_)  # distinct relational value
            O = 0
            for _T in T_:
                if _T is T: continue
                for n,m in zip(_T.N_,_T.m_):
                    if n in m_: O += max(min(m,m_[n]),0)*n.c
            T.olp = O/(M or eps); T.r = T._r+T.olp  # replace, not accumulate per loop

    T_ = []
    for E in sorted(E_, key=lambda E: (-E.w,E.id)):
        T = CL(N_=[E],root=Ft, **{a:copy(getattr(E,a)) for a in ('dTT','m','d','c','r','kern','yx','span','angl')})
        T.fC=bool(fC(E)); T.typ=2 if T.fC else 1; T.nF='Ct' if T.fC else 'Nt'
        T._r=T.r+_r; T.r=T._r; T.wTT=Ft.wTT
        T.m_,T.d_,T.B_,T.olp = [E.m],[E.d],[],0; T_ += [T]

    while T_:
        Tt_,Up,change = [],0,False; r_ = {T:T.r for T in T_}
        for T in T_:  # one expansion per cluster, from the same previous-round state
            in_ = set(T.N_); rim_ = list(dict.fromkeys(L for n in T.N_ for L in n.rim))
            L_,B_ = [],[]
            if T.fC:
                n_,m_,d_ = [],[],[]
                for n in dict.fromkeys(T.N_+[n for L in rim_ for n in L.N_]):
                    m,d = val_(base_comp(T,n)[0],ttcC,1)
                    if m*n.c*wcC > ave*(T.r+n.r+ccC): n_+=[n]; m_+=[m]; d_+=[d]
                old = dict(zip(T.N_,zip(T.m_,T.d_))); new = dict(zip(n_,zip(m_,d_)))
                Up += sum(abs(new.get(n,(0,0))[0]-old.get(n,(0,0))[0]) +
                          abs(new.get(n,(0,0))[1]-old.get(n,(0,0))[1]) for n in old.keys() | new.keys())
            else:
                n_,L_,B_ = list(T.N_),list(T.L_),list(T.B_)
                for L in rim_:
                    m,d = nt_vt(*L.N_)
                    if m > ave*(T.r-1):
                        n = L.N_[0] if L.N_[1] in in_ else L.N_[1]
                        if n not in in_:
                            span = np.sqrt(len(in_))
                            if span > 3:
                                iM = sum(l.m for l in n.rim if any(_n in in_ for _n in l.N_))
                                if iM/(span*decay) < ave*(T.r-1): continue
                            n_ += [n]
                        L_ += [L]
                    elif d > avd*(T.r-1): B_ += [L]
                n_,L_,B_ = [list(dict.fromkeys(F_)) for F_ in (n_,L_,B_)]
                m_,d_ = [n.m for n in n_],[n.d for n in n_]
            change |= set(n_) != in_
            if n_: Tt_ += [(T,n_,m_,d_,L_,B_)]

        T_ = []
        for T,n_,m_,d_,L_,B_ in Tt_:
            T.N_,T.m_,T.d_,T.L_,T.B_ = n_,m_,d_,L_,B_
            T_ += [T]
        merged_ = []
        for T in T_:  # merge intersecting CNs; CC overlap remains alternative membership
            if not T.fC:
                for G in list(merged_):
                    if not G.fC and set(T.N_) & set(G.N_):
                        for a in ('N_','L_','B_'): setattr(T,a,list(dict.fromkeys(getattr(T,a)+getattr(G,a))))
                        merged_.remove(G); change=True
                T.m_,T.d_ = [n.m for n in T.N_],[n.d for n in T.N_]
                T.B_ = [L for L in T.B_ if len(set(L.N_) & set(T.N_)) == 1]
            reform(T); merged_ += [T]

        T_ = merged_; overlap(T_)
        keep_ = []
        for T in T_:
            w,c = (wcC,ccC) if T.fC else (wcN,ccN)
            if T.w*T.c*w*((len(T.N_)-1)*wL) > ave*T.r*c: keep_ += [T]
        change |= len(keep_) != len(T_); T_ = keep_
        dr = sum(abs(T.r-r_[T]) for T in T_)
        if not change and dr <= ave and Up*wcC <= avd*(_r+ccC*len(T_)): break

    # Publish memberships once; caller installs Ct and schedules sub+/agg+.
    C_ = [T for T in T_ if T.fC]; Ct = CF(N_=C_,nF='Ct',root=Ft.root,wTT=Ft.wTT)
    med_ = []
    for C in C_:
        C.root=Ct
        C.m_,C.d_ = map(list,zip(*(val_(base_comp(C,n)[0],ttcC,1) for n in C.N_)))
        med_ += [C.N_[np.argmax(C.m_)]]  # existing nodes, unchanged composition depth
    if C_: Ct.dTT,Ct.c,Ct.r = sum_vt(C_); Ct.m,Ct.d = val_(Ct.dTT,ttcC,1)

    G_ = []
    for T in T_:
        if T.fC: continue
        Ft_ = []
        for N_,nF in zip((T.N_,T.L_,T.B_),('Nt','Lt','Bt')):
            F = CF(N_=N_,nF=nF)
            if N_:
                F.dTT,F.c,F.r = sum_vt(N_); F.m,F.d = val_(F.dTT,ttcN,1)
                F.r += _r+T.olp  # preserve cost through later fork recomputation
            Ft_ += [F]
        c_ = list(dict.fromkeys(C for n in T.N_ for C in n.C_)); ct = CF(N_=c_,nF='Ct')
        if c_: ct.dTT,ct.c,ct.r = sum_vt(c_); ct.m,ct.d = val_(ct.dTT,ttcC,1)
        G = comb_Ft(*Ft_,ct,Ft,wTT=ttcN)  # composition / H added only for CNs
        G.olp=T.olp; G_ += [G]
    return G_, Ct, list(dict.fromkeys(med_))  # separate composition depths for cross-comp
```

frame_2D_alg/agg_recursion.py
boris-kz, line 86: so the root process now should be comp_N_, we can unpack it in cross_comp:

def cross_comp(pL_,r, tnF=None,root=2,fall=0):  # incremental-distance cross_comp, max dist depends on prior match

    L_,N_,mrg_ = [],[],[]
        if _N != N and fall or m>0 and gv_(m*(lc*wN/(lr*cN)) - ave*(r+cN)):  # marginal -gV
            L_+= [Link]; N_+= [_N,N]
            if _N.root_ and gv_(Link.m*wF- ave*(Link.r+cF)):
                add2F(_N,N,1); mrg_ += [N]  # may not already be merged
                for n in N.N_:
                    for rt in n.root_:  # [C,m,d]
                        if rt[0] is N: rt[0] = _N  # keep m,d positions
                if Link in N.rim: N.rim.remove(Link)  # packed if full
                Link.N_ = [_N,_N]  # replaces the merged N
                for pt in pL_[i+1:]:  # dist, dy_dx, _N,N, lc,lr, pTT,m,d  (replaces in pL_)
                    if pt[2] is N: pt[2] = _N
                    elif pt[3] is N: pt[3] = _N
    N_ = set(N_)
    if mrg_: N_ = list(N_ - set(mrg_))
    for N in set(N_): N.Rt = sum2F(N.rim, root=N, nF='Rt') if N.rim else None
    if L_:
        TT,C,R = sum_vt(L_,wTT=ttN)
        cV = FV_(CoF.get(), TT,C,R)  # +ve Ls for oF, no oF.N_?
        root.L_ = L_; oF_[CoF.get().nF].V_ += [cV]  # +-/ comp
        if gv_(val_(TT,ttcN) * (C*wcN /(r*ccN)) * ((len(L_)-1)*wL) - ave):  # return +ve, store -ve gate Vs
            e_ = get_exemplars({N for L in L_ for N in L.N_}, R,C)  # +ve Ls only
            return cluster_(root.Nt, e_,R,C)  # sum2G -> agg+

boris-kz: Yes: `cross_comp` can absorb the comparison loop, followed by exemplar selection and mixed clustering. The preliminary per-G CC pass is no longer needed.

The important change is **removing the old merge block**. `_N.root_` indicates centroid membership, not that `_N` is a centroid. With medoids as inputs, `add2F(_N,N,1)` combines the original nodes’ member lists and changes their parameters before `cluster_` evaluates CN/CC alternatives. Merging belongs in `cluster_`. 

Lines 3–4 of your snippet:

```python
    L_,N_ = [],[]
```

Delete lines 10–21, retaining the preceding `L_` and `N_` accumulation.

For the tail, preserve the comparison-only return used by `comp_F`: it consumes raw links and adds their derivatives to `dF` and `rL`, rather than forming clusters. Also remove the second `V_` append: after combining the functions, `FV_` already records that value in the same function’s accumulator.  

Lines 22–29, for the `Nt` clustering path:

```python
    if L_:
        TT,C,R = sum_vt(L_,wTT=ttN)
        cV = FV_(CoF.get(), TT,C,R)
        if tnF: return L_,TT,C,R,cV  # comp_F: derivatives only

        N_ = set(N_)
        for N in N_: N.Rt = sum2F(N.rim, root=N, nF='Rt') if N.rim else CF(root=N,nF='Rt')
        Lt = root.Lt
        Lt.N_,Lt.dTT,Lt.c,Lt.r = L_,TT,C,R
        Lt.m,Lt.d = val_(TT,ttX,1)
        if gv_(val_(TT,ttcN) * (C*wcN /(R*ccN)) * ((len(L_)-1)*wL) - ave):
            e_ = get_exemplars(N_,R,C)  # selection by current rim value
            return cluster_(root.Nt,e_,R,C)
```

This uses the resulting `R` in the clustering gate and updates the link-fork values, not just its membership list. The non-`tnF` path requires an actual graph `root`.

Two interfaces still need adjustment: `comp_F` must call the renamed function, and boundary clustering must retain its separate `Bt` destination—`tnF` is not that destination selector. My earlier `cluster_` draft also still requires `fC` and merely returns `G_,Ct,med_`; your four-argument call assumes selection moves inside it. Publishing those outputs and scheduling subsequent cross-comparison remain to be added, so `# sum2G -> agg+` is not implemented by that draft.

boris-kz: sub+ is optional rng+ followed by independently gated reclustering. Existing links already support clustering; rng+ only extends the compared range. Clustering can proceed without new comparisons, provided the existing evidence passes the higher filter.

That changes the current `cross_comp` structure: `if L_:` should control incorporating new links and recording comparison value, **not access to `cluster_`**. The clustering gate must evaluate existing plus newly added evidence, rather than only `TT` from the new batch. Likewise, exemplar selection must retain the existing candidate pool, not reconstruct it solely from new-link endpoints. In sub+, new links extend rather than replace the relevant link set.

Agg+ starts with newly formed graph comparands and establishes their connectivity. Sub+ already has connectivity and re-evaluates grouping at higher `r`, optionally supported by longer-range comparisons. The CN/CC expansion rules remain shared; the prerequisites for entering clustering differ.

kwcckw: > CCs don't add composition depth, they are only used to select next-level exemplars: medoids

When we cluster G and CC from a same set of Ns, CC don't add composition depth as they do not have new H level?

> so the root process now should be comp_N_, we can unpack it in cross_comp:

Okay, and from cluster_ above, node.fC is determined in get_exemplars, and that's mainly determined by node's d?

Also i think it's better to group them into 2 groups based on node.fC and cluster them separately before merge the Gs in the final step?

boris-kz: We don't need node fC, it's a different typ in T and len H in medoid. 
Yes, I think E.d represents local variance, that should favor CC

kwcckw: Okay, right now we need to parse fC function to determine T.fC during the clustering:

     T.fC=bool(fC(E)

I think we can skip that and use a `def fC()` instead?

Or it could be just as simpler as `T.fC = T.d > ave`?

PR #267, opened 2026-09-13
9/13/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -271,35 +278,36 @@ def overlap(T_):  # not reviewed
...
+        Up,change = 0,False; r_ = {T:T.r for T in T_}
+        for T in copy(T_):  # explore T frontier
+            in_ = set(T.N_); rim_ = set(L for n in T.N_ for L in n.rim); L_,B_ = [],[]
             if T.fC:
                 n_,m_,d_ = [],[],[]
-                for n in dict.fromkeys(T.N_+[n for L in rim_ for n in L.N_]):
+                for n in set(T.N_+[n for L in rim_ for n in L.N_]):  # internal + frontier
                     m,d = val_(base_comp(T,n)[0],ttcC,1)
-                    if m*n.c*wcC > ave*(T.r+n.r+ccC): n_+=[n]; m_+=[m]; d_+=[d]
+                    if m*n.c*wcC > ave*(T.r+n.r+ccC): n_+=[n]; m_+=[m]; d_+=[d]  # conditional inclusion of potential members into C now?
```
kwcckw, line 299: I edited the code and not sure it's a good idea to merge both CC and CG processes within the same pipeline since we actually can pack them separately into 2 functions.

Also the centroid clustering here packs member n conditionally, while it's default previously on all frontier since eval is per C, so this eval should be skipped?

frame_2D_alg/agg_recursion.py
```diff
@@ -83,7 +83,7 @@ def cent_TT(dTT, r):  # EM-like weight attr matches | diffs by their match to th
 - forward: selective extend cross-comp, clustering across tiles, re-order centroids by eigenvalues
 - feedback filter updates 
 '''
-def cross_comp(pL_,r, tnF=None,root=2,fall=0, nF='Nt',fC=None):  # fC(E) for clustering
+def cross_comp(root,pL_,r,nF='Nt',tnF=None):  # fall is redundant now?
```
boris-kz, line 86: fall is to skip the eval, which may not be worth the cost

frame_2D_alg/agg_recursion.py
```diff
@@ -94,12 +94,19 @@ def cross_comp(pL_,r, tnF=None,root=2,fall=0, nF='Nt',fC=None):  # fC(E) for clu
...
         if gv_(val_(TT,ttcN) * (C*wcN/(R*ccN)) * ((len(L_)-1)*wL) - ave):
             e_ = get_exemplars(N_,R,C)  # select by current rim value
-            return cluster_(getattr(root,nF),e_,R,C,fC)  # G_,Ct, med_
+            G_, Ct, med_ = cluster_(getattr(root,nF),e_,R,C)
+        if G_:
+            Ft = sum2F(G_, root=root,nF=nF)
+            if nF == 'Nt': Ft.H += [root.Nt]
+            setattr(root,nF,Ft)
+            root.dTT,root.c,root.r = sum_vt([root.Nt,root.Lt,root.Bt]); root.m,root.d = val_(root.dTT,root.wTT,fd=1)
+        # med_ for the next agg+?
```
boris-kz, line 108: next agg+ is for both med_ and G_, med_ is an optional unpacking
but all this can be in cluster_

boris-kz, line 108: next agg+ is for both med_ and G_, med_ is an optional unpacking
but all this can be in cluster_

frame_2D_alg/agg_recursion.py
boris-kz: yes, the frontier should be packed unconditionally.
the point of merging is to track member overlap between the forks
we can do that with separate functions, but then one of them must have a default priority, and I think that should be member-specifc.

boris-kz: **Priority should be member–cluster-specific, not fork-wide.** The same CN and CC can share two members while CN provides stronger support for one and CC for the other.

For each shared member, compare its contribution to that CN—support through that CN’s links—with its match to the CC’s centroid, on a commensurate value/cost basis. A natural rule is to charge the weaker membership for redundancy to the stronger one. That cost belongs to the **membership**, then contributes to the cluster’s `r`; increasing the member’s own `r` would penalize both representations. Both memberships can remain when their additional value justifies their costs.

Separate functions do not inherently require default priority. That happens when one completes and commits its memberships before the other starts, without reconsideration. They can instead advance one expansion each from the same previous-round membership state, followed by joint overlap evaluation and `r` updates. A merged `cluster_` still needs that coordination: merely putting both forks in one loop does not remove first-processed advantage.

My earlier draft missed this distinction: `overlap()` charged both alternatives symmetrically, while the CN’s `m_` merely copied `n.m`, rather than measuring its support within that particular CN. Those values cannot establish member-specific priority. Also, `fC(E)` selects a cluster’s growth rule; it does **not** determine that cluster’s priority over competing memberships.

So the essential shared operation is **member-wise comparison of competing memberships after each expansion**, with separate CN and CC growth rules and recomputed—not repeatedly accumulated—overlap costs.

boris-kz: so we can have common cluster_ that calls loop-wise cluster_N and  cluster_C?

Yes: a common `cluster_` calls **`cluster_N` and `cluster_C` for one expansion/refinement round each**, retaining their working clusters and frontiers between calls.

Both use the same previous-round memberships and costs. `cluster_N` produces proposed memberships with their connectivity support; `cluster_C` produces proposed memberships with their centroid matches. Then `cluster_` jointly evaluates competing memberships per member, recomputes overlap costs, and updates both forks before the next round. **Call order gives neither fork priority.**

Shared membership initialization, overlap accounting, continuation, and final packing belong in `cluster_`. The child functions should not independently clear shared `root_`, claim nodes through global `fin`, or finalize the containing forks. A cluster that stops expanding remains in the overlap evaluation; changed competition may justify revisiting it.

Final packing remains distinct: CNs add composition; CCs remain in `Ct` and select existing medoids.


Yes. `cluster_` owns the common expansion loop; `cluster_N` and `cluster_C` each perform **one expansion/refinement step**, retaining their own cluster state and frontiers between calls.

Both read the same previous-round memberships and `r`. After both return tentative updates, `cluster_` evaluates shared members, assigns member-specific redundancy to weaker memberships, updates cluster `r`, and decides which clusters continue. Sequential function calls then imply no fork priority, provided the first call does not commit shared-state changes before the second.

The division is:

* `cluster_N`: connectivity-based inclusion, merging, and boundary tracking.
* `cluster_C`: centroid-based inclusion and centroid refinement.
* `cluster_`: shared membership accounting, overlap costs, continuation decisions, and final packing.

A cluster that stops expanding must remain available for overlap accounting; it has not disappeared as an alternative representation. Final packing adds composition for CNs, while CCs remain in `Ct` and supply medoids.

kwcckw: > the point of merging is to track member overlap between the forks

I see, so that will be added to G.olp?, which will contribute to the redundancy.

> one of them must have a default priority, and I think that should be member-specifc.

If the priority is based on member (N), the redundancy could be added on both forks, which we need to track both forks' roots from each N?

boris-kz: see astra discussion above.
and here is a new draft:

`agg_recursion`, replace uploaded lines 281–405. This retains `fC(E)` for initial fork selection and uses the Ct-aware `comb_Ft` from the PR. Overlap uses member-specific match/cost as a provisional measure. `cluster_P`, trans-link merging, and recursion scheduling are not included. 

[[Draft source](sandbox:/mnt/data/cluster_draft.py)](sandbox:/mnt/data/cluster_draft.py)

```python
def cluster_(Ft, E_,_r,_c, fC):  # fC(E): CC, else CN; returns G_,Ct,med_

    def pack(N_, nF, root=None):  # summary only, no member or root-value updates
        F = CF(N_=N_,nF=nF,root=root,wTT=Ft.wTT)
        if N_:
            F.dTT,F.c,F.r = sum_vt(N_)
            F.m,F.d = val_(F.dTT,ttcC if nF=='Ct' else ttcN,1)
        return F

    def reform(T):  # temporary CLs; no tentative writes to node.root_
        c_ = np.array([n.c for n in T.N_]); c = c_.sum(); w_ = c_/c
        T._r = _r + sum(n.r*w for n,w in zip(T.N_,w_))
        if T.fC:
            w_ = c_*T.m_; w_ = w_/w_.sum() if w_.sum() else c_/c
            for a in ('dTT','kern','yx','span','m','d'):
                setattr(T,a,sum(getattr(n,a)*w for n,w in zip(T.N_,w_)))
            A_ = [n.angl[0] for n in T.N_ if n.angl is not None]
            T.angl = [sum(A_),np.sign(T.dTT[1] @ ttcC[1])] if A_ else None
            T.c = c
            T.m_,T.d_ = map(list,zip(*(val_(base_comp(T,n)[0],ttcC,1) for n in T.N_)))
            T.w = c_ @ T.m_ / c  # membership value, separate from centroid data
        else:
            T.dTT,T.c,r = sum_vt(T.N_+T.L_+T.B_); T._r = _r+r
            T.m,T.d = val_(T.dTT,ttcN,1); T.w = T.m
            T.m_,T.d_ = [],[]
            for n in T.N_:  # support within this CN, not the node's intrinsic m
                L_ = [L for L in T.L_ if n in L.N_]
                m,d = val_(sum_vt(L_)[0],ttcN,1) if L_ else (0,0)
                T.m_ += [m]; T.d_ += [d]
        T.r = T._r+T.olp

    def overlap(T_):  # stronger membership charges weaker; ties split the charge
        root__ = {}
        for T in T_:
            T.r_ = {}; w,c = (wcC,ccC) if T.fC else (wcN,ccN)
            for n,m in zip(T.N_,T.m_):
                v = max(m,0)*w / (c*(_r+n.r))
                root__.setdefault(n,[]).append((T,v))
        for n,rt_ in root__.items():
            for T,v in rt_:
                T.r_[n] = sum(_v/v if _v>v else .5 if _v==v else 0
                              for _T,_v in rt_ if _T is not T) if v>0 else 0
        for T in T_:
            T.olp = sum(n.c*T.r_[n] for n in T.N_) / sum(n.c for n in T.N_)
            T.r = T._r+T.olp  # replace overlap, don't accumulate it per round

    oldC_ = {C for n in Ft.N_ for C,m,d in n.root_ if getattr(C,'source',None) is Ft}
    G_,C_,root__ = [],[],{}
    for i,E in enumerate(sorted(E_,key=lambda E:E.id)):
        T = CL(root=Ft,wTT=Ft.wTT, **{a:copy(getattr(E,a)) for a in
               ('dTT','m','d','c','r','kern','yx','span','angl')})
        T.i=i; T.fC=bool(fC(E)); T.typ=2 if T.fC else 1; T.nF='Ct' if T.fC else Ft.nF
        T.N_,T.L_,T.B_ = [E],[],[]; T.m_,T.d_ = [E.m],[E.d]
        T._N_ = list(dict.fromkeys(n for L in E.rim for n in L.N_ if n is not E))
        T._r=_r+E.r; T.r=T._r; T.r_={}; T.olp=0; T.fin=0
        (C_ if T.fC else G_).append(T)

    while G_ or C_:
        old_ = {T.i:(set(T.N_),dict(zip(T.N_,zip(T.m_,T.d_))),dict(T.r_)) for T in G_+C_}
        G_ = cluster_N(G_,_r,root__)
        C_ = cluster_C(C_,_r)  # neither call writes shared member state
        T_ = G_+C_
        for T in T_:
            if not T.fin: reform(T)
        root__ = overlap(T_)
        keep_ = []
        for T in T_:
            w,c = (wcC,ccC) if T.fC else (wcN,ccN)
            if T.w*T.c*w*((len(T.N_)-1)*wL) > ave*T.r*c: keep_ += [T]
        if len(keep_) != len(T_): root__ = overlap(keep_)  # removed proposals no longer charge survivors

        for T in keep_:
            N_,md_,r_ = old_[T.i]; md = dict(zip(T.N_,zip(T.m_,T.d_)))
            dr = max((abs(T.r_.get(n,0)-r_.get(n,0)) for n in set(T.r_) | set(r_)),default=0)
            up = sum(abs(md[n][0]-md_.get(n,(0,0))[0]) +
                     abs(md[n][1]-md_.get(n,(0,0))[1]) for n in T.N_)
            if T.fC:
                T.fin = up*wcC <= avd*(T.r+ccC*len(T.N_)) and dr<=ave
            else: T.fin = set(T.N_)==N_ and dr<=ave
        G_ = [T for T in keep_ if not T.fC]; C_ = [T for T in keep_ if T.fC]
        if not any(not T.fin for T in keep_): break

    Ct = pack(C_,'Ct',Ft.root)  # no composition / H increment
    for n in {n for C in list(oldC_)+C_ for n in C.N_}:
        n.root_ = [rt for rt in n.root_ if rt[0] not in oldC_]  # replace this scope only
    for C in C_:
        C.root=Ct; C.source=Ft
    med_ = list(dict.fromkeys(C.N_[np.argmax(C.m_)] for C in C_))

    out_ = []
    for T in G_:
        Ft_ = [pack(N_,nF) for N_,nF in zip((T.N_,T.L_,T.B_),('Nt','Lt','Bt'))]
        for F in Ft_:
            if F: F.r += _r+T.olp  # persistent fork costs, not member n.r
        c_ = list(dict.fromkeys(C for n in T.N_ for C in n.C_))  # lower-level Ct only
        G.r_=T.r_; G.olp=T.olp; out_ += [G]
    return out_,Ct,med_  # caller installs results and schedules sub+/agg+


def cluster_N(G_,_r,root__):  # one CN expansion round, previous-round memberships

    out_ = []
    for _G in G_:
        G = copy(_G); G.N_=list(_G.N_); G.L_=list(_G.L_)
        if not G.fin:
            in_ = set(_G.N_); span = np.sqrt(len(in_))
            rim_ = list(dict.fromkeys(L for n in _G.N_ for L in n.rim))
            for L in rim_:
                if len(set(L.N_) & in_) != 1: continue
                n = next(n for n in L.N_ if n not in in_)
                l_ = [l for l in n.rim if any(_n in in_ for _n in l.N_)]
                v = max(val_(sum_vt(l_)[0],ttcN),0)*wcN / (ccN*(_r+n.r))
                olp = sum(_v/v if _v>v else .5 if _v==v else 0
                          for i,_v in root__.get(n,[]) if i!=G.i) if v>0 else 0
                r = _G._r-1+olp
                m,d = nt_vt(*L.N_)
                if m > ave*r:
                    if span > 3:
                        iM = sum(l.m for l in l_)
                        if iM/(span*decay) < ave*r: continue
                    G.N_ += [n]; G.L_ += [L]
            G.N_ = list(dict.fromkeys(G.N_)); G.L_ = list(dict.fromkeys(G.L_))
        out_ += [G]

    merged_ = []
    for G in out_:  # merge same-fork intersections, including stopped CNs
        while True:
            join_ = [H for H in merged_ if set(G.N_) & set(H.N_)]
            if not join_: break
            for H in join_:
                G.N_ = list(dict.fromkeys(G.N_+H.N_))
                G.L_ = list(dict.fromkeys(G.L_+H.L_))
                merged_.remove(H); G.fin=0
        merged_ += [G]

    for G in merged_:
        in_ = set(G.N_); G.B_ = []
        for L in dict.fromkeys(L for n in G.N_ for L in n.rim):
            m,d = nt_vt(*L.N_)
            if all(n in in_ for n in L.N_):
                if m > ave*(G._r-1) and L not in G.L_: G.L_ += [L]
            elif m <= ave*(G._r-1) and d > avd*(G._r-1): G.B_ += [L]
    return merged_


def cluster_C(C_,_r):  # one CC refinement round; inclusion precedes template evaluation

    out_ = []
    for _C in C_:
        C = copy(_C)
        if not C.fin:
            C.N_ = list(dict.fromkeys(_C.N_+_C._N_))  # unconditional current + frontier
            C.m_,C.d_ = map(list,zip(*(val_(base_comp(_C,n)[0],ttcC,1) for n in C.N_)))
            in_ = set(C.N_)
            C._N_ = list(dict.fromkeys(n for N in C.N_ for L in N.rim
                                      for n in L.N_ if n not in in_))
        out_ += [C]  # cluster_ reforms templates, then evaluates both forks together
    return out_
```

kwcckw: Okay, so the root__ packs each N and their possible G/CC from both forks, and redundancy is the ratio of roots' m. Each T.r_[n] is the priority of each N in the respective Ts in each fork,where higher redundancy means lower priority?

          def overlap(T_):  # stronger membership charges weaker; ties split the charge
              root__ = {}
              for T in T_:
                  T.r_ = {}; w,c = (wcC,ccC) if T.fC else (wcN,ccN)
                  for n,m in zip(T.N_,T.m_):
                      v = max(m,0)*w / (c*(_r+n.r))
                      root__.setdefault(n,[]).append((T,v))
              for n,rt_ in root__.items():
                  for T,v in rt_:
                      T.r_[n] = sum(_v/v if _v>v else .5 if _v==v else 0
                                    for _T,_v in rt_ if _T is not T) if v>0 else 0
              for T in T_:
                  T.olp = sum(n.c*T.r_[n] for n in T.N_) / sum(n.c for n in T.N_)
                  T.r = T._r+T.olp  # replace overlap, don't accumulate it per round

frame_2D_alg/agg_recursion.py
boris-kz: I can't believe the amont of fluff you added here, it should all be in packed in callees
something like this:

def cross_comp(root,pL_,r,nF='Nt',tnF=None, fall=1):

    L_,N_ = [],[]
        if _N != N and (fall or (m>0 and gv_(m*(lc*wN/(lr*cN)) - ave*(r+cN)))):
            Link.rTT = np.abs(pTT-Link.dTT) / eps_(Link.dTT); L_+=[Link]; N_+=[_N,N]  # relative prediction error/oF
    if L_:
        m,d,tt,c,r = sum_vt(L_,wTT=ttN, fm=1); cV = FV_(CoF.get(),tt,c,r)
        if tnF:  # comp_F: no cluster_, recursion
            return L_,m,d,tt,c,r,cV
        Ft = getattr(root,nF)  # Nt|Bt, each should be CN with Lt,Ct?  
        Lt = root.Ft.Lt; Lt.N_,Lt.m,Lt.d,Lt.dTT,Lt.c,Lt.r = L_,m,d,tt,c,r 
        for n_,R in (G_,med_), (root,root.H[0]):  # xcomp per composition level, med_=H[0], transitive to root? 
                cross_comp(R, proj_L_(combinations(n_,2),R,r), r,R.nF)

Astra:
The two recursive paths make sense, but **fork ownership and composition-level ownership need to be explicit before recursion.**

**`Nt` and `Bt` as processing scopes.** Each independently processed fork needs its own members, links, centroids, and hierarchy. Promoting those forks to `CN` is consistent; that storage nesting does not itself add composition. The assignment is therefore `Lt = Ft.Lt`, not `root.Ft.Lt`. Each preserved level used as a recursive root needs the same capabilities.

**`med_` is a selection from the previous composition level, not that level itself.** Before replacing the current members with `G_`, preserve their source scope—including its existing links and `Ct`. Medoids reference members of that scope. Using `H[0]` is possible as a new explicit convention, but the earlier code appends saved levels; it does not establish that index as the immediately preceding level. 

For `Bt`, the preserved scope must belong to that fork, not unconditionally to `root.H`, which is tied to `Nt`. Likewise, the destination of a recursive call cannot generally be inferred from `R.nF`: a container’s label in its parent is not necessarily the name of the child fork being processed.

Your final loop intends these pairings:

```python
        for n_,R in (G_,root), (med_,root.H[0]):
```

That corrects the transposed pairs, but assumes the level-packing contract above has already been implemented. The current [[cluster draft, lines 81–96](sandbox:/mnt/data/cluster_draft.py)](sandbox:/mnt/data/cluster_draft.py) returns its outputs without installing the new level or preserving the previous one.

**The sub+ path still needs clustering outside `if L_:`.** New comparisons update the existing scope; they do not determine whether that scope can be reclustered. Append new links to its existing `Lt`, retain its existing candidate nodes, and evaluate clustering from the combined evidence at the higher filter. The medoid path should exclude already-compared pairs so it remains rng+.

**“Transitive to root” requires an actual update path.** New medoid comparisons change the original members’ rims. Reclustering those members may then change the CNs represented in `G_`. Merely storing the lower scope in `H` does not refresh the parent’s member list or summary values. I would therefore resolve the lower-level medoid work before cross-comparing the resulting `G_`, or explicitly revisit the parent after lower-level changes.

Finally, the `+1` currently affects only the recursion gate. The intended higher filter must also reach `proj_L_`, the recursive call, and clustering when there are no new links.

frame_2D_alg/agg_recursion.py
boris-kz: yes, so we can call it prior()

frame_2D_alg/agg_recursion.py
kwcckw: for n_,R in (G_,root), (med_,root.H[0]):

The `med_` actually are the selected `N_` of current agg+, so the subsequent cross_comp are actually recluster those medoids again, which i think similar with sub+.

Also for the next agg+ clustered Gs via the current medoids, that should have the same composition level as the current agg+ N_'s clustered G, which we need to merge their H?

frame_2D_alg/agg_recursion.py
kwcckw: Alright, and also `list(dict.fromkeys(_C.N_+_C._N_))` could be `list(set(_C.N_+_C._N_))` if the unique order is not important.

frame_2D_alg/agg_recursion.py
boris-kz: med_ is N_ of existing Gs, which is H[0] of new Gs
so their Gs will be the same level as existing Gs, we then need  
Ft.N_ += med_Gs, no direct agg+

frame_2D_alg/agg_recursion.py
boris-kz: right

frame_2D_alg/agg_recursion.py
kwcckw: > med_ is N_ of existing Gs, which is H[0] of new Gs

Those new Gs should be clustered by med_, so med_ should be in new G.N_ rather than new G.H[0]? Since both Gs are the same level, their H should be the same and it depends on their G.N_'s H only?

Also this is adding more overlapping and redundancy since those medoid could be in C, G, or the new Gs of medoids now, we probably need to check that when cluster new G from medoids.

boris-kz: New Gs from G_ are GGs, new Gs from med_ are same level as G_

kwcckw: Okay, so for med Gs fork, we probably need a flag to identify that since they shouldn't replace the root's Ft.

Also this section should be moved to cross_comp?

            Ft = sum2F(G_, root=root,nF=nF)
            if nF == 'Nt': Ft.H += [root.Nt]
            setattr(root,nF,Ft)
            
Since we need to run the cross_comp of medoids first, and then merge medoid's Gs to the current G_ before we sum the merged G_ to the updated root.Ft again.

boris-kz: Add a flag to sum2F, I don't want that trivial crap in cross comp

PR #268, opened 2026-09-14
9/14/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -341,8 +338,24 @@ def priority(T_):  # stronger membership charges weaker; ties split the charge
...
+    cG_ = []
+    if med_:
+        R = F2N(Ft)  # Ft is new Ft.H[0] when new Ft is formed with combined G_ 
+        if gv_(R.m*R.c*wX - ave*(R.r+1+cX)):
+            cG_ = cross_comp(R, proj_L_(combinations(med_,2),R,_r), _r,R.nF)
+    # update root with the combined oG_ and mG_, so this needs to be after xcomp of med_
+    root = Ft.root
+    if G_:=oG_+ (cG_ or []):
+        new_Ft = sum2F(G_, root=root,nF=Ft.nF)
+        if Ft.nF == 'Nt': new_Ft.H += [Ft]  # med_ is already in Ft.N_
```
kwcckw, line 351: > med_=H[0]: sum to root?

We pack Ft.N_ to new Ft.H[0] here when forming new level, and actually med_ is already part of the existing Ft.N_, so there's no need to pack med_ only again?

frame_2D_alg/agg_recursion.py
```diff
@@ -341,8 +338,24 @@ def priority(T_):  # stronger membership charges weaker; ties split the charge
             if F: F.r += _r+T.olp  # persistent fork costs, not member n.r
         c_ = list(set(C for n in T.N_ for C in n.C_))  # lower-level Ct only
         G = comb_Ft(*Ft_,pack(c_,'Ct'),Ft,wTT=ttcN)
-        G.r_=T.r_; G.olp=T.olp; oG_ += [G]
-    return oG_,med_
+        G.r_=T.r_; G.olp=T.olp; oG_ += [G]  
+    cG_ = []
+    if med_:
+        R = F2N(Ft)  # Ft is new Ft.H[0] when new Ft is formed with combined G_
```
kwcckw, line 344: > Add a flag to sum2F, I don't want that trivial crap in cross comp

So that shouldn't be necessary now since the root of med is H[0], which is actually Ft since we may pack Ft to new Ft.H when we have new G_.

frame_2D_alg/agg_recursion.py
```diff
@@ -166,7 +163,7 @@ def comp_F(_F, F, ir=0, rL=None):
             else:
                 if nF != 'Nt': [(F2N(_N),F2N(N)) for _N,N in Np_]  # convert for comp_N_ below
                 if Lt := cross_comp(rL,proj_L_(Np_,2, r),r,nF,tnF=1):  # root = 2 to use N pair span to compute decay?
-                    pass  # all this is in cross_comp:
+                    pass  # all this is in cross_comp: (we return early when tnF=1, do we really need the agg+ for trans comp too?)
```
kwcckw, line 166: We actually return early from cross_comp when tnF = 1:

    if L_:
        m,d,tt,c,r = sum_vt(L_,wTT=ttN, fm=1); cV = FV_(CoF.get(),tt,c,r)
        if tnF:  # comp_F: no cluster_, recursion
            return L_,m,d,tt,c,r,cV

And do we need the same recursive cross_comp for each trans_fork? Else it's actually better to return the params since we just need to sum them into dF and rL.

boris-kz: not a direct agg+. trans-comp is actually the same as med+ xcomp, they both extend across root Gs?

frame_2D_alg/agg_recursion.py
boris-kz: this for new Gs formed by xcomp med_, they are not in GG.N_ yet

frame_2D_alg/agg_recursion.py
boris-kz: but we still need to add resulting derivatives in root, that's not automatic for H levs

frame_2D_alg/agg_recursion.py
```diff
@@ -96,13 +96,10 @@ def cross_comp(root,pL_,r,nF='Nt',tnF=None, fall=1):
             return L_,m,d,tt,c,r,cV
         for N in (N_:= list(set(N_))):
             N.Rt = sum2F(N.rim,root=N,nF='Rt') if N.rim else CF(root=N,nF='Rt')
-        Ft = getattr(root,nF); Lt = Ft.Lt; Lt.N_,Lt.m,Lt.d,Lt.dTT,Lt.c,Lt.r = L_,m,d,tt,c,r  # Nt | Bt
```
boris-kz, line 99: we need that Ft, it can be Nt or Bt

frame_2D_alg/agg_recursion.py
```diff
@@ -96,13 +96,10 @@ def cross_comp(root,pL_,r,nF='Nt',tnF=None, fall=1):
...
+        Lt = root.Lt; Lt.N_,Lt.m,Lt.d,Lt.dTT,Lt.c,Lt.r = L_,m,d,tt,c,r  # Nt | Bt
         if gv_(val_(tt,ttcN) * (c*wcN/(r*ccN)) * ((len(L_)-1)*wL) - ave):
-            G_,med_ = cluster_(Ft, get_exemplars(N_,r,c),r,c)
-        for n_,R in (G_,root), (med_,root.H[0]):  # xcomp per composition level, med_=H[0]: sum to root?
-            if n_ and gv_(R.m*R.c*wX - ave*(R.r+1+cX)):  # mdecay(L_)-decay? eval len n_?
-                cross_comp(R, proj_L_(combinations(n_,2),R,r), r,R.nF)
```
boris-kz, line 104: med_ xcomp should be separate, it's a different level

frame_2D_alg/agg_recursion.py
kwcckw: The new Ft.N_ would be G_, and existing N_ will be packed as Ft.H[0]. For new Gs formed by xcomp med_, those are same level as G_, which will be merged with G_ and packed within the new Ft.N_. For GG_, they will be formed with non merging G_ only (which is much higher level), so how they are actually related to Gs formed by xcom med_ here?

frame_2D_alg/agg_recursion.py
kwcckw: Yes, that should be added to root when we sum the existing G_ and the new G_ formed by med_ to new Ft?

    new_Ft = sum2F(G_, root=root,nF=Ft.nF)

frame_2D_alg/agg_recursion.py
kwcckw: >  trans-comp is actually the same as med+ xcomp, they both extend across root Gs?

I think no since the cross-fork comparison should be still within the root? Since they are the same as med xcomp, then it shouldn't return early, but we only need the L_ instead of G_ here?

frame_2D_alg/agg_recursion.py
kwcckw: Okay, but `Lt = Ft.Lt` should be `Ft.root.Lt ` only if nF = 'Nt'? For Bt, it's not relevant here?

frame_2D_alg/agg_recursion.py
kwcckw: Yes, but we need to run them first so that we can get their med_'s G_ to merge with the existing same level's G_?

frame_2D_alg/agg_recursion.py
boris-kz: I think we need Bt.Lt too

frame_2D_alg/agg_recursion.py
boris-kz: yes so my order should be reversed, but we still need these two calls here

frame_2D_alg/agg_recursion.py
boris-kz: ok I thought this is for trans-links, but there it's just extended clustering

frame_2D_alg/agg_recursion.py
boris-kz: right, that's in sum2G

frame_2D_alg/agg_recursion.py
boris-kz: G_ formed by xcom med_ is merged in g_ before g_ forms GGs

frame_2D_alg/agg_recursion.py
kwcckw: Okay, so G_ should be merged with the first run's med G_ during the second run if there's any med G_ before we parse them to the subsequent cross_comp. In that case, we will need to move the root's Ft and H related ops back to cross_comp now?

frame_2D_alg/agg_recursion.py
kwcckw: Okay, thanks. I was mistaken earlier and thinking we are using only the existing G_ but not the new G_ from med.

But that should be the same since GG.N_ should be merged G_, and GG.H should be G.N_, which could be N_ or med_? If existing G_ and med Gs are the same level, their H should be N_ since N_ included med_?

frame_2D_alg/agg_recursion.py
kwcckw: It clusters between `N.Ft and _N.Ft`, which i think they should be still be still within the root.

frame_2D_alg/agg_recursion.py
boris-kz: probably

frame_2D_alg/agg_recursion.py
boris-kz: right

frame_2D_alg/agg_recursion.py
boris-kz: Ft = getattr(root,nF); Lt = Ft.Lt; Lt.N_,Lt.m,Lt.d,Lt.dTT,Lt.c,Lt.r = L_,m,d,tt,c,r  # Nt | Bt
        for fsub, n_,R in (1,0), (G_,root), (med_,root.H[0]):  # xcomp per composition level, med_=H[0]: sum to root?
                g_ = cross_comp(R, proj_L_(combinations(n_,2),R,r), r,R.nF)
                if fsub and g_: G_+=g_  # same level
        return G_  # for above

kwcckw: This line should be:

    for fsub, n_,R in zip((1,0), (med_,G_), (root.H[0],root))

So that we xcomp the med first.

Also the section above should be in cross_comp right? To pack the new H level into root, that should be done right after the cross_comp too?

frame_2D_alg/agg_recursion.py
kwcckw: So Lt = Ft.Lt only if nF != 'Nt'? Else Lt should be from root?

frame_2D_alg/agg_recursion.py
boris-kz: ok, so we can sum that internally:

def cross_comp(root,pL_,r,nF='Nt',dF=None,rL=None, fall=1):

    L_,N_ = [],[]
        if _N != N and (fall or (m>0 and gv_(m*(lc*wN/(lr*cN)) - ave*(r+cN)))):
            Link.rTT = np.abs(pTT-Link.dTT) / eps_(Link.dTT); L_+=[Link]; N_+=[_N,N]  # relative prediction error/oF
    if L_:
        m,d,tt,c,r = sum_vt(L_,wTT=ttN, fm=1); cV = FV_(CoF.get(),tt,c,r)
        if dF:  # comp_F: no cluster_, recursion
            add2F(dF, CF(N_=L_,m=m,d=d,dTT=tt,c=c,r=r), merge=1); add2F(rL, dF, merge=2)
            return

frame_2D_alg/agg_recursion.py
boris-kz: main Lt can be in both root and Nt

frame_2D_alg/agg_recursion.py
boris-kz: right

frame_2D_alg/agg_recursion.py
kwcckw: But if we update Nt.Lt here, the root.Lt will never get updated?

frame_2D_alg/agg_recursion.py
kwcckw: Okay, so it should be in the else of if fsub:

    if g_:
        if fsub: G_+=g_  # same level
        else:     # pack new root H

frame_2D_alg/agg_recursion.py
boris-kz: so we don't need root.Lt?

frame_2D_alg/agg_recursion.py
boris-kz: that's done in sum2G, in cluster_?

frame_2D_alg/agg_recursion.py
```diff
@@ -341,8 +338,24 @@ def priority(T_):  # stronger membership charges weaker; ties split the charge
...
+            cG_ = cross_comp(R, proj_L_(combinations(med_,2),R,_r), _r,R.nF)
+    # update root with the combined oG_ and mG_, so this needs to be after xcomp of med_
+    root = Ft.root
+    if G_:=oG_+ (cG_ or []):
+        new_Ft = sum2F(G_, root=root,nF=Ft.nF)
+        if Ft.nF == 'Nt': new_Ft.H += [Ft]  # med_ is already in Ft.N_
+        root.dTT,root.c,root.r = sum_vt([root.Nt,root.Lt,root.Bt]); root.m,root.d = val_(root.dTT,root.wTT,fd=1)
+        setattr(root,Ft.nF,new_Ft)
+    setattr(root,'Ct',Ct)
+    # xcomp per composition level, med_=H[0]: sum to root?
```
boris-kz, line 355: so this is now in cross_comp

frame_2D_alg/agg_recursion.py
kwcckw: We need that because root.Lt is packed in comb_Ft, so i think it's better to pack in root for consistency:

    G = CN(Nt=Nt,Lt=Lt,Bt=Bt,Ct=Ct,root=root)
    
We probably will get Nt.Lt in med G fork only, where we parse H[0] as their root. But we should've run F2N on that H level before parsing it as root.

frame_2D_alg/agg_recursion.py
kwcckw: That H append in add_Nt doesn't add new level to root when we call new agg+ or sub+? That covers only the H levels up to the depth when we form the G, but not the new depth in agg+ or sub+.

boris-kz: it does add new level

frame_2D_alg/agg_recursion.py
boris-kz: I think all that should be Nt or Bt -specific

kwcckw: In that case, i guess we should init Lt in Nt, that's including Ct so that we have Lt and Ct per Nt?

frame_2D_alg/agg_recursion.py
kwcckw: Okay sorry, that's per G, but for root, we still need to add the new H level into root when there's new agg+ or sub+?

frame_2D_alg/agg_recursion.py
boris-kz: probably.
this is review of my current push:

Previous-conversation retrieval was partial. Against the available context, these are the main issues in [[agg_recursion(1).py](sandbox:/workspace/scratch/bdbe7fe55808/project_sources/05-agg_recursion-1-.py)](sandbox:/workspace/scratch/bdbe7fe55808/project_sources/05-agg_recursion-1-.py):

1. **New graphs aren’t committed to the parent hierarchy — lines 102–107.** `cluster_` returns `G_`, but neither function installs them in `root.N_` or `root.H`. They survive through members’ `n.root` pointers, but `frame_H` ignores the return at line 893, and `pack_seg` subsequently consumes the old hierarchy.

2. **`comp_F` reverses the sub-comparison gate — lines 166–170.** The `else` belongs to `if gv_(...)`. Consequently, a successful gate skips non-`Lt` sub-comparison, while a failed gate executes it. That `else` needs to belong to `if l`, inside the successful gate.

3. **Recursive fork results enter their parent twice — lines 96 and 146.** `cross_comp` adds `dF` into `rL`. Later, `comp_N` packs that same returned `dF` through `sum2F(dn_,L)`, which adds its statistics into `L` again at line 504. This gives recursively expanded forks extra weight in subsequent comparisons.

4. **`Bt` is also counted twice — lines 569–571.** The loop already incorporates `Bt` into `T`; then `add2F(G,T)` incorporates it into `G`. The following `add2F(G,Bt)` repeats it, changing persistent `G.c`, `G.r`, and `G.dTT`, which feed higher comparisons.

5. **Range recursion repeats existing comparisons — lines 697–707.** There is no existing-pair exclusion. Repeated pairs append additional links to both rims at line 149; `nt_vt` sums those links separately, and `cluster_N` uses that inflated support for inclusion. Increasing `r` doesn’t itself restrict comparisons to new pairs.

6. **Prediction error remains disconnected from feedback — lines 92–94.** `rTT` is stored, but its only reader is `sum_vt(fr=1)`, which has no callers. `FV_` receives actual comparison statistics, so the new prediction error does not reach the function values consumed by `ffeedback`.

Three immediate wiring issues also block this flow:

* **Line 100:** `Ft.Lt` is absent on the ordinary `CF` forks constructed by `comb_Ft`.
* **Line 103:** the three sequences require `zip`; currently each iteration supplies two values for three targets.
* **Line 303:** `fC(E)` has no definition or import in the supplied code.

frame_2D_alg/agg_recursion.py
boris-kz: that should be in sum2G too

kwcckw: Okay, i guess we can include that after the cross_comp in sum2G? This should be the same for Bt fork too:


    if G.Lt:  # sub+
            cross_comp(G,proj_L_(combinations(N_,2),G, lr), lr+1, 'Nt')

frame_2D_alg/agg_recursion.py
kwcckw: > Line 303: fC(E) has no definition or import in the supplied code.

So for fC, we can skip that and eval T.fC using E.d within the cluster_ function?

PR #269, opened 2026-09-15
9/15/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -545,24 +546,31 @@ def sum2G(ft_, fTT, root=None, init=1):  # core clustering function
...
-    Ft_ += [sum2F(list(set(C_)), root.Ct,nF='Ct') if C_ else CF(nF='Ct')]  # add multiple root_ in Cs?
+    Ft_ += [sum2F(list(set(C_)), root.Ct if root else None ,nF='Ct') if C_ else CF(nF='Ct')]  # add multiple root_ in Cs?
     G = comb_Ft(*Ft_, root, wTT=fTT)
     N_ = G.N_; N=N_[0]; r=G.r; Av=ave+avd
-    if G.Lt:  # sub+
+    if G.Nt.Lt:  # sub+
         Lt = G.Lt; L_,lm,lc,lr = Lt.N_,Lt.m,Lt.c,Lt.r  # no levR = 1/len(L_): represented by c
         if gv_(lm*lc*wX - Av* (lr+1+cX)):  # mdecay(L_)-decay?
-            cross_comp(G,proj_L_(combinations(N_,2),G, lr), lr+1, 'Nt')  # sub+, cross_comp  (may get endless recursion sum2G -> cross_comp -> cluster_N)
+            F2N(G.Nt)  # add Lt to Nt
```
kwcckw, line 555: > that should be in sum2G too

So we should add the H level in sum2G, right after the cross_comp here in sum2G? Bt fork should follows the same way?

frame_2D_alg/agg_recursion.py
```diff
@@ -100,7 +100,8 @@ def cross_comp(root,pL_,r,nF='Nt',dF=None,rL=None, fall=1):
         Ft = getattr(root,nF); Lt = Ft.Lt; Lt.N_,Lt.m,Lt.d,Lt.dTT,Lt.c,Lt.r = L_,m,d,tt,c,r  # Nt | Bt
         if gv_(val_(tt,ttcN) * (c*wcN/(r*ccN)) * ((len(L_)-1)*wL) - ave):
             G_,med_ = cluster_(Ft, get_exemplars(N_,r,c),r,c)
-            for fsub, n_,R in (1,0), (med_,G_),(root.H[0],root):  # xcomp per lev, med_=H[-1], no lower levs?
+            if med_ and not root.H: root.H += [CF(root=root,nF=nF)]
```
boris-kz, line 103: this should be in sum2G

kwcckw: For G clustered from PPs, their H should be empty because PP.N_ is empty? So if there's sub+ on those Gs, we don't have any root.H[0]?

boris-kz: it should be added in cluster_ above, that's where all nesting happens

frame_2D_alg/agg_recursion.py
```diff
@@ -139,7 +140,7 @@ def comp_H(_Nt,Nt, Link):  # tentative pre-comp, before comp N_?
         if N.typ < 3:  # L | C | Nt, merge?
             for _n,n in product(_N.N_,N.N_): dn_ += [comp_N(_n,n,r,c, rL=L)]  # CN L.nt, rL spec in comp.N
         else:  # CN
-            for i,(_Ft,Ft, tnF) in enumerate(zip((_N.Nt,_N.Lt,_N.Bt,_N.Ct),(N.Nt,N.Lt,N.Bt,N.Ct),('Nt','Lt','Bt','Ct'))):
+            for i,(_Ft,Ft, tnF) in enumerate(zip((_N.Nt,_N.Nt.Lt,_N.Bt,_N.Ct),(N.Nt,N.Nt.Lt,N.Bt,N.Ct),('Nt','Lt','Bt','Ct'))):
```
boris-kz, line 143: so the principal forks now are Nt,Bt,Ct, while Lt is compared and summed as part of Nt and Bt.

frame_2D_alg/agg_recursion.py
boris-kz: a level is added in cluster_, it's not a default in cross_comp

kwcckw: So the existing cluster_ actually didn't use sum2G, but using comb_Ft only, i guess we need to replace that with sum2G and add new H level after the sum2G in cluster_.

Also for cluster_P, that should be applied to C_, right after the while refinement loop?

frame_2D_alg/agg_recursion.py
kwcckw: Okay, then we not really comp Lt as trans fork here.

frame_2D_alg/agg_recursion.py
boris-kz: yes, sum2G was called in old cluster_N, that got lost

kwcckw: On the cluster_P, would it be better if we apply it to C_ and G_?

Another thing is the fC in cluster_:

    T.fC=bool(fC(E))

That fC should be a function to determine if the exemplar should be cluster by connectivity or centroid clustering, or it could be just a single eval with E.d?

boris-kz: cluster_P is CC only, there is no refine for CN
fC is typ=2

frame_2D_alg/agg_recursion.py
boris-kz: yes, Lt is really parallel to Ft.N_

frame_2D_alg/agg_recursion.py
kwcckw: But why fC is E.typ == 2? Because we won't get any E.typ == 2 in the first iteration, which we need to eval without using their typ here?

boris-kz: I don't think we need it at all, reform should be Cs only:

def cluster_C(_C_,_r):  # one CC refinement round; inclusion precedes template evaluation

    C_ = []
    for _C in _C_:
        C = copy(_C)
        if not C.fin:
            C.N_ = list(set(_C.N_+_C._N_))  # unconditional current + frontier
            C.m_,C.d_ = map(list,zip(*(val_(base_comp(_C,n)[0],ttcC,1) for n in C.N_)))
            in_ = set(C.N_)
            C._N_ = list(set(n for N in C.N_ for L in N.rim for n in L.N_ if n not in in_))
            c_ = np.array([n.c for n in C.N_]); C.c = c = c_.sum(); w_ = c_ / c
            C._r = _r + sum(n.r * w for n, w in zip(C.N_, w_))
            w_ = c_ * C.m_; w_ = w_ / w_.sum() if w_.sum() else c_ / c
            for a in ('dTT','kern','yx','span','m','d'): setattr(C, a, sum(getattr(n, a) * w for n, w in zip(C.N_, w_)))
            A_ = [n.angl[0] for n in C.N_ if n.angl is not None]
            C.angl = [sum(A_), np.sign(C.dTT[1] @ ttcC[1])] if A_ else None
            C.w = c_ @ C.m_ / c  # membership value, separate from centroid data
            C.r = C._r + C.olp
        C_ += [C]  # cluster_ reforms templates, then evaluates both forks together
    return C_

frame_2D_alg/agg_recursion.py
boris-kz, line 754: how does it cancel-out?

frame_2D_alg/agg_recursion.py
```diff
@@ -750,7 +759,7 @@ def proj_TT(L, cos_d, dist, r, pTT, wTT, dec=1, fdec=0):     # accumulate L|N' p
     cos_d = (N.angl[0].dot(A) / ((np.hypot(*N.angl[0]) * dist) or eps)) * N.angl[1] if N.angl else 0  # int x ext angle alignment, mean=0
     iTT, eTT = np.zeros((2,9)),np.zeros((2,9)); c = 0
     wTT = CoF.get().wTT*ttPrj
-    for L in N.L_+ N.B_:  # proj C-to-N links:
+    for L in N.Nt.L_+ N.B_:  # proj C-to-N links:
```
boris-kz, line 762: for L in N.Nt.L_+ N.Bt.L_:  # or Bt.L_ proj cancels Nt.L_ proj?

frame_2D_alg/agg_recursion.py
```diff
@@ -545,24 +546,31 @@ def sum2G(ft_, fTT, root=None, init=1):  # core clustering function
...
+            F2N(G.Bt)
+            if (sub := cross_comp(G, proj_L_(combinations([F2N(L) for L in Bt.N_],2),G,br),br, nF='Bt')):
+                Bt = sum2F(sub, root=G, nF='Bt')
+                Bt.H += [G.Bt]; setattr(G,'Bt',Bt)
         if rroot: Bt.brrw = Bt.m * (rroot.m * (decay * (rroot.span/G.span)))  # external lend only, need to subtract from root?
+    if G.Lt or G.Bt: G.dTT,G.c,G.r = sum_vt([G.Nt,G.Lt,G.Bt]); G.m,G.d = val_(G.dTT,G.wTT,fd=1)  # recompute after deeper sub
     FV_(CoF.get(), G.dTT, G.c, G.r)
     return G
 
 def comb_Ft(Nt, Lt, Bt, Ct, root,wTT):  # from sum2G, default Nt
```
boris-kz, line 571: so all this needs to be revised, I don't think we need to compare between fork types

frame_2D_alg/agg_recursion.py
kwcckw: Okay, reform is after we have the initial `C_`,  and we need to init `C_` with T.fC, which we need to identify that first? From each exemplar, we need to identify if they should be run by connectivity clustering or centroid clustering, and that will be determined by their d?

frame_2D_alg/agg_recursion.py
kwcckw: From my prior comment:

The decay is actually cancel out each other in val_:

vm = (m_/ eps_(m_+ad_)) @ wTT[0]

m_ = L.dTT[0] * dec
ad_ = abs( L.dTT[1] * cos_d * dec)

It's actually (L.dTT[0] * dec) / ([L.dTT[0] * dec + L.dTT[1] * cos_d * dec)

and it becomes (L.dTT[0]) / ([L.dTT[0]  + L.dTT[1] * cos_d) * dec/dec 

So decay cancel out each other.

I guess decay should be applied to cert only?

frame_2D_alg/agg_recursion.py
kwcckw: The Bt.N_ are those negative links, which is still relevant?

frame_2D_alg/agg_recursion.py
kwcckw: Okay, the primary ones are actually just Nt and Bt, which i guess we can skip them now.

frame_2D_alg/agg_recursion.py
boris-kz: We need an option for divisive clustering per loop, forming C: https://chatgpt.com/share/6aaa3ef8-a198-83ea-a475-945bd6b7c3e5

frame_2D_alg/agg_recursion.py
boris-kz: Decay is value-proportional, so it doesn't fully cancel. Also, dTT[1] decay is for d projection only. abs ds without d-specific md should not be projected, so we probably need to project only dTT[0]

frame_2D_alg/agg_recursion.py
boris-kz: in the opposite way, they cancel projection of Nt.L_

PR #270, opened 2026-09-16
9/16/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -271,35 +271,44 @@ def prioritize(T_):  # stronger membership charges weaker; ties split the charge
...
+    for i,E in enumerate(sorted(E_,key=lambda E:E.id)):
+        G_ += [init(E,i)]
     while G_ or C_:
         _T_ = {T.i:(set(T.N_),dict(zip(T.N_,zip(T.m_,T.d_))),dict(T.r_)) for T in G_+C_}
         G_ = cluster_N(G_,_r,root__)
-        C_ = cluster_C(C_,_r)
+        for G in G_:
+            if G.d*(G.c*wcC/(G.r*ccC))*((len(G.N_)-1)*wL) > ave:  # high-variance G eval from claude, and test G.N_ for C_?
+                for e in [e for e in G.N_ if e.exe]:  # only exemplars?
+                    C_ += [init(e,G.i)]
```
kwcckw, line 292: > We need an option for divisive clustering per loop, forming C: https://chatgpt.com/share/6aaa3ef8-a198-83ea-a475-945bd6b7c3e5

So something like this? C_ will be packed from the high variance's G's N_?

Also since sub+ is removed form sum2G now, the sub+ is actually done via the C fork here?

frame_2D_alg/agg_recursion.py
kwcckw, line 755: > Decay is value-proportional, so it doesn't fully cancel. Also, dTT[1] decay is for d projection only. abs ds without d-specific md should not be projected, so we probably need to project only dTT[0]

So TT here could be just `TT = np.array([L.dTT[0] * dec, L.dTT[1] * cos_d])`.

boris-kz: L.dTT[1] is actually a local component of decay, the projetcion of it's deviation should be added to the average decay.  

Yes: average decay supplies the baseline; the projected local deviation adjusts it. Adding the full local component would count its average contribution twice.

But L.dTT[1] contains signed parameter differences. To interpret them as decay, normalize by parameter magnitude and L.span, then project their deviation from the corresponding average along the target direction. Since decay ** ... represents retention, excess predicted loss reduces that factor.

[](sandbox:/workspace/scratch/ad137d8d4d5d/project_sources/05-agg_recursion-1-.py), `proj_N.proj_TT`, lines 742–746:

```python
        m_,d_ = L.dTT; ad_ = np.abs(d_); rdist = dist / L.span
        Dec = dist if fdec else decay ** (1 + rdist*dec)
        ddec_ = (ad_/eps_(m_+ad_) - (1-decay)) * cos_d * rdist  # projected local loss deviation
        dm_ = m_ * (1-Dec+ddec_)  # average loss + local deviation, in match units
        TT = np.array([m_-dm_, d_+np.where(d_<0, -dm_,dm_)])  # transfer lost match to signed difference
```

Syntax and isolated projection checks passed.

kwcckw: Okay, so this `m_-dm_` becomes `m_ - m_*(1-Dec+ddec_)` = `m_(1- (1+Dec-ddec))`, which is actually just `m_ * (Dec - ddec)`? Then actually the Dec is actually subtracting the ddec from dTT[1], rather than add to it?

frame_2D_alg/agg_recursion.py
boris-kz: right, C fork

frame_2D_alg/agg_recursion.py
boris-kz: Dec is subtracting the deviation of ddec, because high ddec should be reducing mdec, dproj interferes with mproj

frame_2D_alg/agg_recursion.py
```diff
@@ -271,35 +271,44 @@ def prioritize(T_):  # stronger membership charges weaker; ties split the charge
             T.r = T._r+T.olp
         return {n:[(T.i,v) for T,v in rt_] for n,rt_ in root__.items()}
 
-    _C_ = {C for n in Ft.N_ for C,m,d in n.root_ if getattr(C,'source',None) is Ft}
-    G_,C_,root__ = [],[],{}
-    for i,E in enumerate(sorted(E_,key=lambda E:E.id)):
+    def init(E,i):
```
boris-kz, line 274: this should be different for Gs and Cs?

frame_2D_alg/agg_recursion.py
boris-kz: inhibition by exemplars is local, so it's not relevant to centroids 

        for G in G_:= cluster_N(G_,_r,root__):  # preselected for G.m
            if G.d * (G.c*wcC/(G.r*ccC)) * ((len(G.N_)-1)*wL) > avd:  # high-variance G
                C_ += [init(N, G.i,fC=1) for N in G.N_ if N.m * (G.c*wcC/(G.r*ccC)) > ave]  # tentative, any high-m N?

frame_2D_alg/agg_recursion.py
boris-kz: also it seems redundant to pack()?

frame_2D_alg/agg_recursion.py
```diff
@@ -271,35 +271,44 @@ def prioritize(T_):  # stronger membership charges weaker; ties split the charge
...
-        G_ = [T for T in next_ if not T.fC]; C_ = [T for T in next_ if T.fC]
+        G_ = [T for T in next_ if T.typ==2]; C_ = [T for T in next_ if T.typ==1]
         if not any(not T.fin for T in next_):
             break
-    Ct = pack(C_,'Ct', Ft.root)  # no composition / H increment
-    if Ft.nF=='Nt': Ft.root.Ct = Ct  # not for Bt?
+    if C_:
+        C_ = cluster_P(C_,Ft)
+        Ct = pack(C_,'Ct', Ft.root)  # no composition / H increment
+        Ft.root.Ct = Ct  # not for Bt? Bt should be the same?
```
boris-kz, line 311: right, Bs may become Cs only in der+, where they become Nts

frame_2D_alg/agg_recursion.py
kwcckw: > if N.m * (G.c*wcC/(G.r*ccC)) > ave

Thanks, so this is reselecting the exemplar using N.m , rather than using the prior exe.

frame_2D_alg/agg_recursion.py
kwcckw: I see,thanks for clarifying.

frame_2D_alg/agg_recursion.py
kwcckw: Okay, i guess we can have the flag fC to differentiate them, but in current cluster_, we actually need m_ and d_ in both G_ and C_.

> also it seems redundant to pack()?

We probably can merge them, since pack init from array of Ns, while init using only a seed.

frame_2D_alg/agg_recursion.py
boris-kz, line 322: why if F?

frame_2D_alg/agg_recursion.py
boris-kz: yes, try that

frame_2D_alg/agg_recursion.py
```diff
@@ -561,13 +575,17 @@ def comb_Ft(Nt, Lt, Bt, Ct, root,wTT):  # assemble core and boundary with separa
         elif Ft is Nt and (L_ := [l for n in Nt.N_ for l in n.L_]):
             Ft.Lt = sum2F(L_,nF='Lt')
         Ft.Lt.root=Ft; Ft.Lt.nF='Lt'; Ft.Lt.wTT=Ft.wTT
-    G.Lt = Nt.Lt  # compatibility alias after replacement; Copy_ must preserve this relation
+    # G.Lt = Nt.Lt  # compatibility alias after replacement; Copy_ must preserve this relation  (no more G.Lt?)
```
boris-kz, line 578: do we still need G.Lt?

kwcckw: It's packed in Nt.Lt now? So Lt is per Nt or Bt fork now, i don't see why we still need it in G.Lt.

frame_2D_alg/agg_recursion.py
kwcckw: Not sure, it was added in the astra version, i guess it's not relevant if the fork is empty?

PR #271, opened 2026-09-17
9/17/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -253,14 +253,20 @@ def nt_vt(n,_n):
...
+            T.kern,T.yx,T.span,T.angl = copy(E.kern),copy(E.yx),E.span,copy(E.angl)
+            T.L_,T.B_,T.m_,T.d_ = [],[],[E.m],[E.d]; T._N_ = list(set(n for L in E.rim for n in L.N_ if n is not E))
+            T._r=_r+E.r; T.r=T._r; T.r_={}; T.olp=0; T.fin=0; T.i=T.id; T.typ=2 if fC else 3
+        return T
 
     def prioritize(T_):  # stronger membership charges weaker; ties split the charge
         root__ = {}
         for T in T_:
-            T.r_ = {}; w,c = (wcC,ccC) if T.typ==1 else (wcN,ccN)
+            T.r_ = {}; w,c = (wcC,ccC) if T.typ==2 else (wcN,ccN)
```
kwcckw, line 269: The existing types here is a bit confusing. For fC, their typ should be 2, while typ should be 3 for connectivity clustering? typ = 1 is for link only?

frame_2D_alg/agg_recursion.py
```diff
@@ -253,14 +253,20 @@ def nt_vt(n,_n):
 
 def cluster_(Ft, E_,_r,_c):  # fC(E): CC, else CN; returns G_,Ct,med_
 
-    def pack(N_, nF, root=None):  # summary only, no member or root-value updates
+    def pack(N_, nF='Nt', root=None, fseed=0, fC=0):  # summary only, no member or root-value updates
         m,d,dTT,c,r = sum_vt(N_, fm=1)
-        return CF(N_= N_,nF=nF,root=root,wTT=Ft.wTT,dTT=dTT,m=m,d=d,c=c,r=r)
+        T = (CF,(CN,CL)[fC])[fseed](N_=N_,nF=nF,root=root,wTT=Ft.wTT,dTT=dTT,m=m,d=d,c=c,r=r)
```
kwcckw, line 258: For init, we should use CL when fC is 1, while CN for the others?

That CF is only relevant to sum Ns to Ft.

frame_2D_alg/agg_recursion.py
```diff
@@ -269,38 +275,40 @@ def prioritize(T_):  # stronger membership charges weaker; ties split the charge
...
-            else:    T.fin = set(T.N_)==N_ and dr<=ave
-        G_ = [T for T in front_ if T.typ==2]; C_ = [T for T in front_ if T.typ==1]
-        if not any(not T.fin for T in front_):  # or remove from front?
-            break
+            Dr = max(Dr,dr); Up += up  # not using Dr now
+            if T.typ==2: T.fin = up*wcC <= avd*(T.r+ccC*len(T.N_)) and dr<=ave
+            else:        T.fin = set(T.N_)==N_ and dr<=ave
+        G_ = [T for T in front_ if T.typ==3]; C_ = [T for T in front_ if T.typ==2]
+        conv = 0  # eval per front_  
+        if Up:
```
kwcckw, line 303: The existing convergence eval is per T in front_, i think it should be per front_ to stop the iterations, which is similar with the cluster_P?

frame_2D_alg/agg_recursion.py
boris-kz: It's confusing because our forks keep changing.
sorry my Lt packing in Nt and Bt was wrong, Bt is also a links fork, it should only get Lt and own Bt when converted to Nt in der+.
So CN should still have Nt,Lt,Bt,Ct, because top Lt and Bt are not packed in N_ nodes

frame_2D_alg/agg_recursion.py
boris-kz: yes, that's for global convergence

frame_2D_alg/agg_recursion.py
boris-kz: you use typ 2 for CC, that's not CL?

boris-kz: astra:
<html><head></head><body><p><code inline="">CL</code> can represent a centroid with <code inline="">typ=2</code>: the class supplies its fields, while <code inline="">typ</code> controls comparison. The problematic change is using <code inline="">CN</code> for temporary connectivity clusters without updating initialization and copying.</p><p>Reviewed head <code inline="">27b0c91</code>. Three problems introduced by these changes:</p><ol><li><p><strong>CN seeds lose their member.</strong> At <a href="https://github.com/kwcckw/CogAlg/blob/27b0c91e5a932b9d2cb653f0dfaab0cba50d31fa/frame_2D_alg/agg_recursion.py#L258-L263">L258–263</a>, <code inline="">CN(N_=[E])</code> creates an empty <code inline="">Nt</code>; inherited <code inline="">CF.__init__</code> sees the existing <code inline="">N_</code> property and skips assignment. <code inline="">cluster_N</code> receives an empty seed, then <code inline="">prioritize</code> divides by zero. Explicitly assign <code inline="">T.N_</code> after construction.</p></li><li><p><strong>Rollback no longer restores the previous graph.</strong> <a href="https://github.com/kwcckw/CogAlg/blob/27b0c91e5a932b9d2cb653f0dfaab0cba50d31fa/frame_2D_alg/agg_recursion.py#L339">L339</a> shallow-copies CN, sharing <code inline="">Nt</code>, <code inline="">Lt</code>, and <code inline="">Bt</code>. Assigning or extending their member lists changes <code inline="">_G</code> too. Consequently, restoring <code inline="">_front_</code> restores changed memberships with old summaries and scores. Copy the mutable forks while retaining member identities.</p></li><li><p><strong>Global convergence ignores continuing overlap changes.</strong> <a href="https://github.com/kwcckw/CogAlg/blob/27b0c91e5a932b9d2cb653f0dfaab0cba50d31fa/frame_2D_alg/agg_recursion.py#L293-L311">L293–311</a> computes <code inline="">Dr</code> but excludes it from termination. Unchanged membership scores can produce <code inline="">Up=0</code>, <code inline="">T.fin=False</code>, yet <code inline="">conv=1</code>. That skips further expansion under the changed overlap costs. Removed members/proposals also contribute nothing to <code inline="">Up</code>.</p></li></ol><p>The discussion exposes three existing inconsistencies:</p><ol start="4"><li><p><strong>Your revised fork ownership is not implemented.</strong> <a href="https://github.com/kwcckw/CogAlg/blob/27b0c91e5a932b9d2cb653f0dfaab0cba50d31fa/frame_2D_alg/agg_recursion.py#L323-L327">L323–327</a> omits <code inline="">T.L_</code> and passes forks positionally as:</p>
comb_Ft parameter | Actual contents
-- | --
Nt | T.N_
Lt | T.B_
Bt | T.C_

<p>Internal links disappear from this handoff; boundary links supply core-link orientation. <code inline="">comb_Ft</code>, <code inline="">cross_comp</code>, and <code inline="">proj_N</code> also retain the earlier nested-link ownership.</p></li><li><p><strong>Centroids lose <code inline="">typ=2</code> downstream.</strong> <a href="https://github.com/kwcckw/CogAlg/blob/27b0c91e5a932b9d2cb653f0dfaab0cba50d31fa/frame_2D_alg/agg_recursion.py#L477-L492">L477–492</a> selects <code inline="">CL</code> using local <code inline="">typ=2</code>, but never passes that tag into construction. <code inline="">cluster_P</code> therefore produces <code inline="">typ=1</code> centroids, making subsequent <code inline="">comp_N</code> skip centroid subcomparison.</p></li><li><p><strong>High-variance Gs repeatedly seed the same members.</strong> <a href="https://github.com/kwcckw/CogAlg/blob/27b0c91e5a932b9d2cb653f0dfaab0cba50d31fa/frame_2D_alg/agg_recursion.py#L284-L287">L284–287</a> creates fresh Cs every round; <code inline="">set(C_)</code> cannot deduplicate distinct objects. These duplicates increase overlap charges and register as new change in <code inline="">Up</code>. Seeding needs remembered member identity or an explicit reseeding criterion.</p></li></ol><p>Targeted execution confirmed the empty seed, shared-fork mutation, lost centroid type, and premature convergence.</p></body></html>`CL` can represent a centroid with `typ=2`: the class supplies its fields, while `typ` controls comparison. The problematic change is using `CN` for temporary connectivity clusters without updating initialization and copying.

Reviewed head `27b0c91`. Three problems introduced by these changes:

1. **CN seeds lose their member.** At [[L258–263](https://github.com/kwcckw/CogAlg/blob/27b0c91e5a932b9d2cb653f0dfaab0cba50d31fa/frame_2D_alg/agg_recursion.py#L258-L263)](https://github.com/kwcckw/CogAlg/blob/27b0c91e5a932b9d2cb653f0dfaab0cba50d31fa/frame_2D_alg/agg_recursion.py#L258-L263), `CN(N_=[E])` creates an empty `Nt`; inherited `CF.__init__` sees the existing `N_` property and skips assignment. `cluster_N` receives an empty seed, then `prioritize` divides by zero. Explicitly assign `T.N_` after construction.

2. **Rollback no longer restores the previous graph.** [[L339](https://github.com/kwcckw/CogAlg/blob/27b0c91e5a932b9d2cb653f0dfaab0cba50d31fa/frame_2D_alg/agg_recursion.py#L339)](https://github.com/kwcckw/CogAlg/blob/27b0c91e5a932b9d2cb653f0dfaab0cba50d31fa/frame_2D_alg/agg_recursion.py#L339) shallow-copies CN, sharing `Nt`, `Lt`, and `Bt`. Assigning or extending their member lists changes `_G` too. Consequently, restoring `_front_` restores changed memberships with old summaries and scores. Copy the mutable forks while retaining member identities.

3. **Global convergence ignores continuing overlap changes.** [[L293–311](https://github.com/kwcckw/CogAlg/blob/27b0c91e5a932b9d2cb653f0dfaab0cba50d31fa/frame_2D_alg/agg_recursion.py#L293-L311)](https://github.com/kwcckw/CogAlg/blob/27b0c91e5a932b9d2cb653f0dfaab0cba50d31fa/frame_2D_alg/agg_recursion.py#L293-L311) computes `Dr` but excludes it from termination. Unchanged membership scores can produce `Up=0`, `T.fin=False`, yet `conv=1`. That skips further expansion under the changed overlap costs. Removed members/proposals also contribute nothing to `Up`.

The discussion exposes three existing inconsistencies:

4. **Your revised fork ownership is not implemented.** [[L323–327](https://github.com/kwcckw/CogAlg/blob/27b0c91e5a932b9d2cb653f0dfaab0cba50d31fa/frame_2D_alg/agg_recursion.py#L323-L327)](https://github.com/kwcckw/CogAlg/blob/27b0c91e5a932b9d2cb653f0dfaab0cba50d31fa/frame_2D_alg/agg_recursion.py#L323-L327) omits `T.L_` and passes forks positionally as:

   | `comb_Ft` parameter | Actual contents |
   | ------------------- | --------------- |
   | `Nt`                | `T.N_`          |
   | `Lt`                | `T.B_`          |
   | `Bt`                | `T.C_`          |

   Internal links disappear from this handoff; boundary links supply core-link orientation. `comb_Ft`, `cross_comp`, and `proj_N` also retain the earlier nested-link ownership.

5. **Centroids lose `typ=2` downstream.** [[L477–492](https://github.com/kwcckw/CogAlg/blob/27b0c91e5a932b9d2cb653f0dfaab0cba50d31fa/frame_2D_alg/agg_recursion.py#L477-L492)](https://github.com/kwcckw/CogAlg/blob/27b0c91e5a932b9d2cb653f0dfaab0cba50d31fa/frame_2D_alg/agg_recursion.py#L477-L492) selects `CL` using local `typ=2`, but never passes that tag into construction. `cluster_P` therefore produces `typ=1` centroids, making subsequent `comp_N` skip centroid subcomparison.

6. **High-variance Gs repeatedly seed the same members.** [[L284–287](https://github.com/kwcckw/CogAlg/blob/27b0c91e5a932b9d2cb653f0dfaab0cba50d31fa/frame_2D_alg/agg_recursion.py#L284-L287)](https://github.com/kwcckw/CogAlg/blob/27b0c91e5a932b9d2cb653f0dfaab0cba50d31fa/frame_2D_alg/agg_recursion.py#L284-L287) creates fresh Cs every round; `set(C_)` cannot deduplicate distinct objects. These duplicates increase overlap charges and register as new change in `Up`. Seeding needs remembered member identity or an explicit reseeding criterion.

Targeted execution confirmed the empty seed, shared-fork mutation, lost centroid type, and premature convergence.

frame_2D_alg/agg_recursion.py
kwcckw: Okay, but we removed that B fork in sum2G now, so we won't get converted Bt anymore?

frame_2D_alg/agg_recursion.py
kwcckw: If i'm not mistaken, typ 0 = CF, typ1 = CL, typ 2 = CC (using CL), typ 3 = CN?  So CC is a typ 2 CL, which we already decided that previously.

frame_2D_alg/agg_recursion.py
boris-kz: need to add that option in cross_comp

kwcckw: Okay, i guess we only need to add Bt fork recursion in cross_comp since C fork is done via med_.

boris-kz: Actually this link sub+ is connectivity clustering per G, not extensible centroid clustering.
So it should still be in sum2G

boris-kz: def sum2G(ft_, fTT, root=None, init=1):  # finalize cluster

    if not init:
        N_,_,ntt,nc,nr = ft_[0]; N_+=root.N_; ntt+=root.Nt.dTT; nc+=root.Nt.c; nr+=root.Nt.r; ft_[0] = N_,_,ntt,nc,nr
    Ft_ = []
        if ft: n_,_,tt,c,r = ft; Ft_+= [CF(N_=n_,nF=nF,dTT=tt,m=(vt:=val_(tt,wTT,1))[0],d=vt[1],c=c,r=r)]
        else:  Ft_ += [CF()]
    Ft_ += [sum2F(list(set(C_)), root.Ct if root else None ,nF='Ct') if C_ else CF(nF='Ct')]  # add multiple root_ in Cs?
    if Lt:= G.Lt:  # rng+'sub+
            cross_comp(G, proj_L_(combinations(N_,2),G, lr), lr+1, 'Nt')  # replace G Nt,Lt
    if Bt:= G.Bt:  # der+'sub+
        bd,bc,br = Bt.d,Bt.c,Bt.r; rroot = root.root if root.root else 0
        if N.typ!=1 and gv_(bd*bc*wX - Av*(br+1+cX)):  # no ddfork, eval len?
            cross_comp(F2N(G.Bt), proj_L_(combinations([F2N(L) for L in Bt.N_],2),G,br),br, nF='Bt')
        if rroot: Bt.brrw = Bt.m * (rroot.m * (decay * (rroot.span/G.span)))  # external lend, subtract from root?
    if G.Lt or G.Bt: G.dTT,G.c,G.r = sum_vt([G.Nt,G.Lt,G.Bt]); G.m,G.d = val_(G.dTT,G.wTT,fd=1)  # recompute after deeper sub
    FV_(CoF.get(), G.dTT, G.c, G.r)
    return G

boris-kz: astra:
The revised role for `sum2G` makes sense: **finalize each G, including its local link `sub+`.** The [[posted version](https://github.com/boris-kz/CogAlg/pull/271#discussion_r4041910579)](https://github.com/boris-kz/CogAlg/pull/271#discussion_r4041910579) still needs four connections resolved:

1. **The called operation remains broader than local connectivity clustering.** `cross_comp` invokes `cluster_`, which runs centroid refinement, then recursively compares medoids and returned Gs. `cluster_N` also follows member rims without restricting candidates to the enclosing G. Moving this call into `sum2G` therefore does not establish the proposed scope. Local `sub+` needs an explicit scope and clustering mode. [[Current calls](https://github.com/kwcckw/CogAlg/blob/27b0c91e5a932b9d2cb653f0dfaab0cba50d31fa/frame_2D_alg/agg_recursion.py#L99-L105)](https://github.com/kwcckw/CogAlg/blob/27b0c91e5a932b9d2cb653f0dfaab0cba50d31fa/frame_2D_alg/agg_recursion.py#L99-L105)

2. **Comment L21 targets the wrong fork after promotion.** `F2N(G.Bt)` moves the original boundary members into the promoted graph’s `Nt`. Passing `nF='Bt'` selects its newly created, empty boundary fork. Use `Nt` internally while retaining the outer attachment as `G.Bt`. Recursive dispatch also needs attention: `F2N` leaves the promoted object’s `nF='Bt'`. The existing `cross_comp` access to `Ft.Lt` must follow the restored graph-level `Lt` ownership.

3. **Recompute the promoted boundary graph before its parent.** After correcting that routing, `cross_comp` updates the promoted graph’s internal fork summaries. Its own `dTT,c,r,m,d` remain unchanged. Comment L23 then reads those stale values through `G.Bt`. Aggregate the promoted graph first, then aggregate `G`, excluding `Ct` at both levels. [[Fork summary update](https://github.com/kwcckw/CogAlg/blob/27b0c91e5a932b9d2cb653f0dfaab0cba50d31fa/frame_2D_alg/agg_recursion.py#L329-L332)](https://github.com/kwcckw/CogAlg/blob/27b0c91e5a932b9d2cb653f0dfaab0cba50d31fa/frame_2D_alg/agg_recursion.py#L329-L332)

4. **`cluster_` currently bypasses this finalization.** Its output Gs go directly through `comb_Ft`, so restoring `sub+` only in `sum2G` will not process those Gs. That output path must invoke the intended finalization too. [[L322–328](https://github.com/kwcckw/CogAlg/blob/27b0c91e5a932b9d2cb653f0dfaab0cba50d31fa/frame_2D_alg/agg_recursion.py#L322-L328)](https://github.com/kwcckw/CogAlg/blob/27b0c91e5a932b9d2cb653f0dfaab0cba50d31fa/frame_2D_alg/agg_recursion.py#L322-L328)

For `H`, this gives a precise distinction: an unprocessed `Bt` stays flat; successful `der+` develops the promoted graph’s `Nt.H`.

kwcckw: > cross_comp(F2N(G.Bt), proj_L_(combinations([F2N(L) for L in Bt.N_],2),G,br),br, nF='Bt')

So we may get Nt, Lt, Bt and Ct per Bt now.

Also since we have this `if Lt:= G.Lt:  # rng+'sub+`, wouldn't this be partly redundant to the existing C fork in cluster_? Since the C fork is unpacking high variance G.N_ and recluster them again.

frame_2D_alg/agg_recursion.py
boris-kz: right

frame_2D_alg/agg_recursion.py
boris-kz: that C fork is for clustering via existing rims, this is extended cross_comp

kwcckw: I see, but in the extended cross_comp, the number of pairs could be the same from a same set of Ns since the pairs are determined by their distance too? So that max in `if dist < max:` should be incremented with extend cross_comp?

    for _N, N in pairs:  # -> all-to-all pre-links
        if len(_N.H) != len(N.H): continue  # or comp x agg Lev?
        else:
            dy_dx = _N.yx - N.yx; dist = np.hypot(*dy_dx)  # rim angl is not canonic
            if dist < max:
                pTT = proj_V(_N, N, dist, dy_dx, root.m if root != 2 else decay ** (dist / ((_N.span + N.span) / 2)), r)  # based on current rim
                m, d = val_(pTT, ttN, 1)
                if fall or m > ave:
                    lc = min(_N.c, N.c); lr = r + (N.r + _N.r) / 2  # +|-match certainty
                    pL_ += [[dist, dy_dx, _N, N, lc, lr, pTT, m, d]]

boris-kz: yes, something like max*nsub, default=1

kwcckw: The nsub should be a local var and we need to parse it to the cross_comp? Because in sub+, the N_'s nsub should be remained the same.

frame_2D_alg/agg_recursion.py
```diff
@@ -269,38 +275,40 @@ def prioritize(T_):  # stronger membership charges weaker; ties split the charge
...
+            if _Up:  # first loop doesn't have prior
+                dUp_ += [_Up-Up]; dUp_ = dUp_[-5:]  # same with cluster_P, only last 5 loops
+                if sum(dUp_) < 0:  # break and preserve the last front_ before the divergence
+                    G_ = [T for T in _front_ if T.typ==3]; C_ = [T for T in _front_ if T.typ==2]; conv = 1
+                else: conv = Up <= avd
```
boris-kz, line 308: it should be cost per loop, not avd.

kwcckw: Sorry, that's a typo, it should be ave. We actually have Dr and Up now, so the cost should be based on the change in redundancy or m? Something like `(Up < ave or Dr < ave)`? Or both must be true?

kwcckw: The cost should be inverse of Up and Dr since we need to minimize them to converge?

boris-kz: we need a process cost per loop, so a loop should probably be a function
Dr would be added to that cost.

frame_2D_alg/agg_recursion.py
```diff
@@ -269,38 +275,40 @@ def prioritize(T_):  # stronger membership charges weaker; ties split the charge
...
-        if not any(not T.fin for T in front_):  # or remove from front?
-            break
+            Dr = max(Dr,dr); Up += up  # not using Dr now
+            if T.typ==2: T.fin = up*wcC <= avd*(T.r+ccC*len(T.N_)) and dr<=ave
+            else:        T.fin = set(T.N_)==N_ and dr<=ave
+        G_ = [T for T in front_ if T.typ==3]; C_ = [T for T in front_ if T.typ==2]
+        conv = 0  # eval per front_  
+        if Up:
+            if _Up:  # first loop doesn't have prior
+                dUp_ += [_Up-Up]; dUp_ = dUp_[-5:]  # same with cluster_P, only last 5 loops
```
boris-kz, line 305: this dUp is only needed if we want to project Up over multiple loops

frame_2D_alg/agg_recursion.py
boris-kz: astra:
 **One complete refinement round is a useful oF boundary**: expansion, centroid refinement, overlap priority, and pruning. Register it as a module-level function so the current tracer gives each round its own call record.

`Up > loop_cost + Dr` is coherent with two qualifications:

* **`loop_cost` should reflect that round’s workload.** The current `fc` estimates static code complexity; executed comparisons and callee work determine the process cost as the frontier changes. [[Cost construction](https://github.com/kwcckw/CogAlg/blob/27b0c91e5a932b9d2cb653f0dfaab0cba50d31fa/frame_2D_alg/meta_code.py#L144-L167)](https://github.com/kwcckw/CogAlg/blob/27b0c91e5a932b9d2cb653f0dfaab0cba50d31fa/frame_2D_alg/meta_code.py#L144-L167)
* **Redefine `Dr` as the change in redundancy cost charged against those updates.** Currently it is the maximum absolute change in membership redundancy. A reduction in overlap therefore produces positive `Dr` too. For your gate, increased redundancy raises the cost; reduced redundancy provides a credit. It also needs weighting onto the same scale as `Up`. [[Current calculation](https://github.com/kwcckw/CogAlg/blob/27b0c91e5a932b9d2cb653f0dfaab0cba50d31fa/frame_2D_alg/agg_recursion.py#L293-L298)](https://github.com/kwcckw/CogAlg/blob/27b0c91e5a932b9d2cb653f0dfaab0cba50d31fa/frame_2D_alg/agg_recursion.py#L293-L298)

The last round’s `Up` can estimate the next round’s update value. `dUp` becomes useful when projecting how that value changes over further rounds.

kwcckw: > this dUp is only needed if we want to project Up over multiple loops

Yes, that's for the case of divergence. After we pack the loop into a new function, this dUp should be packing each loop Up from the function?

boris-kz: divergence is just negative Up, then we need to pack _T_ instead of T_
same for accumulating it over multiple loops, we don't need that now.

astra draft:

`agg_recursion.py`, before L335:

```python
def cluster_loop(G_,C_,root__,_r, seed_,pack,prioritize):  # one round; helpers retain cluster_ scope

    F = CoF.get()
    old_ = {T.i:(T.typ,{n:(m,d,T.r_.get(n,0)) for n,m,d in zip(T.N_,T.m_,T.d_)}) for T in G_+C_}
    for T in G_+C_: T.fin=0  # continuation is global
    G_ = cluster_N(G_,_r,root__)
    for G in G_:
        if G.d * (G.c*wcC/(G.r*ccC)) * ((len(G.N_)-1)*wL) > avd:
            for N in G.N_:
                if N not in seed_ and N.m * (G.c*wcC/(G.r*ccC)) > ave:
                    C_ += [pack([N],root=G.root,fseed=1,fC=1)]; seed_.add(N)
    C_ = cluster_C(C_,_r)
    T_ = G_+C_; root__ = prioritize(T_); front_ = []
    for T in T_:
        w,c = (wcC,ccC) if T.typ==2 else (wcN,ccN)
        if T.w*T.c*w*((len(T.N_)-1)*wL) > ave*T.r*c: front_ += [T]
    if len(front_) != len(T_): root__ = prioritize(front_)

    Up,Dr = 0,0; new_ = {T.i:T for T in front_}
    for i in old_.keys() | new_.keys():
        T = new_.get(i); typ = T.typ if T is not None else old_[i][0]
        md_ = old_.get(i,(typ,{}))[1]
        md = {n:(m,d,T.r_[n]) for n,m,d in zip(T.N_,T.m_,T.d_)} if T is not None else {}
        w,c = (wcC,ccC) if typ==2 else (wcN,ccN)
        for n in md_.keys() | md.keys():  # includes removed memberships and proposals
            _m,_d,_rn = md_.get(n,(0,0,0)); m,d,rn = md.get(n,(0,0,0))
            Up += n.c*w*(abs(m-_m)+abs(d-_d))
            Dr += ave*n.c*c*(rn-_rn)  # signed redundancy cost, in update-value units

    F.fc = oF_[F.nF].fc; call_ = list(F.call_)
    while call_:  # static AST estimates, once per executed traced call
        f = call_.pop(); F.fc += oF_[f.nF].fc; call_ += f.call_
    F.w = Up-Dr; oF_[F.nF].V_ += [F.w]
    G_ = [T for T in front_ if T.typ==3]; C_ = [T for T in front_ if T.typ==2]
    return G_,C_,root__,gv_(Up - ave*F.fc - Dr)
```

L279:

```python
    G_,C_,root__,seed_ = [],[],{},set()
```

Replace L282–311:

```python
    while G_ or C_:
        G_,C_,root__,v = cluster_loop(G_,C_,root__,_r,seed_,pack,prioritize)
        if not v: break
```

[](sandbox:/workspace/scratch/26919bb9fa77/cluster_loop_draft.py) includes the registration, coefficient-unpacking, and seed-initialization edits.

kwcckw: Alright, and for this convergence eval in the draft above:

   return G_,C_,root__,gv_(Up - ave*F.fc - Dr)

That gv_ is redundant? Since it's just an eval for break?

boris-kz: This is for loop oF eval, same as any oF

kwcckw: From here:

    while G_ or C_:
        G_,C_,root__,v = cluster_loop(G_,C_,root__,_r,seed_,pack,prioritize)
        if not v: break
        
That's for convergence eval?

PR #272, opened 2026-09-18
9/18/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -587,11 +587,12 @@ def sum2G(ft_, fTT, root=None, init=1):  # finalize cluster
     if Lt:= G.Lt:  # rng+'sub+
         lm,lc,lr = Lt.m,Lt.c,Lt.r  # no levR = 1/len(L_): represented by c
         if gv_(lm*lc*wX - Av* (lr+1+cX)):  # mdecay(L_)-decay?
-            cross_comp(G, proj_L_(combinations(N_,2),G,lr), lr+1,'Nt')  # replace G Nt,Lt?
+            # or pack nsub into L or pL since L is always per level?
+            cross_comp(G, proj_L_(combinations(N_,2),G,lr,nsub+1), lr+1,'Nt',nsub=nsub+1)  # replace G Nt,Lt?
```
kwcckw, line 591: Right now i'm parsing the nsub as local var, but i think we can pack them in proj_L, into the pLs? Because pLs are per level, so their nsub is unique in each level of Ls.

frame_2D_alg/agg_recursion.py
```diff
@@ -277,21 +277,22 @@ def cluster_loop(G_,C_,root__,_r, seed_,pack,prioritize):  # one round; helpers
         w,c = (wcC,ccC) if typ==2 else (wcN,ccN)
         for n in md_.keys() | md.keys():  # includes removed memberships and proposals
             _m,_d,_rn = md_.get(n,(0,0,0)); m,d,rn = md.get(n,(0,0,0))
-            Up += n.c*w*(abs(m-_m)+abs(d-_d))
+            Up += n.c*w*(abs(m-_m)+abs(d-_d))  # d is redundant?
```
boris-kz, line 280: abs d is not redundant because it accumulates differently

frame_2D_alg/agg_recursion.py
boris-kz: that nsub can probably be represented by r, something like L.r-G.r?

frame_2D_alg/agg_recursion.py
```diff
@@ -277,21 +277,22 @@ def cluster_loop(G_,C_,root__,_r, seed_,pack,prioritize):  # one round; helpers
...
             Dr += ave*n.c*c*(rn-_rn)  # signed redundancy cost, in update-value units
 
+    # this call_ params update shouldn't be here?
     F.fc = oF_[F.nF].fc; call_ = list(F.call_)
     while call_:  # static AST estimates, once per executed traced call
         f = call_.pop(); F.fc += oF_[f.nF].fc; call_ += f.call_
     F.w = Up-Dr; oF_[F.nF].V_ += [F.w]
     G_ = [T for T in front_ if T.typ==3]; C_ = [T for T in front_ if T.typ==2]
-    return G_,C_,root__,gv_(Up - ave*F.fc - Dr)
+    return G_,C_,root__,gv_(Up - ave*F.fc - Dr)  # Up - ave*F.fc + Dr? Both Up and Dr should be small for convergece?
```
boris-kz, line 289: (ave*F.fc + Dr), it's a cost param

frame_2D_alg/agg_recursion.py
kwcckw: L.r doesn't represent the depth?

Lr is computed using `lr = r + (N.r + _N.r) / 2`, so the difference is actually just ` (N.r + _N.r) / 2`?

boris-kz: r should be incremented with depth

kwcckw: Then Lr could be r + 1? But G.r could be larger than Lr, since it's summed from all Fts in sum2G:

    if G.Lt or G.Bt: G.dTT,G.c,G.r = sum_vt([G.Nt,G.Lt,G.Bt]);

I guess fsub could be just using Lr?

boris-kz: Lr is computed from nodet rs. I guess we can keep separate nsub, renamed nexp

frame_2D_alg/agg_recursion.py
```diff
@@ -277,21 +277,22 @@ def cluster_loop(G_,C_,root__,_r, seed_,pack,prioritize):  # one round; helpers
         w,c = (wcC,ccC) if typ==2 else (wcN,ccN)
         for n in md_.keys() | md.keys():  # includes removed memberships and proposals
             _m,_d,_rn = md_.get(n,(0,0,0)); m,d,rn = md.get(n,(0,0,0))
-            Up += n.c*w*(abs(m-_m)+abs(d-_d))
+            Up += n.c*w*(abs(m-_m)+abs(d-_d))  # d is redundant?
             Dr += ave*n.c*c*(rn-_rn)  # signed redundancy cost, in update-value units
 
+    # this call_ params update shouldn't be here?
```
boris-kz, line 283: astra:
To estimate the **whole round’s processing cost**, including its callees.

[`[build](https://github.com/kwcckw/CogAlg/blob/27b0c91e5a932b9d2cb653f0dfaab0cba50d31fa/frame_2D_alg/meta_code.py#L153)`](https://github.com/kwcckw/CogAlg/blob/27b0c91e5a932b9d2cb653f0dfaab0cba50d31fa/frame_2D_alg/meta_code.py#L153) charges a registered function call only `3`; the callee’s body cost is excluded from the caller’s `fc`. That traversal adds each executed descendant call’s body cost.

But this is general oF accounting. It belongs in `CoF.traced`, accumulating callee costs as calls return. Putting that traversal specifically inside `cluster_loop` was ad hoc.

We can remove it from `cluster_loop` **after moving callee-cost accumulation into `CoF.traced`**. That isn’t implemented yet; simply deleting the block would omit callee costs from the loop’s process cost.

frame_2D_alg/agg_recursion.py
boris-kz, line 263: what is this seed_?

boris-kz: def cluster_(Ft, E_,_r,_c):  # fC(E): CC, else CN; returns G_,Ct,med_

    def pack(N_, nF='Nt', root=None, fseed=0, fC=0):  # summary only, no member or root-value updates
        m,d,dTT,c,r = sum_vt(N_, fm=1)
        T = (CF,(CN,CL)[fC])[fseed](nF=nF,root=root,wTT=Ft.wTT,dTT=dTT,m=m,d=d,c=c,r=r); T.N_ = N_
        if fseed:  # summary only if not fT, no member or root-value updates
            E = N_[0]; T.m,T.d = E.m,E.d  # raw vals for base_comp in cluster_C
            T.kern,T.yx,T.span,T.angl = copy(E.kern),copy(E.yx),E.span,copy(E.angl)
            T.L_,T.B_,T.m_,T.d_ = [],[],[E.m],[E.d]; T._N_ = list(set(n for L in E.rim for n in L.N_ if n is not E))
            T._r=_r+E.r; T.r=T._r; T.r_={}; T.olp=0; T.fin=0; T.i=T.id
            T.typ = 3-fC
        return T
    def prioritize(T_):  # stronger membership charges weaker; ties split the charge
        root__ = {}
        for T in T_:
            T.r_ = {}; w,c = (wcC,ccC) if T.typ==2 else (wcN,ccN)
            for n,m in zip(T.N_,T.m_):
                v = max(m,0)*w / (c*(_r+n.r)); root__.setdefault(n,[]).append((T,v))
        for n,rt_ in root__.items():
            for T,v in rt_: T.r_[n] = sum(_v/v if _v>v else .5 if _v==v else 0 for _T,_v in rt_ if _T is not T) if v>0 else 0
        for T in T_:
            T.olp = sum(n.c*T.r_[n] for n in T.N_) / sum(n.c for n in T.N_)
            T.r = T._r+T.olp

    def extend(G_,C_,root__,_r):  # one round; helpers retain cluster_ scope
        F = CoF.get()
        old_ = {T.i:(T.typ,{n:(m,d,T.r_.get(n,0)) for n,m,d in zip(T.N_,T.m_,T.d_)}) for T in G_+C_}
        for T in G_+C_: T.fin=0  # continuation is global
        G_ = cluster_N(G_,_r,root__)
        for G in G_:
            if G.d * (G.c*wcC/(G.r*ccC)) * ((len(G.N_)-1)*wL) > avd:
                for N in G.N_:
                    if N not in N_ and N.m * (G.c*wcC/(G.r*ccC)) > ave:
                        C_ += [pack([N],root=G.root,fseed=1,fC=1)]; N_.add(N)
        C_ = cluster_C(C_,_r)
        T_ = G_+C_; root__ = prioritize(T_); front_ = []
        for T in T_:
            w,c = (wcC,ccC) if T.typ==2 else (wcN,ccN)
            if T.w*T.c*w*((len(T.N_)-1)*wL) > ave*T.r*c: front_ += [T]
        if len(front_) != len(T_): root__ = prioritize(front_)
        Up,Dr = 0,0; new_ = {T.i:T for T in front_}
        for i in old_.keys() | new_.keys():
            T = new_.get(i); typ = T.typ if T is not None else old_[i][0]
            md_ = old_.get(i,(typ,{}))[1]
            md = {n:(m,d,T.r_[n]) for n,m,d in zip(T.N_,T.m_,T.d_)} if T is not None else {}
            w,c = (wcC,ccC) if typ==2 else (wcN,ccN)
            for n in md_.keys() | md.keys():  # includes removed memberships and proposals
                _m,_d,_rn = md_.get(n,(0,0,0)); m,d,rn = md.get(n,(0,0,0))
                Up += n.c*w*(abs(m-_m)+abs(d-_d))  # d is redundant?
                Dr += ave*n.c* c* (rn-_rn)  # signed redundancy cost in update-value units
        F.w = Up-Dr; oF_[F.nF].V_ += [F.w]
        G_ = [T for T in front_ if T.typ==3]; C_ = [T for T in front_ if T.typ==2]
        return G_,C_, root__, gv_(Up- (ave*F.fc +Dr))

    _C_ = {C for n in Ft.N_ for C,m,d in n.root_ if getattr(C,'source',None) is Ft}
    C_,root__, v = [],[],{},1
    N_ = Ft.N_  # all evaluated nodes, extended via front?
    G_ = [pack([E],root=Ft,fseed=1) for i,E in enumerate(sorted(E_,key=lambda E:E.id))]
    while (G_ or C_) and v:
        G_,C_,root__,v = extend(G_,C_,root__,_r)  # frontier expansion loop
    if C_:  # add eval here
        C_ = cluster_P(C_,Ft)
        Ct = pack(C_,'Ct', Ft.root); Ft.root.Ct = Ct  # Bs may form Cs in der+, as Ns
    for n in {n for C in list(_C_)+C_ for n in C.N_}:
        n.root_ = [rt for rt in n.root_ if rt[0] not in _C_]  # replace this scope only
    for C in C_:  # not reviewed
        C.root=Ct
    med_= list(set(C.N_[np.argmax(C.m_)] for C in C_))
    oG_ = []
    for T in G_:
        Ft_ = [pack(N_,nF) for N_,nF in zip((T.N_,T.L_,T.B_),('Nt','Lt','Bt'))]
        for F in Ft_:
            if Ft: F.r += _r+T.olp  # persistent fork costs, not member n.r (why if Ft?)
        G.r_=T.r_; G.olp=T.olp; oG_ += [G]
    if oG_:
        TT,C,R = sum_vt(oG_)
        lev = Copy_(Ft); lt = Copy_(Ft.Lt,root=lev); lev.Lt = lt # packed N_|L_, Lt for med_ xcomp
        Ft.H += [lev]; Ft.N_ = oG_; Ft.dTT,Ft.c,Ft.r = TT,C,R; Ft.m,Ft.d = val_(TT,ttcN,1)
    return oG_,med_

boris-kz: Astra: There are six corrections to make here.

1. **Nested `extend` does not get its own oF.** The current parser discovers top-level functions, and tracing wraps module functions. Here `CoF.get()` returns the enclosing traced caller; `F.w` and `V_` therefore update that caller. The new `gv_` also lacks its generated gate index. Keep `extend` registered at module level, or add support for nested functions. Callee-cost accumulation must reach `F.fc` **before this gate executes**. [[Current tracing and parsing](https://github.com/kwcckw/CogAlg/blob/27b0c91e5a932b9d2cb653f0dfaab0cba50d31fa/frame_2D_alg/meta_code.py#L121-L175)](https://github.com/kwcckw/CogAlg/blob/27b0c91e5a932b9d2cb653f0dfaab0cba50d31fa/frame_2D_alg/meta_code.py#L121-L175)

2. **Initialize a separate centroid-seed set.**

   ```python
   C_,root__,v,N_ = [],{},1,set()  # N_: nodes already used as centroid seeds
   G_ = [pack([E],root=Ft,fseed=1) for E in sorted(E_,key=lambda E:E.id)]
   ```

   Your assignment has four values for three variables. Also, `N_=Ft.N_` excludes existing members from seeding, and it is a list without `.add()`. This set records **seeding**, not all evaluated or frontier nodes.

3. **`Copy_` loses `_r`.** In `cluster_N`, boundary collection reads `G._r` before the summary refresh assigns it. Preserve it:

   ```python
   G = [Copy_](https://github.com/kwcckw/CogAlg/blob/27b0c91e5a932b9d2cb653f0dfaab0cba50d31fa/frame_2D_alg/agg_recursion.py#L617-L637)(_G); G.N_=list(_G.N_); G.L_=list(_G.L_); G._r=_G._r
   ```

   This follows from the current [`Copy_`](https://github.com/kwcckw/CogAlg/blob/27b0c91e5a932b9d2cb653f0dfaab0cba50d31fa/frame_2D_alg/agg_recursion.py#L617-L637); a targeted expansion reproduced the missing-attribute error.

4. **For membership improvement, omit `Δd` and make `Up` signed.** `d` is not mathematically redundant with `m`, but membership priority and selection here use match. Changes in difference should not automatically count as membership improvement.

   ```python
   Up += n.c*w*(m-_m)
   Dr += ave*n.c*c*(rn-_rn)
   ```

   Two qualifications: seed scores must use the same definition as subsequent membership scores—currently raw `E.m` becomes internal-link support for Gs—and negative `Up` must return the previous proposals if you want the rollback described in your [[updated discussion](https://github.com/boris-kz/CogAlg/pull/271#discussion_r4042939835)](https://github.com/boris-kz/CogAlg/pull/271#discussion_r4042939835). The current `old_` stores scores only; it cannot restore the previous proposals.

5. **`if Ft` should be `if F`.**

   ```python
   for F in Ft_:
       if F: F.r += _r+T.olp
   ```

   This charges each populated output fork. Testing the input `Ft` charges empty output forks too. Separately, `pack([])` needs an empty-summary case: `sum_vt([],fm=1)` divides by zero redundancy.

6. **The final centroid bookkeeping still conflicts with `cluster_P`.** Restore `C.source=Ft`, since `_C_` depends on it. But also fix the complete mutation sequence: `cluster_P` clears members’ `root_`, and its `sum2F` calls already append output centroid memberships. Your subsequent append therefore duplicates those entries, while unrelated memberships have already been erased. Preserve other scopes and install each new membership once. If no centroids survive, replace the previous `Ct` with an empty fork as well. [[Current mutation path](https://github.com/kwcckw/CogAlg/blob/27b0c91e5a932b9d2cb653f0dfaab0cba50d31fa/frame_2D_alg/agg_recursion.py#L408-L516)](https://github.com/kwcckw/CogAlg/blob/27b0c91e5a932b9d2cb653f0dfaab0cba50d31fa/frame_2D_alg/agg_recursion.py#L408-L516)

frame_2D_alg/agg_recursion.py
kwcckw: Alright, so it should be per L, L.nexp, and i think we can parse it via L_[0].nexp+1 to proj_L _ here so that it's packed in pL_, which it will init the new nexp with the new L in the next depth.

frame_2D_alg/agg_recursion.py
kwcckw: > But this is general oF accounting. It belongs in CoF.traced, accumulating callee costs as calls return

Yes, I think it should be in traced, right after the function calling process.

frame_2D_alg/agg_recursion.py
kwcckw: > what is this seed_?

It's the seed for cluster_C, and we need a other var since C_ pack new instance with pack function.

frame_2D_alg/agg_recursion.py
boris-kz: ok, that's another return

frame_2D_alg/agg_recursion.py
boris-kz: that's `_C_`?

kwcckw: Those seed_ are in G.N_ , so it won't be in the same level as C_ or G_, and we won't have _C_ here?

Also most of the astra comments above are not relevant or are already discussed. I'm travelling today and will check again the rest much later.

frame_2D_alg/agg_recursion.py
```diff
@@ -601,12 +602,13 @@ def comb_Ft(Nt, Lt, Bt, Ct, root,wTT):  # assemble core and boundary with separa
 
     G = CN(Nt=Nt,Lt=Lt,Bt=Bt,Ct=Ct,root=root,wTT=wTT)
     for F in Nt,Lt,Bt,Ct: F.root=G
-    r = sum_vt([Nt,Nt.Lt,Bt,Bt.Lt])[2]
+    add2F(G,G.Nt,merge=2)  # sum base params?
```
boris-kz, line 605: this is copy, but we need sum_F_ anyway

frame_2D_alg/agg_recursion.py
boris-kz: ok, I pushed a bunch of edits

frame_2D_alg/agg_recursion.py
kwcckw: But that's conditional? 

   if G.Lt or G.Bt: G.dTT,G.c,G.r = sum_vt([G.Nt,G.Lt,G.Bt]); G.m,G.d = val_(G.dTT,G.wTT,fd=1)

PR #273, opened 2026-09-19
9/19/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -289,33 +289,34 @@ def prioritize(T_):  # stronger membership charges weaker; ties split the charge
...
         else:     _G_,_C_,_root__ = G_,C_,root__
     if C_:  # add eval here
-        C_ = cluster_P(C_,Ft)
+        C_ = cluster_P(C_,Ft.root)
         Ct = pack(C_,'Ct', Ft.root); Ft.root.Ct = Ct  # Bs may form Cs in der+, as Ns
     for n in {n for C in list(_C_)+C_ for n in C.N_}:
         n.root_ = [rt for rt in n.root_ if rt[0] not in _C_]  # replace this scope only
     for C in C_:  # not reviewed
-        C.root=Ct
+        C.root=C
```
kwcckw, line 297: Sorry, this is a typo, it should be Ct.

frame_2D_alg/agg_recursion.py
```diff
@@ -289,33 +289,34 @@ def prioritize(T_):  # stronger membership charges weaker; ties split the charge
...
+            if F: F.r += _r+T.olp  # persistent fork costs, not member n.r (why if Ft?)
         c_ = list(set(C for n in T.N_ for C in n.C_))  # lower-level Ct only
         G = comb_Ft(*Ft_,pack(c_,'Ct'),Ft,wTT=ttcN)
         G.r_=T.r_; G.olp=T.olp; oG_ += [G]
     if oG_:
         TT,C,R = sum_vt(oG_)
-        lev = Copy_(Ft); lt = Copy_(Ft.Lt,root=lev); lev.Lt = lt # packed N_|L_, Lt for med_ xcomp
+        lev = Copy_(Ft); lt = Copy_(Ft.root.Lt,root=lev); lev.Lt = lt # packed N_|L_, Lt for med_ xcomp
         Ft.H += [lev]; Ft.N_ = oG_; Ft.dTT,Ft.c,Ft.r = TT,C,R; Ft.m,Ft.d = val_(TT,ttcN,1)
+        for C in C_: C.source = Ft  # for _C_ in cluster_
```
kwcckw, line 312: So i think we should assign source here, so that we can retrieve it as prior `_C_` above?

     _C_ = list({C for n in Ft.N_ for C, m, d in n.root_ if getattr(C, 'source', None) is Ft})

frame_2D_alg/agg_recursion.py
```diff
@@ -560,16 +561,16 @@ def sum2G(ft_, fTT, root=None, init=1, nexp=1):  # finalize cluster
     C_= [c for N in Ft_[0].N_ for c in N.C_]  # splice centroids
     Ft_ += [sum2F(list(set(C_)), root.Ct if root else None ,nF='Ct') if C_ else CF(nF='Ct')]  # add multiple root_ in Cs?
     G = comb_Ft(*Ft_, root, wTT=fTT)
-    N_ = G.N_; N=N_[0]; r=G.r; Av=ave+avd
+    N_ = G.N_; N=N_[0]; r=G.r; Av=ave+avd; nexp = L_[0].nexp if (L_:= (G.Lt.N_ + G.Bt.N_)) else 1
```
kwcckw, line 564: I moved nexp to pL via proj_L_, so we nexp is retrieved from L now.

frame_2D_alg/agg_recursion.py
```diff
@@ -97,10 +97,10 @@ def cross_comp(root,pL_,r,nF='Nt',dF=None,rL=None, fall=1, nexp=1):
         for N in (N_:= list(set(N_))):
             N.Rt = sum2F(N.rim,root=N,nF='Rt') if N.rim else CF(root=N,nF='Rt')
         if gv_(val_(tt,ttcN) * (c*wcN/(r*ccN)) * ((len(L_)-1)*wL) - ave):
-            G_,med_ = cluster_(root, get_exemplars(N_,r,c),r,c)
+            G_,med_ = cluster_(root.Nt, get_exemplars(N_,r,c),r,c)
             for fsub,n_,R in ((1,med_,root.H[0]), (0,G_,root)):  # sub+ xcomp new G_, G_'N_| root H[-1] only?
                 if n_ and gv_(R.m*R.c*wX - ave*(R.r+1+cX)):  # mdecay(L_)-decay? eval len n_?
-                    g_ = cross_comp(R, proj_L_(combinations(n_,2),R,r, nexp), r,R.nF)
+                    g_ = cross_comp(R, proj_L_(combinations(n_,2),R,r, nexp+1), r,R.nF)  # increase nexp in agg+ too?
```
boris-kz, line 103: agg+ nexp should be 1, it's the base links

frame_2D_alg/agg_recursion.py
```diff
@@ -83,10 +83,10 @@ def cent_TT(dTT, r):  # EM-like weight attr matches | diffs by their match to th
 - forward: selective extend cross-comp, clustering across tiles, re-order centroids by eigenvalues
 - feedback filter updates 
 '''
-def cross_comp(root,pL_,r,nF='Nt',dF=None,rL=None, fall=1, nexp=1):
+def cross_comp(root,pL_,r,nF='Nt',dF=None,rL=None, fall=1):
 
     L_,N_ = [],[]
-    for dist, dy_dx, _N,N, lc,lr, pTT,m,_ in pL_:
+    for dist, dy_dx, _N,N, lc,lr, pTT,m,_,nexp in pL_:
```
boris-kz, line 89: we don't need nexp here?

frame_2D_alg/agg_recursion.py
```diff
@@ -97,10 +97,10 @@ def cross_comp(root,pL_,r,nF='Nt',dF=None,rL=None, fall=1, nexp=1):
         for N in (N_:= list(set(N_))):
             N.Rt = sum2F(N.rim,root=N,nF='Rt') if N.rim else CF(root=N,nF='Rt')
         if gv_(val_(tt,ttcN) * (c*wcN/(r*ccN)) * ((len(L_)-1)*wL) - ave):
-            G_,med_ = cluster_(root, get_exemplars(N_,r,c),r,c)
+            G_,med_ = cluster_(root.Nt, get_exemplars(N_,r,c),r,c)
```
boris-kz, line 100: that Nt | Bt is the root here?

frame_2D_alg/agg_recursion.py
kwcckw: Yes, this is no longer needed since agg+ nexp is 1. Originally it's intended to parse into the agg+ below.

frame_2D_alg/agg_recursion.py
kwcckw: We still need to parse root.Nt because Bt is the converted root?  We should convert Bt via F2N before the B fork, so in B fork, the root here is Bt, and we pack Bt's Nt, Lt, Ct or Bt, as in the prior discussion?

boris-kz: but then the root is always Nt, so we don't need to specify it?

kwcckw: Yes, so even for Bt fork, It's actually parsing Bt.Nt.N_, which is those B_. We probably can parse root and get the Ft within cluster_.

frame_2D_alg/agg_recursion.py
```diff
@@ -266,7 +266,7 @@ def prioritize(T_):  # stronger membership charges weaker; ties split the charge
         root__ = {}
         for T in T_:
             T.r_ = {}; w,c = (wcC,ccC) if T.typ==2 else (wcN,ccN)
-            for n,m in zip(T.N_,T.m_):
+            for n,m in zip(T.N_,T.m_):  # T is G_ + C_, but both sum m differently? cluster_N uses L that comp from comp_N, while cluster_C uses base_comp only
```
boris-kz, line 269: yes, this redundancy between forks is not correct, not sure how to define it

frame_2D_alg/agg_recursion.py
boris-kz, line 288: I think we need to separate extend for cluster_N and cluster_C, they don't interact

astra:
`extend` currently imposes two couplings:

* **Lines 314–318:** Gs seed Cs. This is a one-way dependency; move it into `cluster_` between the two calls.
* **Lines 323–340:** joint `prioritize(G_+C_)` and `upV`. Centroid memberships increase `G.r` and enter `root__`, affecting G survival and next-round node inclusion. The shared `upV` also lets either process prolong the other.

So `cluster_` should:

1. Run `cluster_N` to completion.
2. Seed Cs from the accepted Gs, retaining existing Cs.
3. Run `cluster_C` to completion, with `cluster_P` selection belonging to that stage.
4. Pack Gs, Ct, and medoids.

Keep overlap accounting within each representation. CN link support and CC centroid match describe different relationships; their sharing a node does not itself make them redundant. Each loop then evaluates its own updates against its own processing cost.

boris-kz: so this is back to separate cluster_N and cluster_C, unpacking cluster_.
but Gs can probably be terminated early if their internal variance exceeds frontier?

kwcckw: With separated clustering functions, do we back to the prior structure where only one fork per level? 

Or we still run them in sequence where we reselect medoid from G.N_, perform centroid clustering to form C, and then merge the resulting G_ and C_?

boris-kz: that medoid clustering part should be in cluster_C, after C_ is complete. 
so the next agg+ will be on combined G_

kwcckw: Okay basically the process is still the same, it's just we don't have refinement on the cluster_N now, and we still need that agg+ based on medoids?

boris-kz: yes, I think we can have nested calls:
cluster_N -> sum2G,rng+ -> batched cluster_C -> cluster_P -> cluster_N(med_) -> G_+=g_ 
-> cross_comp( combined G_)   

Yes. `cluster_N` becomes the recursive driver; `cluster_` and `extend` are unnecessary.

* `sum2G` packs accepted CNs, with internal `rng+` evaluated there.
* `cluster_C` processes the resulting batch; `cluster_P` follows if overlap justifies it.
* Their `med_` seed another connectivity pass, producing **same-level** `g_`.
* Append `g_` before cross-comparing the combined Gs.

The return boundary matters: `cluster_N(med_)` should return those same-level Gs before performing their higher aggregation, so the parent can cross-compare the combined set.

This is a nested handoff between clustering stages; each stage retains its own continuation test.

kwcckw: > sum2G,rng+ -> batched cluster_C -

This batched cluster_C is per G? Or we should merge all selected G.N_ for the cluster_C process?

>  cluster_P -> cluster_N(med_)

This will be an individual clustering process now? Or we still need to call cross_comp with the medoids?

boris-kz: merge all selected G.N_ for the cluster_C, then cluster_N(med_), then combined_G_ cross_comp

kwcckw: > merge all selected G.N_ for the cluster_C, then cluster_N(med_), then combined_G_ cross_comp

Alright, so this is much cleaner without the cross_comp of meda now.

PR #274, opened 2026-09-20
9/20/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -86,17 +86,18 @@ def cent_TT(dTT, r):  # EM-like weight attr matches | diffs by their match to th
 def cross_comp(root, pL_,r, dF=None,rL=None, fall=1):
 
     L_,N_ = [],[]
-    for dist, dy_dx, _N,N, lc,lr, pTT,m,_ in pL_:
+    for dist, dy_dx, _N,N, lc,lr, pTT,m,_,nexp in pL_:
         if _N != N and (fall or (m>0 and gv_(m*(lc*wN/(lr*cN)) - ave* (r+cN)))):
             Link = comp_N(_N,N, lr,lc, full=not dF, A=dy_dx, span=dist, rL=root)
-            Link.rTT = np.abs(pTT-Link.dTT) / eps_(Link.dTT); L_+=[Link]; N_+=[_N,N]  # relative prediction error/oF
+            Link.nexp=nexp; Link.rTT = np.abs(pTT-Link.dTT) / eps_(Link.dTT); L_+=[Link]; N_+=[_N,N]  # relative prediction error/oF
```
kwcckw, line 92: Sorry, we still need nexp to assign them in L, so that we can use this L.nexp in sum2G later.

frame_2D_alg/agg_recursion.py
```diff
@@ -304,13 +305,23 @@ def trans_cluster(G):
...
+                C += Ct.C; R += Ct.r  # not sure
+                if g_:= cluster_N(root, med_, Ct.r,Ct.c): G_ += g_;  # increment R and C frpm g_ here?
         if gv_(M* C* wX - ave * (R+1+cX)):  # mdecay(L_)-decay? eval len n_?
             G_ = cross_comp(root, proj_L_(combinations(G_,2),root,R), R,root.nF)  # agg+
     return G_
 
 def cluster_C(Ft, E_,_r,_c):  # form centroids by clustering exemplar surround via rims of new member nodes, within root
 
-    N_= copy(Ft.N_); _C_=[]  # revert if 0 clusters?
+    N_= copy(Ft.N_); _C_=[]  # revert if 0 clusters? (current cluster_C's scope is limited to selected G's N_ only?)
```
kwcckw, line 324: Right now cluster_C uses selected G's selected N_ as exemplar, but for the scope, it may extend to those non-selected G and G.N_?

frame_2D_alg/agg_recursion.py
```diff
@@ -86,17 +86,18 @@ def cent_TT(dTT, r):  # EM-like weight attr matches | diffs by their match to th
...
-            Link.rTT = np.abs(pTT-Link.dTT) / eps_(Link.dTT); L_+=[Link]; N_+=[_N,N]  # relative prediction error/oF
+            Link.nexp=nexp; Link.rTT = np.abs(pTT-Link.dTT) / eps_(Link.dTT); L_+=[Link]; N_+=[_N,N]  # relative prediction error/oF
     if L_:
         Lt = root.Lt = sum2F(L_,root, nF='Lt'); m,d,tt,c,r = Lt.m,Lt.d,Lt.dTT,Lt.c,Lt.r; FV_(CoF.get(),tt,c,r)
         if dF:  # comp_F: no cluster_,recursion
             add2F(dF,CF(N_=L_,m=m,d=d,dTT=tt,c=c,r=r),merge=1); add2F(rL,dF,merge=2); return
         for N in (N_:= list(set(N_))):
             N.Rt = sum2F(N.rim,root=N,nF='Rt') if N.rim else CF(root=N,nF='Rt')
         if gv_(val_(tt,ttcN) * (c*wcN/(r*ccN)) * ((len(L_)-1)*wL) - ave):
+            for N in N_: N.fin = 0  # reset from sub+
```
boris-kz, line 100: this reset should be in sub+

frame_2D_alg/agg_recursion.py
```diff
@@ -304,13 +305,23 @@ def trans_cluster(G):
             root.N_ = G_; root.dTT=TT; root.c=C; r=root.r=R; root.m=M, root.d=D
         FV_(CoF.get(), *sum_vt(G_)[:-1],R)
         # add cluster_C, cluster_P, cluster_N(med_)
+        for G in G_:
+            N_ = []
+            if G.d * (G.c*wcC/(G.r*ccC)) * ((len(G.N_)-1)*wL) > avd:  # high variance
+                N_ += [N for G in G_ for n in G.N_ if N.m * (G.c*wcC/(G.r*ccC)) > ave]  # or use get_exemplar?
+            if Ct:= cluster_C(root, N_, R,C):
+                med_ = [C.N_[int(np.argmax(C.m_))] for C in Ct.N_]
```
boris-kz, line 313: this should be packed in cluster_C

kwcckw: Okay right thanks, that's including the reset of fin and exe assignment.

frame_2D_alg/agg_recursion.py
boris-kz: I think primary process is global cluster_C(G_), which is creating med_ for agg+ cluster_N
sub+ should only extend xcomp rng and der, which add internal nesting 
but rng+ links should extend beyond G and be used for G merging?

boris-kz: astra: 
The added block still mixes **global centroid selection** with **subdivision inside Gs**.

Your [[latest comment](https://github.com/boris-kz/CogAlg/pull/274#discussion_r4056684263)](https://github.com/boris-kz/CogAlg/pull/274#discussion_r4056684263) needs one distinction: if `cluster_C(G_)` clusters **Gs themselves**, its medoids are Gs, and `cluster_N(med_)` produces the next composition level. My earlier endorsement of `G_ += g_` assumed centroids over **G members**, producing additional same-level Gs.

The substantive issues in the edits:

1. **Batch `cluster_C` outside `for G in G_`.** Each call resets membership state, so overlapping calls can erase earlier centroid memberships. Meanwhile `G_ += g_` extends the very list being iterated, sending appended graphs through the same block. Collect the seeds first, then perform one centroid pass. [[L308–317](https://github.com/kwcckw/CogAlg/blob/5081374d2a33c4f3bb2070993e134b3c042239ee/frame_2D_alg/agg_recursion.py#L308-L317)](https://github.com/kwcckw/CogAlg/blob/5081374d2a33c4f3bb2070993e134b3c042239ee/frame_2D_alg/agg_recursion.py#L308-L317)

2. **Reset the actual clustering scope, not only the original seeds.** Centroid output calls `comb_Ft → add_Nt`, which sets every member’s `fin=1`. A selected medoid can be an absorbed frontier node absent from the original `N_`; line 314 leaves it finished, and `cluster_N` immediately skips it. Moving medoid preparation into `cluster_C`, as discussed, should include this reset. [[L313–317](https://github.com/kwcckw/CogAlg/blob/5081374d2a33c4f3bb2070993e134b3c042239ee/frame_2D_alg/agg_recursion.py#L313-L317)](https://github.com/kwcckw/CogAlg/blob/5081374d2a33c4f3bb2070993e134b3c042239ee/frame_2D_alg/agg_recursion.py#L313-L317)

3. **The recursive return still violates the proposed same-level append.** `cluster_N(med_)` performs its own final `cross_comp` and returns that result. If retaining `G_ += g_`, that recursive branch must return before higher aggregation. Also, line 319 still passes `root.nF` into `dF`: this selects derivative-only comparison, then passes a string to `add2F`. Remove that argument. [[L317–320](https://github.com/kwcckw/CogAlg/blob/5081374d2a33c4f3bb2070993e134b3c042239ee/frame_2D_alg/agg_recursion.py#L317-L320)](https://github.com/kwcckw/CogAlg/blob/5081374d2a33c4f3bb2070993e134b3c042239ee/frame_2D_alg/agg_recursion.py#L317-L320)

4. **Restoring `Link.nexp` does not yet restore range expansion.** `proj_L_` takes `max` fourth; the calls pass `nexp+1` and `nexp` there. Consequently, emitted links retain `nexp=1`, and subsequent `rng+` keeps using radius 2. Pass `nexp=nexp+1` and `nexp=nexp` explicitly. [[L538–546](https://github.com/kwcckw/CogAlg/blob/5081374d2a33c4f3bb2070993e134b3c042239ee/frame_2D_alg/agg_recursion.py#L538-L546)](https://github.com/kwcckw/CogAlg/blob/5081374d2a33c4f3bb2070993e134b3c042239ee/frame_2D_alg/agg_recursion.py#L538-L546)

On **rng+ crossing G boundaries**: currently `combinations(G.N_,2)` cannot generate an external endpoint, regardless of range. Existing rims can reach outside, but that is different from creating new cross-boundary comparisons. Those comparisons can support merging; their results should be reconciled at the shared parent, where both G memberships and summaries can be updated.

Mechanical leftovers: line 311 selects the old `N` while iterating `n`; `Ct.C` should not be used—and merely changing it to `Ct.c` would count an overlapping centroid representation as additional evidence in the aggregation gate.

kwcckw: > I think primary process is global cluster_C(G_), which is creating med_ for agg+ cluster_N

If we cluster G_ with centroid clustering, the medoids C.N_ are actually Gs, so it's basically just selecting the exemplars from G_ to cluster_N? 

> sub+ should only extend xcomp rng and der, which add internal nesting

For sub+, if it's just extending xcomp rng and der, we parse the same set of N_, but the cluster_C should be applied to the combined G.N_ instead?

> but rng+ links should extend beyond G and be used for G merging?

That's probably similar to trans cluster, where we probably can merge their root (G) via the N's rim.

boris-kz: Sub+ can add nesting in both Nt and Bt, regardless of rng+. rng+ can merge Gs, and sub+ can lead to unpacking existing Gs, before next agg+.
That medoid selection can seed get_exemplars, so sparsity is increased both spatially and associatively?

boris-kz: Astra: associative compression followed by local redundancy suppression:

* cluster_C selects medoids representing shared attributes, potentially across distant Gs.
* get_exemplars(med_) then selects among those representatives using their lateral support and overlapping rims.

That increases sparsity in both senses, although get_exemplars is link-overlap-based, not purely spatial. Nearby but associatively distinct medoids should not be suppressed merely because they are nearby.

Your clarification⁠￼ also separates structural revision from aggregation: sub+ can refine Nt/Bt and unpack Gs; rng+ can merge Gs; agg+ operates on the resulting active representation.

Consequently, medoid selection should use that revised population, after unpacking and merging. Otherwise a selected medoid may represent a G that no longer exists as an active unit. Unpacked children replace their parent in that population; merged Gs replace their constituents. This is not unconditional G_ += g_.

kwcckw: > That increases sparsity in both senses, although get_exemplars is link-overlap-based, not purely spatial.

Wouldn't that be partly redundant since cluster_C uses links in extending the frontier, and we feed the medoids to get_exemplars again?

boris-kz: the process maybe partly redundant, but the results are not

kwcckw: Okay, so for `I think primary process is global cluster_C(G_), which is creating med_ for agg+ cluster_N`, the agg+ comp and cluster only between med, which is mainly for noise and redundancy reduction?

boris-kz: med_ is the seeds, agg+ proceeds through their rims

boris-kz: Yes, but sub+ is also needed to refine base Gs

frame_2D_alg/agg_recursion.py
```diff
@@ -528,12 +539,11 @@ def sum2G(ft_, fTT, root=None, init=1):  # finalize cluster
     if Lt:= G.Lt:  # rng+'sub+
         lm,lc,lr = Lt.m,Lt.c,Lt.r  # no levR = 1/len(L_): represented by c
         if gv_(lm*lc*wX - Av* (lr+1+cX)):  # mdecay(L_)-decay?
-            # or pack nexp into L or pL since L is always per level?
-            cross_comp(G, proj_L_(combinations(N_,2),G,lr,nexp+1), lr+1,'Nt')  # replace G Nt,Lt?
+            cross_comp(G, proj_L_(combinations(N_,2),G,lr,nexp+1), lr+1)  # replace G Nt,Lt?
```
boris-kz, line 542: so here we need to exclude old Ls from comp_N, but include them in cluster_N if >new threshold

frame_2D_alg/agg_recursion.py
kwcckw: > med_ for agg+ cluster_N

Okay thanks. So the agg+ here is referring to the cluster_N only right? So if there's no new comp, it's actually cluster new nodes into the medoid via their existing rim, while the refinement is done via the sub+ in sum2G after the medoid form G.

frame_2D_alg/agg_recursion.py
kwcckw: The old Ls is referring to the same pair of N's link?

I guess we can identify that from their nexp? So new links should have the matching nexp with the current level nexp, while old links' nexp will be lower than that.

frame_2D_alg/agg_recursion.py
boris-kz: cluster_N is always the first step, agg+ is the full cycle

frame_2D_alg/agg_recursion.py
boris-kz: this is all in sub+, and nexp is for comp_N_, it doesn't necessarily map to clustering levels

frame_2D_alg/agg_recursion.py
kwcckw: Sorry, i misunderstood earlier.

The current agg+ in cluster_N:

        if gv_(M* C* wX - ave * (R+1+cX)):  # mdecay(L_)-decay? eval len n_?
            G_ = cross_comp(root, proj_L_(combinations(G_,2),root,R), R,root.nF)  # agg+
            
The G_ in `combinations(G_,2)` should be the medoids from cluster_C right? In the new agg+, they may comp with each other to form new links, but clustering may reuse the prior agg+ links. 

> this is all in sub+, and nexp is for comp_N_, it doesn't necessarily map to clustering levels

Yes, i guess we just need to check if `_N` in `N.rim` before comp them to skip old links, and clustering with all rims to include the new and old links.

frame_2D_alg/agg_recursion.py
boris-kz, line 364: so it's something like this:

    if oC_:
        _m,_d,_tt,_c,_r = sum_vt(oC_, fm=1)
        FV_(CoF.get(),_tt,_c,_r)  # r per deeper cross_comp?
        if gv_((_m *_c *wcC) / (_r+ccC) * ((len(oC_)-1)*wL) - ave):
            sum2F(oC_,root, nF='Ct')  # may need some additions
            for N in N_: N.fin, N.exe = 0, 0
            for n in [N for C in oC_ for N in C.N_]:  # exemplar V + sum n match_dev to Cs, m* ||C rvals
                n.exe = (n.d if n.typ==1 else n.m) + sum(rt[1] for rt in n.root_) > ave
                # or exemplars in med_?:
            med_ = [o.N_[int(np.argmax(o.m_))] for o in oC_]
            if gv_((_m *_c *wX) / (_r+cX) * ((len(med_)-1)*wL) - ave):
                cross_comp(root, proj_L_(combinations(med_,2),root,_r), _r)  
                # selective agg+, no return?

kwcckw: That oC_ is Ct.N_ from cluster_C?

So this `sum2F(oC_,root, nF='Ct')  # may need some additions` shouldn't be needed since this will be done in cluster_C?

boris-kz: This is all in cluster_C

kwcckw: If the cross_comp in cluster_C, the agg+ in cluster_N is no longer needed?

boris-kz: Not directly

PR #275, opened 2026-09-21
9/21/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -357,14 +357,16 @@ def cluster_C(root, E_,_r,_c):  # form centroids by clustering exemplar surround
...
+            if root.Ct: Ct.H += [root.Ct]
+            root.Ct = Ct; Ct.root = root
             for N in N_: N.fin, N.exe = 0, 0
             for n in [N for C in oC_ for N in C.N_]:  # exemplar V + sum n match_dev to Cs, m* ||C rvals
                 n.exe = (n.d if n.typ==1 else n.m) + sum(rt[1] for rt in n.root_) > ave
                 # or exemplars in med_?:
             med_ = [o.N_[int(np.argmax(o.m_))] for o in oC_]
             if gv_((_m *_c *wX) / (_r+cX) * ((len(med_)-1)*wL) - ave):
-                cross_comp(root, proj_L_(combinations(med_,2),root,_r), _r)
+                cross_comp(F2N(Ct), proj_L_(combinations(med_,2),root,_r), _r)  # parse Ct as root, same as B fork?
```
kwcckw, line 369: So for medoids, the root should be converted Ct, as in the B fork? Or it should stick with prior scheme where it should be root.H[0]?

frame_2D_alg/agg_recursion.py
```diff
@@ -357,14 +357,16 @@ def cluster_C(root, E_,_r,_c):  # form centroids by clustering exemplar surround
         _m,_d,_tt,_c,_r = sum_vt(oC_, fm=1)
         FV_(CoF.get(),_tt,_c,_r)  # r per deeper cross_comp?
         if gv_((_m *_c *wcC) / (_r+ccC) * ((len(oC_)-1)*wL) - ave):
-            sum2F(oC_,root, nF='Ct')  # may need some additions
+            Ct = sum2F(oC_,root, nF='Ct')  # may need some additions
+            if root.Ct: Ct.H += [root.Ct]
```
boris-kz, line 361: this should be added to sum2F, we can't keep adding the clutter on top

frame_2D_alg/agg_recursion.py
boris-kz: it's medoids, in root.N_, not CT. root.H[0] is for sub+, in sum2G only

kwcckw: Okay, so cross_comp of medoids should be using root here, but we have another agg+ of G_ after centroid clustering, which may add new level, while medoid Gs should be the same level as G_, so how do we ensure their H are packing the correct level when we have 2 separate cross_comp but using the same root?

frame_2D_alg/agg_recursion.py
kwcckw: Alright, we should add the level the same way in cluster_N when we have new top level from G_.

frame_2D_alg/agg_recursion.py
```diff
@@ -262,6 +261,7 @@ def trans_cluster(G):
                 L.Nt,L.Bt, L.Ct = CF(),CF(),CF()
             # merge roots
     G_, Gt_, in_ = [],[],set()  # root attrs, add prelink pL_,pN_? include merged Cs, in feature space for Cs
+    for N in root.N_: N.fin = 0
```
boris-kz, line 264: the reset should be after each run, not before

frame_2D_alg/agg_recursion.py
```diff
@@ -302,12 +302,12 @@ def trans_cluster(G):
         if gv_(M * (C*wcN /(_r+R+ccN)) * ((len(G_)-1)*wL) - ave):  # reform root,Nt, no other forks yet:
             if root.nF == 'Nt' or root.nF == 'Ct':  # concat C_ if Ct is hierarchical, same as Nt it maps to?  not revised
                 root.H += [Copy_(root)]  # add H for Nt, also Ct?
-            root.N_ = G_; root.dTT=TT; root.c=C; root.r=R; root.m=M, root.d=D
+            root.N_ = G_; root.dTT=TT; root.c=C; root.r=R; root.m=M; root.d=D
+            if D * (C*wcC / R*ccC) * ((len(G_)-1)*wL) > avd:  # high variance (this must be after reform, so that root.N_ packs all G_?)
```
boris-kz, line 306: right, and cross_comp should be removed

frame_2D_alg/agg_recursion.py
boris-kz: every pack in sum2G -> add_Nt adds a new level:  

    for N in N_:
        G.root_ += [rt for rt in N.root_]  # Ct || Nt  (use rt for consistency, to be used in get_exemplars:  rc = sum(r[0].c for r in n.root_))
        G.kern += N.kern*w; yx = N.yx; yx_+=[yx]  # * w?
        add_H(G.H, N.H, G); add_H(G.Ct.H, N.Ct.H, G.Ct)
    if (n_:=[n for N in N_ for n in N.N_]): G.H += [sum2F(n_, G)]  # new top lev  (n_ shouldn't be summed back to G here?)

frame_2D_alg/agg_recursion.py
```diff
@@ -518,12 +520,12 @@ def sum2G(ft_, fTT, root=None, init=1):  # finalize cluster
...
         lm,lc,lr = Lt.m,Lt.c,Lt.r  # no levR = 1/len(L_): represented by c
         if gv_(lm*lc*wX - Av* (lr+1+cX)):  # mdecay(L_)-decay?
-            cross_comp(G, proj_L_(combinations(N_,2),G,lr,nexp+1), lr+1)  # exclude old Ls from comp_N, include in cluster_N?
+            cross_comp(G, proj_L_(combinations(N_,2),G,lr,nexp=nexp+1), lr+1)  # exclude old Ls from comp_N, include in cluster_N?
             # unpack Gs if V < r+1: redundancy to stronger sub G_|C_? or partial unpack: exclude stronger sub G_|C_?
     if Bt:= G.Bt:  # der+'sub+
         bd,bc,br = Bt.d,Bt.c,Bt.r; rroot = root.root if root.root else 0
         if N.typ!=1 and gv_(bd*bc*wX - Av*(br+1+cX)):  # no ddfork, eval len?
-            cross_comp(F2N(G.Bt), proj_L_(combinations([F2N(L) for L in Bt.N_],2),G,br,nexp),br)
+            cross_comp(F2N(G.Bt), proj_L_(combinations([F2N(L) for L in Bt.N_],2),G,br,nexp=nexp),br)
```
boris-kz, line 528: nexp is 1 here

frame_2D_alg/agg_recursion.py
```diff
@@ -518,12 +520,12 @@ def sum2G(ft_, fTT, root=None, init=1):  # finalize cluster
     if Lt:= G.Lt:  # rng+'sub+
         lm,lc,lr = Lt.m,Lt.c,Lt.r  # no levR = 1/len(L_): represented by c
         if gv_(lm*lc*wX - Av* (lr+1+cX)):  # mdecay(L_)-decay?
-            cross_comp(G, proj_L_(combinations(N_,2),G,lr,nexp+1), lr+1)  # exclude old Ls from comp_N, include in cluster_N?
+            cross_comp(G, proj_L_(combinations(N_,2),G,lr,nexp=nexp+1), lr+1)  # exclude old Ls from comp_N, include in cluster_N?
```
boris-kz, line 523: here we need to cluster_N with old links too, regardless of xcomp

frame_2D_alg/agg_recursion.py
```diff
@@ -357,14 +357,16 @@ def cluster_C(root, E_,_r,_c):  # form centroids by clustering exemplar surround
...
+            Ct = sum2F(oC_,root, nF='Ct')  # may need some additions
+            if root.Ct: Ct.H += [root.Ct]
+            root.Ct = Ct; Ct.root = root
             for N in N_: N.fin, N.exe = 0, 0
             for n in [N for C in oC_ for N in C.N_]:  # exemplar V + sum n match_dev to Cs, m* ||C rvals
                 n.exe = (n.d if n.typ==1 else n.m) + sum(rt[1] for rt in n.root_) > ave
                 # or exemplars in med_?:
             med_ = [o.N_[int(np.argmax(o.m_))] for o in oC_]
             if gv_((_m *_c *wX) / (_r+cX) * ((len(med_)-1)*wL) - ave):
-                cross_comp(root, proj_L_(combinations(med_,2),root,_r), _r)
```
boris-kz, line 367: get exemplars(med_)?

frame_2D_alg/agg_recursion.py
kwcckw: But this is adding new level to newly formed G, but not the existing root.H? We may need to add new root.H level in both sub+ or agg+?

frame_2D_alg/agg_recursion.py
kwcckw: I placed it here because we only need it in cluster_N, if we should reset it after each run, that actually will be right before every call of cross_comp.

frame_2D_alg/agg_recursion.py
kwcckw: Do you mean the cross_comp after the cluster_C below should be removed?

        if gv_(M*C* wX - ave * (R+1+cX)):  # M,C,R not affected by clustering? mdecay(L_)-decay? eval len G_?
            G_ = cross_comp(root, proj_L_(combinations(G_,2),root,R), R)  # agg+

frame_2D_alg/agg_recursion.py
kwcckw: Yes thanks, i overlooked this part.

frame_2D_alg/agg_recursion.py
kwcckw: G.N_ should be already include old links' nt because G.N_ should check and cluster all old+new links' nt?

frame_2D_alg/agg_recursion.py
kwcckw: Yes, and I guess we need to add eval to make sure we have at least 2 exemplars for the proj_L_ here?

frame_2D_alg/agg_recursion.py
boris-kz: yes, but xcomp has a different eval

frame_2D_alg/agg_recursion.py
boris-kz: it should be a termination sequence in all clustering

frame_2D_alg/agg_recursion.py
boris-kz: yes, it's in CC

frame_2D_alg/agg_recursion.py
boris-kz: sub+ is nested in agg+, so the levels added in in sub+ should already be in N.H

boris-kz: This was astra posting for me:

Sub+ levels do reach the node used by the next agg+. In the current head (`3251297`), `sum2G` calls `cross_comp(G, ...)` at L523. Accepted `cluster_N` then updates that same `G.N_` and `G.H` at L303–305. `CN.H` is a property of `CN.Nt.H`, so this survives the return from `sum2G`. When that G becomes a member N of a higher graph, `add_Nt` inherits `N.H` at L560.

That establishes inheritance. Preserving the previous level when replacing the current `root.N_` is a separate operation: the members' histories do not by themselves replace that snapshot. We should add levels for actual composition transitions, not an additional level merely because execution passes from sub+ back to agg+.

The medoid call at L369 should therefore use `root`. With `F2N(Ct)`, successful reclustering assigns the medoid graphs to `Ct.N_`. Since the parent's `C_` aliases that field, subsequent `sum2G` reads those graphs as centroids at L516. Keep the centroid representation in `root.Ct` and aggregate the selected original nodes under `root`. Remove the separate G_ cross-comp at L309–310 as already discussed.

Other implications of the edits:

- **Old links:** L523 still makes clustering depend on the cross-comp gate, and `cross_comp` only reaches `cluster_N` under `if L_` and its new-link evaluation. After the new existing-pair exclusion, an empty new-link batch skips clustering entirely. Existing links in node rims cannot help if the call never happens. Clustering needs its own evaluation over the intended node set and accumulated links.
- **Reset scope:** moving the `fin` reset to termination needs to account for nested calls. L295 enters `sum2G` while the outer seed loop is still active; an inner reset of shared member flags can make L266 revisit an already processed outer seed. Finish the outer flood-fill batch before recursive packing, or preserve its visit state across the inner run. Reset timing alone does not separate those scopes.
- **Hierarchy-only packing:** the concern at L561 is real: `sum2F(n_, G)` calls `add2F(G,F,2)`, changing G's totals before appending the level. `comb_Ft` then derives `G.m`, which the ensuing projection reads. Packing already represented descendants into H should not also accumulate their totals into G.
- **Keep the H-preservation edit in F2N.** `Nt.H = F.H` must occur before changing the class, because `F.H` subsequently resolves through the new Nt. The named `nexp=nexp+1` correction is also needed; the new derivation's range should start at the default 1.

boris-kz: astra:

[[Posted the review in the discussion.](https://github.com/boris-kz/CogAlg/pull/275#discussion_r4061364093)](https://github.com/boris-kz/CogAlg/pull/275#discussion_r4061364093)

Sub+ updates the same `G` returned by `sum2G`; `G.H` aliases `G.Nt.H`. Higher aggregation inherits those levels through `add_Nt`. Preserving the previous level when replacing `root.N_` remains a separate operation.

Main corrections:

* **L369:** use `root` for medoid aggregation. `F2N(Ct)` lets reclustering overwrite centroid membership with connectivity graphs.
* **L309–310:** remove the duplicate G cross-comparison.
* **L523:** evaluate clustering independently of new comparisons; otherwise an empty new-link batch prevents reclustering existing links.
* **L264/L295:** nested runs share `fin`. Moving reset to termination alone can make the still-active outer loop revisit processed nodes.
* **L561:** `sum2F(n_, G)` also accumulates descendants into G’s totals. Packing their hierarchy should avoid that additional accumulation.

The `F2N` hierarchy-preservation edit and named `nexp` argument correction are valid.

frame_2D_alg/agg_recursion.py
boris-kz: there are bigger issues to consider here

frame_2D_alg/agg_recursion.py
kwcckw: Okay, so i think we actually need to run `add_Nt` after we reassign `root.G_` in cluster_N:

    root.N_ = G_; root.dTT=TT; root.c=C; root.r=R; root.m=M; root.d=D; add_Nt(root)

In that way all the prior H levels should be reformed and G_ will become the top level in root.

frame_2D_alg/agg_recursion.py
kwcckw: > here we need to cluster_N with old links too, regardless of xcomp

Do you mean we should run cluster_N with old links in the else section of xcomp?

But this G is already formed with old links+new links, and why we need recluster with old links again?

frame_2D_alg/agg_recursion.py
kwcckw: Yes, but in term of eval, get_exemplars should be applied with the med_ above so that we can eval with the final exemplars params. Also we probably need to recomp the values with the final exemplars only before the cross_comp eval?

    _m,_d,_tt,_c,_r = sum_vt(exemplars, fm=1)

frame_2D_alg/agg_recursion.py
boris-kz: it's done in sum2G forming root

frame_2D_alg/agg_recursion.py
boris-kz: not else, we may just skip xcomp and go striaght to clustering

frame_2D_alg/agg_recursion.py
kwcckw: Okay right, that add_Nt is redundant since H is formed in add_Nt via comb_Ft of sum2G, while any other new level will be added in cluster_N anyway: `root.H += [Copy_(root)]  # add H for Nt, also Ct?`.

frame_2D_alg/agg_recursion.py
kwcckw: Okay, so this is actually similar to one of our prior scheme. If there's more Gs with meaningful old links, it should take a shortcut and cluster with old links? Since those Gs shouldn't have any new rim yet and those old links should be from existing N_'s rim?

frame_2D_alg/agg_recursion.py
boris-kz: so,

    if G_:
        for G in G_: trans_cluster(G)  # splice trans_links, merge L.N_.roots
        C = sum([g[1] for g in Gt_]); TT=np.zeros((2,9)); R=0; _r+=1  # + wC?
        for tt,c,r in Gt_: w=c/C; TT+=tt*w; R+=r*w; 
        M,D = val_(TT* root.wTT*ttcN, fd=1)
        if gv_(M * (C*wcN /((_r+R)*ccN)) * ((len(G_)-1)*wL) - ave): 
            sum2F(G_,root, nF='Nt')  # reform root,Nt, no other forks yet, redundant to lower levs
            cluster_C(root, G_,R,C)  # default-> med_'agg+, or if D * (C*wcC / (R*ccC)) * ((len(G_)-1)*wL) > avd:  # high variance?
        FV_(CoF.get(), *sum_vt(G_)[:-1],R)
    return G_
?

boris-kz: that's astra again:

Calling `cluster_C` by default here fits the intended sequence: CN -> CC -> medoids -> agg+. Once the separate G_ cross-comp is removed, keeping the high-D condition would terminate that route for low-D batches. Also, this D comes from the averaged derivative tuple; it does not directly measure dispersion between the Gs around a centroid.

The proposed `sum2F(G_,root,nF='Nt')` needs replacement semantics first. Currently it reaches `add2F(root,F,2)`, which adds totals but skips updating `root.N_`. The returned F is discarded. Then `cluster_C` copies the old `root.N_` and initializes those nodes' `_root_`, while its exemplar loop uses G_. Newly formed Gs can consequently reach `E._root_ += ...` without that initialization.

So the helper must install G_ as the current Nt, preserve the displaced level once, and replace its summary without accumulating the same represented content again. Until that is implemented, retain the explicit root-replacement block.

One other change, at L298–300: Gt_ contains values captured before `sum2G` finishes recursive refinement and before `trans_cluster`. Evaluate the resulting graphs:

```python
        TT,C,R = sum_vt(G_); _r+=1
```

That gives the following gate and `cluster_C` the returned graphs' current totals.

sum2G: packs Nt,Lt,Bt,Ct into a new G, then evaluates sub+ comparisons of its members and boundary links.
Here: packs G_ as the existing root’s new Nt, then runs cluster_C → med_ → agg+ cross-comparison.

The member Gs already contain their other forks; this step forms only the enclosing Nt.

There is also creation versus replacement: sum2G creates a graph, whereas this operation replaces root.Nt, preserving the previous level without adding its totals again.

boris-kz: astra draft to pack current Nt before adding a new one in sub+:

    fC = any(m_)
    typ = 0 if nF=='Nt' else 2 if fC else N.typ  # Nt: summary only
    if fC: w_ *= np.array(m_)/sum(m_)  # differential initialization to break symmetry
    cls_ = [CF,CL,CL,CN]  # typ=2 CCs
        if i:
            TT += n.dTT*w; R+=n.r*w; n_ += (n.N_ if merge else [n])
            if typ:  # links and higher
                if n.angl is not None: angl = copy(n.angl[0]) if angl is None else angl+n.angl[0]
                if typ==3: box=extend_box(box,n.box)
        else:  # init
            if typ:
                if typ==3: box=copy(n.box)
    F = (cls_[typ])(dTT=TT, c=C, r=R, nF=nF); F.N_ = n_
    if fC:
        F.m_,F.d_ = m_,d_  # for CCs only
        F.m, F.d = sum(m*c for m,c in zip(m_,c_))/C, sum(d*c for d,c in zip(d_,c_))/C
    else:
        F.m, F.d = val_(TT, fd=1)
    F.w = F.m if fC else 0
    if typ:
        if typ==3:
            F.box = box
            F.Nt.dTT = copy(TT); F.Nt.c = C; F.Nt.r = R; F.Nt.m,F.Nt.d = F.m,F.d
    if root is not None:
        F.wTT = root.wTT
        if nF=='Nt':
            F.H = copy(root.H)
            if root.Nt: F.H += [Copy_(root.Nt, root=root)]  # preserve previous top level
            root.Nt = F; F.root = root
            root.dTT=copy(F.dTT); root.m,root.d,root.c,root.r = F.m,F.d,F.c,F.r
        elif nF not in ('Ct','Rt'): add2F(root,F,2)
    if froot == 1:
    if fC:
        for N, m, d in zip(N_, m_, d_): N.root_ += [[F,m,d]]
    return F

Three remaining issues:

1. **`add_Nt` still double-counts descendants** (line 548). `sum2F(n_, G)` calls `add2F(G,F,2)`, adding the descendants’ totals back into `G`. For graphs without `Lt` or `Bt`, `sum2G` never corrects this before returning. A focused probe produced `G.c=14` with `G.Nt.c=7`. Preserve the history level without accumulating its summary into `G`.

2. **The final CC summary is discarded** (`cluster_C`, line 352). `sum2F(oC_,root,nF='Ct')` neither assigns `root.Ct` nor updates it internally. Consequently, later `sum2G` reads stale `N.C_`, and `comp_N` sees an empty or outdated centroid fork. Assign the returned summary to `root.Ct` and set its parent.

3. **The aggregation gate uses pre-recursion totals** (`cluster_N`, line 296). `Gt_` stores the candidate totals calculated before `sum2G` performs sub-recursion. The gate and subsequent CC receive those old totals, while `sum2F(G_,root,nF='Nt')` commits the current graph totals. Compute `TT,C,R` from the resulting `G_`.

On **CC priority**: freshly formed graphs have empty rims. Therefore, this unconditional `cluster_C(root,G_,...)` starts with singleton candidates and cannot expand across graphs unless it escalates to `cluster_P`. Local CC does not provide inter-graph clustering before xcomp establishes those connections; it need not be a mandatory prerequisite for agg+.

The snapshot preservation and first two findings were checked with focused helper probes.

kwcckw: In the updated sum2F code above, i think this H level should be enabled for both nF==Nt and nF==Bt? But they should be still using root.Nt: 

      if root is not None:
          F.wTT = root.wTT
          if nF=='Nt' or nF == 'Bt':
              F.H = copy(root.H)
              if root.Nt: F.H += [Copy_(root.Nt, root=root)]  # preserve previous top level
              root.Nt = F; F.root = root
              root.dTT=copy(F.dTT); root.m,root.d,root.c,root.r = F.m,F.d,F.c,F.r

frame_2D_alg/agg_recursion.py
boris-kz: exemplars are clustering seeds, they should not be used in xcomp

frame_2D_alg/agg_recursion.py
boris-kz: it's any typ=3, whatever the nF?
but as stated above, current CC can't be done before xcomp, it requires rims.
so it can only run on the same nodes as CN, in sub+?

kwcckw: > add_Nt still double-counts descendants (line 548). sum2F(n_, G) calls add2F(G,F,2), adding the descendants’ totals back into G

I think only kern is redundant since we include kern in add_Nt, i guess we should remove that `G.kern += N.kern*w; yx = N.yx; yx_+=[yx]  # * w?` in add_Nt.

kwcckw: > it's any typ=3, whatever the nF?

B fork has typ == 1 since it's negative L:

    cross_comp(F2N(G.Bt), proj_L_(combinations([F2N(L) for L in Bt.N_],2),G,br,nexp=nexp),br)

So i think we still need to check with their nF instead.

> but as stated above, current CC can't be done before xcomp, it requires rims.
so it can only run on the same nodes as CN, in sub+?

Yes, unless it's an all to all clustering method, as in cluster_P, and the current cluster_C uses rim as frontier. In sub+, it's actually a refinement since it's reusing the old links only from existing nodes anyway.

So in the suggested code above:

    if gv_(M * (C*wcN /((_r+R)*ccN)) * ((len(G_)-1)*wL) - ave): 
        sum2F(G_,root, nF='Nt')  # reform root,Nt, no other forks yet, redundant to lower levs
        cluster_C(root, G_,R,C)  # default-> med_'agg+, or if D * (C*wcC / (R*ccC)) * ((len(G_)-1)*wL) > avd:  # high variance?
        
That cluster_C should be using combined G_'s N_ instead?

boris-kz: the links are converted to CN before xcomp
CC should be in sub+ only, in sum2G.
probably as an alternative to CN, if high D

kwcckw: > the links are converted to CN before xcomp

Yes, but we don't change their typ. We would need to change their typ if we check their typ here. 

> CC should be in sub+ only, in sum2G.
probably as an alternative to CN, if high D

So in sub+, teh resulting CC should replace G.N_?

boris-kz: Yes, replace typ. No, C_ never replaces G N_, it's only to get medoids.
I will check tomorrow

PR #276, opened 2026-09-22
9/22/2026

kwcckw:


frame_2D_alg/agg_recursion.py
```diff
@@ -476,7 +476,7 @@ def add2F(F, n, merge=0):  # unpack for batching in sum2F
     F.m, F.d = val_(F.dTT,fd=1)
     if merge <2:
         F.N_ += (n.N_ if merge else [n])
-    if hasattr(F,'H') and getattr(n,'H',None): add_H(F.H, n.H, F)  # redundant to add_Nt?
+    if fH and hasattr(F,'H') and getattr(n,'H',None): add_H(F.H, n.H, F)  # redundant to add_Nt?
```
kwcckw, line 479: This add_H is redundant if we call add2F after the add_Nt in comb_Ft, while the other sections still need it. So i added fH to exclude this add_H from the add2F in comb_Ft instead.

frame_2D_alg/agg_recursion.py
```diff
@@ -506,7 +506,14 @@ def sum2G(ft_, fTT, root=None, init=1):  # finalize cluster
     if Lt:= G.Lt:  # rng+'sub+
         lm,lc,lr = Lt.m,Lt.c,Lt.r  # no levR = 1/len(L_): represented by c
         if gv_(lm*lc*wX - Av* (lr+1+cX)):  # mdecay(L_)-decay? sub+:
-            cross_comp(G, proj_L_(combinations(N_,2), G,lr, nexp=nexp+1), lr+1)
+            cluster_C(G, N_,lr,lc,nexp+1)  # cross_comp within the cluster_C
```
kwcckw, line 509: > Yes, replace typ. No, C_ never replaces G N_, it's only to get medoids.
I will check tomorrow

I guess we can have cluster_C here, but the xcomp in cluster_C may replace G.N_ with the deeper graphs formed from the medoids now.

frame_2D_alg/agg_recursion.py
```diff
@@ -506,7 +506,14 @@ def sum2G(ft_, fTT, root=None, init=1):  # finalize cluster
...
+            # cross_comp(G, proj_L_(combinations(N_,2), G,lr, nexp=nexp+1), lr+1)
+        elif gv_(lm*lc*wcN - Av*(lr+1+ccN)):        # CN: default, cluster old+new links regardless of xcomp
```
kwcckw, line 511: >  eval cluster with old+new links, regardless of xcomp:
>  CN/M, CC/D, parallel because it's all sub+, or CC -> med_ for CN?
>  unpack Gs if V < r+1: redundancy to stronger sub G_|C_? or partial unpack: exclude stronger sub G_|C_?

So it will be something like this? We apply cluster_N to N_ to form sub Gs, and unpack sub G if their g.m < g.r+1?

frame_2D_alg/agg_recursion.py
boris-kz: then it should be in else to add_Nt?

frame_2D_alg/agg_recursion.py
boris-kz: the eval above is for xcomp, CC | CN selection is separate

frame_2D_alg/agg_recursion.py
boris-kz: it should be CC if D else CN?

frame_2D_alg/agg_recursion.py
```diff
@@ -157,7 +157,7 @@ def comp_F(_F, F, ir=0, rL=None):
             L = len(Np_)-1
             tt = (rL.dTT*rL.c + ddTT*dF.c) / (rL.c+dF.c)
             if gv_(val_(tt,ttF) * (c/r) * (wF*L) - ave* (r+cF*L)):
-                if l: L_= [L for Np in Np_ for L in comp_F(*Np, r,rL=dF).N_]; TT,C,R = sum_vt(L_, wTT=ttF)
+                if l: L_= [L for Np in Np_ for L in comp_F(*Np, r,rL=dF).N_]  # TT,C,R = sum_vt(L_, wTT=ttF)  (we don't need this TT, C and R? Same as the L_)
```
boris-kz, line 160: don't remember where that came from, but we need an eval for xcomp here?

frame_2D_alg/agg_recursion.py
kwcckw: But add_Nt is optional to pack geometric params only? So it shouldn't be default in the else section?

frame_2D_alg/agg_recursion.py
kwcckw: > it should be CC if D else CN?

Okay right, so high variance G should run cluster_C to further refinement their nodes, while we need another separate eval for CN direct clustering. To eval for CN direct clustering, it should be eval based on their Rt to check on their new and old links?

frame_2D_alg/agg_recursion.py
kwcckw: But the eval is already based on gv_ above?

    if gv_(val_(tt,ttF) * (c/r) * (wF*L) - ave* (r+cF*L)):
    
It should be more direct without xcomp again if the gv_ eval is true, else we rerun xcomp on those pairs?

boris-kz: without xcomp, nothing happens

frame_2D_alg/agg_recursion.py
boris-kz: we need a conditional call to add_Nt inside add2F?

frame_2D_alg/agg_recursion.py
boris-kz: it's after but not dependent on xcomp

boris-kz: just tried Opus 5.5, it worked for 20 min:

I couldn't post the reply because this session doesn't have GitHub API access, so it's below ready to paste. Your two replies there set the structure, and the PR code breaks both. CC runs inside the xcomp check, and CN is the `elif` of that same check, so both depend on xcomp. I tested the key claims on small made-up graphs with a placeholder `sum2G`. The fixes compile.

---

The CC|CN choice should come after the xcomp block and not depend on it:

```python
if Lt:= G.Lt:  # rng+'sub+
    lm,lc,lr = Lt.m,Lt.c,Lt.r
    if gv_(lm*lc*wX - Av* (lr+1+cX)):  # xcomp eval only
        cross_comp(G, proj_L_(combinations(N_,2), G,lr, nexp=nexp+1), lr+1, fsub=1)  # +rim links, no cluster_N
    # recluster by old+new internal links, regardless of xcomp:
    nS = set(N_); tt,c,r = sum_vt(list({l for n in N_ for l in n.rim if l.N_[0] in nS and l.N_[1] in nS}))
    m,d = val_(tt,fd=1); r+=1  # +sub level
    if   gv_(d*c*wcC - Av*(r+ccC)): cluster_C(G, get_exemplars(N_,r,c), r,c, nexp+1)  # CC if D
    elif gv_(m*c*wcN - Av*(r+ccN)): cluster_N(G, get_exemplars(N_,r,c), r,c)  # else CN, reforms G.Nt
```

1. **`cross_comp` needs a `fsub=0` argument:** use `if not fsub and gv_(...): return cluster_N(...)`. Otherwise its own `cluster_N`, which is gated only by the new links, keeps CN dependent on xcomp and runs in addition to the CC|CN choice. The Bt der+ call and the CC medoid call keep `fsub=0`, because they cluster into a new level.

2. **Which links to evaluate:**
   - Not `Rt`: rims include links to nodes outside G.
   - Not `G.Lt`: it holds only the flood-fill links, and `cross_comp` replaces it with the new Lt.
   - Use G's internal rim links: all old plus new, including the negative ones that carry D.
   - With `fsub`, it may be better to extend `root.Lt` (`add2F(_Lt,Lt,1)`) than to replace it, since the new links are on the same level.

3. **`fin` is a global flag, and both settings break something:**
   - With `fin=1` in `add_Nt` (master), nested `cluster_N` skips every seed. So does frame agg+ after `sum2G(Fr)`, so both are dead.
   - With `fin=0` (this PR), a later exemplar that's already in a formed G re-clusters that G's members through the new sub+ links, which creates a duplicate G.
   - Also with `fin=0`, the nested flood-fill pulls in nodes outside G through their rims.
   - Both failures reproduce in a small test.
   - Fix: make `fin` local to each `cluster_N` pass, like `in_`, and add a scope check. Changed lines in `cluster_N`:
   ```python
   G_, Gt_, in_, fin_, nS = [],[],set(),set(), set(root.N_)
   if N in fin_ or (root.root and not N.exe): continue
   N_ = [N]; L_,B_ = [],[]; fin_.add(N)
   _N = L.N_[0] if L.N_[1] in fin_ else L.N_[1]; in_.add(L)
   if _N not in fin_ and _N in nS:
   N_ += [_N]; L_ += [L]; fin_.add(_N)
   _L_+= [l for l in _N.rim if l not in in_ and ((l.N_[0] in fin_) ^ (l.N_[1] in fin_))]
   ```
   After that, the `fin` assignments in `add_Nt` and `cluster_C` no longer matter.

4. **`sum2F`'s new `if typ == 3:` never runs for `nF='Nt'`**, because `typ` is 0 there. Consequences:
   - `cluster_N` no longer reforms `root.Nt`, so the `G_` from frame agg+ and from CC medoids is lost.
   - `add_Nt`'s `sum2F(n_, G)` now takes that branch whenever members' members are CNs. It replaces `G.Nt` with the grandchildren, which inverts the levels (confirmed in a test).
   - Fix: revert to `if nF=='Nt'` with `[Copy_(root.Nt, root=root)]`. In `add_Nt`, use `G.H += [sum2F(n_)]` with no root, so it isn't summed into G.
   - The manual pack in the CN branch is then redundant: `cluster_N`'s reform already puts N_ into `H[-1]` and sets `G.N_` to the sub-Gs.

5. **CC:**
   - Seed it with `get_exemplars`, not `N_`; right now every node is a seed.
   - Your point about medoids holds. The medoid `cross_comp → cluster_N(G)` would reform `G.Nt` with medoid graphs, which is the CC → med_ → CN chain.
   - After that, non-medoid members survive only in `H[-1]` and `root_`, because `oC_` is thrown away (the result of `sum2F(oC_,root,nF='Ct')` isn't assigned).
   - CC's frontier also comes from rims, so in sub+ it reaches nodes outside G and resets their `root_`/`_root_`.
   - Open question: keep CC inside G (filter by `nS`), or let it extend beyond G without those resets?

6. **Unpacking:** sub-Gs with V < r+1 must be unpacked before `cluster_N`'s reform (`sum2F(G_,root,nF='Nt')`), not after it. Members that aren't in any sub-G drop out of `G.N_` unless they're kept ("partial unpack").

7. **Scale:** `vd` from `val_` grows as Σ|d|·w·vm, while `vm` is bounded. So an absolute D check placed ahead of CN will mostly pre-empt CN. `sum_vt(..., fm=1)` puts m and d on comparable scales.

---

Sources: [[PR #276 thread r4074697683](https://github.com/boris-kz/CogAlg/pull/276/changes#r4074697683)](https://github.com/boris-kz/CogAlg/pull/276/changes#r4074697683), [[agg_recursion.py at PR head 45ac825](https://github.com/boris-kz/CogAlg/blob/45ac8259077a582c830395453805260f4d76821e/frame_2D_alg/agg_recursion.py)](https://github.com/boris-kz/CogAlg/blob/45ac8259077a582c830395453805260f4d76821e/frame_2D_alg/agg_recursion.py)

frame_2D_alg/agg_recursion.py
boris-kz: I think it should be

            if gv_(val_(tt,ttF) * (c/r) * (wF*L) - ave* (r+cF*L)):  # any sub-comp?
                if l:
                    rL.Nt = sum2F([L for Np in Np_ for L in comp_F(*Np, r,rL=dF).N_], root=rL)
                else:
                    if nF != 'Nt': [(F2N(_N),F2N(N)) for _N,N in Np_]  # convert for comp_N_ below
                    cross_comp(rL, proj_L_(Np_,2,r), r,dF,rL)  # root=2: N pair span computes decay?
?

frame_2D_alg/agg_recursion.py
boris-kz: so this fork should be in cross_comp, instead of CN only?

frame_2D_alg/agg_recursion.py
boris-kz: astra:
The nested `else` is correct. **The recursive results must populate `dF.N_`, not `rL.Nt`.**

In [[your proposed block](https://github.com/boris-kz/CogAlg/pull/276#discussion_r4077105905)](https://github.com/boris-kz/CogAlg/pull/276#discussion_r4077105905), replace the `if l` branch at line 160 with:

```python
                if l:
                    dF.N_ = [L for Np in Np_ for L in comp_F(*Np, r,rL=dF).N_]
```

Keep your relocated `else`.

The consumption chain is `comp_F` returns `dF` → `comp_N` packs it into `L.H[-1]` → `trans_cluster` reads each `tFt.N_`. Assigning `rL.Nt` leaves the returned `dF.N_` empty, so those trans-links cannot trigger merges. An isolated check produced eight links: your assignment exposed zero to that traversal; this replacement exposed all eight.

Also, **the extra `sum2F(..., root=rL)` would repeat accumulation**. Recursive `cross_comp` already updates the enclosing `dF` through its `rL` argument. There is an existing duplication to resolve there too: line 95 adds the new-link summary into `rL`, then line 97 adds `dF`, which already includes those links; `comp_N` subsequently sums the returned fork again. That changes the parent link’s `c` and `dTT`, hence its later clustering value.

opus 5.5:

The `else` move is right, but the Lt line isn't. At the moment `else` pairs with the `gv_` check, so `cross_comp` runs only when the check fails. When it passes, the non-Lt forks get nothing. That's a lost indent: the Feb–Mar versions had both branches inside the check.

I checked the problems below on two small made-up graphs, each with two members and one internal link.

**Problems with `rL.Nt = sum2F(...)`:**

1. **It produces the Nt fork's output again.**
   - The zip pairing ("same forks") dates from February, when `Lt.N_` held cross-fork dFs lined up by fork type; those now live in `Xt`.
   - `Lt.N_` is now links in flood-fill order, so zip pairs them arbitrarily.
   - Links have `nF='Nt'` by default, so `comp_F(_link, link)` goes down into their endpoints. `.N_` then collects member×member links.
   - In the test these were exactly the same 4 pairs that `comp_F(_N.Nt, N.Nt)` produces.
   - The actual der+ part, the link-vs-link ddTT in each dF_i, survives only when that endpoint sub-comp passes. Otherwise it's dropped.
2. **Nothing reads `rL.Nt`, and it gets overwritten.** `trans_cluster` resets `L.Nt,L.Bt,L.Ct` on links, and `F2N` overwrites `F.Nt`. In the recursive calls `rL` is the parent dF, where `.Nt` is a stray attribute. The other forks put their output in `dF.N_`, which goes through comp_N's `dn_` into `L.H[-1]`.
3. **It crashes on an empty list.** When `proj_L_` drops all endpoint pairs, `sum2F([])` raises IndexError at `N_[0]`. In the test, graphs far apart triggered it.

**Proposed:** drop the Lt special case and treat Lt as a link fork, the same as Bt:
```python
    if _F.nF == F.nF:  # sub-comp
        _N_,N_=_F.N_,F.N_; nF=F.nF
        if  _N_ and N_:
            Np_ = list(product(_N_,N_))  # pairs; Lt,Bt: links, Ct: centroids
            L = len(Np_)-1
            tt = (rL.dTT*rL.c + ddTT*dF.c) / (rL.c+dF.c)
            if gv_(val_(tt,ttF) * (c/r) * (wF*L) - ave* (r+cF*L)):  # any sub-comp?
                if nF != 'Nt': [(F2N(_N),F2N(N)) for _N,N in Np_]  # convert for comp_N below
                cross_comp(rL, proj_L_(Np_,2,r), r,dF,rL)  # -> dF.N_
```
In the test this gives link×link der+ in `dF.N_` (`comp_N` with `full=0`, i.e. `comp_derT` of `dTT[1]`), stored in the same place as the other forks. This also removes the unused `TT,C,R` that kwcckw asked about. The alternative, following your comment "no comp F.Lt: included in F.dTT?", is to skip the Lt sub-comp: `if _F.nF == F.nF and nF != 'Lt'`.

**Related changes this needs:**

4. **`cross_comp`'s dF path overwrites the parent link.** It runs `root.Lt = sum2F(L_, root, nF='Lt')` before the `if dF:` check. In the test this set `L.Lt` to the sub-links, and each fork overwrites it, so the last one wins. It also blends into `L`, then `add2F(rL,dF,2)` blends again, and comp_N's `sum2F(dn_,L)` a third time. Once the recursion is gone, `rL` is always that `L`, so:
   ```python
   Lt = sum2F(L_, None if dF else root, nF='Lt'); m,d,tt,c,r = Lt.m,Lt.d,Lt.dTT,Lt.c,Lt.r; FV_(CoF.get(),tt,c,r)
   if dF: add2F(dF,Lt,merge=1); return  # sub-links -> dF.N_; dF -> L once, via comp_N sum2F(dn_,L)
   root.Lt = Lt
   ```
   The cost is that the `tt` check for later forks no longer includes earlier forks' sub-comp results. That's minor.
5. **`trans_cluster` needs an Lt guard.** The comment "Lt doesn't form trans-links" is true now only because Lt's `dF.N_` is empty. With link×link links in it, add `if tFt.nF == 'Lt': continue`. Otherwise the root lookup on F2N'd links (`.root.root`) would merge graphs two levels up.
6. **The PR's `F.typ = 3` in F2N sends links back into their endpoints.** F2N'd links then pass `N.typ > 1` in `comp_N` and sub-compare their Nt, i.e. their endpoints. In the test that re-created the same 4 member pairs under the link-link comparison. With master's F2N, links keep their own typ and don't go down.

kwcckw: > The recursive results must populate dF.N_, not rL.Nt.

I would think the same. So the if section should be packing dF.N_ and their params?

        if gv_(val_(tt,ttF) * (c/r) * (wF*L) - ave* (r+cF*L)):  # any sub-comp?
            if l:
                L_= [L for Np in Np_ for L in comp_F(*Np, r,rL=dF).N_] 
                dF.N_,dF.dTT,dF.c,dF.r = L_,sum_vt(L_, wTT=ttF) 
            else:
                if nF != 'Nt': [(F2N(_N),F2N(N)) for _N,N in Np_]  # convert for comp_N_ below
                cross_comp(rL, proj_L_(Np_,2,r), r,dF,rL)  # root=2: N pair span computes decay?

frame_2D_alg/agg_recursion.py
kwcckw: Based on suggestion from opus 5.5:

      if Lt:= G.Lt:  # rng+'sub+
          lm,lc,lr = Lt.m,Lt.c,Lt.r
          if gv_(lm*lc*wX - Av* (lr+1+cX)):  # xcomp eval only
              cross_comp(G, proj_L_(combinations(N_,2), G,lr, nexp=nexp+1), lr+1, fsub=1)  # +rim links, no cluster_N
          # recluster by old+new internal links, regardless of xcomp:
          nS = set(N_); tt,c,r = sum_vt(list({l for n in N_ for l in n.rim if l.N_[0] in nS and l.N_[1] in nS}))
          m,d = val_(tt,fd=1); r+=1  # +sub level
          if   gv_(d*c*wcC - Av*(r+ccC)): cluster_C(G, get_exemplars(N_,r,c), r,c, nexp+1)  # CC if D
          elif gv_(m*c*wcN - Av*(r+ccN)): cluster_N(G, get_exemplars(N_,r,c), r,c)  # else CN, reforms G.Nt
          
          
Shouldn't be the cross_comp is after the cluster_C and cluster_N fork since the cross_comp should be using the sub Gs or Cs?

If the fork is in cross_comp, then it's actually more straight forward since we can just eval for xcomp here, and the forking will be done in cross_comp, as in one of our prior scheme. And now both forks should run recursively, and CC shouldn't be the termination fork right?

frame_2D_alg/agg_recursion.py
kwcckw: If we move it in add2F, i guess the add_Nt is only valid when we add Nt into G, but not the other Ct, Lt or Bt.

frame_2D_alg/agg_recursion.py
boris-kz: opus proposed removing l above:

    if _F.nF == F.nF:  # sub-comp
        _N_,N_=_F.N_,F.N_; nF=F.nF
        if  _N_ and N_:
            Np_ = list(product(_N_,N_))  # pairs; Lt,Bt: links, Ct: centroids
            L = len(Np_)-1
            tt = (rL.dTT*rL.c + ddTT*dF.c) / (rL.c+dF.c)
            if gv_(val_(tt,ttF) * (c/r) * (wF*L) - ave* (r+cF*L)):  # any sub-comp?
                if nF != 'Nt': [(F2N(_N),F2N(N)) for _N,N in Np_]  # convert for comp_N below
                cross_comp(rL, proj_L_(Np_,2,r), r,dF,rL)  # -> dF.N_

frame_2D_alg/agg_recursion.py
boris-kz: actually, we can unpack it in sum2F, 
astra:

    fC = any(m_)
    typ = 0 if nF=='Nt' else 2 if fC else N.typ  # Nt: summary only
    if fC: w_ *= np.array(m_)/sum(m_)  # differential initialization to break symmetry
    cls_ = [CF,CL,CL,CN]  # typ=2 CCs
        if i:
            TT += n.dTT*w; R+=n.r*w; n_ += (n.N_ if merge else [n])
            if 0 < typ < 3:
                if n.angl is not None: angl = copy(n.angl[0]) if angl is None else angl+n.angl[0]
        else:
            if 0 < typ < 3:
    F = (cls_[typ])(dTT=TT, c=C, r=R, nF=nF, typ=typ); F.N_ = n_
    if fC: F.m_,F.d_ = m_,d_; F.m,F.d = sum(m*c for m,c in zip(m_,c_))/C, sum(d*c for d,c in zip(d_,c_))/C; F.w=F.m
    else:  F.m, F.d = val_(TT,fd=1)
    if typ:
        if typ==3:
            N_ = F.N_; c_ = np.array([n.c for n in N_]); w_ = c_/c_.sum()
            yx_ = np.array([n.yx for n in N_]); F.yx = yx_.mean(axis=0)
            F.span = w_ @ np.hypot(*(yx_-F.yx).T) if len(N_)>1 else N_[0].span
            F.kern = np.zeros(4); F.root = root; angl = None
            for N,w in zip(N_,w_):
                N.fin = 0; N.root = F; F.root_ += N.root_
                F.kern += N.kern*w; F.box = extend_box(F.box,N.box)
                if N.angl is not None: angl = copy(N.angl[0]) if angl is None else angl+N.angl[0]
                add_H(F.H,N.H,F); add_H(F.Ct.H,N.Ct.H,F.Ct)
            if (n_:=[n for N in N_ for n in N.N_]):
                F.H += [sum2F(n_,nF='Nt')]; F.H[-1].root = F  # lower level, no member or root-value updates
            F.Nt.dTT = copy(TT); F.Nt.c = C; F.Nt.r = R; F.Nt.m,F.Nt.d = F.m,F.d
        else:
    if root is not None:
        F.wTT = root.wTT
        if nF=='Nt':
            F.H = copy(root.H) + [Copy_(root.Nt, root=root)]  # previous top level
            root.Nt = F; F.root = root; root.dTT=copy(F.dTT); root.m,root.d,root.c,root.r = F.m,F.d,F.c,F.r
        elif nF not in ('Ct','Rt'): add2F(root,F,2)
    if froot == 1 and typ!=3:
    if fC:
        for N, m, d in zip(N_, m_, d_): N.root_ += [[F,m,d]]
    return F

frame_2D_alg/agg_recursion.py
boris-kz: no, xcomp here is optional, to add rng+ Ls

frame_2D_alg/agg_recursion.py
boris-kz, line 86: update:

def cross_comp(root, pL_,r, dF=None,rL=None, fall=1, fsub=0):  # recursion root

    L_,N_ = [],[]
    for dist, dy_dx, _N,N, lc,lr, pTT,m,_,nexp in pL_:
        if _N != N and (fall or (m>0 and gv_(m*(lc*wN/(lr*cN)) - ave* (r+cN)))):
            Link = comp_N(_N,N, lr,lc, full=not dF, A=dy_dx, span=dist, rL=root)
            Link.nexp=nexp; Link.rTT = np.abs(pTT-Link.dTT) / eps_(Link.dTT)  # relative prediction error/oF
            L_+=[Link]; N_+=[_N,N]
    if L_:
        Lt = sum2F(L_,None if dF else root,nF='Lt')
        m,d,tt,c,r = Lt.m,Lt.d,Lt.dTT,Lt.c,Lt.r; v=FV_(CoF.get(),tt,c,r)
        if dF:  # comp_F: caller accumulates returned dF
            add2F(dF,Lt,merge=1); return
        root.Lt = Lt
        for N in (N_:= list(set(N_))): N.Rt = sum2F(N.rim,root=N,nF='Rt') if N.rim else CF(root=N,nF='Rt')
        if not fsub and v > ave and gv_(m * (c*wcN/(r*ccN)) * ((len(L_)-1)*wL) - ave):
            return cluster_N(root,get_exemplars(N_,r,c),r,c)  #-> CC,agg+

Opus:

Yes, the structure works. I ran it on the same small test graphs:

- **dF path (called from `comp_F`):** the link×link results go only into `dF.N_`, and `L.Lt` stays unset.
- **`fsub=1` path:** it sets `root.Lt`, updates `Rt` and skips `cluster_N`.

Four points:

1. **`rL` is now unused.** Drop it from the signature and change the call in `comp_F` to `cross_comp(rL, proj_L_(Np_,2,r), r, dF)`. If you drop it but leave `comp_F`'s call unchanged, the fifth positional argument lands in `fall`.

2. **The `cluster_N` gate has changed in two ways.**
   - **Weights:** `m` is `Lt.m` from `sum2F`'s `val_(TT,fd=1)`, which uses the default weights. Master used `val_(tt,ttcN)`, so the gate no longer uses `cluster_N`'s learned attribute weights.
   - **`v > ave`:** it isn't wrapped in `gv_`, so when it blocks, nothing is recorded in `gV_` for feedback. It's also mostly redundant. `v = val_(tt*ttX)*(c/r)` grows with `c`, so it exceeds `ave` whenever net match is positive. That's the same condition the `gv_` tests.

   Either use one gate:
   ```python
   if not fsub and gv_(val_(tt,ttcN) * (c*wcN/(r*ccN)) * ((len(L_)-1)*wL) - ave):
   ```
   or, if `v` is meant to mean "xcomp paid off", wrap it in `gv_` as well.

3. **With `fsub`, `root.Lt = Lt` replaces `G.Lt` with only the new nexp+1 links.** The links that formed G then drop out of three places: `G.Lt` itself, G's final `sum_vt([G.Nt,G.Lt,G.Bt])`, and later `comp_F` on `G.Lt`. If sub+ should add to `G.Lt` instead:
   ```python
   root.Lt = add2F(root.Lt,Lt,1) if fsub and root.Lt else Lt
   ```

4. **This assumes `comp_F` no longer has the recursive Lt branch.** If that branch stays, its inner dF values no longer reach the parent dF. They only got there through the `add2F(rL,dF,2)` you removed.

kwcckw: > no, xcomp here is optional, to add rng+ Ls

Okay, so we still need the CN and CC forks in sum2G, and cross_comp remains the same with cluster_N only?

      if   gv_(d*c*wcC - Av*(r+ccC)): cluster_C(G, get_exemplars(N_,r,c), r,c, nexp+1)  # CC if D
      elif gv_(m*c*wcN - Av*(r+ccN)): cluster_N(G, get_exemplars(N_,r,c), r,c)  # else CN, reforms G.Nt
     
But both forks above may reform G.N_ via the deeper recursion in sum2G?

boris-kz: Xcomp is CN only because that's always the starting point, but in sub+ we already have current-level N_

kwcckw: Yes, let me check the updated code again. So actually CC mainly is selecting medoid for xcomp now.

